PR #6, opened 2025-07-10
Support for Whisper model in vLLM (WIP)

rebel-eunji:
### 🚀 Pull Request Summary

- Whisper model is now supported in vLLM!
- It will work after additional patch in optimum-rbln (@rebel-kblee)
---

---

### ✅ Changes

* [x] Feature
* [ ] Bug Fix
* [ ] Documentation
* [ ] Refactor
* [ ] Optimization
* [ ] Other (please describe):

---

---

### 🧪 How to Test

1. Run `python examples/optimum/run_whisper.py`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📋 Checklist

* [ ] I’ve added unit tests or updated existing ones.
* [ ] I’ve updated relevant documentation.
* [ ] I’ve checked compatibility (e.g., hardware/backend/platform).

---

### 💬 Notes
The decoding steps differ slightly from optimum-rbln when running the same prompt, because the decoding process itself is different.
In optimum-rbln (and transformers), three tokens are processed together in the first decoding step.
However, this is not supported in vLLM, as vLLM Engine V0 always schedules only one token per step.
Therefore, we follow the vLLM style. It generates the tokens correctly, but the results and exit points may differ slightly from optimum-rbln (and Transformers).

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/whisper.py
+++ b/vllm_rbln/model_executor/models/optimum/whisper.py
@@ -97,7 +97,7 @@ class RBLNOptimumWhisperForConditionalGeneration(RBLNOptimumModelBase, RBLNOptim
 
         finished_requests_ids = model_input.finished_requests_ids
         running_requests_ids = model_input.running_requests_ids
-        # request_nums = input_ids.shape[0]
+        request_nums = input_ids.shape[0]
 
         table_ids = self.get_table_id(is_prompt, finished_requests_ids, running_requests_ids)
         valid_block_ids = torch.tensor(table_ids)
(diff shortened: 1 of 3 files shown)


vllm_rbln/model_executor/models/optimum/whisper.py
```diff
@@ -0,0 +1,154 @@
...
+                available_ids = set(range(self.decoder_batch_size)) - used_ids
+                assert len(available_ids) > 0
+                table_id = min(available_ids)
+            return [table_id]
+        else:
+            table_ids = []
+            for request_id in running_requests_ids:
+                table_id = self.table_mapping[request_id]
+                table_ids.append(table_id)
+            return table_ids
```
rebel-kblee, line 73: can make it to general format?

vllm_rbln/model_executor/models/optimum/whisper.py
```diff
@@ -0,0 +1,154 @@
...
+        input_ids = kwargs.pop("input_ids")
+        cache_position = kwargs.pop("cache_position")
+        block_tables = kwargs.pop("block_tables")
+
+        if is_prompt:
+            _ = self.model.encoder(input_features=input_features,
+                                   block_tables=block_tables)
+            # FIXME Is `is_languaged_detected` required
+            # unless the time stamp is not supported in vLLM?
+            self.model.is_language_detected = True
```
rebel-kblee, line 112: why this configuration set to True?

rebel-eunji: Thanks for the clear explanation you gave in person. Language detection task is not enabled in vLLM, so I'll remove it.

rebel-kblee (approved):
LGTM 👍

PR #57, opened 2025-08-19
gha: do not use cuda torch

dtrifiro:
- Set `PIP_EXTRA_INDEX_URL` to install torch cpu version
- use `vllm@git+https://` schema to force rebuilding vllm


dtrifiro:
@rebel-jonghewk marked this as ready and rebased it on the latest main, put some other ideas in #67, still as draft

rebel-eunji:
We previously used vLLM CUDA and Torch CUDA because using vLLM on CPU required building from source, which caused inconvenience for users.
However, we have now started using prebuilt vLLM CPU and Torch CPU instead.
Thank you for your contribution. I'll close this PR.

PR #67, opened 2025-08-28
gha: do not use cuda torch improvements

dtrifiro:
- using uv instead of pip
- adding `--no-binary=vllm` to force a vllm rebuild

### DIFF
--- a/.github/workflows/rbln_optimum_ci.yaml
+++ b/.github/workflows/rbln_optimum_ci.yaml
@@ -71,18 +71,21 @@ jobs:
 
       - name: Install local vllm-rbln package and dependencies
         run: |
-          pip install packaging setuptools wheel simphile pynvml huggingface_hub setuptools_scm fire
+          pip install uv
+          uv pip install packaging setuptools wheel simphile pynvml huggingface_hub setuptools_scm fire
 
-          pip install \
+          uv pip install \
             --force-reinstall \
             --no-cache-dir \
+            --no-binary "vllm" \
             --constraint <(echo "optimum-rbln==$VERSION") \
                 dist/vllm_rbln*.whl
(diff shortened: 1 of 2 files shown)


dtrifiro:
Based on #57

.github/workflows/rbln_optimum_ci.yaml
```diff
@@ -71,10 +71,21 @@ jobs:
...
-          pip install packaging setuptools wheel simphile pynvml huggingface_hub setuptools_scm fire
-          VERSION=${{ inputs.optimum_rbln_version }}
-          pip install --force-reinstall --no-cache-dir dist/vllm_rbln*.whl --constraint <(echo "optimum-rbln==$VERSION")
+          pip install uv
+          uv pip install packaging setuptools wheel simphile pynvml huggingface_hub setuptools_scm fire
+
+          uv pip install \
+            --force-reinstall \
+            --no-cache-dir \
+            --no-binary "vllm" \
```
dtrifiro, line 80: I guess this works well in CI scenarios, but we need a way to make sure that when users install the plugin from pypi, they have

```
VLLM_TARGET_DEVICE=cpu
UV_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu
UV_INDEX_STRATEGY="unsafe-best-match"`
```

set and use `--no-binary=vllm`.

Perhaps we can provide a vllm cpu wheel that you can install via an extra index url that could simplify this process. More thought needed.

rebel-jonghewk (approved):
https://github.com/rebellions-sw/vllm-rbln/actions/runs/17910125135
LGTM. works well with our current ci.

rebel-eunji:
Finally, we removes dependencies on cuda torch. Thanks. #439

PR #112, opened 2025-10-16
other: Script for debugging and metrics

rebel-eunji:
### 🚀 Summary of Changes

- Debugging
  - Given the length of prompt and generation step, we can validate the result
- Metrics
  - Given the length of prompt and generation step, we can measure the latency and throughput

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`bug-fix`)

---

### 🧪 How to Test

1. Run
- `python scripts/validation/compare_logprobs_advanced.py`
- `python scripts/performance/measure_latency_advanced.py`

---

---

### 📋 Checklist

---

### 💬 Notes
- TODO
  - Add device time
  - Add TTFT
  - Add warmup option
---

### DIFF
--- a/scripts/performance/measure_latency_advanced.py
+++ b/scripts/performance/measure_latency_advanced.py
@@ -12,14 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 # Reference - https://github.com/vllm-project/vllm/blob/v0.9.1/benchmarks/benchmark_throughput.py
 import argparse
 import os
 import time
 import urllib.request
-from multiprocessing import get_context
-from multiprocessing.queues import Queue as MPQueue
 from typing import TYPE_CHECKING, Any
 
 import torch
@@ -31,6 +28,7 @@ if TYPE_CHECKING:
 
 MODEL_NAME = "meta-llama/Llama-3.2-1B"
 PREFILL_CHUNK_SIZE = 128
+
 
 def get_wiki_prompt():
     wiki_txt_url = "https://raw.githubusercontent.com/huggingface/optimum-neuron/refs/heads/main/benchmark/text-generation/performance/wiki.txt"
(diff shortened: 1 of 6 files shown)


rebel-eunji:
@huijjj 
Thank you for your review.

The purpose of this test is a bit different from what you described.
It aims to measure the **average** metrics (not each logging step) of the prefill and decode phases **separately** over the multiple iteration.
Therefore, reusing the built-in metric logging in vLLM may not be suitable for our purpose.

huijjj:
> > > @huijjj Thank you for your review.
> > > The purpose of this test is a bit different from what you described. It aims to measure the **average** metrics (not each logging step) of the prefill and decode phases **separately** over the multiple iteration simply (without any additional service like Prometheus, Grafana..) Therefore, reusing the built-in metric logging in vLLM may not be suitable for our purpose.
> > 
> > 
> > I see, your goal is to measure prefill and decode separately within a single LLMEngine .generate call. I’m not entirely convinced this change is necessary. It might be better to achieve that without modifying the core path. For instance we could approximate by running the same workload twice: (a) with the output length fixed to 1, and (b) with the normal output length, then comparing the results. It’s not perfect as scheduler and KV-cache effects could skew the comparison, but it could provide us good enough results while keeping the core untouched.
> > For benchmarking, our team have been using the scripts in the [benchmark-script branch](https://github.com/rebellions-sw/vllm-rbln/tree/benchmark-script) for performance reporting. Most of that code is copy-pasted from vLLM’s benchmarking scripts, per our prior agreement, so we can produce comparable and objective results. In this PR I see new benchmarking scripts and metrics, could you share the context for introducing them? Do you intend for us to use these for future measurements and reporting?
> 
> Hello, and thank you for sharing your idea and benchmark. After discussing it with our team, we’ve decided to proceed with this approach for our implementation. We need to measure **fine-grained** latency and throughput **directly**, rather than relying on indirect methods for internal development.
> 
> The vLLM’s metric manager is tightly coupled with the vLLM engine and includes vLLM operations. I think it is good for service monitoring, but not for development of device operation. In addition, we plan to measure the latency of running operations at the RBLN device level in a more low-level manner. Therefore, modifying the core components is unavoidable, but I’ll do my best to minimize the changes. And this metric is used solely for internal development and is not associated with official benchmarking or reporting.

@rebel-eunji 
Understood. As performance evaluation is an important part of our collaboration, I wanted to clarify to ensure we're on the same page. Thank you for the kind and thorough explanation!

Issue #166, opened 2025-11-24
[feature] LoRA support on V1


junstar92:
LoRA (Low-Rank Adaptation) support for vLLM-RBLN V1 on RBLN hardware.

### Graph Compilation Constraints

**Pre-compilation requirement**: LoRA weights must be prepared before graph compilation
- Currently, there is no direct method to access device memory within the compiled graph at runtime
- All LoRA weight structures must be known during the compilation phase
- Dynamic LoRA loading after compilation is not supported now

### LoRA Masking Strategy

**Fixed Graph with Masking**: Always compute for N LoRA (for multi-lora) adapters with masking to apply only active ones
- Maintains consistent computation graph and tensor shapes across requests
- Uses masking mechanism to enable/disable specific LoRA adapters
- Accepts computational overhead to preserve graph stability
- Ensures predictable memory layout and execution flow

### Implementation Details

**Pure Matmul Implementation**: Initially implementing with standard matrix multiplication operations only
- No custom kernels in the first iteration
- **Future Consideration**: Custom kernel optimization may be needed for better performance

## Expected Limitations

- V1 only
- Single LoRA: Multi-LoRA not supported due to prefill scheduling constraints
- No warmup: Warmup must be disabled; first request must include LoRA request
- Static Compilation: Cannot add new LoRA adapters after graph compilation

PR #215, opened 2025-12-20
fix: incorrect LoRA logics and some test cases

junstar92:
### 🚀 Summary of Changes

This PR includes modifications to LoRA test cases and minor code fixes.

- The existing tests contained several cases that were not suitable for RBLN devices. These have been updated to properly reflect RBLN-specific characteristics and requirements (e.g., `torch.compile` with rbln backend).
- While examining the tests, which cover multi-LoRA scenarios in addition to single LoRA, I discovered a logic error that would occur when multi-LoRA is applied. Although this issue does not affect the current single LoRA implementations, the fix has been included to ensure correctness for future multi-LoRA use cases. This fix does not impact existing functionality.

---

* Related to #166 

---

### ✅ Type of Change

* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`bug-fix`)
* [x] ❓ Other (`other`): test

---

### 🧪 How to Test

Run `pytest tests/torch_compile/v1/lora`

---

---

### 📋 Checklist

---

### 💬 Notes

While examining the code `run_lora_test.py`, I found that the results were very inaccurate when loading the mode in BF16(original mode dtype) instead of FP32, so I added an environment variable to the example code to load the model in FP32.

---


junstar92:
@rebel-jiwoopark I changed the base branch from `dev` to `dev-0.12`.

junstar92:
For vLLM v0.12, some logics need to be fixed. I’ll let you know once it’s done.

junstar92:
@rebel-jiwoopark I completed the changes for v0.12 and confirmed that all tests passed.
Also, you can run the example code `examples/experimental/run_lora_test.py`.

.github/workflows/rbln_dispatch_trigger_on_pr_ci.yaml
```diff
@@ -13,7 +13,7 @@ permissions:
 
 jobs:
   sync-and-call:
-    runs-on: runner-vllm-ci
+    runs-on: ubuntu-latest-rbln
```
rebel-jiwoopark: This looks like dev branch changes that are unrelated to the current PR. Could you limit the scope of this PR to LoRA only, so that the CI workflow–related parts are not affected?

junstar92: It looks like this was accidentally included during the rebase. I removed it.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1845,7 +1853,8 @@ def execute_model(
                     else:  # decode
                         logits = self.compute_logits(hidden_states)
                         logits = logits[logits_indices]
-                    logits = self.logits_processor._gather_logits(logits)
+                    if envs.VLLM_RBLN_LOGITS_ALL_GATHER:
```
rebel-jiwoopark: Why Could you clarify why this condition is needed?

junstar92: When the`VLLM_RBLN_LOGITS_ALL_GATHER` environment variable is enabled, custom logit functions are applied. In this case, unlike the original vLLM implementation, the `_get_logits` function no longer calls the gather operation internally.

When LoRA is enabled, model execution does not include logit computation and gathering (`compute_logits`). If `VLLM_RBLN_LOGITS_ALL_GATHER` is disabled in this scenario, the gather operation is applied twice, which leads to incorrect behavior.

rebel-jiwoopark: I understood. Thank you for the detailed explanation.

PR #291, opened 2026-01-20
fix(model): pad image tokens using an out-of-vocab index

rebel-eunji:
### 🚀 Summary of Changes

The benchmark random dataset uses pad_token_id (0) as a valid token ID, which conflicts with the assumption that pad_token_id is not used as input. Previously, pad_token_id was only used for padding image tokens.
So this PR adopted a new token, out-of-vocab index to pad image tokens and Gemma 3 works correctly with the benchmark code.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`bug-fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/gemma3.py
+++ b/vllm_rbln/model_executor/models/optimum/gemma3.py
@@ -15,7 +15,6 @@ from typing import Any, Optional
 
 import torch
 import vllm.envs as envs
-from transformers import AutoTokenizer
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
 from vllm.model_executor.models.gemma3_mm import (Gemma3DummyInputsBuilder,
@@ -38,6 +37,7 @@ logger = init_logger(__name__)
 
 IMG_PAD_TOKEN_ID = -1
 PAD_TOKEN_ID = 0
+
 
 class RBLNGemma3MultiModalProcessor(Gemma3MultiModalProcessor):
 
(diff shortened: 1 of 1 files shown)


vllm_rbln/model_executor/models/optimum/gemma3.py
```diff
@@ -92,9 +95,15 @@ def __init__(
         vllm_config: VllmConfig,
     ) -> None:
         super().__init__(vllm_config=vllm_config)
+        model_path = self.vllm_config.model_config.model
+        tokenizer = AutoTokenizer.from_pretrained(model_path,
```
rebel-thkim: Wouldn't this be an overhead?

rebel-eunji: I agree. So I’ll just use a global variable instead of extracting the vocabulary size from the tokenizer.

vllm_rbln/model_executor/models/optimum/gemma3.py
```diff
@@ -36,6 +35,9 @@
 
 logger = init_logger(__name__)
 
+IMG_PAD_TOKEN_ID = -1
+PAD_TOKEN_ID = 0
+
```
rebel-thkim: This is code that will be reverted someday, right? 
IMG_PAD_TOKEN_ID = -1 should be fine, but I'm not sure if PAD_TOKEN_ID should be hardcoded.

PR #376, opened 2026-02-13
update: support data parallel without expert parallel

rebel-wonsubkim:
* support data parallel without expert parallel for MoE model
* fix up sampler warm-up for pipeline parallel
* disable dp metadata if ep is disabled


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3202,7 +3200,8 @@ def maybe_randomize_inputs(self, input_ids: torch.Tensor):
          - during profile_run
          - during DP rank dummy run
         """
-        dp_size = self.vllm_config.parallel_config.data_parallel_size
+        dp_size = self.vllm_config.parallel_config.data_parallel_size \
+                  and self.vllm_config.parallel_config.enable_expert_parallel
```
rebel-ykchoi: I think it needs to be fixed like below.
dp_ep_enable = self.vllm_config.parallel_config.data_parallel_size > 1 \
                  and self.vllm_config.parallel_config.enable_expert_parallel

dp_size > 1 --> dp_ep_enable

PR #384, opened 2026-02-18
feature: compile optimum model in vLLM if not present

rebel-eunji:
### 🚀 Summary of Changes

Previously, the system required a precompiled model and failed otherwise.
With this PR, the model is compiled on demand when not found.

FIXME
- prevent to use compiled models compiled by previous compiler, optimum-rbln.
- enable to compile pooling models with flash attention

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`bug-fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/model_base.py
+++ b/vllm_rbln/model_executor/models/optimum/model_base.py
@@ -186,9 +186,7 @@ class RBLNOptimumModelBase(nn.Module):
                 batch_size=self.scheduler_config.max_num_seqs,
                 block_size=get_attn_block_size(self.vllm_config),
                 max_model_len=self.model_config.max_model_len,
-                tp_size=self.vllm_config.additional_config.get(
-                    "tensor_parallel_size", 1
-                ),
+                tp_size=envs.VLLM_RBLN_TP_SIZE,
             )
             cached_model_path = os.path.join(
                 envs.VLLM_CACHE_ROOT,
@@ -205,9 +203,7 @@ class RBLNOptimumModelBase(nn.Module):
                     batch_size=self.scheduler_config.max_num_seqs,
                     block_size=get_attn_block_size(self.vllm_config),
                     max_model_len=self.model_config.max_model_len,
-                    tp_size=self.vllm_config.additional_config.get(
-                        "tensor_parallel_size", 1
-                    ),
+                    tp_size=envs.VLLM_RBLN_TP_SIZE,
                     model_path=str(cached_model_path),
                 )
             else:
--- a/vllm_rbln/model_executor/models/optimum/model_base.py
+++ b/vllm_rbln/model_executor/models/optimum/model_base.py
@@ -180,7 +180,9 @@ class RBLNOptimumModelBase(nn.Module):
                 batch_size=self.scheduler_config.max_num_seqs,
                 block_size=get_attn_block_size(self.vllm_config),
                 max_model_len=self.model_config.max_model_len,
-                tp_size=self.vllm_config.additional_config.get("tensor_parallel_size", 1),
+                tp_size=self.vllm_config.additional_config.get(
+                    "tensor_parallel_size", 1
+                ),
                 model_path=model_path,
             )
             self.vllm_config.model_config.model = model_path
--- a/vllm_rbln/model_executor/models/optimum/model_base.py
+++ b/vllm_rbln/model_executor/models/optimum/model_base.py
@@ -25,13 +25,22 @@ from vllm.model_executor.models.interfaces_base import VllmModelForTextGeneratio
 from vllm.v1.sample.metadata import SamplingMetadata
 
 import optimum.rbln
+import vllm_rbln.rbln_envs as envs
 from optimum.rbln.transformers.models.decoderonly import (
     decoderonly_runtime_utils as runtime_utils,
 )
 from vllm_rbln.utils.optimum.common import select_bucket_size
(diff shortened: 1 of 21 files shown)


vllm_rbln/utils/optimum/registry.py
```diff
@@ -124,3 +137,77 @@ def get_rbln_model_info(config: PretrainedConfig) -> tuple[str, str]:
...
+    max_model_len: int,
+    tp_size: int,
+    model_path: str,
+) -> Any:
+    architectures = getattr(config, "architectures", [])
+    model_name, model_cls_name = get_rbln_model_info(
+        config
+    )  # check if the model is supported and get model info
+    default_param: dict[str, Any] = {
+        "export": True,
```
rebel-kblee: need this?

vllm_rbln/utils/optimum/registry.py
```diff
@@ -124,3 +137,77 @@ def get_rbln_model_info(config: PretrainedConfig) -> tuple[str, str]:
...
+            hf_model_name,
+            **default_param,
+        )
+    elif is_pooling_arch(config):
+        model_cls_name = _RBLN_SUPPORTED_MODELS[architectures[0]][1]
+        model_cls = getattr(optimum.rbln, model_cls_name)
+        assert model_cls is not None
+        default_param["rbln_batch_size"] = batch_size
+        default_param["rbln_max_seq_len"] = max_model_len
+        if architectures[0] == "Qwen3Model" and block_size != max_model_len:
```
rebel-kblee, line 178: I don’t think we need to branch specifically for qwen3model—is there anything else we should consider?

vllm_rbln/utils/optimum/registry.py
rebel-eunji: Thanks. @rebel-kblee explained me that `export` is deprecated so I've removed it.

vllm_rbln/utils/optimum/registry.py
rebel-eunji: As far as I know, the Qwen3 pooling model includes a kvcache_partition_len field, whereas other models do not.

rebel-eunji: https://github.com/RBLN-SW/rbln-model-zoo/blob/40681c4831d3d30d9a93fd500dde4fd480bf35ae/huggingface/transformers/sentence-similarity/qwen3-embedding-0.6b/compile.py

vllm_rbln/platform.py
```diff
@@ -233,10 +233,6 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             scheduler_config.scheduler_cls = (
                 "vllm_rbln.v1.core.optimum_scheduler.RBLNOptimumScheduler"
             )
-
-            assert vllm_config.parallel_config.tensor_parallel_size == 1, (
```
rebel-jonghewk, line 241: @rebel-eunji  Could you share the reasoning behind removing this?

vllm_rbln/model_executor/models/optimum/model_base.py
```diff
@@ -145,39 +160,82 @@ def _resolve_kvcache_num_blocks(self) -> int:
...
+        # If it does not exist, compile and save it to the cache for future use.
+        if not is_compiled_model:
+            model_path_name = generate_model_path_name(
+                self.model_config.model,
+                batch_size=self.scheduler_config.max_num_seqs,
+                block_size=get_attn_block_size(self.vllm_config),
+                max_model_len=self.model_config.max_model_len,
+                tp_size=self.vllm_config.additional_config.get(
+                    "tensor_parallel_size", 1
+                ),
```
rebel-jonghewk: I think we shouldn't use tensor_parallel_size from vllm config. I think we should stick with VLLM_RBLN_TP_SIZE

vllm_rbln/utils/optimum/registry.py
rebel-kblee: Yes, that’s right. However, as I expect there will be more and more models like this, I was wondering if there is a specific way to distinguish between them.

vllm_rbln/platform.py
rebel-eunji: I've modified to use VLLM_RBLN_TP_SIZE.

vllm_rbln/utils/optimum/registry.py
rebel-eunji: Following our discussion, we have reached the following conclusions:

- Block size logic: We need a more generalized logic to specify block sizes as the number of supported models continues to grow.

- Reference Architecture: It is recommended to refer to vLLM's Qwen3Model, as its architecture is versatile enough to support both pooling and generation models.

- Current Status & Next Steps: Although the current if branch is suboptimal, it compiles successfully in eager mode. We plan to address and refine this in the upcoming PR."

rebel-jonghewk (approved):
LGTM. I think it's better to merge first and see if anything surfaces in CI. @rebel-eunji Do you know if our daily CI covers this usage?

rebel-eunji:
@rebel-jonghewk 
Our daily CI currently assumes that the optimal models have already been compiled.
However, I think it would be a good idea to skip the compilation step and run only the inference step in the daily CI. (dispatch)

PR #417, opened 2026-02-27
fix: consider kv cache 64 align to estimate kv cache memory size

rebel-wonsubkim:
Problem - in opt model test, kv cache oom occurred

Solution - when calculating kv cache available memory, SHOULD consider device kv cache 64 memory align
if head_size is not aligned to 64 (=128B), SHOULD adjust available memory for kv cache

### DIFF
--- a/vllm_rbln/v1/worker/rbln_worker.py
+++ b/vllm_rbln/v1/worker/rbln_worker.py
@@ -14,6 +14,7 @@
 """A RBLN worker class."""
 
 import copy
+import math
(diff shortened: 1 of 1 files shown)


rebel-jaehwang (approved):

rebel-jiwoopark (approved):
lgtm

rebel-jiwoopark:
@rebel-wonsubkim If there are no problems, we can go ahead and merge this.

Issue #398, opened 2026-02-20
[Speculative Decoding] Full support plan 


rebel-jinhwan:
### Features
- [x] https://github.com/RBLN-SW/vllm-rbln/pull/270
- [x] https://github.com/RBLN-SW/vllm-rbln/pull/341
- [x] https://github.com/RBLN-SW/vllm-rbln/pull/346
- [x] https://github.com/RBLN-SW/vllm-rbln/issues/324
- [x] https://github.com/RBLN-SW/vllm-rbln/issues/399
- [x] https://github.com/RBLN-SW/vllm-rbln/pull/440
- [x] https://github.com/RBLN-SW/vllm-rbln/pull/454

### Quality & Performance Assurance
- [ ] n-gram (link here)
- [ ] Medusa (link here)
- [ ] EAGLE (link here)
- [ ] Suffix Decoding (link here)
- [ ] Batched Attention
- [ ] DP + EP

PR #434, opened 2026-03-05
fix(scheduler): adjust max_model_len temporarily for speculative decoding

junstar92:
### 🚀 Summary of Changes

- Add `RBLNScheduler.__init__` to patch vllm_config when speculative decoding is enabled. Reduce max_model_len by num_speculative_tokens to avoid over-accessing KV cache.
- Use deepcopy to avoid mutating the original vllm_config.
- Reduce max_model_len when warming up the model.

---

* Resolves #
* Related to #398  

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `examples/experimental/run_specdec_ngram_test.py` after setting `max_tokens` to `max_model_len` and `ignore_eos` in sampling parameter.
2. Verify output: generates text sequences without error
3. Edge case tested: None

---

---

### 📋 Checklist

---

### 💬 Notes

To resolve runtime error at `max_model_len` boundary due to accessing multi-block of KV caches and benchmark initially, I simply reduced `max_model_len` in the scheduler. After initial benchmarks, we will revisit this issue.

---

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -1796,23 +1796,23 @@ class RBLNModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         )
         self._execute_dummy_requests(so, cso, self.prefill_intermediate_tensors)
 
+        # FIXME(RBLN): At the moment, a single request can’t access multiple
+        # blocks per layer under the current implementation, so we’re forced
+        # to reduce the warmup length to a reasonable value for now.
+        # This is mainly because we still have to run the computation over
+        # the padded tokens in speculative decoing scenario as well.
+        decode_max_seq_len = self.max_model_len // 2
+
         # compile decode graph considering decode batch buckets
         for batch_bucket_size in self.bucketing_manager.decode_batch_buckets:
-            decode_max_seq_len = self.max_model_len // 2
-
             dummy_decode_requests: list[NewRequestData] = []
             dummy_decode_num_scheduled_tokens: dict[str, int] = {}
-            num_speculative_tokens = (
-                self.speculative_config.num_speculative_tokens
-                if self.speculative_config is not None
-                else 0
-            )
             for _ in range(batch_bucket_size):
                 self._add_dummy_requests(
                     requests=dummy_decode_requests,
                     num_scheduled_tokens=dummy_decode_num_scheduled_tokens,
-                    total_tokens=decode_max_seq_len - 1 - num_speculative_tokens,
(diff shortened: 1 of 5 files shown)


huijjj (approved):
How about temporarily enabling decode graph warmup with a shorter sequence length, then unblocking the part we currently gate in platform.py for now with the max model len workaround introduced in this PR, and revisiting or addressing it properly later?

junstar92:
> How about temporarily enabling decode graph warmup with a shorter sequence length, then unblocking the part we currently gate in platform.py for now with the max model len workaround introduced in this PR, and revisiting or addressing it properly later?

@huijjj This PR change alone is not sufficient to make the warmup work, and errors still occur. It would be better to address this while updating the runner and warmup together.

huijjj:
As warmup issue roots from the same problem addressed from this PR(long seq len) warmup will can be enabled simply reducing the seq len of the dummy inputs we use during warmup to a fairly small number such as (max_model_
len // 2).

junstar92:
@huijjj I see. I also reduced max_model_len when warming up the model.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1799,7 +1799,7 @@ def warm_up_model(self) -> None:
 
         # compile decode graph considering decode batch buckets
         for batch_bucket_size in self.bucketing_manager.decode_batch_buckets:
-            decode_max_seq_len = self.max_model_len
+            decode_max_seq_len = self.max_model_len // 2
```
rebel-jinhwan: It would be good to add comments that describe why this workaround is added.

junstar92: Max model length is not necessary to compile the decode graph. This is not meant to be a permanent workaround, and the warp-up logic will be refactored as part of the model runner refactor.

junstar92: Just to add a bit more context, this isn't really a workaround. Since any value works as long as it compiles, I'm not sure if adding a comment here would be necessary.

huijjj: Since this is just a rough, temporary fix, it felt a bit awkward to leave a permanent public record of it, but we’ve added a brief comment for now as you requested.

rebel-jinhwan (approved):
LGTM 👍

junstar92:
@rebel-jiwoopark @rebel-jinhwan 
I close this PR and handle it in #440 with related fixes.

PR #437, opened 2026-03-06
fix: guide what to do on oom

rebel-jaehwang:
Available memory estimation for kv cache can be slightly off, leading to OOM. Guide users to reduce the number of blocks.

Example:
```
RuntimeError: Not enough memory for 800 blocks of KV cache. Try reducing the number of blocks by setting --num-gpu-blocks-override.
```

Note: ~~This requires defaulting VLLM_RBLN_COMPILE_STRICT_MODE to true~~ This requires `"mode": "strict"`, so that we can catch compile failure.

### 🧪 How to Test

Run
```
VLLM_RBLN_USE_VLLM_MODEL=1 python examples/experimental/offline_inference_basic.py
```
with the following patch
```diff
diff --git a/examples/experimental/offline_inference_basic.py b/examples/experimental/offline_inference_basic.py
index 36f7c34..3c6e959 100644
--- a/examples/experimental/offline_inference_basic.py
+++ b/examples/experimental/offline_inference_basic.py
@@ -48,7 +48,8 @@ def main():
         block_size=args.block_size,
         enable_chunked_prefill=True,
         max_num_batched_tokens=128,
-        gpu_memory_utilization=0.9,
+        gpu_memory_utilization=1, # 428 blocks → SYS_EBUSY
+        # num_gpu_blocks_override=800, # → SYS_ENOMEM
         enable_expert_parallel=args.enable_expert_parallel,
     )
```

---

### DIFF
--- a/tests/torch_compile/common/test_rbln_envs.py
+++ b/tests/torch_compile/common/test_rbln_envs.py
@@ -22,8 +22,8 @@ def test_rbln_envs():
         got {rbln_envs.VLLM_RBLN_COMPILE_MODEL}"
     )
 
-    assert not rbln_envs.VLLM_RBLN_COMPILE_STRICT_MODE, (
-        f"Expected VLLM_RBLN_COMPILE_STRICT_MODE to be False, \
+    assert rbln_envs.VLLM_RBLN_COMPILE_STRICT_MODE, (
+        f"Expected VLLM_RBLN_COMPILE_STRICT_MODE to be True, \
         got {rbln_envs.VLLM_RBLN_COMPILE_STRICT_MODE}"
     )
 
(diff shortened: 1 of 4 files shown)


rebel-eunji:
@rebel-jonghewk 
fyi. This PR affects the sampler in optimum. The sampler in optimum is currently compiled without strict mode as `VLLM_RBLN_COMPILE_STRICT_MODE` is 0 by default.

rebel-jonghewk (approved):
lgtm

rebel-jaehwang:
Interesting. Sampler compilation does fail due to oom in some optimum tests. Is this expected? @rebel-eunji @rebel-jonghewk  https://github.com/RBLN-SW/vllm-rbln/actions/runs/22753987341/job/65996243177?pr=437#step:11:6960

rebel-eunji:
That’s exactly what I was worried about. 🥲 The sampler falls back to the CPU if compilation fails. I spoke with @rebel-jaehwang in person, and we agreed to strictly compile the model, but not the sampler.

PR #442, opened 2026-03-09
feature: sub-block prefix caching

rebel-jaehwang:
## 🚀 Summary of Changes
see the rendered design doc: https://github.com/RBLN-SW/vllm-rbln/blob/ab677e4d49ad9f09b41146cb131922728746c336/docs/sub_block_prefix_caching.md
 
**Configuration**: Default on. Disable with env var `VLLM_RBLN_SUB_BLOCK_CACHE=0`.

---

### unit tests
```
pytest tests/torch_compile/v1/core/test_sub_block_prefix_caching.py
```

### e2e example

without sub-block caching (block_size=1024)
```
VLLM_RBLN_SUB_BLOCK_CACHE=0 VLLM_RBLN_SAMPLER=0 VLLM_RBLN_USE_VLLM_MODEL=1 python examples/experimental/prefix_caching.py
```
```
...
Prefix cache hit tokens during cached generate(): 4096
```

with sub-block caching (block_size=1024, sub_block_size=128):
```
VLLM_RBLN_SAMPLER=0 VLLM_RBLN_USE_VLLM_MODEL=1 python examples/experimental/prefix_caching.py
```
```
...
Prefix cache hit tokens during cached generate(): 6144
```

### DIFF
--- a/examples/experimental/prefix_caching.py
+++ b/examples/experimental/prefix_caching.py
@@ -204,11 +204,12 @@ def main():
         model="Qwen/Qwen3-0.6B",
         max_num_seqs=2,
         max_model_len=2048,
-        block_size=256,
         enable_chunked_prefill=True,
         max_num_batched_tokens=128,
         num_gpu_blocks_override=32,
     )
+    # Can't directly pass this value due to validation
+    args.block_size = 1024  # type: ignore
 
     # Create an LLM without prefix caching as a baseline.
     args.enable_prefix_caching = False
--- a/vllm_rbln/rbln_envs.py
+++ b/vllm_rbln/rbln_envs.py
@@ -46,6 +46,7 @@ if TYPE_CHECKING:
     VLLM_RBLN_DECODE_BATCH_BUCKET_MANUAL_BUCKETS: list[int] = []
     VLLM_RBLN_USE_CUSTOM_KERNEL: bool = False
     VLLM_RBLN_AUTO_PORT: bool = True
+    VLLM_RBLN_SUB_BLOCK_CACHE: bool = True
 
 
 def get_dp_impl() -> str:
@@ -256,6 +257,11 @@ environment_variables = {
     ),
     "VLLM_RBLN_PROFILER": (
         lambda: os.environ.get("RBLN_PROFILER", "False").lower() in ("true", "1")
+    ),
+    # Enable sub-block prefix caching.
+    # Sub-block size equals max_num_batched_tokens (prefill chunk size).
+    "VLLM_RBLN_SUB_BLOCK_CACHE": lambda: (
+        os.environ.get("VLLM_RBLN_SUB_BLOCK_CACHE", "True").lower() in ("true", "1")
     ),
 }
 
--- a/vllm_rbln/v1/core/rbln_kv_cache_manager.py
+++ b/vllm_rbln/v1/core/rbln_kv_cache_manager.py
@@ -1,0 +1,622 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Sub-block prefix caching for RBLN KV cache management.
+
+See docs/sub_block_prefix_caching.md for design overview and rationale.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from vllm.v1.core.kv_cache_manager import KVCacheBlocks, KVCacheManager
+from vllm.v1.core.kv_cache_utils import (
+    BlockHash,
+    hash_block_tokens,
+    make_block_hash_with_group_id,
+)
+from vllm.v1.kv_cache_interface import (
+    ChunkedLocalAttentionSpec,
+    FullAttentionSpec,
+    KVCacheSpec,
+    SlidingWindowSpec,
+)
+
+from vllm_rbln.logger import init_logger
+
+if TYPE_CHECKING:
+    from vllm.v1.core.kv_cache_utils import KVCacheBlock
+    from vllm.v1.kv_cache_interface import KVCacheConfig
+    from vllm.v1.request import Request
+
+# Attention spec types whose KV cache stores per-token data and can be
+# partially copied at sub-block granularity.  Types NOT in this set (e.g.
+# MambaSpec — stores recurrent state per block, not per-token KV) are
+# incompatible with sub-block caching.
+_SUB_BLOCK_ELIGIBLE_SPECS: tuple[type[KVCacheSpec], ...] = (
+    FullAttentionSpec,  # includes MLAAttentionSpec (subclass)
+    SlidingWindowSpec,
+    ChunkedLocalAttentionSpec,
+)
+
+logger = init_logger(__name__)
+
+
+@dataclass
+class KVCacheCopyOp:
+    """Describes a sub-block KV cache copy from a cached block to a new block."""
+
+    group_id: int
+    # NOTE: While pending, it holds a ref count of the source block to prevent eviction.
+    src_block_id: int
+    dst_block_id: int
+    # Number of tokens to copy (= num_matched_sub_blocks * sub_block_size).
+    num_tokens: int
+
+
+class SubBlockHasher:
+    """Computes chained sub-block hashes from token IDs.
+
+    Uses the same ``hash_block_tokens`` as upstream, but at sub-block
+    granularity.
+    """
+
+    def __init__(
+        self,
+        hash_fn: Callable,
+        sub_block_size: int,
+    ) -> None:
+        self.hash_fn = hash_fn
+        self.sub_block_size = sub_block_size
+
+    def hash_tokens(
+        self,
+        token_ids: Sequence[int],
+        *,
+        parent_hash: BlockHash | None = None,
+        num_hashed_tokens: int = 0,
+    ) -> list[BlockHash]:
+        """Return sub-block hashes for *full* sub-blocks in ``token_ids``.
+
+        Args:
+            token_ids: Full token sequence of the request.
+            parent_hash: Hash of the last sub-block before the range we
+                are hashing (``None`` for the very first sub-block).
+            num_hashed_tokens: Number of tokens already hashed (i.e. the
+                start offset into ``token_ids``).
+
+        Returns:
+            List of ``BlockHash`` values, one per full sub-block starting
+            from ``num_hashed_tokens``.
+        """
+        sbs = self.sub_block_size
+        hashes: list[BlockHash] = []
+        start = num_hashed_tokens
+        for i in range(start, len(token_ids) - sbs + 1, sbs):
+            parent_hash = hash_block_tokens(
+                self.hash_fn,
+                parent_hash,
+                token_ids[i : i + sbs],
+            )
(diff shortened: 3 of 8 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1357,6 +1363,7 @@ def _compile_model(self, model):
             "process_group_dict": process_group_dict,
             "guard_filter_fn": torch.compiler.keep_tensor_guards_unsafe,
             "mode": "strict",
+            "_runtime_holder": self.runtime_holder,
```
huijjj, line 1454: Could you elaborate on the runtime_holder in the compile options, and also explain how _copy_kv_cache works in the runtime it holds(probably d2d copy)? Also is this feature only available in the nightly build of the new rebel-compiler?

huijjj: Also, is this feature to be release before the version bump up? If not please consider changing the base branch to dev-0.18.

rebel-jaehwang: Yes, it's for d2d copy. Ideally we should use torch-rbln for that, but it's not mature enough yet. `_copy_kv_cache` is the specialized runtime method for bulk d2d copy with `mark_static_address`ed tensors. Using this requires access to the `rebel.Runtime` object, which is not exposed in normal way for `torch.compile`-based workflow.  `_runtime_holder` is basically a backdoor to the runtime objects created by rbln backend under the hood. At this moment, this feature is only available in the nightly build, and it will be included in the next release. 

> Also, is this feature to be release before the version bump up? If not please consider changing the base branch to dev-0.18.

👍

huijjj: thanks for the detailed explanation

huijjj:
There are a lot of changes here, so I’m still digesting the full flow.

In a previous scheduler change, I had allocate_slot happen speculatively even when the request would not actually be scheduled in the next iteration. That caused block hashes to be written for blocks that had not been computed yet, and I handled it with undo_uncomputed_block_caching in the scheduler.

I see that _maybe_evict_cached_block is monkey patched here so that corresponding sub-blocks can also be cleared when a physical block is evicted. But what happens in the case where the request length is such that the physical block is not yet full, so it never gets a hash and _maybe_evict_cached_block is never triggered, while sub-block cache entries may already exist for not-yet-computed tokens? Is that scenario accounted for?

vllm_rbln/v1/core/rbln_kv_cache_manager.py
```diff
@@ -0,0 +1,658 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-eunji, line 1: Could you split the code?
I think we should keep the original functions of kv_cache_manager in this file and move the custom functions and classes into separate files.

vllm_rbln/rbln_envs.py
```diff
@@ -257,6 +258,11 @@ def get_decode_batch_bucket_manual_buckets() -> list[int]:
     "VLLM_RBLN_PROFILER": (
         lambda: os.environ.get("RBLN_PROFILER", "False").lower() in ("true", "1")
     ),
+    # Enable sub-block prefix caching.
+    # Sub-block size equals max_num_batched_tokens (prefill chunk size).
+    "VLLM_RBLN_SUB_BLOCK_CACHE": lambda: (
+        os.environ.get("VLLM_RBLN_SUB_BLOCK_CACHE", "true").lower() in ("true", "1")
```
rebel-eunji: To align with other env vars,
```suggestion
        os.environ.get("VLLM_RBLN_SUB_BLOCK_CACHE", "True").lower() in ("true", "1")
```

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -433,9 +498,23 @@ def schedule(self) -> SchedulerOutput:
                         request.num_external_computed_tokens = ext_tokens
                         num_external_computed_tokens = ext_tokens
 
+                    # NOTE(RBLN): Arbitrate between sub-block match and KV connector.
+                    sub_block_match, num_sub_block_tokens = self._try_sub_block_match(
+                        request,
+                        num_new_local_computed_tokens,
+                        num_external_computed_tokens,
+                    )
+                    if num_sub_block_tokens > 0 and num_external_computed_tokens > 0:
```
rebel-jinhwan, line 516: According to the design document, it seems like the larger value between the number of sub block tokens and external computed tokens is used as the matched tokens, but here it looks like sub block tokens are always being chosen. Is my understanding correct?

rebel-jaehwang:
> I see that _maybe_evict_cached_block is monkey patched here so that corresponding sub-blocks can also be cleared when a physical block is evicted. But what happens in the case where the request length is such that the physical block is not yet full, so it never gets a hash and _maybe_evict_cached_block is never triggered, while sub-block cache entries may already exist for not-yet-computed tokens? Is that scenario accounted for?

@huijjj Good question. IIRC, the scenario you're describing is: speculative `allocate_slots` creating sub-block cache entries for a not-yet-full block. This does not happen.

There are two places where sub-block cache entries are created, both of which guard against indexing sub-blocks with not-yet-computed tokens:
* `allocate_slots` only attempts to index full blocks (`_index_newly_cached_blocks`). The speculatively indexed sub-blocks from that full blocks are cleared later by `_maybe_evict_cached_block` in `undo_uncomputed_block_caching`.
* partially filled blocks are only indexed when a request is being `free`ed (`_index_partial_block`), at which point `num_computed_tokens` is known.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-jaehwang: comparing sub-block match vs. external match happens in `_try_sub_block_match`. This if branch is for the case where sub-block match won.

examples/experimental/prefix_caching.py
rebel-eunji, line 297: What about setting `envs.VLLM_RBLN_SAMPLER=0` in this example?

vllm_rbln/v1/core/rbln_kv_cache_manager.py
```diff
@@ -0,0 +1,658 @@
...
+    num_matched_sub_blocks: int
+    group_matches: tuple[_GroupPartialMatch, ...]
+
+
+@dataclass(slots=True)
+class _GroupPartialMatch:
+    """Per-group partial match info."""
+
+    src_block_id: int
+    src_block: KVCacheBlock
```
rebel-eunji: nit: It holds both `src_block_id` and `src_block`, but `src_block.block_id` already gives you the ID, so storing both is redundant. Keeping only src_block looks better since src_block is passed to touch() and free_blocks().

vllm_rbln/v1/core/rbln_kv_cache_manager.py
```diff
@@ -0,0 +1,658 @@
...
+    # Index of the block (in the request's block list) that will receive
+    # the copied sub-blocks.
+    dst_block_index: int
```
rebel-eunji: nit: `dst_block_index` is a little bit misleading. It's the position within the **request's** allocated block list. How about renaming it to something like  `dst_req_block_idx`?

vllm_rbln/v1/core/rbln_kv_cache_manager.py
```diff
@@ -0,0 +1,658 @@
...
+        # We enforce the same condition here.
+        max_total = request.num_tokens - 1
+        if num_computed_tokens + extra_tokens > max_total:
+            capped = max_total - num_computed_tokens
+            if capped <= 0:
+                return None
+            num_matched = capped // self.sub_block_size
+            if num_matched == 0:
+                return None
+            extra_tokens = num_matched * self.sub_block_size
```
rebel-eunji: refactoring suggestion

```suggestion
        # NOTE: Upstream guarantees num_computed_tokens <= num_tokens - 1,
        # because at least the last token must be recomputed for logits.
        # Cap the match so we don't exceed that budget.
        token_budget = (request.num_tokens - 1) - num_computed_tokens
        max_sub_blocks_allowed = max(0, token_budget // self.sub_block_size)
        num_matched = min(min_matched, max_sub_blocks_allowed)

        if num_matched == 0:
            return None

        extra_tokens = num_matched * self.sub_block_size
```

huijjj:
> > I see that _maybe_evict_cached_block is monkey patched here so that corresponding sub-blocks can also be cleared when a physical block is evicted. But what happens in the case where the request length is such that the physical block is not yet full, so it never gets a hash and _maybe_evict_cached_block is never triggered, while sub-block cache entries may already exist for not-yet-computed tokens? Is that scenario accounted for?
> 
> @huijjj Good question. IIRC, the scenario you're describing is: speculative `allocate_slots` creating sub-block cache entries for a not-yet-full block. This does not happen.
> 
> There are two places where sub-block cache entries are created, both of which guard against indexing sub-blocks with not-yet-computed tokens:
> 
> * `allocate_slots` only attempts to index full blocks (`_index_newly_cached_blocks`). The speculatively indexed sub-blocks from that full blocks are cleared later by `_maybe_evict_cached_block` in `undo_uncomputed_block_caching`.
> * partially filled blocks are only indexed when a request is being `free`ed (`_index_partial_block`), at which point `num_computed_tokens` is known.

I see, the design here is quite elegant, great job.

My last concern is about sub-block matching for blocks that are not free. 

Especially in our setup, where prefill batch size is capped to 1 and decode/prefill mixed batching does not happen, it seems worth considering sub-block matching for not yet free-ed blocks as well if the goal is to maximize prefix cache hit rate. It might make the implementation quite messy but have you considered that approach?

rebel-jaehwang:
> Especially in our setup, where prefill batch size is capped to 1 and decode/prefill mixed batching does not happen, it seems worth considering sub-block matching for not yet free-ed blocks as well if the goal is to maximize prefix cache hit rate. It might make the implementation quite messy but have you considered that approach?

My understanding is that eagerly indexing (not deferring to `free()`) the last partial block's sub-blocks is beneficial only when a new prefill is scheduled while there is an earlier running request with shared sub-block prefix that is not already indexed.
The earlier running request can be in either
* paused decoding stage (which happens in rbln scheduler), or
* actively running prefill or decode stage (which happens in the upstream scheduler).

Such a situation would arise when multiple requests with a new prefix arrive in burst. I think such cases are quite rare for long-running servers, which is why I initially avoided implementing eager indexing of the last partial block.

But eager indexing doesn't seem particularly difficult. I'm looking into it.

---

Aside:
While thinking about this, I noticed that the current implementation is not compatible with multi-prefill batch.
Suppose there are two requests
* R1: `[a, b], [c, d]`
* R2: `[a, b], [c, e]`

where each letter means a sub-block and `[...]` is a full block.

Suppose both requests are scheduled for prefill in the same step
* R1 gets blocks B1 (`[a, b]`) and B2 (`[c, d]`).
* R2 get full block hit for B1 and gets a new block B3 (`[c, e]`).
  It gets sub-block hit for `c`, scheduling copy from B2 to B3.
* Before executing the model, the copy is executed.
  **Since kv is not written on B2 yet, this copies uninit data. This is wrong**.
* During attn, if the kernel writes kv for all requests in batch before proceeding to attn, R2 reusing B1 is correct.

Maybe I should fix this first. Maybe sub-block indexing should be deferred to `RBLNScheduler.update_from_output`.

rebel-jaehwang:
> speculative allocate_slots creating sub-block cache entries for a not-yet-full block. This does not happen.

Added test for this. See:
* "test(sub-block): interaction with undo_uncomputed_block_caching"

> I noticed that the current implementation is not compatible with multi-prefill batch.

Fixed. See:
* fix(sub-block): compatibility with multi-prefill batch

> > it seems worth considering sub-block matching for not yet free-ed blocks as well if the goal is to maximize prefix cache hit rate.
>
> [...] But eager indexing doesn't seem particularly difficult. I'm looking into it.

Implemented. See:
* feature(sub-block): eagerly index partially filled block


---

will rebase on dev-0.18 soon

rebel-jaehwang:
Rebased on dev-0.18 and changed base branch. Also squashed test-only commits into each feature/fix commits

The last 0.13 version is at b20537a217c7f8b476d85e95c1053ea32514baf8

rebel-jinhwan (approved):
LGTM

PR #454, opened 2026-03-12
feature(specdec): support Medusa on RBLN backend

junstar92:
### 🚀 Summary of Changes

- Introducing `RBLNMedusaProposer` subclassing vllm's `MedusaProposer` to support RBLN backend compilation
  - Compile `_propose` method with `CompileContext` for weight sharing
  - Replace vllm's `MedusaProposer` with `RBLNMedusaProposer` as the Medusa drafter
- Simplify hidden state indexing in the decode phase: replace offset-based calculating using `spec_decode_metadata` with direct indexing based on `sampled_token_ids` lengths
- Enforce `draft_model_config.dtype = torch.float` when `speculative_config` is present and `VLLM_RBLN_ENFORCE_MODEL_FP32` is set, following the same pattern as lora dtype enforcement

---

* Resolves #
* Related to #398 

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `examples/experimental/run_specdec_medusa_test.py`.
2. Verify output: generates text sequences with acceptance tokens and no errors.
3. Edge case tested: None

---

---

### 📋 Checklist

---

### 💬 Notes

- `_propose` combines both the Medusa model inference and `compute_logits` into a single compiled function. **Compiling them separately was attempted but resulted in logits values that differed from the uncompiled baseline for unknown reasons**.

---

### DIFF
--- a/vllm_rbln/v1/spec_decoding/medusa.py
+++ b/vllm_rbln/v1/spec_decoding/medusa.py
@@ -25,7 +25,7 @@ from vllm.v1.spec_decode.medusa import MedusaProposer
 import vllm_rbln.rbln_envs as envs
 
 
-class RblnMedusaProposer(MedusaProposer):
+class RBLNMedusaProposer(MedusaProposer):
     def __init__(self, vllm_config: VllmConfig, runner) -> None:
         super().__init__(vllm_config, runner.device)
 
--- a/vllm_rbln/v1/spec_decoding/medusa.py
+++ b/vllm_rbln/v1/spec_decoding/medusa.py
@@ -35,7 +35,19 @@ class RBLNMedusaProposer(MedusaProposer):
 
     def load_model(self, target_model: nn.Module) -> None:
         super().load_model(target_model)
-        self._propose = self._compile_model(self._propose)  # type: ignore
+
+        def model_wrapper(target_hidden_states: torch.Tensor) -> torch.Tensor:
+            hidden_states = self.model(target_hidden_states)
+            logits = self.model.compute_logits(hidden_states)
+            return logits
+
+        if (
+            self.vllm_config.speculative_config.enforce_eager
+            or not envs.VLLM_RBLN_COMPILE_MODEL
+        ):
+            self.model_executable = model_wrapper
+        else:
+            self.model_executable = self._compile_model(model_wrapper)
 
     def _compile_model(self, model: nn.Module):
         TP = get_tp_group()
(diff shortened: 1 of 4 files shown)


vllm_rbln/v1/spec_decoding/medusa.py
```diff
@@ -0,0 +1,96 @@
...
+from vllm.forward_context import set_forward_context
+from vllm.v1.sample.metadata import SamplingMetadata
+from vllm.v1.spec_decode.medusa import MedusaProposer
+
+import vllm_rbln.rbln_envs as envs
+
+
+class RBLNMedusaProposer(MedusaProposer):
+    def __init__(self, vllm_config: VllmConfig, runner) -> None:
+        super().__init__(vllm_config, runner.device)
```
rebel-jinhwan: should we pass runner instead of runner.device?

junstar92: No, we don't need to pass `runner`. That part is incorrect, and I'll revise it so that it accepts only `device`.

vllm_rbln/v1/spec_decoding/medusa.py
```diff
@@ -0,0 +1,96 @@
...
+        # Compute argmax for each Medusa head and stack into a single tensor
+        # Shape: [batch_size, num_heads]
+        draft_tokens = torch.stack([logit.argmax(dim=-1) for logit in logits], dim=1)
+
+        return draft_tokens
+
+    @torch.inference_mode()
+    def dummy_run(self, batch_size: int) -> None:
+        hidden_states = torch.zeros(
+            (batch_size, self.hidden_size),
```
rebel-jinhwan, line 99: i'm not entirely sure about the details here, but since this part differs from upstream, dont we need to use the decode phase max batch size here?

junstar92: `dummy_run` is not currently used. The reason I added it anyway is for warmup. At the moment, warmup is tied to the scheduler and works by executing the runner executable, so in practice, the draft model warmup compilation also happens when `propose` is actually called.
This approach does not seem very robust, so I thought it would be better to perform the warmup compilation through `dummy_run`, similar to upstream. That is why I added this path. When calling `dummy_run`, you can just pass the decode-phase max batch size. It is possible that you may eventually need compilation for every batch for performance reasons.

If you would prefer not to keep this, I can just remove it.

junstar92:
I added an additional commit to support eager mode.

vllm_rbln/v1/spec_decoding/medusa.py
rebel-jinhwan: iiuc, current warming up may include duplicated base model forward (or, tied with scheduler), so you're explaining why dummy_run is needed in this code. I think warming up through dummy run seems like a significant change. how about create a issue for this so we can review the proposal together beforehand?

rebel-jinhwan (approved):
LGTM

vllm_rbln/v1/spec_decoding/medusa.py
junstar92: In my opinion, even if this is handled in a separate issue, it will likely need to be modified again once the overall code refactoring (especially the runner refactoring) takes place. I expect the warmup-related changes to be addressed naturally during the runner refactoring process.

PR #459, opened 2026-03-16
feature(specdec): support Eagle on RBLN backend

junstar92:
### 🚀 Summary of Changes

This PR updates the RBLN speculative decoding path for EAGLE/EAGLE3 on top of the `dev-0.18` branch.

- Added `RBLNEagleProposer` and aligned the RBLN padded EAGLE flow with the upstream `vLLM v0.18.0` interface while preserving the RBLN-specific input shape and attention metadata handling.
- Extracted the torch-native padded EAGLE helpers into `vllm_rbln/v1/spec_decoding/utils.py`:
  - `eagle_prepare_next_token_padded`
  - `eagle_prepare_inputs_padded`
- Updated the proposer/runner contract to propagate `num_rejected_tokens_gpu` and use the padded EAGLE input preparation path consistently.
- Added an RBLN-specific `set_inputs_first_pass()` path instead of relying on the upstream implementation, because the RBLN input layout and attention path are different from upstream.
- Updated the model runner to use `RBLNEagleProposer`, initialize drafter attention backends, and handle EAGLE/EAGLE3-specific hidden-state/logits plumbing.
- Added Eagle3 auxiliary hidden-state handling from config when available.
- Consolidated the experimental scripts by removing the separate Eagle3 example and extending `examples/experimental/run_specdec_eagle_test.py` to support both:
  - `--method eagle`
  - `--method eagle3`
- Added unit tests for:
  - padded Eagle helper functions
  - `RBLNEagleProposer` functional behavior
  - Eagle3-specific proposer behavior

---

* Resolves #
* Related to #398 

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run the unit tests: `pytest tests/torch_compile/unit/v1/specdec/test_utils.py tests/torch_compile/unit/v1/specdec/test_eagle.py`
2. Run the experimental EAGLE example: `python examples/experimental/run_specdec_eagle_test.py --method eagle`
3. Run the experimental EAGLE3 example: `python examples/experimental/run_specdec_eagle_test.py --method eagle3`

Verify output:
- Unit tests should pass.
- The EAGLE example should run and print generated outputs plus speculative decoding metrics.
- The EAGLE3 example is currently subject to the known compiler issue described below.

---

---

### 📋 Checklist

---

### 💬 Notes

- The upstream `set_inputs_first_pass()` implementation could not be used directly because `vllm-rbln` has different input-shape and attention-metadata assumptions from upstream.
- The padded EAGLE helper logic was moved into `utils.py` so it can be tested independently from the proposer implementation.
- The experimental Eagle and Eagle3 scripts were unified into a single entry point with a `--method` argument.
- Eagle3 compilation is still blocked by a compiler-side issue even with compiler `v0.10.2`, so additional investigation from the Rebellions side is likely still required.

---


junstar92:
@rebel-jinhwan @rebel-jiwoopark
Not right now, but if we share the CompileContext later, I'm not sure if the KV cache will be shared between the base and draft models. If we share the CompileContext, should the KV cache be shared as well?

rebel-sjlee:
https://github.com/rebellions-sw/fsw-inference/issues/127

rebel-wonsubkim:
@junstar92 if we share same compile context between target model and eagle model, kv caches will be shared as well since both models have same static address tensor registered in same CompileContext.

junstar92:
@rebel-wonsubkim Thanks for your answer. It seems like there are no issues after adding `mark_static_address`, even without sharing a compile context. So, I think it is okay not to share the compile context.

junstar92:
@rebel-jinhwan @rebel-jiwoopark I updated this PR for v0.18.1 base.

I also considered adding E2E coverage for EAGLE/EAGLE3 based on the current experimental script. However, the models currently used there are not native vLLM-supported EAGLE checkpoints and require extra conversion. Because of that, they do not seem ideal for stable automated E2E coverage. Would using the 8B model from the official vLLM EAGLE example be acceptable for the E2E tests?

rebel-jinhwan:
@junstar92 yes, I think using vllm official example is better

junstar92:
@rebel-jinhwan Testing with the 8B model requires parallelism, but EAGLE currently hits a compiler error when used with TP. Should this case also be marked as xfail for now?

rebel-jinhwan:
EAGLE3 is verified at `rebel-compiler==0.10.3.dev56+gad6ab56a`

junstar92:
@rebel-jinhwan I just added the end-to-end test script for eagle/eagle3.

junstar92:
@rebel-jinhwan I also added correctness tests for eagle and eagle3. The eagle test is also failing, and I couldn't run the eagle3 test since it doesn't compile in my environment.

rebel-jinhwan:
@junstar92 I don't know why, but it works when I add `from examples.experimental.run_specdec_eagle_test import _map_external_eagle_state_dict` to `utils.py`. (I restricted `max_tokens` to 32 because it seems there is a bit of precision problem)

fyi. https://github.com/RBLN-SW/vllm-rbln/blob/jinhwan/support-specdec-eagle/tests/torch_compile/e2e/v1/spec_decode/test_eagle_correctness.py

junstar92:
@rebel-jinhwan So, that means you want me to set `max_tokens` to 32?

rebel-jinhwan:
@junstar92 It is enough to test upto 32 tokens at the moment, but i cant find the reason why i have to add this line https://github.com/RBLN-SW/vllm-rbln/blob/60e100e4b6b837d369d18aec41ceba2e23f6fd84/tests/torch_compile/e2e/v1/spec_decode/utils.py#L90

two functions seem exactly same :(

junstar92:
Well, I can't reproduce that. Also, in my case, Ealge keeps failing the test for me when `max_token` is set to 32. Is this just my compiler issue?

rebel-jinhwan:
@junstar92 would you try to remove compiler caches??

junstar92:
I always set `VLLM_DISABLE_COMPILE_CACHE` to '1', but do I need to remove caches?

junstar92:
I removed caches, but the test is still failing.

rebel-jinhwan:
Tests are all passed at compiler version: 0.10.3.dev136+g52719190

FYI. this patch fixes torch.compile bugs that choose torch.jit.trace instead of torch.export

rebel-jinhwan (approved):
LGTM

PR #473, opened 2026-03-23
fix: don't halve # of cpus twice

rebel-jaehwang:
Problem: set_cpu_affinity already halves the number of cpus for machine with hyperthreading. So halving again results in using only half of physical cores.

Solution: Don't halve again in init_device.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_worker.py
+++ b/vllm_rbln/v1/worker/rbln_worker.py
@@ -198,9 +198,8 @@ class RBLNWorker(WorkerBase):
             self.parallel_config,
         )
 
-        # Use half of allocated CPUs to avoid oversubscription
         allocated_cpus = len(os.sched_getaffinity(0))
-        num_threads = max(2, allocated_cpus // 2)
+        num_threads = max(2, allocated_cpus)
         set_omp_num_threads(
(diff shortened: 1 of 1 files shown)


vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -198,9 +198,8 @@ def init_device(self) -> None:
             self.parallel_config,
         )
 
-        # Use half of allocated CPUs to avoid oversubscription
         allocated_cpus = len(os.sched_getaffinity(0))
-        num_threads = max(2, allocated_cpus // 2)
+        num_threads = max(2, allocated_cpus)
```
rebel-yskim, line 202: Halving allocated_cpus, which was already halved due to disabling HT, was actually expected behavior: it was an optimal point of RBLN NPU's perf (conservative number of threads).

It would be good to verify the performance first. If this PR improves performance (if you already checked), LGTM.

rebel-jaehwang: Oh, I thought this was mistake. In retrospect, it does seem to make sense to use only half the physical cpus so that the other half can be used for other python processes.

PR #476, opened 2026-03-24
fix(NUMA): Cpu split after warmup

rebel-yskim:
### 🚀 Summary of Changes

This PR moves CPU affinity setting & # of thread setting after warm_up.
This enables rbln compiler to use full CPU resources, including HT.

Jira Issue: SR-56

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/v1/worker/rbln_worker.py
+++ b/vllm_rbln/v1/worker/rbln_worker.py
@@ -58,6 +58,7 @@ from vllm_rbln.logger import init_logger
 from vllm_rbln.v1.worker.rbln_model_runner import RBLNModelRunner
 from vllm_rbln.v1.worker.utils import (
     estimate_available_memory,
+    get_rbln_planned_affinity_cpu_count,
     set_cpu_affinity,
     set_omp_num_threads,
 )
@@ -102,7 +103,8 @@ class RBLNWorker(WorkerBase):
 
         # Buffers saved before sleep
         self._sleep_saved_buffers: dict[str, torch.Tensor] = {}
-        self._rbln_cpu_threading_configured = False
+        self._rbln_host_threads_before_compile_ready = False
+        self._rbln_cpu_affinity_applied = False
 
         profiler_config = vllm_config.profiler_config
         # Set up profiler if profiling is enabled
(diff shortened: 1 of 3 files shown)


vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -331,6 +304,46 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+        # potentially change torch num thread to other value
+        numba.set_num_threads(torch.get_num_threads())
+
+        # Code below restores torch num thread to its original value
+        # before numba.set_num_threads
+        torch.set_num_threads(numba.get_num_threads())
+
+        self._rbln_cpu_threading_configured = True
+
     def compile_or_warm_up_model(self) -> None:
```
rebel-jonghewk, line 363: can you add a logging message that cpu_threading is not configured for `"skipping compile_or_warm_up_model"` cases?

rebel-jaehwang:
iirc changing num threads can trigger torch.compile recompilation. Can you confirm that this doesn't happen? If recompilation does not happen, this might be due to deleting guards (`guard_filter_fn` option for torch.compile`). Is this actually safe?

rebel-yskim:
> iirc changing num threads can trigger torch.compile recompilation. Can you confirm that this doesn't happen? If recompilation does not happen, this might be due to deleting guards (`guard_filter_fn` option for torch.compile`). Is this actually safe?

Agreed—num_threads changes can cause Dynamo recompiles when guarded; we hit that before. The rebel_compiler change limits the bump to inside the RBLN backend and restores using RBLN_NUM_THREADS or entry threads, while persisting the policy count for runtime so guards stay aligned. Affinity isn’t in that global state, so we pin CPUs after dummy run; we’re not relying on guard_filter_fn to drop num_threads guards for safety.

vllm_rbln/v1/worker/rbln_worker.py
rebel-yskim: instead of logging, I've also appended CPU affinity setting logic in here. How about this way?

vllm_rbln/v1/worker/utils.py
```diff
@@ -286,38 +286,60 @@ def get_autobind_cpu_ids(
...
-        local_rank: Local rank of the worker.
-        parallel_config: Parallel configuration.
-    """
-    # Setup thread affinity based on NUMA nodes
+) -> str:
+    """CPU set string that ``set_cpu_affinity`` will use (comma list, ``all``, or ``nobind``)."""
     if envs.VLLM_RBLN_NUMA and platform.system() == "Linux":
         cpu_arch = current_platform.get_cpu_architecture()
         if cpu_arch in (CpuArchEnum.POWERPC, CpuArchEnum.S390X):
-            # For S390X/POWERPC SMT-8/4/2
```
rebel-jaehwang, line 305: please revive the comment

vllm_rbln/v1/worker/utils.py
```diff
@@ -286,38 +286,60 @@ def get_autobind_cpu_ids(
...
-            local_omp_cpuid = get_autobind_cpu_ids(
+            return get_autobind_cpu_ids(
                 rank,
                 local_rank,
                 parallel_config,
                 lambda cpus: [cpu for cpu in cpus if cpu.id % 8 < 4],
             )
-        elif cpu_arch == CpuArchEnum.X86:
-            # For x86 SMT-2, use 1 CPU per core
```
rebel-jaehwang, line 313: this comment too

rebel-jaehwang (approved):

rebel-jonghewk (approved):

PR #499, opened 2026-03-31
fix(scheduler): use prefill_token_budget for new request chunk size

rebel-jaehwang:
### 🚀 Summary of Changes
Problem:
When a new prefill request is scheduled, the RBLN scheduler kicks out all running decode requests and restores the full token budget. However, num_new_tokens was clipped using the already-reduced token_budget before the kick-out, causing the first prefill chunk to be short by the number of tokens consumed by decode requests (e.g. 127 instead of 128).

This off-by-one misaligned all subsequent chunk positions, eventually triggering a device runtime abort (SYS_TASK_ABORTED).

Solution:
Use prefill_token_budget instead of token_budget when computing num_new_tokens for new prefill requests.

### 🧪 How to Test
```
pytest tests/torch_compile/v1/core/test_scheduler.py::test_new_prefill_uses_full_budget_when_decode_running
```

### DIFF
--- a/tests/torch_compile/v1/core/test_scheduler.py
+++ b/tests/torch_compile/v1/core/test_scheduler.py
@@ -88,6 +88,44 @@ def test_schedule_chunked_prefill():
     assert output.num_scheduled_tokens[request.request_id] == 1
 
 
+def test_new_prefill_uses_full_budget_when_decode_running():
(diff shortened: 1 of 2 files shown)


huijjj (approved):
Thanks

PR #506, opened 2026-04-03
other: add medusa spec decode test coverage

junstar92:
### 🚀 Summary of Changes

- Add Medusa proposer unit tests under `tests/torch_compile/unit/v1/spec_decode/test_medusa.py`
- Add a Medusa spec decode e2e smoke test under `test/torch_compile/e2e/v1/spec_decode/test_medusa_basic.py`
- Add test-only helpers under `test/torch_compile/e2e/v1/spec_decode/utils.py` to convert external Medusa adapters into a vLLM-compatible format

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run the unit tests: `pytest tests/torch_compile/unit/v1/spec_decode/test_medusa.py`
2. Run the e2e tests: `pytest tests/torch_compile/e2e/v1/spec_decode/test_medusa_basic.py`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/torch_compile/e2e/v1/spec_decode/test_medusa_basic.py
+++ b/tests/torch_compile/e2e/v1/spec_decode/test_medusa_basic.py
@@ -1,0 +1,82 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""End-to-end smoke tests for Medusa speculative decoding."""
+
+from __future__ import annotations
+
+import pytest
+from vllm import LLM, SamplingParams
+
+from .utils import (
+    DEFAULT_MEDUSA_MODEL_ID,
+    DEFAULT_MODEL_ID,
+    ensure_converted_medusa_adapter,
+)
+
+PROMPTS = [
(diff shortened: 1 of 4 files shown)


rebel-jinhwan:
I think this e2e test is too simple to find any mismatch bug. I found tpu-inference is doing correctness and performance test. I think this approach seems good for us too
https://github.com/vllm-project/tpu-inference/blob/443447126329daa17f8e0e8c4318eaec2d4f0270/tests/e2e/test_speculative_decoding.py#L281

junstar92:
@rebel-jinhwan I think this is sufficient for now. Since the draft model has not been optimized at all, I do not expect it to pass if correctness is judged based on execution time. If a failure is acceptable, I am fine with proceeding as is.

For now, I’ll check next week whether it is actually faster.

rebel-jinhwan:
Thanks, we should ensure correctness even if there's no speedup

junstar92:
@rebel-jinhwan Just to make sure we’re aligned: when you say “correctness,” do you mean “speedup”?

junstar92:
Sorry, I was viewing this on mobile and think I only looked at the performance-related functions. I understand it now as a comparison against the base model results.

junstar92:
@rebel-jinhwan I added the correctness test as you suggested, but it appears to be failing. Could you take a look?
Do you also need a performance-side test?

rebel-jinhwan:
@junstar92 I see. I'll check it out. Performance test seems not urgent at the moment. but i think we should add it later to ensure the performance gain

junstar92:
Okay, then, I'll do the same for eagle/eagle3 and wrap it up.

rebel-jinhwan:
Tests are all passed at `compiler version: 0.10.3.dev136+g52719190`

FYI. this patch fixes torch.compile bugs that choose torch.jit.trace instead of torch.export

rebel-jinhwan (approved):
LGTM

PR #525, opened 2026-04-13
refactor(scheduler): delay caching instead of undoing

rebel-jaehwang:
### 🚀 Summary of Changes

Problem:
The scheduler speculatively cached blocks during allocate_slots, then
had to undo the caching (via undo_uncomputed_block_caching) in three
places: spec_decode_cap trimming, prefill over-allocation for chunked
prefill, and prefill preempting running decodes. This was error-prone
and coupled the scheduler to KVCacheManager internals (block_pool,
num_cached_block).

Solution:
Pass delay_cache_blocks=True to all allocate_slots calls so no blocks
are cached during allocation. A single finalization loop after all
scheduling decisions calls cache_blocks and schedule_sub_block_indexing
for each actually-scheduled request. This eliminates
undo_uncomputed_block_caching.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

---

### 🧪 How to Test

Run pytests.

### DIFF
--- a/vllm_rbln/v1/core/rbln_scheduler.py
+++ b/vllm_rbln/v1/core/rbln_scheduler.py
@@ -777,6 +777,10 @@ class RBLNScheduler(Scheduler):
                     )
                     scheduled_encoder_inputs.pop(req.request_id, None)
                     undo_uncomputed_block_caching(req, self.kv_cache_manager)
+                    if isinstance(self.kv_cache_manager, RBLNKVCacheManager):
+                        self.kv_cache_manager._pending_indexing.pop(
+                            req.request_id, None
+                        )
 
                 scheduled_running_reqs.clear()
                 token_budget = prefill_token_budget
--- a/vllm_rbln/v1/core/rbln_scheduler.py
+++ b/vllm_rbln/v1/core/rbln_scheduler.py
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import itertools
 import time
 from dataclasses import dataclass, field
 
 from vllm.distributed.ec_transfer.ec_connector.base import ECConnectorMetadata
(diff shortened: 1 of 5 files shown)


vllm_rbln/v1/core/rbln_scheduler.py
huijjj, line 77: shouldn't this if branch be removed and make RBLNKVCacheManager our only kv cache manager and support all cases to properly remove undo_uncomputed_block_caching?

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -807,6 +761,26 @@ def schedule(self) -> RBLNSchedulerOutput:
...
+            self.kv_cache_manager.cache_blocks(
+                req,
+                # Cap at req.num_tokens to exclude unverified spec decode
+                # draft tokens, matching the upstream allocate_slots behavior.
+                min(
+                    req.num_computed_tokens + num_scheduled_tokens[req.request_id],
+                    req.num_tokens,
+                ),
+            )
+            if isinstance(self.kv_cache_manager, RBLNKVCacheManager):
```
huijjj, line 781: as mentioned above, this should be an assertion that always holds

huijjj (changes requested):

vllm_rbln/v1/core/rbln_scheduler.py
rebel-jaehwang: The changes to the schedular are compatible with both upstream KVCacheManager and RBLNKVCacheManager. The latter required a bit of change for compatibility with `delay_cache_blocks=True`.

huijjj: didn't knew `delay_cache_blocks` was a feature introduced in upstream, thanks

huijjj (approved):

PR #527, opened 2026-04-14
fix(config): ensure max_num_batched_tokens >= max_source_positions for enc-dec

rebel-jonghewk:
## Summary
- Fix `ValueError: Chunked MM input disabled but max_tokens_per_mm_item (1500) is larger than max_num_batched_tokens (448)` when initializing Whisper models
- `sync_vllm_from_rbln_config` was setting `max_num_batched_tokens = dec_max_seq_len` (448), but vllm's `MultiModalBudget` now validates that `max_num_batched_tokens >= max_tokens_per_mm_item` (1500 for whisper audio) when chunked MM input is disabled
- For encoder-decoder models, now uses `max(max_model_len, max_source_positions)` as `max_num_batched_tokens` to satisfy the encoder's token budget

## Root Cause
vllm added a `MultiModalBudget` validation in the v1 engine that checks encoder input size against `max_num_batched_tokens`. The RBLN config sync was overriding `max_num_batched_tokens` to the decoder length (448) without considering the encoder's larger input size (1500).

## Test plan
- [ ] Run Whisper CI script (`whisper-small`) to verify model initializes and transcribes correctly
- [ ] Run non-encoder-decoder models to verify no regression

### DIFF
--- a/vllm_rbln/utils/optimum/configuration.py
+++ b/vllm_rbln/utils/optimum/configuration.py
@@ -32,6 +32,7 @@ from vllm_rbln.utils.optimum.rbln_params import (
 )
 from vllm_rbln.utils.optimum.registry import (
     get_rbln_model_info,
+    is_enc_dec_arch,
     is_generation_arch,
     is_multi_modal,
 )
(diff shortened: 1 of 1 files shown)


rebel-seinpark (approved):
LGTM! But is there any effects to chunked-prefill ? As I know, in vllm-rbln, chunked-prefill use `max_num_batched_tokens` as chunked prefill size.

rebel-jonghewk:
> LGTM! But is there any effects to chunked-prefill ? As I know, in vllm-rbln, chunked-prefill use `max_num_batched_tokens` as chunked prefill size.

I think it's safe because its prefill doesn't actually chunk the decoder — it runs the encoder in a single call and returns a dummy logit.

PR #528, opened 2026-04-14
fix(gemma3): use non-negative sentinel for IMG_PAD_TOKEN_ID

rebel-jonghewk:
## Summary

Fixes `OverflowError` during Gemma3 multimodal inference caused by `IMG_PAD_TOKEN_ID = -1` being passed to vllm's `FastIncrementalDetokenizer`, which expects u32 values.

### Root Cause

`IMG_PAD_TOKEN_ID` was set to `-1` and inserted into `prompt_token_ids` for image boundary padding. When vllm's `FastIncrementalDetokenizer` passes these IDs to the tokenizer's Rust `DecodeStream`, it expects unsigned 32-bit integers, causing an overflow.

### Solution

Use `PAD_TOKEN_ID` (0) directly instead of a separate sentinel:

- **Non-negative** — fits within u32 range (no `OverflowError`)
- **Within vocab range** — passes vllm's `InputProcessor._validate_prompt` (no `ValueError`)
- **Safe as a sentinel** — the HF tokenizer does not produce token 0 (`<pad>`) in single-sequence encoding, so `(input_ids != PAD_TOKEN_ID)` correctly masks only the image padding positions in the attention mask
- **Removes `postprocess_input_ids()`** — replacing the sentinel with `PAD_TOKEN_ID` is now a no-op

Also fixes a bug in `prepare_vllm_for_compile` where `block_size is None` was dead code (pydantic's validator converts `None` → `16` before the check runs). Changed to use `user_specified_block_size` so the auto-default to 4096 for compilation actually works.

### Testing

- Gemma3 multimodal (image + text) inference: no OverflowError or ValueError
- Text-only inference: no regression
- Compile-on-the-fly with HuggingFace model ID: works without explicit `--block-size`

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/gemma3.py
+++ b/vllm_rbln/model_executor/models/optimum/gemma3.py
@@ -33,7 +33,12 @@ from .optimum_attention import HybridAttentionImageManager, HybridAttentionImage
 
 logger = init_logger(__name__)
 
-IMG_PAD_TOKEN_ID = -1
+# Sentinel token ID for image boundary padding. Must be a non-negative value
+# that fits within u32 range, since vllm's FastIncrementalDetokenizer passes
+# prompt_token_ids to tokenizers.DecodeStream which expects u32 values.
+# Using a value beyond the vocab size (262208) so it won't collide with real tokens.
+# This is replaced with PAD_TOKEN_ID before model execution.
+IMG_PAD_TOKEN_ID = 2**30
 PAD_TOKEN_ID = 0
 
 
--- a/vllm_rbln/model_executor/models/optimum/gemma3.py
+++ b/vllm_rbln/model_executor/models/optimum/gemma3.py
@@ -33,12 +33,6 @@ from .optimum_attention import HybridAttentionImageManager, HybridAttentionImage
 
 logger = init_logger(__name__)
 
-# Sentinel token ID for image boundary padding. Must be a non-negative value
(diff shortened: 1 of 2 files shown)


rebel-seinpark:
I think 2**30 is too big to it. It raise an error `ValueError: Token id 1073741824 is out of vocabulary`

rebel-jonghewk:
You're right, `2**30` would fail vllm's `InputProcessor._validate_prompt` (`input_processor.py:582-585`) which checks `max(prompt_token_ids)` against the vocab size before our `postprocess_input_ids()` gets a chance to replace it.

I've revised the approach: instead of using a separate sentinel, we now use `PAD_TOKEN_ID` (0) directly. This works because:

1. **Non-negative** — no u32 overflow in `DecodeStream`
2. **Within vocab range** — passes `_validate_prompt`
3. **Safe as a sentinel** — the HF tokenizer does not produce token 0 (`<pad>`) in single-sequence encoding, so `(input_ids != PAD_TOKEN_ID)` correctly masks only the image padding positions in the attention mask
4. **Simpler** — `postprocess_input_ids()` becomes a no-op and is removed

Also included a fix for a `prepare_vllm_for_compile` bug: the `block_size is None` check was dead code because pydantic's validator already converts `None` → `16` before it runs. Changed to use `user_specified_block_size` so auto-defaulting to 4096 for compilation works correctly.

Verified with Gemma3 multimodal (image + text) and text-only inference — no errors.

rebel-seinpark (approved):
LGTM

PR #512, opened 2026-04-07
fix: use container-aware thread count to avoid host op performance degradation in pods

rebel-jonghewk:
### 🚀 Summary of Changes

> *What does this PR do? What feature, fix, or improvement does it bring?*

---

### 📌 Related Issues / Tickets

* Resolves # `Jira Issue: SR-98`

---

## Problem

In container environments, `os.cpu_count()` reports all host CPUs (e.g., 96) while the pod is restricted to a cpuset subset (e.g., 32). Libraries like OpenMP, MKL, and NumPy use `os.cpu_count()` to spawn threads, causing massive thread contention on the limited cpuset — resulting in 30-60x slower host ops compared to bare metal on the same server.

Additionally, `set_cpu_affinity()` further narrows the K8s-assigned cpuset
(e.g., 32 → 16) by applying NUMA-aware physical-core-only selection on top
of an already-configured cpuset, which is unnecessary and harmful in
container environments.

## Solution

Detect container environments by comparing `os.sched_getaffinity(0)` with
`os.cpu_count()`. When a mismatch is detected:

- **Skip `set_cpu_affinity()`**: trust the K8s cpuset configuration
- **Set thread count to `allocated_cpus // 2`**: use physical cores only, excluding HT siblings
- **Set all thread pool env vars**: `OMP_NUM_THREADS`, `MKL_NUM_THREADS`,
  `OPENBLAS_NUM_THREADS`, `NUMEXPR_MAX_THREADS`, `RBLN_NUM_THREADS`

On bare metal, set NUMA-aware binding similar to rbln_worker.py (type2).

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
[...]

PR #529, opened 2026-04-14
fix: skip NUMA cpu affinity on bare metal to prevent 32x latency regression

rebel-jonghewk:
## Summary
- **Removes `set_cpu_affinity` call from the bare metal path** in `optimum_worker.py`, which was restricting the process to a single NUMA node (~32 of 128 cores) and causing 32x latency regression (0.29s → 9.5s) with CPU util drop (387% → 54%)
- **Sets all thread pool env vars** (`OMP_NUM_THREADS`, `MKL_NUM_THREADS`, `OPENBLAS_NUM_THREADS`, `NUMEXPR_MAX_THREADS`, `RBLN_NUM_THREADS`) consistently, matching the container path — the old bare metal path only set `RBLN_NUM_THREADS`
- **Uses `allocated_cpus // 2`** for thread count (physical cores only), consistent with `rbln_worker.py` and the container path — the old bare metal path used `allocated_cpus` (no halving)

## Context
Commit `25dd74f` (PR #512) fixed container pod performance but introduced a severe regression on bare metal (128-core) servers. The container path (4/32 core pods) works correctly; only the bare metal path (`allocated_cpus == reported_cpus`) is affected.

## Root Cause
Three issues in the bare metal `else` branch of `init_device`:
1. `set_cpu_affinity` binds the entire process (including asyncio, GC, IPC) to one NUMA node's physical cores
2. `OMP_NUM_THREADS` env var was not set → OpenMP initialized with 128 threads, then `torch.set_num_threads(32)` reduced active threads, creating thread pool mismatch
3. Thread count used `max(2, allocated_cpus)` instead of `max(2, allocated_cpus // 2)`, inconsistent with `rbln_worker.py`

## Test plan
- [ ] Verify 128-core bare metal: latency should return to ~0.29s baseline
- [ ] Verify 32-core container: no regression from the #512 improvement
- [ ] Verify 4-core container: no regression from the #512 improvement
- [ ] Check logs for `"Bare metal detected (N CPUs). Skipping set_cpu_affinity, setting threads to M"`

### DIFF
--- a/vllm_rbln/v1/worker/optimum_worker.py
+++ b/vllm_rbln/v1/worker/optimum_worker.py
@@ -40,7 +40,7 @@ from vllm_rbln.logger import init_logger
 from vllm_rbln.utils.optimum.cache_blocks import sync_num_blocks
 from vllm_rbln.utils.optimum.rbln_params import get_rbln_params
 from vllm_rbln.v1.worker.optimum_model_runner import RBLNOptimumModelRunner
-from vllm_rbln.v1.worker.utils import set_cpu_affinity, set_omp_num_threads
+from vllm_rbln.v1.worker.utils import set_omp_num_threads
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


rebel-shshin (approved):
👍

PR #435, opened 2026-03-06
model: support minimax 

rebel-kblee:
### 🚀 Summary of Changes

- support minimax
- add fp 8 quantization logic

> *What does this PR do? What feature, fix, or improvement does it bring?*

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #534, opened 2026-04-15
fix(model): make MoE expert_map and tokens_mask compatible with torch.export

rebel-jonghewk:
## Summary
- Replace `.tolist()` + `torch.tensor()` pattern for `expert_map_const` with `.clone().detach().to()` to avoid graph-breaks under `torch.export`
- Add `device` parameter to `get_tokens_mask()` to prevent device mismatch during export tracing
- Applies to both `fused_moe/layer.py` (unquantized optimized path) and `mxfp4.py` (quantized path)

## Test plan
- [ ] Verify `torch.export.export` succeeds on MoE models (e.g. Qwen3 MoE) without graph-break errors
- [ ] Verify MoE inference output is unchanged

### DIFF
--- a/vllm_rbln/model_executor/layers/fused_moe/layer.py
+++ b/vllm_rbln/model_executor/layers/fused_moe/layer.py
@@ -241,7 +241,7 @@ def unquantized_fused_moe_method_rbln(
     return final_hidden_states.reshape(orig_shape)
 
 
-def get_tokens_mask(num_tokens: int, left=1.0, right=0.0):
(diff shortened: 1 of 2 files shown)


rebel-jonghewk:
Converting to draft — these changes caused an output mismatch on some MoE models and were reverted in RBLN-SW/vllm-rbln@a9fbb53344806a670fde9eaf6e92094ea3d06eaf (from #435).

We need to investigate the root cause before landing this. @rebel-kblee will take a look.

PR #543, opened 2026-04-19
feature: support for rejection sampler

rebel-eunji:
### 🚀 Summary of Changes

This PR adds support for running the rejection sampler on RBLN devices.

The flow is split between **`RBLNRejectionSampler.forward`** (logits prep) and
**`RBLNRejectionSampler.rejection_sample`** (the actual accept/reject + scatter),
with the heavy lifting delegated to the `torch.ops.rbln.rejection_sample` NPU primitive.

### Step 0 · Prepare target distribution (`forward`)

1. Sample the **bonus token** from `bonus_logits` via the base sampler.
2. Take `target_logits`, apply logits processors, then `apply_sampling_constraints`:
   - **greedy** → collapse to a **one-hot** distribution (`argmax` set to `1.0`, rest `0`).
   - **non-greedy** → divide by per-token `temperature` (expanded batch→token).
     `top_k` / `top_p` are **not** applied here — they are passed into the kernel
     and applied during sampling.
3. Build `target_probs`: the one-hot tensor for greedy, else `softmax(target_logits)`.

### Step 1 · Build NPU primitive inputs (packed-then-padded `[B*K]` layout)

`rejection_sample` initializes the output buffer and reshapes inputs to the kernel's contract:

- `output_token_ids = [batch_size, max_spec_len + 1]` (int64) filled with `PLACEHOLDER_TOKEN_ID`.
- `active_mask = [n > 0 for n in num_draft_tokens]` — batch-space mask of rows that have any draft.
- `reshaped_draft_token_ids` → `[B*K]` int32, valid drafts copied into the **front**
  (`[:N]`, `N = sum(num_draft_tokens)`), tail zero-padded.
- `reshaped_target_probs` → `[B*K, V]`, same packed-then-padded copy.
- `draft_per_batch` → `[B, K]` per-batch padded view of drafts (placeholder elsewhere),
  so it aligns row-by-row with `output_token_ids` for the scatter in Step 3.
- `cu_num_draft_tokens` cast to int32 (kernel contract).

### Step 2 · Call `torch.ops.rbln.rejection_sample`

Invoke the compiled kernel with
`(reshaped_draft_token_ids, reshaped_target_probs, cu_num_draft_tokens_i32, top_k, top_p)`.
It samples per row (`apply_top_k_top_p` + `random_sample`), compares against the draft,
and returns:

| output | shape | meaning |
|---|---|---|
| `recovered_token_ids` | `[B, K]` int64 | sampled token written **only at the first-reject slot** per row; other slots are filler |
| `num_accepted` | `[B]` int64 | count of consecutive accepts before the first reject (`== num_draft_tokens[i]` when fully accepted) |

> Note: the kernel emits a scalar `num_accepted` count, not per-position acceptance
> flags. There is no chain-AND / `cumprod` step anymore — the count already encodes
> "stop at first reject".

### Step 3 · Compose per-position output (cols `0 … K-1`, batch space)

Using position indices `positions = arange(K)` and `num_accepted` per batch:

| mask | how | meaning |
|---|---|---|
| `all_accepted_active` | `(num_accepted == num_draft_tokens) & active_mask` | row accepted **all of its own** drafts |
| `accepted_pos_mask` | `positions < num_accepted` | accepted positions |
| `recovered_pos_mask` | `(positions == num_accepted) & active_mask & ~all_accepted_active` | exactly the first-reject slot (skipped for inactive / fully-accepted rows) |

Two masked writes (no Python loop):

- **3a** `torch.where(accepted_pos_mask, draft_per_batch, …)` → accepted positions get the draft token unchanged.
- **3b** `torch.where(recovered_pos_mask, recovered_token_ids, …)` → first-reject position gets the recovered token.

Everything else stays `PLACEHOLDER`.

### Step 4 · Scatter the bonus token

| sub-step | rows | col | value |
|---|---|---|---|
| 4a | active ∩ all-accepted | `num_draft_tokens[i]` (`== K` only for full rows) | `bonus` |
| 4b | inactive (no drafts) | `0` | `bonus` |

`bonus` is cast to the output dtype first (boolean-mask `index_put` does not auto-cast).
Finally the buffer is cast to int32 to stay consistent with the CPU sampler.

## Behavior matrix

| case | cols `0 … K-1` | bonus placement |
|---|---|---|
| fully accepted | `[draft₀, …, draft_{n-1}]` | `bonus` at col `n` (`= num_draft_tokens[i]`) |
| partial accept (reject at `i`) | `[draft₀, …, draft_{i-1}, recovered, PLACEHOLDER, …]` | no bonus |
| no drafts (inactive) | `[bonus, PLACEHOLDER, …]` | bonus already at col 0 |

## Why each piece

- **`num_accepted` count instead of per-position flags**: the kernel stops at the
  first reject by construction, so a single count fully describes the accept/reject
  boundary — position masks (`positions < / == num_accepted`) reconstruct the layout
  without `cumprod`.
- **packed-then-padded `[B*K]` input**: matches the NPU primitive's fixed-shape
  contract; `draft_per_batch` bridges back to the per-batch-padded output layout.
- **masked `torch.where` writes**: pick the right token per position without a Python loop.
- **per-row all-accept (`num_draft_tokens[i]`, not `max_spec_len`)**: a row may have
  fewer drafts than `max_spec_len`, so the bonus lands right after *its own* last draft.
- **bonus only when fully accepted**: if the chain broke, the recovered token already
  replaces the rejected draft — no extra token.

## Known issues cc. @rebel-jongho 
- wrapping cpu-level torch.compile in rbln_sampler, rbln_rejection_sampler
- pooling tensor buffers
- device tensor
- managing torch.register

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `python examples/experimental/run_specdec_ngram_test.py` (simple one)
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


tests/torch_compile/e2e/v1/spec_decode/test_rejection_sampler_correctness.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-jongho: nit : 2026??

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -17,29 +17,146 @@
...
+    top_k: torch.Tensor | None,
+    top_p: torch.Tensor | None,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """FakeTensor body — only output shape/dtype matters for torch.compile."""
+    B = cu_num_draft_tokens.shape[0]
+    K = draft_token_ids.shape[0] // B
+    device = draft_token_ids.device
+    out_tokens = torch.empty((B, K), dtype=torch.int64, device=device)
+    num_accepted = torch.empty((B,), dtype=torch.int64, device=device)
+    return out_tokens, num_accepted
```
rebel-jongho: could you move this ops to rebel-compiler, so that we don't need to maintain two repositories?

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -147,108 +270,167 @@ def forward(
...
+        src_offset = 0
+        for i, n in enumerate(num_draft_tokens):
+            if n == 0:
+                continue
+            draft_per_batch[i, :n] = draft_token_ids[src_offset : src_offset + n]
+            src_offset += n
+
+        # NPU primitive expects cu_num_draft_tokens in int32; vLLM hands it
+        # over as int64 (inherited from num_draft_tokens cumsum).
+        cu_num_draft_tokens_i32 = cu_num_draft_tokens.to(torch.int32)
```
rebel-jongho: is it neccesary?

it seems like that cast is already included in custom op converter

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -147,108 +270,167 @@ def forward(
...
+        assert draft_probs is None or draft_probs.is_contiguous()
+        assert target_probs.is_contiguous()
+        assert bonus_token_ids.is_contiguous()
+        assert target_probs.shape == (num_tokens, vocab_size)
+
+        # Output buffer (batch space). Unwritten slots stay as PLACEHOLDER.
+        output_token_ids = torch.full(
+            (batch_size, max_spec_len + 1),
+            PLACEHOLDER_TOKEN_ID,
+            dtype=torch.int64,  # Consistent with SamplerOutput.sampled_token_ids.
```
rebel-jongho, line 238: I'm not sure but 

can't we cache this buffer at the first call?

rebel-jongho: if whole sampler is being wrapped within `torch.compile`, then skip this comment

tests/torch_compile/e2e/v1/spec_decode/test_rejection_sampler_correctness.py
rebel-eunji: It is because currently the checking logic only allows 2025, not 2026.
I will fix it in another PR.

rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/655

vllm_rbln/v1/sample/rbln_rejection_sampler.py
rebel-eunji: I've made a related compiler PR.

vllm_rbln/v1/sample/rbln_rejection_sampler.py
rebel-eunji: That's right. Thanks. I've removed it.

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -147,108 +190,163 @@ def forward(
...
-        batch_size,
-        device,
-    )
+        # 3b) First-reject position: write the NPU-recovered token.
+        recovered_pos_mask = (
+            (positions == num_accepted_per_batch.unsqueeze(1))
+            & active_mask.unsqueeze(1)  # To skip inactive row (num_draft_tokens == 0)
+            & ~all_accepted_active.unsqueeze(1)  # all-accept -> no recovery
+        )  # (B, K)
+        output_token_ids[:, :max_spec_len] = torch.where(
```
rebel-wonsubkim, line 352: in another PR, we SHOULD check/verify device tensor behavior for this torch operation
since there are many torch operations after rejection sampler compilation, It is better to use host tensor in my opinion

rebel-wonsubkim (approved):
LGTM
except for rejection sample compilation, there are many torch operations for RBLN rejection sampler
I think that it is better to use host tensor for RBLN rejection sampler

rebel-jonghewk:
@rebel-eunji Is this tested with device tensor?

rebel-eunji:
I'll check the internal pr ci again. The failed ci related with sampler passed in my local server.

junstar92:
@rebel-eunji It looks like the following tests are failing:
- `test_uniform_num_draft_tokens_eager`
- `test_varying_num_draft_tokens_eager`

My current guess is that this is caused by the CPU implementation of `torch.ops.rbln.rejection_sample` returning `torch.empty`, as seen in `rebel.core_ori.op.torch_custom_ops.random.py`.

Could this be due to a compiler version difference? If not, could you help me understand what else might be causing the issue? I'm currently using compiler version `0.15.0.dev143`.

PR #544, opened 2026-04-20
other(bumpup): Make vllm-rbln type2 compatible to upstream v0.19.1

rebel-jinhwan:
### 🚀 Summary of Changes

Make `vllm-rbln` work against both **vLLM 0.18.0** (the current stable baseline) and **vLLM 0.19.1** (next upstream) from the same source tree, so we can validate compatibility before the bump and avoid a long-lived divergent branch later.

---

 ### 🧑‍💻 Optimum side changes                                              

  * `examples/optimum/run_qwen3_reranker.py` — `from vllm.inputs.data import TokensPrompt` → `from vllm.inputs import TokensPrompt`. 0.19 reorganized `vllm/inputs/`: `data.py` is gone and `TokensPrompt` lives in `llm.py`; both versions
  re-export it from `vllm.inputs.__init__`, so the package-level import works on both.
  * `vllm_rbln/v1/worker/optimum_input_batch.py` — same `_make_prompt_token_ids_*` fix above. Without this, every `warm_up=False` sampler test on 0.19 fails with `AttributeError`, and `RBLNOptimumModelRunner.execute_model` cannot build sampling metadata for the optimum compile path.
  * `.github/workflows/rbln_optimum_ci.yaml` — matrix-ifies the three RSD jobs (`RSD 1` / `RSD 4` / `RSD 8`) over `vllm_versions_stable`. Each job now derives the vllm-cpu wheel index URL from `matrix.vllm_version` and pins  `vllm==<matrix.vllm_version>` at install time, mirroring the pytest workflow. Job names now include the vllm version. Optimum CI deliberately runs on the **stable** slice only.

  ### 🧪 Test changes

  * `tests/v1/worker/test_sampler.py::test_sampler_logits_reshape_prevents_torch_compile_recompile` — used to assert `compile_counter.frame_count == 3` (a magic number that comes out as 2 on 0.19 because of upstream compile changes).
  Rewrite to snapshot the frame count after the first step and assert the second step (with a different logits stride) doesn't increase it. This tests the real invariant (no sampler recompile on stride change) without coupling to a
  vllm-internal count.
  * `tests/entrypoints/openai/test_basic.py::test_request_cancellation` — was parametrized with `--disable-frontend-multiprocessing`, a 0.18-only CLI flag that 0.19 removes. The parametrize block forced the server into a 0.18-only mode that doesn't exist on 0.19 (`unrecognized arguments` at server startup). Drop the parametrize so the test runs with default server args; the cancellation invariant is what we actually care about.

  ### 🤖 CI: run everything against multiple vllm versions

  The versions tested are now declared once and consumed everywhere:

  * `.github/ci-versions.json` — single source of truth, split into `stable` (must pass) and `dev` (failure-allowed):
    ```json
    { "vllm_versions": { "stable": ["0.18.0+cpu"], "dev": ["0.19.1+cpu"] } }
    ```
  * `.github/workflows/rbln_utils_get_ci_versions.yaml` — reusable workflow that reads the JSON and emits three outputs: `vllm_versions_all`, `vllm_versions_stable`, `vllm_versions_dev`.

  Consumers:

  | Workflow | Matrix slice | Failure policy |
  |---|---|---|
  | `rbln_trigger_fsw_pr_ci.yaml` | `_all` | both required (one dispatch per version) |
  | `rbln_trigger_fsw_merge_ci.yaml` | `_all` | both required |
  | `rbln_vllm-rbln_pytest.yaml` (pytest + pytest-arc) | `_all` | `continue-on-error: contains(_dev, matrix.vllm_version)` — dev failures don't gate the PR |
  | `rbln_optimum_ci.yaml` (RSD 1 / 4 / 8) | `_stable` only | required |

  Practically, every test in `tests/v1/{core,worker}` and `tests/entrypoints/openai` is exercised against both `0.18.0+cpu` and `0.19.1+cpu` on every PR. Optimum CI (the heavy decoder-only / VL examples) runs only on the stable version. Adding/removing a vllm version is a one-line edit to `ci-versions.json` and propagates to all four workflows automatically.

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)
* [x] ❓ Other (`other`):  vLLM version compatibility / CI matrix expansion

---

### 🧪 How to Test

```bash
export UPSTREAM_VERSION=0.19.1
uv pip install "vllm==${UPSTREAM_VERSION}" \
    --extra-index-url "https://wheels.vllm.ai/${UPSTREAM_VERSION}/cpu" \
    --extra-index-url https://download.pytorch.org/whl/cpu

pytest -s -v tests/torch_compile/unit
pytest -s -v tests/torch_compile/e2e
```
---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/worker/optimum_input_batch.py
```diff
@@ -101,7 +101,11 @@ def _make_sampling_metadata_rbln(self, bucket_size: int) -> SamplingMetadata:
             # step pooling during the sampling/pooling process.
             # Hence copy these tensors only when there are requests which
             # need penalties/step_pooler to be applied.
-            prompt_token_ids = self._make_prompt_token_ids_tensor()
+            # vLLM 0.19 renamed _make_prompt_token_ids_tensor to _cpu_tensor.
+            if hasattr(self, "_make_prompt_token_ids_cpu_tensor"):
+                prompt_token_ids = self._make_prompt_token_ids_cpu_tensor()
+            else:
+                prompt_token_ids = self._make_prompt_token_ids_tensor()
```
rebel-eunji, line 121: Thanks 👍

rebel-jaehwang: For version-specific code paths like this, let's use a fixed comment format such as `NOTE(0.18): ...` so that we can conveniently identify dead code when dropping the old version.

rebel-eunji: 👍

rebel-jaehwang (changes requested):

rebel-jaehwang (approved):

rebel-eunji:
@rebel-seinpark
Could you review the CI workflow changes?

rebel-eunji (approved):

PR #546, opened 2026-04-20
fix(core): account for draft model memory in speculative decoding

junstar92:
### 🚀 Summary of Changes

This PR improves speculative decoding stability and memory estimation by accounting for draft model memory in available memory estimation, aligning spec decode paths with the updated attention build interface, and cleaning up the related test/runtime setup.

- Update available memory estimation to account for draft model memory usage, including draft model weights and related runtime/kernel memory, and add coverage for the updated logic.
- Remove outdated example-side `gpu_memory_utilization` constraints and clean up unnecessary environment variable usage that became redundant after the memory estimation fix.
- Fix the EAGLE failure introduced by the attention build interface change.
- Consolidate `spec_decode` test environment setup and fix the Medusa test issue where `RBLN_USE_CUSTOM_KERNEL` was not applied to the actual test process by isolating Medusa e2e execution in a spawned subprocess.
- Adjust related unit/e2e tests to match the updated execution path and interfaces.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run the updated speculative decoding test suite.

- `pytest tests/torch_compile/unit/v1/spec_decode/test_eagle.py`
- `pytest tests/torch_compile?e2e/v1/spec_decode`

2. Verify that EAGLE and Medusa engines initialize successfully and that the Medusa e2e tests no longer fail due to `RBLN_USE_CUSTOM_KERNEL` not being applied in the test process.
3. Verify that draft-model-aware available memory estimation is used and that the worker logs reflect draft model kernel size / adjusted available memory during initialization.

As part of validation, the updated paths were confirmed to work correctly under the default `gpu_memory_utilization=0.9` setting.

---

---

### 📋 Checklist

---

### 💬 Notes

This PR combines closely related spec decode fixes that were validated together: draft-aware memory estimation, EAGLE compatibility after the attention interface update, and test process isolation for Medusa custom-kernel execution.

EAGLE3 test results could not be verified in this environment because the compiler version currently in use does not support compiling EAGLE3.

Known limitation
- The shared kernel-size estimator still assumes a separate embedding term, so Medusa draft memory is underestimated by roughly one head-size matrix.
- `num_runtimes` is still estimated from the target-model path only, so Medusa/EAGLE-specific draft runtime overhead is not yet modeled precisely.

---

### DIFF
--- a/examples/experimental/run_specdec_eagle_test.py
+++ b/examples/experimental/run_specdec_eagle_test.py
@@ -18,6 +18,7 @@ import re
 from pathlib import Path
 from typing import Any
 
+os.environ["RBLN_USE_CUSTOM_KERNEL"] = "1"
 os.environ["VLLM_RBLN_USE_VLLM_MODEL"] = "1"
 os.environ["VLLM_RBLN_COMPILE_STRICT_MODE"] = "1"
 os.environ["VLLM_DISABLE_COMPILE_CACHE"] = "1"
--- a/examples/experimental/run_specdec_eagle_test.py
+++ b/examples/experimental/run_specdec_eagle_test.py
@@ -18,14 +18,10 @@ import re
 from pathlib import Path
 from typing import Any
 
-os.environ["RBLN_USE_CUSTOM_KERNEL"] = "1"
 os.environ["VLLM_RBLN_USE_VLLM_MODEL"] = "1"
 os.environ["VLLM_RBLN_COMPILE_STRICT_MODE"] = "1"
 os.environ["VLLM_DISABLE_COMPILE_CACHE"] = "1"
 os.environ["VLLM_RBLN_ENABLE_WARM_UP"] = "1"
-os.environ["VLLM_RBLN_SAMPLER"] = "0"
-# vLLM(v0.10.2) bug: speculative decoding works only in multi-processing.
-# os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
 
 import torch
 from huggingface_hub import hf_hub_download
@@ -230,7 +226,6 @@ def main():
         },
         disable_log_stats=False,
         tensor_parallel_size=args.tensor_parallel_size,
-        gpu_memory_utilization=0.5,
     )
 
     sampling_params = SamplingParams(temperature=0.1, top_p=0.9, max_tokens=128)
--- a/tests/torch_compile/e2e/v1/spec_decode/utils.py
+++ b/tests/torch_compile/e2e/v1/spec_decode/utils.py
@@ -16,12 +16,8 @@
 
 from __future__ import annotations
 
-import importlib
 import json
-import multiprocessing
-import os
 import re
-import traceback
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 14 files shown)


examples/experimental/run_specdec_eagle_test.py
```diff
@@ -18,14 +18,10 @@
 from pathlib import Path
 from typing import Any
 
-os.environ["RBLN_USE_CUSTOM_KERNEL"] = "1"
```
rebel-jinhwan, line 21: please remain this line

tests/torch_compile/e2e/v1/spec_decode/utils.py
```diff
@@ -275,11 +279,63 @@ def ensure_vllm_compatible_eagle_draft_model(
...
+        ),
+    )
+    proc.start()
+    proc.join()
+
+    if proc.exitcode != 0:
+        tb = ""
+        if not error_queue.empty():
+            tb = error_queue.get()
+        raise AssertionError(f"Spawned process exited with code {proc.exitcode}\n{tb}")
```
rebel-jinhwan: I don't know details about it. but what is difference with `VLLM_WORKER_MULTIPROC_METHOD =spawn` ?

examples/experimental/run_specdec_eagle_test.py
junstar92: ok, but are there any reasons?

tests/torch_compile/e2e/v1/spec_decode/utils.py
junstar92: In the case of Medusa, an error occurs if the `RBLN_USE_CUSTOM_KERNEL` environment variable is not set. I initially tried setting this variable in `__init__.py` or somewhere by using `monkeypatch`, but it was not reflected in the process running `pytest` (this test).

To address this, I modified the test to run in a spawned process and ensured that the environment variable is set when that process is created.

I'm not entirely sure what role `VLLM_WORKER_MULTIPROC_METHOD` plays, but it seems unrelated to solving this issue by switching the worker process to `spawn`.

If you have a better way to ensure the environment variable is properly applied, let me know, and I'll try it.

examples/experimental/run_specdec_eagle_test.py
rebel-jinhwan: i think it'd be good to ensure users to use custom kernel when using spec dec

vllm_rbln/v1/worker/utils.py
```diff
@@ -31,6 +31,45 @@
...
+            )
+
+    lm_heads_params = align(vocab_size, 64) * hidden_size
+    lm_heads_nbytes = (
+        align_2MB(lm_heads_params * default_bits_per_param // 8 / tp_size) * tp_size
+    )
+    word_embedding_params = lm_heads_params
+    params = n_model_params - lm_heads_params - word_embedding_params
+    layer_nbytes = align_2MB(params * nbits_per_param // 8 / num_layers) * num_layers
+    return int(layer_nbytes + lm_heads_nbytes)
```
rebel-jinhwan: nit. it seems int(*) don't need here

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -345,7 +345,8 @@ def determine_available_memory(self) -> int:
                 draft_nbits_per_param = 16
                 draft_packed_num_elems = 1
 
-            for value in draft_model.parameters():
+            draft_params_dict = dict(draft_model.named_parameters())
+            for key, value in draft_params_dict.items():
```
junstar92: is there any problem?

rebel-jinhwan: I failed at unit test when using .parameters() because the mock has no parameters.return_value

junstar92: Then, would it make sense to extend the mock instead?

rebel-jinhwan: could you explain why we should use .parameters()? because target model is also using .named_parameters()

junstar92: We’re not using key, so there isn’t really a reason to use named_parameters. Also, I didn’t modify the other code because I don’t own it.

rebel-jinhwan: I see, then please fix test codes and turn to parameters()

junstar92: It doesn't seem like any errors are occurring - could you point out which unit test is failing? The tests I added appear to be passing.

junstar92: I found it, it seems like something is wrong with running the test.

rebel-jinhwan (approved):
lgtm

junstar92:
There were additional changes (#535) to the memory estimation in `origin/dev`, which caused a conflict during rebase. I've resolved it by incorporating the incoming changes as much as possible.

In particular, I updated the kernel size calculation for the draft model to use bytes instead of the number of parameters, aligning it with the revised implementation.

It would be great to get a review from the original contributor to ensure everything aligns as expected.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -293,14 +294,123 @@ def determine_available_memory(self) -> int:
...
+                if value.is_floating_point():
+                    draft_n_model_bytes += value.numel() * value.element_size()
+                else:
+                    draft_n_model_bytes += int(
+                        value.numel()
+                        * draft_packed_num_elems
+                        * draft_ratio
+                        * draft_nbits_per_param
+                        // 8
+                    )
```
rebel-jinhwan: how about de-duplicating with draft model's quant param calculation?

junstar92: This is something I thought about when adding this - are there any draft models currently using or planning to use FP8 or NVFP4? If not, it might make sense to remove this for now.

rebel-jinhwan: yeah.. we are not planning now. I think we can skip draft quantization for now

junstar92: okay, then I'll remove it and revisit it if needed later.

PR #548, opened 2026-04-20
feature(core): "rbln" device tensor

rebel-jonghewk:
### Summary of Changes

Introduces an opt-in end-to-end RBLN device-tensor flow, gated behind the new
`VLLM_RBLN_USE_DEVICE_TENSOR` environment variable (default `False`). Existing
behavior is preserved when the flag is unset; setting `VLLM_RBLN_USE_DEVICE_TENSOR=1`
activates the new path.

**What the flag enables**

- **Platform (`vllm_rbln/platform.py`):** sets `RblnPlatform.device_name`,
  `device_type`, `dist_backend` to the RBLN values and rewires
  `vllm_config.device_config` accordingly. Without the flag, the platform keeps
  the legacy CPU device.
- **Model runner (`v1/worker/rbln_model_runner.py`):**
  - Skips `CompileContext` construction; the sampler no longer receives a
    `compile_context`; the compile options dict omits `compile_context`.
  - `_prepare_inputs` uses the `cpu` buffer of `query_start_loc` /
    `block_table` / `input_ids` / `positions` so view / pad / clone stay on
    CPU until `.to(self.device)` just before the model executable.
  - `_pad_sampling_metadata` pads sampling tensors to the padded logits
    bucket when logits are kept at `batch_bucket_size`.
  - `_bookkeeping_sync` slices `sampled_token_ids` to `num_reqs` and handles
    the empty-logits case produced by non-final chunked-prefill chunks.
  - `_allocate_kv_cache_tensors` allocates on `self.device` (not `cpu` /
    `meta`).
  - `initialize_kv_cache_tensors` skips `mark_static_address` (no
    `CompileContext` to attach to).
- **Flash attention (`v1/attention/backends/flash_attention.py`):** uses the
  pre-existing CPU copies of `query_start_loc` / `seq_lens` to avoid an extra
  D2H sync. All arithmetic stays on CPU until `.to(self.device)` at the metadata
  constructor.
- **Sampler (`v1/sample/rbln_sampler.py`, `v1/sample/ops/penalties.py`):** drops
  the legacy `inspect`-based `CompileContext` options and builds penalty tensors
  on the caller-provided device.

**Device-agnostic hunks (no flag needed)**

Changes to `fp8.py`, `mxfp4.py`, and `fused_moe/layer.py` (device-aware
`get_tokens_mask(device=...)` and the precomputed `_expert_map_list`) are pure
refactors — they are no-ops when tensors already live on CPU, so they are not
gated.

### Related Issues / Tickets

Related to the ongoing RBLN device-tensor migration.

### Type of Change

- [x] Feature (`feature`)
- [x] Core engine changes (`core`)

### How to Test

1. Default path (flag off) — legacy CompileContext / device="cpu" path:
   ```
   python examples/experimental/offline_inference_basic.py
   ```
2. Device-tensor path (flag on):
   ```
   VLLM_RBLN_USE_DEVICE_TENSOR=1 python examples/experimental/offline_inference_basic.py
   ```
3. Verify parity between both modes on a short prompt (greedy sampling).

### Notes

- Default remains the legacy path to avoid regressions; the flag is the sole
  kill-switch.
- Scratch scripts (`test_llama.py`, `test_llama_batch.py`) were removed from
  the PR in a follow-up commit; local debugging copies are kept on disk only.
- `examples/experimental/data_parallel.py` was updated to import
  `get_open_port` from `vllm.utils.network_utils` (vllm 0.18+ no longer
  re-exports it from `vllm.utils`).


vllm_rbln/model_executor/layers/fused_moe/layer.py
rebel-ykchoi, line 38: I noticed expert_map_const is being generated for the FuseMoE layer. Could you clarify why we're additionally defining layer._expert_map_list instead of leveraging that constant?

rebel-jonghewk: @rebel-ykchoi I rebased onto dev, so those fixes are no longer necessary. I've reverted them — they were for enabling torch.export.

rebel-jaehwang:
mismatch
* `VLLM_RBLN_USE_DEVICE_TENSOR=1 VLLM_RBLN_USE_VLLM_MODEL=1 python examples/experimental/offline_inference_basic.py`
* torch-rbln 0.2.0
* rebel_compiler 0025ff7012

---

cause: export failure. fix: https://github.com/rebellions-sw/rebel_compiler/pull/10693

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1840,6 +1913,14 @@ def _sample(
...
         # Sample the next token and get logprobs if needed.
         sampling_metadata = self.input_batch.sampling_metadata
+        if (
+            envs.VLLM_RBLN_USE_DEVICE_TENSOR
+            and logits is not None
+            and logits.shape[0] > self.input_batch.num_reqs
+        ):
+            sampling_metadata = self._pad_sampling_metadata(
+                sampling_metadata, logits.shape[0]
+            )
```
rebel-jaehwang, line 1926: review from copilot

**padded `SamplingMetadata` breaks speculative decoding in device-tensor mode**:
`RBLNModelRunner._sample()` now pads `sampling_metadata` whenever `logits.shape[0] > num_reqs` (`vllm_rbln/v1/worker/rbln_model_runner.py:1916-1923`). That condition is also true for speculative decoding, where logits are shaped `[num_draft_tokens + batch_size, vocab]`. But `_pad_sampling_metadata()` rebuilds the metadata at that larger length (`vllm_rbln/v1/worker/rbln_model_runner.py:1843-1896`), so `temperature` / `top_k` / `top_p` no longer match the bonus-token logits batch that `RBLNRejectionSampler` passes into `self.sampler()` (`vllm_rbln/v1/sample/rbln_rejection_sampler.py:87-99`). Random speculative decoding will therefore hit batch-dimension mismatches during sampling. The helper also drops `spec_token_ids`, so penalties / bad-word filtering lose the draft-token context even when execution does not crash.

---

we'll take care of this after this pr is merged

vllm_rbln/platform.py
```diff
@@ -210,6 +220,8 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 )
 
                 RblnPlatform.device_type = "rbln"
+                if envs.VLLM_RBLN_USE_DEVICE_TENSOR:
+                    RblnPlatform.dist_backend = "rbln-ccl"
```
rebel-eunji: This looks like a duplicate of line 171.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -4097,8 +4194,21 @@ def _allocate_kv_cache_tensors(
...
-            tensor = torch.zeros(kv_cache_tensor.size, dtype=torch.int8, device=device)
+            if envs.VLLM_RBLN_USE_DEVICE_TENSOR:
+                device = self.device
+                tensor = torch.empty(
+                    kv_cache_tensor.size, dtype=torch.int8, device=device
+                )
+            else:
+                device = (
+                    "cpu"
+                    if envs.VLLM_RBLN_USE_CUSTOM_KERNEL
```
rebel-eunji: Why is the "to CPU" logic under VLLM_RBLN_USE_CUSTOM_KERNEL added in this device tensor PR?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1307,10 +1314,12 @@ def _prepare_inputs(
                 logits_indices
             )
 
+        if logits_indices.device != self.device:
+            logits_indices = logits_indices.to(self.device)
```
rebel-eunji: It looks like this if condition isn't necessary.

rebel-jonghewk: Isn't the device of logits_indices cpu before this?

vllm_rbln/platform.py
rebel-jonghewk: I think enforce_eager has different purpose with device tensor.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-jonghewk: Applied!

rebel-jaehwang (approved):

vllm_rbln/model_executor/layers/fused_moe/layer.py
rebel-daeyang: Looks like `device` needs to be passed here as well:
`tokens_mask = get_tokens_mask(num_tokens, device=router_logits.device)`

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -1091,16 +1111,29 @@ def build(
...
+            query_start_loc_cpu = common_attn_metadata.query_start_loc_cpu
+            _seq_lens_cpu = common_attn_metadata._seq_lens_cpu
+            seq_lens_cpu = (
+                _seq_lens_cpu[:num_reqs]
+                if _seq_lens_cpu is not None
+                else seq_lens[:num_reqs].cpu()
+            )
+            query_seq_lens = query_start_loc_cpu[1:] - query_start_loc_cpu[:-1]
+            num_computed_tokens_cpu = seq_lens_cpu - query_seq_lens
+            seq_idx = positions[query_start_loc[:num_reqs].to("cpu")].view(-1, 1)
```
rebel-daeyang: nit: I'm curious if there's any benefit to using `query_start_loc[:num_reqs].to("cpu")`. If not, I think going with `query_start_loc_cpu[:num_reqs]` would be better.

rebel-daeyang (approved):
lgtm

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-eunji:
```suggestion
        logits_indices = logits_indices.to(self.device)
```

vllm_rbln/platform.py
rebel-eunji:
```suggestion
```
We've already synced on this in person.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-eunji: We've already synced on this in person.

rebel-eunji (approved):
lgtm

rebel-jonghewk:
FYI @rebel-seinpark — I'm planning to merge this PR today or tomorrow. We'll need to adjust our CI branch accordingly.

PR #549, opened 2026-04-20
fix(whisper): use vLLM block_tables as decoder batch position

rebel-seinpark:
### 🚀 Summary of Changes

The Whisper integration reused AttentionManager from sliding-window models for `input_block_ids`, which allocated a local slot via `min(available_ids)`. In sequential mode `attention_manager.pop` emptied the table after each request, so `min` always returned 0 and every new request was placed at decoder batch position 0. 

Meanwhile vLLM's LRU BlockPool kept advancing its block id (0, 1, 2, ...), so the encoder wrote cross-KV to a different slot than the one the decoder batch was going to read from. The decoder then cross-attended to stale KV from sample 0 for every subsequent sequential request.

Drop AttentionManager in Whisper and derive the batch position directly from `block_tables`. This keeps vLLM block id, decoder batch position, and cross-KV slot consistent across sequential, batched, and batches that exceed `batch_size`.

**Fix**
- In `whisper.py`, drop the `AttentionManager` / `InnerAttentionStrategy` import, the init in `__init__`, the `add(...)` call, and the `table_ids` variable.
- Replace them with `valid_block_ids = block_tables.flatten().to(torch.int16)` so that vLLM block id, decoder batch position, and cross-KV slot all share the same numbering.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/whisper.py
+++ b/vllm_rbln/model_executor/models/optimum/whisper.py
@@ -27,13 +27,6 @@ from vllm.utils.jsontree import json_map_leaves
 
 from .base import ModelInputForRBLN
 from .model_base import RBLNOptimumDecoderMixin, RBLNOptimumModelBase
-from .optimum_attention import (
-    AttentionManager,
-    InnerAttentionEntry,
-    InnerAttentionStrategy,
-    InnerR1,
-    InnerR2,
-)
 
 logger = init_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


rebel-eunji (approved):
LGTM

PR #571, opened 2026-04-27
fix(sampler): decouple greedy and topk-topp sampling

rebel-eunji:
## AS-IS

- If top_k and top_p are both None (all greedy), use greedy sampling
- Greedy case also uses the topk_topp torch op

## TO-BE

- If top_k and top_p are both None, use random sampling
- Greedy case uses a separate greedy (argmax) torch op

## NOTE
- Handling for the case where top_k/top_p are not None but some requests are greedy (greedy + top_k/top_p mixed):
    - Same approach as the rejection sampler — set temperature to an epsilon value so the computation behaves similarly to greedy @rebel-magunudo

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -329,8 +415,13 @@ def apply_temperature(
...
         # in-place division triggers buffer key error
         # in torchinductor
+        # NOTE:
+        # Greedy requests use a small temperature (1e-3) so softmax collapses
+        # to a near one-hot at argmax. _SAMPLING_EPS (1e-5) is too small here —
+        # it pushes logits past softmax's safe exp range and overflows.
+        _GREEDY_EPS = 1e-3
         if not all_random:
-            temp = torch.where(temp < _SAMPLING_EPS, 1.0, temp)
+            temp = torch.where(temp < _SAMPLING_EPS, _GREEDY_EPS, temp)
```
rebel-eunji: Based on a few test cases, _SAMPLING_EPS turned out to be too small considering that logits usually fall in the -20 to +20 range. To address this, I added a separate _GREEDY_EPS with a value of 1e-3. 
@rebel-jongho

rebel-jongho: nit: Since `temp` usually stands for 'temporary,' let's spell out `temperature` to keep the code clear

rebel-jongho (approved):
Just out of curiosity, since the NPU supports greedy sampling, can't we just use the upstream `sample` function as is, not overriding?

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -229,6 +233,88 @@ def __init__(
...
+            sampling_metadata.top_k,
+            sampling_metadata.top_p,
+        )
+
+        assert greedy_sampled is None, (
+            "Upstream vLLM runs greedy and random sampling "
+            "separately and merges the results, "
+            "but vLLM RBLN processes greedy and random requests together: "
+            "greedy requests are routed through the random-sampling path "
+            "with a very small temperature value."
```
rebel-eunji, line 315: @rebel-jongho Upstream vLLM invokes both the greedy and random sampling ops, then merges the results.
### diff
- https://github.com/vllm-project/vllm/blob/694d9a81bbb07977e7a72a597acb44f6a848f774/vllm/v1/sample/sampler.py#L252
   - Upstream vLLM invokes the greedy op whenever not all random (i.e., when at least some requests are greedy), whereas our implementation only invokes it when all_greedy.
- https://github.com/vllm-project/vllm/blob/694d9a81bbb07977e7a72a597acb44f6a848f774/vllm/v1/sample/sampler.py#L288-L296
    - I added assert like this line.

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/571/changes/a5a15200db272e325ad40b757d93d36a47023331

rebel-eunji:
The PR CI failed because the latest compiler is not built on server.

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -137,6 +130,19 @@ def rbln_top_k_top_p_sample(
...
+# NOTE(eunji): argmax op is registered in the compiler
+
+
+def rbln_greedy_sample(logits: torch.Tensor) -> torch.Tensor:
+    """
+    Implementation of RBLN greedy sampling.
+    To avoid self parameter issues when torch.compile is used,
+    we define this as a static method.
+    """
+    sampled = torch.ops.rbln.argmax(logits)
```
huijjj, line 142: @rebel-eunji 
It seems like we do not have this yet, would it be possible after the next release?

cc: @pei0033 @junstar92

rebel-eunji: Sorry for the inconvenience.
This will be available starting from compiler v0.10.4.
Until then, please temporarily replace `torch.ops.rbln.argmax(logits)` with `torch.argmax(dim=-1)`.

huijjj: Thank you for letting us know

PR #581, opened 2026-04-29
feature(runtime): async exec + async memcpy

rebel-jonghewk:
### 🚀 Summary of Changes

Drives vLLM's per-step inference through the new async runtime path so dispatch and host-side sync are decoupled. Inputs/outputs land on rbln-device tensors and the compiled model runs via `run_async()` → `synchronize()` instead of the blocking `run()`.

**Required companion PRs (won't build/run without):**

- rebel_compiler — https://github.com/rebellions-sw/rebel_compiler/pull/10435
  `RuntimeInstance::RunAsync` / `Synchronize`, `Stream` abstraction, async memcpy family (`rbln_memcpy_{h2v,v2h,v2v}_async`).
- torch-rbln — https://github.com/RBLN-SW/torch-rbln/pull/13
  Wires `non_blocking=True` host↔device copies through async memcpy bindings, exposes `c10::rbln::synchronize`.

**Follow-up:** `torch.cuda.Stream`-style explicit user streams (multiple in-flight queues per device, parallel dispatch across independent flows) will be tackled in the next PR. This PR uses a single default stream per Context — enough for implicit dep chaining, not yet for explicit parallelism.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

```bash
VLLM_RBLN_USE_DEVICE_TENSOR=1 python examples/experimental/offline_inference_basic.py \
    --model Qwen/Qwen3-0.6B --max-num-seqs 1 --block-size 1024
```

Expected: model loads, warm-up completes, prompts generate. The async path is exercised by every non-blocking host↔device copy and every `model_executable(...)` call (which now dispatches via `run_async()`).

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-daeyang (approved):
lgtm

huijjj:
Is this functionality available with rebel-compiler 0.10.3?

rebel-jaehwang:
The rebel_compiler PR hasn't been merged yet

huijjj:
I don't see any envvars or torch.compile options specific to async runtime here, are we ditching sync runtime for good? or is it just default chaning to async runtime and sync runtime can be used with envvars or options?

rebel-jonghewk:
> are we ditching sync runtime for good? or is it just default changing to async runtime and sync runtime can be used with envvars or options?

Not ditching sync — async just becomes the default dispatch, and sync stays fully available as an escape hatch.

The async/sync toggle lives at the **rebel_compiler runtime layer**, not in vllm-rbln, which is why there's no vllm-specific envvar or `torch.compile` option here:

- **`RBLN_RUNTIME_FORCE_SYNC=1`** (rebel_compiler, rebellions-sw/rebel_compiler#10435) makes `RuntimeInstance::Run()` wait after every op — i.e. the legacy per-op synchronous behavior. Unset (default) → ops dispatch async and are waited once at the end (caller-visible semantics stay sync).

Since the knob is at the runtime layer it applies to vllm-rbln regardless, no separate option needed.

For **async memcpy** specifically it's opt-in per copy: `non_blocking=True` takes the async path, otherwise copies stay synchronous — so async isn't forced there either.

TL;DR: default → async; sync preserved via `RBLN_RUNTIME_FORCE_SYNC=1` (mainly an escape hatch / debugging lever).

vllm_rbln/platform.py
```diff
@@ -229,10 +229,6 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         parallel_config = vllm_config.parallel_config
         scheduler_config = vllm_config.scheduler_config
 
-        if scheduler_config.async_scheduling:
-            scheduler_config.async_scheduling = False
-            logger.warning("Async scheduler not supported on RBLN.")
-
```
rebel-ykchoi, line 235: This PR drops the platform-level guard that used to force scheduler_config.async_scheduling = False for the RBLN backend (it now only stays disabled on the optimum path). On vLLM 0.22 that guard was effectively what kept async scheduling off for the vLLM-model path — with it gone, async scheduling becomes the default for that path, since VllmConfig.__post_init__ auto-enables it whenever there is no incompatible option (and the mp executor we force does support it).

My concern is that we don't actually have an RBLNAsyncScheduler. We pin scheduler_config.scheduler_cls to RBLNScheduler, which subclasses the plain synchronous Scheduler rather than vLLM's AsyncScheduler. So enabling async_scheduling doesn't route us through the async-aware scheduler that upstream expects for this mode; it just flips the flag while the scheduler itself never got the corresponding async support. The remaining # TODO(RBLN): support async scheduling notes and the commented-out RBLNAsyncScheduler selection in the tests suggest this was always meant to land together with a real async scheduler implementation.

Given that, I don't think we should remove the async_scheduling = False override until RBLNAsyncScheduler actually exists. Could we either:

keep forcing async_scheduling = False on the vLLM-model path as well (not just optimum) until the async scheduler is implemented, or
gate the removal behind an explicit opt-in (e.g. an env var) so the default stays on the well-tested synchronous path?

rebel-jonghewk: @rebel-ykchoi  Good catch — agreed. We don't have an RBLNAsyncScheduler; RBLNScheduler is still the synchronous Scheduler subclass, so removing the override just flips the flag without the async-aware scheduler upstream expects for this mode.

Two questions on scope: (1) is there already a plan to implement RBLNAsyncScheduler, or (2) should I include the async scheduler work in this PR?

rebel-jonghewk: cc @rebel-yskim

rebel-jinhwan: fyi. @rebel-taesoo

rebel-yskim: fyi, I'm creating RBLNAsyncScheduler to support functionality in async scheduler, so I think we don't need to create RBLNAsyncScheduler in this PR + keep async_scheduling as false.

PR #608, opened 2026-05-15
fix(disagg_encoder): handle mixed input scenario and fix potential memory leak

rebel-yskim:
### 🚀 Summary of Changes

This PR fixes bug in disaggregated encoder serving: `txt only input` + `txt + image input` scenario.
It also fixes the potential memory leak bug.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Please refer to [docs](https://docs.rbln.ai/latest/software/model_serving/vllm_support/features/disaggregated-encoder.html) 

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/v1/core/optimum_scheduler.py
+++ b/vllm_rbln/v1/core/optimum_scheduler.py
@@ -136,6 +136,8 @@ class RBLNOptimumScheduler(Scheduler):
                 config=self.vllm_config, role=ECConnectorRole.SCHEDULER
             )
 
+        self._pending_free_mm_hashes: list[str] = []
+
         num_gpu_blocks = self.cache_config.num_gpu_blocks
         assert num_gpu_blocks is not None and num_gpu_blocks > 0
 
@@ -553,7 +555,7 @@ class RBLNOptimumScheduler(Scheduler):
             # It contains the request IDs that are finished in between
             # the previous and the current steps.
             finished_req_ids=self.finished_req_ids,
-            free_encoder_mm_hashes=[],
+            free_encoder_mm_hashes=self._pending_free_mm_hashes,
             new_block_ids_to_zero=None,  # It is used for Mamba models
             block_table_dict=block_table_dict,
             cached_block_table=cached_block_table,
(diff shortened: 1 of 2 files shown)


vllm_rbln/v1/core/optimum_scheduler.py
rebel-eunji, line 473: Shouldn't we free the encoder_cache when a request is preempted?

rebel-yskim: I think preemtion does not guarantee the free of request. Does it?

rebel-yskim: To clarify a bit: preemption itself isn't a free event — the request just moves back to the waiting queue with `PREEMPTED` status. The actual free path is `_free_request`, which `finish_requests` invokes for requests in any state (RUNNING / WAITING / PREEMPTED), so a preempted-then-aborted request still gets its encoder cache cleaned up — nothing leaks across preemption.

Re: upstream's `_preempt_request` calling `encoder_cache_manager.free()` — that's a **soft** ref-count decrement, not equivalent to our `encoder_cache.pop()` on the worker. Upstream's `free()` only removes the request's reference and moves the entry into a `freeable` OrderedDict; the actual eviction (appending to `self.freed`, which is what reaches the worker via `free_encoder_mm_hashes`) only fires from `can_allocate()` under capacity pressure, LRU-style. On resume, `check_and_update_cache()` reclaims from `freeable` with no re-encoding.

Our `encoder_cache` on the EC consumer is a flat dict with no ref counting and no `freeable` buffer — `_pending_free_mm_hashes` triggers an immediate hard pop. Mirroring upstream's "free on preempt" call naively here would (1) lose the encoder output and force an EC re-pull on resume, and (2) invalidate entries still shared by other running requests with the same `mm_hash`.

That said, you're right that there's a remaining concern: we don't enforce a `cache_size` budget on the consumer-side cache, so under preemption-heavy or unique-`mm_hash`-heavy workloads it can grow large (bounded by in-flight + preempted requests, not unbounded). Longer term I think this calls for a dedicated cache manager that ports upstream's pattern (ref count + soft freeable + LRU + scheduler-side budget + worker free-signaling) — ideally sharing the abstraction with whatever disagg-KV cache manager lands later, since they have the same producer/consumer + async-pull shape. For this PR I'd leave that as follow-up; this change closes the immediate leak.

rebel-jonghewk (approved):

rebel-eunji (approved):
lgtm 👍

PR #609, opened 2026-05-15
fix(platform): set device attrs at class definition for spawn compatibility

rebel-jaehwang:
### 🚀 Summary of Changes
Problem:
RblnPlatform mutated its class attributes (device_name, device_type, dist_backend) inside check_and_update_config based on env vars.  Under VLLM_WORKER_MULTIPROC_METHOD=spawn the worker subprocess re-imports the module fresh and does not inherit these mutations.

Solution:
Match the upstream pattern: compute device attrs once at class definition, never mutate at runtime.
The enforce_eager-without-device-tensor case still depends on vllm_config, so isolate it in maybe_override_eager_mode_device and re-invoke it from `RblnWorker.__init__` so spawn subprocesses converge to the same device_type as the parent.

---

* Resolves https://github.com/rebellions-sw/LMCache-RBLN/pull/14#discussion_r3216803651

### DIFF
--- a/vllm_rbln/platform.py
+++ b/vllm_rbln/platform.py
@@ -54,13 +54,17 @@ register_backend(name="bypass", compiler_fn=bypass_backend)
 class RblnPlatform(Platform):
     _enum = PlatformEnum.OOT
 
-    # TODO(jiwoo.park): GroupCoordinator uses the device_name
-    # when torch.device(device_name) is called.
-    # But we don't support the 'rbln'' device yet.
-    # To support this, we must use PyTorch-RBLN
+    # Compute device_name/device_type/dist_backend once at class definition
+    # from env vars so that subprocesses spawned under
+    # VLLM_WORKER_MULTIPROC_METHOD=spawn (which re-import this module fresh)
+    # observe identical values to the parent without any extra plumbing.
+    _USE_DEVICE_TENSOR: bool = (
+        envs.VLLM_RBLN_USE_VLLM_MODEL and envs.VLLM_RBLN_USE_DEVICE_TENSOR
+    )
     plugin_name: str = "rbln"
-    device_name: str = "cpu"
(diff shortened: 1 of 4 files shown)


vllm_rbln/platform.py
```diff
@@ -149,6 +153,26 @@ def validate_and_setup_prerequisite(cls, vllm_config: VllmConfig) -> None:
...
+        But the standard ``enforce_eager`` configuration lives on ``model_config``,
+        which isn't available at class-definition time.
+
+        TODO: This path itself is a temporary workaround.
+        Remove this when it's retired.
+        """
+        if (
+            envs.VLLM_RBLN_USE_VLLM_MODEL
+            and not envs.VLLM_RBLN_USE_DEVICE_TENSOR
+            and vllm_config.model_config.enforce_eager
```
rebel-jinhwan: why enforce_eager needs to be part of this condition?

rebel-jaehwang: The goal of this function was to support the legacy path where `enforce_eager` is set without `VLLM_RBLN_USE_DEVICE_TENSOR`. This mode requires `device_type="rbln"`.

On second thought, it seems better to just error out if `VLLM_RBLN_USE_DEVICE_TENSOR`. See the second commit.

You can test this with:
```
VLLM_RBLN_SAMPLER=0 python examples/experimental/eager_basic.py
```
(`VLLM_RBLN_SAMPLER=0` because sampler compilation currently doesn't seem to handle `enforce_eager` properly)

rebel-jinhwan (approved):
LGTM. Please fix PR tests

rebel-jaehwang:
[the sampler test failures](https://github.com/RBLN-SW/vllm-rbln/actions/runs/26017420025/job/76470737611?pr=609) is not related to this pr (discussed offline)

PR #616, opened 2026-05-20
refactor(attn): remove in-graph dtype casts on attn custom op args

rebel-thkim:
## Summary
- Compiler-side patch now handles dtype casting for attention custom op args, so the explicit `.to(torch.int16)` / `.to(torch.int32)` casts in `flash_attention.py` (on `seq_lens`, `block_tables`, `cache_seq_lens`, `num_computed_tokens`, `dyn_size_for_partitions`) are no longer needed and are removed.
- Also drops the conditional `cache_seq_lens.to(torch.int32) if is_batch_attention_opt and b_size > 1 else ...` branch — it becomes a single unconditional pass-through.
- Cleans up the now-stale `# RBLN - seq_lens tensor dtype SHOULD be int16` comment and a trailing-whitespace line introduced during the cleanup.

Native dtypes flowing in from vLLM (`seq_lens`/`block_tables`/`_seq_lens_cpu` are int32, `slot_mapping`/`positions` are int64) are already compatible with the kernels, which do their own `cast(tl.int32)` internally.

## Test plan
- [ ] CI passes (ruff / mypy / typos / license checks)
- [ ] e2e: flash attention path — prefill + decode
- [ ] e2e: causal attention path
- [ ] e2e: sliding-window attention path (covers `cache_seq_lens` / `cache_offsets`)
- [ ] e2e: batch attention opt path (`VLLM_RBLN_BATCH_ATTN_OPT=1`, batch > 1) — verifies removed `cache_seq_lens.to(torch.int32)` branch
- [ ] Confirm compiler-side casting patch is landed and aligned with this PR


rebel-kblee (approved):
lgtm!

PR #619, opened 2026-05-22
feature: introduce mega cache

rebel-jonghewk:
### 🚀 Summary of Changes

Persist torch.compiler mega-cache so the rbln backend's `.rbln` blobs ship as a single atomic bundle per **(model, config-signature, local_rank)** under `VLLM_CACHE_ROOT/rbln`.

**Why atomic:** the current per-file `cache_dir` scheme leaves partial artifacts on disk if compile fails mid-way (e.g. prefill succeeds, decode fails, mismatched shared state across programs, ...). On the next run, those orphan `.rbln` files cache-hit and silently desync from the freshly compiled siblings. Mega-cache fixes this by writing `mega_cache.bin` **only after `warm_up_model()` completes** — every program (prefill, decode, compute_logits, all bucket sizes) must have compiled successfully before the bundle becomes durable.

**Why config-signature in the path:** a bundle keyed only by `(model, local_rank)` is shared by runs with *different* compile configs (different decode batch buckets, dtype, TP, `VLLM_RBLN_*` toggles, ...). torch's `save_cache_artifacts()` only serializes graphs freshly compiled in the current session; `load_cache_artifacts()` populates the caches but does **not** re-record loaded artifacts. So loading a stale bundle turns would-be-misses into un-re-recorded hits, and the next `save()` overwrites the file with a **partial** bundle — dropping graphs the run actually used (e.g. a hit `prefill`) and forcing recompiles on later runs. Keying the path by a config signature keeps each config's snapshot in its own file, so a config change never clobbers nor falsely reuses another config's bundle.

**What's added**
- `v1/worker/mega_cache.py`:
  - `config_signature(vllm_config)` = `vllm_config.compute_hash()` + the full env registry (vLLM + `VLLM_RBLN_*`, resolved values) + rebel **major.minor**. Errs toward over-inclusion: a spurious factor only costs a recompile, a missed one would poison a bundle. A patch-level rebel bump stays cache-compatible; a minor bump invalidates.
  - `bundle_path(model, sig)` inserts `sig` as a path component; `load`/`save` take `sig`.
  - rebel import made lazy (module now imports without a built rebel extension → unit-testable).
- `v1/worker/rbln_model_runner.py`: pass `mega_cache_only=True` to `torch.compile`, compute the signature once in `warm_up_model` and bracket it with `mega_cache.load/save`. No more permanent per-file `.rbln` under `VLLM_CACHE_ROOT/rbln/` — only the bundle.
- `tests/.../test_mega_cache.py`: version parsing (patch-stable / minor-invalidates), bundle-path signature isolation, and `config_signature` behavior against the real env registry (decode bucket strategy / manual buckets invalidate; identical config → identical signature).

**Depends on**
- rebellions-sw/rebel_compiler#11035 (`update(runtime): introduce mega cache`)

* Related to rebellions-sw/rebel_compiler#11035

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

### 🧪 How to Test

1. Cold start: `vllm serve <model>` → check that only `~/.cache/vllm/rbln/<model>-<hash>/<sig>/rank0/mega_cache.bin` exists (no orphan `*.rbln` next to it).
2. Warm restart (same config): rerun the same model → log shows `Loaded rbln mega-cache bundle from ...`, warm-up skips re-compile, bundle file is left untouched.
3. Config change: rerun with a different `VLLM_RBLN_DECODE_BATCH_BUCKET_STRATEGY` (or bucket set) → a new `<sig>/` directory is created; the previous config's bundle is preserved, not clobbered.
4. Failure injection: kill the worker mid warm-up → bundle file is unchanged (no partial persistence).
5. TP=N: confirm each rank writes to its own `rank{N}/` subdirectory.

### 📋 Checklist


rebel-jonghewk:
@rebel-eunji do you remember what problem we ran into when we cached the sampler (the reason it was disabled)? This PR will cache every torch.compiled artifact, including the sampler.

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,203 @@
...
+
+    if hasattr(vllm_envs, "compile_factors"):
+        try:
+            from vllm.config.utils import hash_factors
+
+            factors = vllm_envs.compile_factors()
+            for name in _PER_WORKER_ENV_DROP:
+                factors.pop(name, None)
+            return hash_factors(factors)
+        except Exception:  # pylint: disable=broad-exception-caught
```
rebel-yskim: **[Nit]** This `except Exception: pass` silently swallows a real failure in `compile_factors()` / `hash_factors()` and falls straight through to the weaker fallback signature. Consider logging what was swallowed (`logger.warning(...)`) so a genuine breakage in the primary path is visible instead of degrading silently into a different signature.

> yskim-review-bot

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,203 @@
...
+            pass
+    from vllm_rbln import rbln_envs
```
rebel-yskim: **[Nit]** The fallback only hashes `rbln_envs.environment_variables`, not vLLM's own compile-relevant env vars, so on this path a vLLM-level env change won't invalidate the signature (under-inclusion). That runs opposite to the over-inclusion philosophy stated in the PR description ("a missed factor poisons the bundle"). Worth confirming the old-vLLM fallback is actually needed — if not, dropping it removes the weaker path entirely.

> yskim-review-bot

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,203 @@
...
+    _coord_store_port), so the hash changes every launch. Neutralize every
+    "port"-named parallel field around the call — ports never affect the compiled
+    graph — so the signature is launch-stable."""
+    import dataclasses
+
+    pc = getattr(vllm_config, "parallel_config", None)
+    saved: dict = {}
+    if pc is not None and dataclasses.is_dataclass(pc):
+        for field in dataclasses.fields(pc):
+            if "port" in field.name.lower():
```
rebel-yskim, line 106: **[Nit]** Blanking every parallel-config field whose name contains `port` removes it from the signature — the opposite direction from the over-inclusion policy used elsewhere. It is safe today because ports never affect the compiled graph, but a future graph-relevant `*port*` field would silently collide distinct configs onto one bundle. A one-line comment nailing down that assumption would guard against a regression here.

> yskim-review-bot

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,203 @@
...
+
+def save(model: str, sig: str) -> None:
+    """Persist artifacts atomically. Call only after warm-up succeeds."""
+    if envs.VLLM_DISABLE_COMPILE_CACHE:
+        return
+    from rebel.core import mega_cache as rbln_mega_cache
+
+    path = bundle_path(model, sig)
+    try:
+        rbln_mega_cache.flush_to_bundle()
```
rebel-yskim, line 194: **[Nit]** `save()` depends on `load()` having called `rbln_mega_cache.set_dir(cache_root())` earlier — it never sets the dir itself. This holds today because `warm_up_model` always runs load→save, but the ordering coupling is implicit. Calling `set_dir(cache_root())` here too would make `save()` self-contained.

> yskim-review-bot

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,203 @@
...
+    """Hash of everything that changes the compiled graphs, keying the bundle.
+
+    vLLM config hash + vLLM's worker-aligned compile env factors + rebel
+    major.minor. Env factors drop per-worker vars (LOCAL_RANK, ports, ...) and
+    per-launch open ports are blanked, so all TP/DP ranks across launches share
+    one signature and the rank subdir isolates their shards.
+    """
+    cfg = _stable_compute_hash(vllm_config)
+    env = _compile_env_factors()
+    rebel = _rebel_major_minor()
```
rebel-yskim: **[Nit]** The local `rebel` shadows the module name and actually holds a version string, not the module. `rebel_ver` (or similar) reads clearer.

> yskim-review-bot

rebel-eunji:
It looks like samplers aren't covered by the mega cache. Is that intentional?

rebel-jonghewk:
> It looks like samplers aren't covered by the mega cache. Is that intentional?

Yes, that's intentional for the time being. Type1 will be affected by this, but I don't expect it to cause any problems for Type1 — could you confirm?

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,187 @@
...
+            version = ""
+    match = re.match(r"\s*(\d+)\.(\d+)", version)
+    return f"{match.group(1)}.{match.group(2)}" if match else "unknown"
+
+
+# Per-worker identity that must NOT enter the bundle signature: the rank subdir
+# separates workers and DP replicas compile identically, so the DP-degree lives
+# in parallel_config.compute_hash, not the per-worker index. vLLM's
+# compile_factors already drops LOCAL_RANK / ports but keeps VLLM_DP_RANK*.
+_PER_WORKER_ENV_DROP = frozenset({"VLLM_DP_RANK", "VLLM_DP_RANK_LOCAL"})
```
rebel-eunji, line 59: nit: I think a few more env vars could be excluded from hash generation.
- VLLM_RBLN_ENABLE_WARM_UP
- VLLM_RBLN_METRICS, VLLM_RBLN_METRICS_FILE
- VLLM_RBLN_NUMA
- VLLM_RBLN_PROFILER

rebel-jonghewk: Thanks. I'll apply it.

rebel-eunji:
I checked mega-cache with type1, and it doesn't work well.

In optimum_model_runner, the sampler is compiled with both the default backend and the rbln backend. The default backend records new `aot_autograd` and `inductor` artifacts on every run. This makes `save_cache_artifacts()` return non-None, so `mega_cache.save()` runs even on a warm start.

The problem is that the rbln ops are served from the cache (cache hit) and are not re-recorded, so they are excluded from `save_cache_artifacts()`. As a result, the existing mega-cache file is overwritten with a new one that holds only the default-backend artifacts and no rbln ops. The mega-cache file ends up with no rbln ops — effectively a dummy, and useless.

rebel-jonghewk:
> I checked mega-cache with type1, and it doesn't work well.
> 
> In optimum_model_runner, the sampler is compiled with both the default backend and the rbln backend. The default backend records new `aot_autograd` and `inductor` artifacts on every run. This makes `save_cache_artifacts()` return non-None, so `mega_cache.save()` runs even on a warm start.
> 
> The problem is that the rbln ops are served from the cache (cache hit) and are not re-recorded, so they are excluded from `save_cache_artifacts()`. As a result, the existing mega-cache file is overwritten with a new one that holds only the default-backend artifacts and no rbln ops. The mega-cache file ends up with no rbln ops — effectively a dummy, and useless.

@rebel-eunji So for now leave it like this? (exclude sampler)

rebel-eunji:
@rebel-jonghewk Yes, I think sampler caching would be better handled in a follow-up PR.
There is a minor issue besides the ones mentioned above.
In rbln_model_runner, warm-up currently only happens with greedy sampling, not with random sampling.

rebel-eunji (approved):
lgtm

rebel-yskim (approved):
LGTM

PR #623, opened 2026-05-26
fix(sampler): guard padded-row sampling tensors against torch.empty garbage

rebel-eunji:
## Summary
- `RBLNInputBatch` only filled `top_k` and `temperature` defaults — `top_p`, `presence`, `frequency`, `repetition_penalties` were left as raw `torch.empty()` allocations. When `num_reqs < bucket_size`, pad rows of those tensors carried uninitialized garbage that propagated through penalty / top_k / top_p ops, producing NaN logits and out-of-vocab (=`vocab_size`) sampled tokens which then triggered out-of-bounds `IndexError` in downstream `torch.gather` (logprobs).
- Add `test_no_nan_logits_with_padded_bucket` that monkey-patches `torch.empty` during runner construction to return NaN-filled float tensors — the test now depends on the init guards rather than lucky allocator state.
- Extend the sampler CI step to run tests not carrying the `warm_up` parametrize id (previously filtered out by `-k "warm_upTrue/False"`).

## Test plan
- [x] `pytest tests/v1/worker/test_sampler.py::test_no_nan_logits_with_padded_bucket` — 16/16 combos pass with guards in place.
- [x] Verified the test correctly fails when any one of the new guards (`top_p`, `repetition_penalties`, `presence_penalties`, `frequency_penalties`) is removed.
- [x] PR CI sampler job picks up the new test (via the added third `pytest -k` line).

### DIFF
--- a/.github/workflows/rbln_vllm-rbln_pytest.yaml
+++ b/.github/workflows/rbln_vllm-rbln_pytest.yaml
@@ -330,6 +330,7 @@ jobs:
             "sampler")
               pytest -n auto -k "warm_upTrue" tests/v1/worker/test_sampler.py -vv --durations 0
               pytest -n auto -k "warm_upFalse" tests/v1/worker/test_sampler.py -vv --durations 0
+              pytest -n auto -k "not warm_upTrue and not warm_upFalse" tests/v1/worker/test_sampler.py -vv --durations 0
               ;;
           esac
 
--- a/.github/workflows/rbln_vllm-rbln_pytest.yaml
+++ b/.github/workflows/rbln_vllm-rbln_pytest.yaml
@@ -328,9 +328,9 @@ jobs:
               pytest tests/entrypoints/openai -vv --durations 0
               ;;
             "sampler")
-              pytest -n auto -m "param_sweep and warm_up" tests/v1/worker/test_sampler.py -vv --durations 0
-              pytest -n auto -m "param_sweep and not warm_up" tests/v1/worker/test_sampler.py -vv --durations 0
-              pytest -n auto -m "not param_sweep" tests/v1/worker/test_sampler.py -vv --durations 0
+              pytest -n auto -k "warm_up_true" tests/v1/worker/test_sampler.py -vv --durations 0
+              pytest -n auto -k "warm_up_false" tests/v1/worker/test_sampler.py -vv --durations 0
+              pytest -n auto -k "not warm_up_true and not warm_up_false" tests/v1/worker/test_sampler.py -vv --durations 0
               ;;
           esac
 
--- a/.github/workflows/rbln_vllm-rbln_pytest.yaml
+++ b/.github/workflows/rbln_vllm-rbln_pytest.yaml
@@ -205,7 +205,7 @@ jobs:
     strategy:
       fail-fast: ${{ inputs.fail_fast }}
       matrix:
-        test_type: 
+        test_type:
           [
             {
               name : "openai",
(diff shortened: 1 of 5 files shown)


rebel-eunji:
It seems like https://github.com/RBLN-SW/vllm-rbln/pull/621 resolves the error that this PR is resolving. I'll take a look..

vllm_rbln/v1/worker/optimum_input_batch.py
```diff
@@ -43,13 +43,27 @@ def __init__(self, *args, **kwargs):
...
+            #
+            # Why: apply_penalties does `logits -= penalty * mask` and
+            # `logits = logits / repetition_penalty` over the full padded
+            # batch. If a pad row's penalty is NaN (from torch.empty), then
+            # NaN * 0 = NaN contaminates every position; if repetition is 0,
+            # the division yields inf. Either way the pad row leaves
+            # apply_penalties unusable for the rest of the sampler pipeline.
+            self.frequency_penalties_cpu_tensor.fill_(0.0)
+            self.presence_penalties_cpu_tensor.fill_(0.0)
+            self.repetition_penalties_cpu_tensor.fill_(1.0)
```
rebel-jonghewk, line 67: Should `self.frequency_penalties`, `self.presence_penalties`, and `self.repetition_penalties` also be filled in?

.github/workflows/rbln_vllm-rbln_pytest.yaml
```diff
@@ -330,6 +330,7 @@ jobs:
             "sampler")
               pytest -n auto -k "warm_upTrue" tests/v1/worker/test_sampler.py -vv --durations 0
               pytest -n auto -k "warm_upFalse" tests/v1/worker/test_sampler.py -vv --durations 0
+              pytest -n auto -k "not warm_upTrue and not warm_upFalse" tests/v1/worker/test_sampler.py -vv --durations 0
```
rebel-jonghewk: instead of this can it be replaced with  pytest marker?

vllm_rbln/v1/worker/optimum_input_batch.py
rebel-eunji: They are generated with torch.empty. (https://github.com/vllm-project/vllm/blob/bcf2be96120005e9aea171927f85055a6a5c0cf6/vllm/v1/worker/gpu_input_batch.py#L184)
This can generate a tensor with dummy values, which is then used as an operand as shown below, resulting in NaN logits.
So this PR fills the tensors with default value
```
# Repetition penalty
if token appeared in prompt or output:
    if logits[token] > 0:
        logits[token] /= penalty
    else:
        logits[token] *= penalty
```

```
# Frequency Penalty
logits -= frequency_penalties.unsqueeze(dim=1) * output_bin_counts
```

```
# Presence Penalty
logits -= presence_penalties.unsqueeze(dim=1) * output_mask
```

rebel-eunji: So they are all required to prevent nan logits.

.github/workflows/rbln_vllm-rbln_pytest.yaml
rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/623/changes/02b82ac716db3f1fd1b14eba7e105c539bedf386

pyproject.toml
```diff
@@ -76,6 +76,16 @@ rbln = "vllm_rbln:register"
...
+[tool.pytest.ini_options]
+markers = [
+    "bucketing: sampler bucket-size computation (test_get_bucket_sizes)",
+    "sampler_mode: sampler mode / structured-output forward path",
+    "param_sweep: full sampling-parameter combination sweep (large)",
+    "warm_up: param_sweep cases running with VLLM_RBLN_ENABLE_WARM_UP=True",
+    "nan_guard: padded-bucket NaN / out-of-vocab guard",
+    "recompile_guard: sampler torch.compile recompile guard",
+]
+
```
rebel-eunji: When using pytest markers, it's recommended to register them under [tool.pytest.ini_options]. Otherwise, pytest emits a warning.

@rebel-jonghewk @rebel-seinpark

vllm_rbln/v1/worker/optimum_input_batch.py
rebel-jonghewk: No I mean `repetition_penalties_cpu_tensor` is filled but `repetition_penalties` doesn't seem to be filled here.

pyproject.toml
rebel-eunji: I think removing the markers from pyproject.toml looks better. What do you think?

vllm_rbln/v1/worker/optimum_input_batch.py
rebel-eunji: I see. `repetition_penalties_cpu_tensor` are sliced and copied into `repetition_penalties`, so it shouldn't cause any real issue. Still, I'll initialize `repetition_penalties` with a default value for robustness.

pyproject.toml
rebel-jonghewk: got it. so.. remove the markers entirely (decorators and registration) and select with -k ?

rebel-eunji: Yes, let’s revisit this issue later.

rebel-jonghewk (approved):
lgtm

PR #625, opened 2026-05-27
other(platform): default max_num_seqs to 1

rebel-eunji:
### 🚀 Summary of Changes

vLLM upstream has an [`apply_config_platform_defaults`](https://github.com/vllm-project/vllm/blob/683033d4ba60a2861ddcc5df9bb9d2e6ca7c935c/vllm/platforms/interface.py#L423) class method, but it has a major flaw — there's no way to tell apart the default 256 and a user-specified 256. To work around this, I added a new method that gets called in `pre_register_and_update`.

- Reference
    - https://github.com/vllm-project/vllm/blob/683033d4ba60a2861ddcc5df9bb9d2e6ca7c935c/vllm/engine/arg_utils.py#L2282

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/entrypoints/llm/test_default_max_num_seqs.py
+++ b/tests/entrypoints/llm/test_default_max_num_seqs.py
@@ -14,7 +14,7 @@
 
 """Offline LLM(...) coverage for the RBLN max_num_seqs default.
 
-`a860d0d6` makes an unset max_num_seqs resolve to 1 on RBLN (upstream vLLM
+An unset max_num_seqs resolve to 1 on RBLN (upstream vLLM
 defaults to 256), while leaving an explicitly-set value untouched. This module
 exercises the `LLM(...)` (UsageContext.LLM_CLASS) entry point; the `vllm serve`
 counterpart lives in tests/entrypoints/openai/test_default_max_num_seqs.py. Both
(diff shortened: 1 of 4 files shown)


rebel-jonghewk:
At minimum, I'd recommend unit tests covering:
  - Unset max_num_seqs resolves to 1 for both LLM(...) and serve
  - An explicitly set value (e.g. 4) is preserved
  - Idempotency — calling it twice still wraps the original only once

rebel-eunji:
@rebel-seinpark
I added some new pytests. They're pretty lightweight, so I just ran them against the actual compiled model to test the entrypoints.

tests/entrypoints/llm/test_default_max_num_seqs.py
```diff
@@ -0,0 +1,101 @@
...
+    from vllm_rbln.platform import RblnPlatform
+
+    # Apply the override exactly as pre_register_and_update does in production.
+    RblnPlatform._override_default_max_num_seqs()
+
+    # model is never loaded (no create_model_config call); max_num_batched_tokens
+    # is pinned so the stub model_config is never dereferenced.
+    engine_args = EngineArgs(
+        model=MODEL_NAME,
+        max_num_seqs=ORIGINAL_MAX_NUM_SEQS,
```
rebel-jonghewk, line 106: how about adding case when max_num_seqs is None

rebel-jonghewk (approved):

tests/entrypoints/llm/test_default_max_num_seqs.py
rebel-eunji: I've added the test in LLM (offline)
https://github.com/RBLN-SW/vllm-rbln/pull/625/changes/f2b66506fe627c1f332a1d2b630ff4ecbb9b1a07
And in vllm serve (online) , max_num_seqs=none is not allowed.

junstar92:
@rebel-eunji 
Just to clarify, is the main reason for setting the default `max_num_seqs` to `1` to distinguish an unset value from an explicitly provided value?

I was thinking that vLLM already prints non-default args, so if a user explicitly sets `max_num_seqs`, we should be able to see it there. Is there another case this default override is intended to cover?

PR #630, opened 2026-05-29
fix(test): replace exact match test in spec decode by logprob similarity test

rebel-jinhwan:
### 🚀 Summary of Changes

Replace the exact-match correctness test for speculative decoding with a
noise-aware, **every-position** check, and parametrize it over batch sizes
**1 / 8 / 16**.

Greedy spec decode can flip the argmax at a bf16 **near-tie**, producing a
different but equally valid continuation, so exact text/token equality is too
strict. We instead teacher-force the spec-generated tokens back through the
base model and verify, at every position, that each spec token is the base
model's greedy choice (or a near-tie within a small logprob gap).

---

## The Phenomenon

Prompt `The capital of France is` (greedy) diverges between base and spec
decoding one token after `" Paris"`:

* **Base:** `" Paris. The city is located in the"`
* **Spec:** `" Paris, and it is the most visited"`

Raw logits at the branch point:

| Token | Base (decode) | Spec (verify) |
| --- | --- | --- |
| `.` | **23.625** | 23.5 |
| `,` | 23.5 | 23.5 |
| others (top-k) | identical | identical |

Base picks `.` (23.625 > 23.5). In spec, `.` rounds down one step to 23.5, ties
with `,`, and the argmax tie-break (lower index) picks `,`.

## Root Cause: bf16 Rounding Noise (Not a Bug)

`Paris` is causal, so its logit should be identical in both paths. But base
(batch 1) and verify (batch N) run **different compiled kernels**, and bf16
addition is non-associative — a different accumulation order rounds the same
value to an adjacent bf16 grid point. At magnitude ~23 the bf16 step (ULP) is
exactly `0.125`, which is the observed shift. Only `.` moved (by 1 ULP) while
every other top-k logit stayed bit-identical — the fingerprint of rounding, not
a logic bug. 

This is not a quality regression: `.` and `,` are equally valid continuations.

---

## Verification: `assert_spec_matches_base_within_noise`

Exact match is too strict, and comparing only "up to the first divergence" is
blind to anything after the cascade. So:

1. Generate the sequence with spec decode.
2. **Teacher-force** it through the base model: for each position `i`, feed
   `prompt + spec[:i]` and read the base model's greedy choice + logprobs there
   (all prefixes in one batched `generate`). Base and spec are now aligned on
   the same trajectory at **every** position.
3. A position passes if the spec token is the base argmax, or its logprob gap
   from the argmax is **≤ 0.2** (a near-tie). Any larger gap fails as a real
   regression — reporting every offending position.

Notes:
* The gap is taken **within the base distribution** at one position, so the
  softmax `logsumexp` cancels and `logprob_gap == raw logit_gap`.
* Tolerance `0.2` is set from data: the largest gap observed across all
  Medusa/Eagle **batch 1/8/16** runs was `0.125` (= 1 bf16 ULP, eagle3 only;
  Medusa/eagle had zero divergence). `0.2` sits above 1 ULP and below the next
  step (`0.25`); a real bug diverges by ≥ ~1 nat.
* Teacher forcing uses the base **prefill** kernel (itself ~1 ULP off the verify
  kernel), so the tolerance is still required. We use generation logprobs
  because RBLN's `prompt_logprobs` path is currently broken (worth a separate
  fix).

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

### 🧪 How to Test

```
pytest -s -v tests/torch_compile/e2e/v1/spec_decode
```

Expected: 9 cases pass — `test_medusa_matches_base_generation[1/8/16]` and
`test_eagle_matches_base_generation[eagle-8 | eagle3-32][1/8/16]`.

### 📋 Checklist

### 💬 Notes

RBLN `prompt_logprobs` is broken (surfaced while building this test); worked
around here, worth fixing separately.

### DIFF
--- a/tests/torch_compile/e2e/v1/spec_decode/test_eagle_correctness.py
+++ b/tests/torch_compile/e2e/v1/spec_decode/test_eagle_correctness.py
@@ -22,6 +22,7 @@ import pytest
 from vllm import LLM, SamplingParams
 
 from .utils import (
+    assert_spec_matches_base_within_noise,
     ensure_vllm_compatible_eagle_draft_model,
     get_default_eagle_test_model_ids,
 )
@@ -79,6 +80,9 @@ def test_eagle_matches_base_generation(method: str, max_tokens: int) -> None:
         top_p=1.0,
         max_tokens=max_tokens,
         ignore_eos=True,
+        # Needed for the bf16 near-tie check in
+        # assert_spec_matches_base_within_noise.
+        logprobs=20,
     )
 
     base_llm = _build_base_llm(method)
@@ -100,8 +104,13 @@ def test_eagle_matches_base_generation(method: str, max_tokens: int) -> None:
                 base_output.outputs[0].finish_reason
                 == eagle_output.outputs[0].finish_reason
             )
-            assert base_output.outputs[0].text == eagle_output.outputs[0].text
-            assert base_output.outputs[0].token_ids == eagle_output.outputs[0].token_ids
+            # Greedy spec decode may flip a near-tie argmax due to bf16 ULP
+            # differences between the base decode and EAGLE verify kernels.
+            # Require a match up to the first divergence and that the
+            # divergence (if any) is genuine floating-point noise.
+            assert_spec_matches_base_within_noise(
+                base_output.outputs[0], eagle_output.outputs[0]
+            )
     finally:
         del eagle_llm
         gc.collect()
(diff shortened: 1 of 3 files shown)


rebel-jaehwang (approved):

rebel-wonsubkim (approved):
LGTM
exact match test --> logprob similiary test

PR #646, opened 2026-06-05
feature(warmup): offload host regions during device_tensor warm-up

rebel-jonghewk:
### 🚀 Summary of Changes

Enable RBLN file offloading around the model-load and warm-up phases when `VLLM_RBLN_USE_DEVICE_TENSOR` is set, so host-resident regions backing RBLN tensors are paged to disk and host RAM pressure drops.

- **1. load_model offload:** wrap `model_loader.load_model()` / `load_weights()` in `torch.rbln.offload()`. Weight host backings are written to file at load time instead of staying resident.
- **2. warm_up offload:** wrap `_warm_up_model_inner()` in `torch.rbln.offload()`. Keeps offloading active through compile (which fires on the first dummy run) so the just-offloaded weights are not migrated back to host.
- **Guard:** both use `VLLM_RBLN_USE_DEVICE_TENSOR and has_torch_rbln`, else `nullcontext()` — no behavior change in cpu mode or when `torch-rbln` is absent.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 💬 Notes

Two one-block changes in `rbln_model_runner.py` (`load_model`, `warm_up_model`). Depends on the rebel-compiler runtime path being offload-aware for copies that read offloaded weight regions during compile.


rebel-eunji (approved):
lgtm 👍

PR #648, opened 2026-06-05
fix(attn): normalize seq_lens to int32 so custom kernel compiles

rebel-jinhwan:
### 🚀 Summary of Changes

PR #616 removed the in-graph dtype casts on attention custom op args. Since `positions` are int64, `seq_idx` (and the `seq_lens_tensor` / `dyn_size_for_partitions` derived from it) became int64. With RBLN_USE_CUSTOM_KERNEL=1 attention lowers `seq_lens` through the triton kernels' `rtosa.to_dynamic_index`, which rejects int64 operands and fails compilation:
```
  flash_causal_attention.py: 'rtosa.to_dynamic_index' op operand #0 must be
  ... 32-bit signless integer ... but got 'tensor<1x4xi64>'
```
`block_tables` is natively int32 so it passed; only the int64 `seq_lens` broke, and only on the custom-kernel path (rbln_custom_ops accepts int64).

Normalize `seq_idx` to int32 in the metadata builder. This runs eagerly during metadata construction, so the dtype is correct before the tensors enter the compiled graph -- honoring #616's "no in-graph casts" intent while giving the triton kernels a valid dtype. int32 covers the full position range (unlike the original int16, which overflows at max_model_len >= 32768).

Verified on NPU: both RBLN_USE_CUSTOM_KERNEL=0 and =1 compile and generate correctly with identical output.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/v1/attention/backends/flash_attention.py
+++ b/vllm_rbln/v1/attention/backends/flash_attention.py
@@ -1129,6 +1129,9 @@ class RBLNFlashAttentionMetadataBuilder(
             num_computed_tokens = seq_lens - query_seq_lens
             seq_idx = positions[query_start_loc[:num_reqs]].view(-1, 1)
 
+        # custom (triton) kernel's to_dynamic_index rejects int64 seq_lens
+        seq_idx = seq_idx.to(torch.int32)
+
         cu_prefix_query_lens = None
         prefix_kv_lens = None
         suffix_kv_lens = None


rebel-jaehwang (approved):

huijjj:
It looks like you forgot to adjust the inputs for the SWA kernels, see https://github.com/RBLN-SW/vllm-rbln/commit/9610cdf5957d9f9bc0b819748dd8ca2ca6314cb3

PR #604, opened 2026-05-14
feature(spec_dec): implement spec decode backfill for fixed length drafts (full_spec only)

rebel-wonsubkim:
### 🚀 Summary of Changes

> *What does this PR do? What feature, fix, or improvement does it bring?*
* resolves kv cache block constraint in speculative decoding
   * attention kernel cannot handle multi-block kv cache update
   * since speculative decoding can have multiple query length for drafts, above problem can happen
   * final solution - change attention kernel to remove constraint
   * intermediate solution1 - change attention kernel to invalidate update beyond current kv cache block
   * intermediate solution2 - vllm workaround to prevent following situation by adjusting spec decoding query length
* current solution is intermediate solution2
   * adjust actual number of drafts by keeping full_spec via sliding window
   * if full_spec is across kv cache block boundary, sliding window is triggered to adjust number of drafts
---

### 📌 Related Issues / Tickets

* Resolves #
   * https://github.com/rebellions-sw/fsw-inference/issues/255
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #649, opened 2026-06-06
feature: cross-block no-spec fallback for variable-length spec decoding proposers

rebel-wonsubkim:
### 🚀 Summary of Changes

  Motivation

  Query backfill (#604) keeps every decode step at a single (batch, num_spec+1) shape by prepending past tokens. But variable-length proposers 
  (ngram/suffix) can need backfill that crosses a KV block boundary: right after entering a new block, a short draft would pull past tokens from the
  previous block, which the single-block decode path can't express → silent KV corruption. (Fixed-length proposers never hit this — num_spec+1 <= 
  block_size keeps their backfill in-block.)

  Key idea
  
  - Detect cross-block in the scheduler (desired_slide > tokens_used_in_block) and fall back to no-spec (query_len=1) for that step.
  - Cross-DP collective: any rank tripping it → every rank drops to no-spec via the step_no_spec_required OR-reduce, so all ranks keep the same decode
  shape.
  - Compile the no-spec graph in warmup. A decode graph's compile key has two axes — input_ids shape and max_pads_across_dp size (num_padded). The warmup
  dummy drives the same runtime path (RBLNSchedulerOutput(step_no_spec_required=True)) so num_padded matches too; setting query_len=1 alone would still
  hot-path recompile at runtime.
  - Two runtime asserts guard the invariant: no backfill window may cross a block boundary, and the no-spec scrub must clear all slide + force
  query_len=1.
  
  Verification

  Full model (MiniMax-M2.5, DP4/EP4, ngram num_spec=3), isl=512/osl=2048: cross-block no-spec fired 322× organically, with 0 recompile / 0 hot-path / 0 
  assert, outputs coherent.
  
  ▎ Full motivation & key idea are kept in the rbln_scheduler.py module docstring.

host tensor - 검증 완료
```
RBLN_WEIGHT_FREE=0 VLLM_RBLN_USE_DEVICE_TENSOR=0 VLLM_RBLN_SAMPLER=0 \
vllm serve MiniMaxAI/MiniMax-M2.5 --port 8000 --data-parallel-size 4 --enable-expert-parallel \
--max-model-len 196608 --block-size 1024 --enable-chunked-prefill \
--max-num-batched-tokens 512 --max-num-seqs 8 --gpu-memory-utilization 0.8 --trust-remote-code \
--speculative-config '{"method": "ngram", "num_speculative_tokens": 3, "prompt_lookup_max": 5, "prompt_lookup_min": 2}'`
```
device tensor - 검증 중
```
RBLN_WEIGHT_FREE=1 VLLM_RBLN_USE_DEVICE_TENSOR=1 VLLM_RBLN_SAMPLER=0 \
vllm serve MiniMaxAI/MiniMax-M2.5 --port 8000 --data-parallel-size 4 --enable-expert-parallel \
--max-model-len 196608 --block-size 1024 --enable-chunked-prefill \
--max-num-batched-tokens 512 --max-num-seqs 8 --gpu-memory-utilization 0.8 --trust-remote-code \
--speculative-config '{"method": "ngram", "num_speculative_tokens": 3, "prompt_lookup_max": 5, "prompt_lookup_min": 2}'`
```

```
vllm bench serve \
    --model MiniMaxAI/MiniMax-M2.5 --backend vllm \
    --base-url http://127.0.0.1:8000 --endpoint /v1/completions \
    --dataset-name random --random-input-len 512 --random-output-len 2048 \
    --num-prompts 320 --max-concurrency 32 --request-rate 4 \
    --percentile-metrics ttft,tpot,itl,e2el --temperature 0
```

---

* Resolves #
* Related to #604

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-jinhwan (approved):
LGTM

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -893,8 +1008,40 @@ def schedule(self) -> RBLNSchedulerOutput:
...
+                    # block above, so apply the SAME sliding-window backfill via
+                    # the shared `_decide_spec_slide`. With spec decode on, every
+                    # decode query window is unified to num_spec_tokens+1; a
+                    # freshly promoted req carries 0 drafts, so without this
+                    # backfill its window stays length 1 while running peers run
+                    # at num_spec+1 and the runner would zero-pad it (garbage KV)
+                    # instead of re-feeding real past tokens. num_new_tokens == 1
+                    # (asserted) so the advance is never boundary-trimmed
+                    # (new_n == 1).
+                    if not is_prefill(request) and self.num_spec_tokens > 0:
```
rebel-ykchoi: The promotion path assumes decode-ready state. If that assumption is violated, we should fail loudly before the request is emitted as a decode-batch member. `assert not is_prefill` is the right primary guard; if not is_prefill as a backfill-only filter is too weak and can mask a much worse failure mode.

rebel-ykchoi (changes requested):

rebel-ykchoi (approved):
Thanks.

PR #651, opened 2026-06-08
refactor: reorganize and clean up the RBLN plugin structure

junstar92:
### 🚀 Summary of Changes

Over time, RBLN-specific logic has been added inline throughout the upstream execution flow. This has made the code harder to follow, blurred the line between what is reused from upstream, what is patched, and what is forked, and made maintenance difficult. On top of that, closely following the upstream structure has meant a high review-and-fix cost every time we sync to a new upstream version.

This PR addresses that structural debt: it pulls RBLN-specific logic out of the upstream flow, gives each component a clear responsibility, and reorganizes the whole plugin around the way RBLN actually runs. External behavior stays the same as `dev`.

- Make the boundaries between reuse, patch, RBLN-specific policy, and the upstream contract explicit in the code, so the intent and scope of each change are easy to see
- Replace monkey-patching with a declarative patch registry / descriptor layer, and remove patches that relied on import side effects
- Split the `set_forward_context` path and the scheduler's RBLN-specific policy into a dedicated implementation and separate methods
- Rebuild `rbln_model_runner.py` as an RBLN-owned runner instead of a fork of the GPU runner (redefined input-prep -> forward -> sampling pipeline, plus reusable device-tensor buffers)
- Align compilation, attention, and quantization with vLLM's native interfaces and drop unused fields/code, lowering the cost of upstream syncs

#### Add a patch registry / descriptor layer and remove import-side-effect patches

- New `vllm_rbln/patches/`. Patches are registered declaratively with `@register_patch(...)`, recoding what is patched, why, and under which constraint. Registration (import) and application (setattr) are separated(`apply_registrations()` -> `apply_registered_patches()`), with support for ordering, conditions, and verification. Implementations live next to their matching upstream vLLM paths, but the actual application is driven by the registry.
- Removed the old "apply patches by importing modules" approach in `register_ops()`. Only the unavoidable imports remain; the other lines are gone and now go through the registry.
- Runtime assignment (`FusedMoE.__init__ = ...`) is replaced with proper subclasses + OOT registration, and patch files are moved into `patches/`.

#### Clean up the forward-context path

- The `set_forward_context` path did not actually need monkey-patching, so it is split into a dedicated `vllm-rbln` implementation instead of an upstream patch (it is only used in the real model call and already behaves differently from upstream).

#### Separate the scheduler's RBLN-specific policy

- `schedule()` keeps the upstream flow, but RBLN-specific policy that used to be scattered as inline conditions is pulled into separate methods, making diffs easier to read and unit tests easier to write.

#### Rebuild the model runner as RBLN-owned

Removes the upstream-GPU scaffolding and redefines the **input-prep -> forward -> sampling pipeline** around RBLN run flow.

(a) Input preparation / padding

- `_prepare_inputs` is greatly reduced (~448 -> ~125 lines): inline sliding-window backfill and mrope position computation are removed, and it now focuses on computing `logits_indices` / `spec_decode_metadata` / `query_lengths` and filling positions via `index_select`.
- Attention metadata construction is pulled out into a separate `_build_attention_metadata`.
- Batch/token padding that used to be inline right before forward (`rbln_utils.pad(...)` + manual `.to(device)`) is split: the padding size is decided in `_determine_batch_padding` (bucketing + DP metadata), and the actual padding and device copy are delegated to `InputStager`.
- Sampling metadata is now clearly handled as `pad -> sampler -> depad` (`_pad_sampling_metadata` / `_depad_sampler_output`).

(b) Device-tensor memory buffer (new `input_stager.py`)

- Replaces the boilerplate of allocating fresh input tensors on every forward (`view`/`to(long)` -> `pad()` (new alloc) -> manual `.to(device)`) with pre-allocated, reusable buffers.
- Made up of `InputLayout` (describes the target padded shape), `InputBuffer` (holds reusable device tensors), `StagedModelInputs` (`.as_kwargs()` expands the model call arguments), and `InputStager` (caches buffers by shape/dtype, does `fill_(pad)` and then `copy_(non_blocking=True)` for the real region only).

(c) Simplified inference / execute path

- `execute_model` = `_update_states` -> `_process_kv_cache_copy_ops` -> `_prepare_inputs` -> `_determine_batch_padding` -> `_build_attention_metadata` -> `_preprocess(stage)` -> forward -> postprocess, with sampling split out into `sample_tokens`
- Removed the in-forward LoRA machinery (TODO feature)
- Removed the async-scheduling scaffolding (unsupported currently)
- Replaced the `rebel.capture_reports()` + `StepReport` plumbing with a `performance_ctx.profile(...)_` context for metrics.

(d) Removed unused code/fields

- Module symbols/classes, dummy-run methods, vLLM interfaces (from the GPU runner) that are not used right now

#### Consolidate the compilation path

- Moved `torch.compile` logic into `vllm_rbln/compilation/` (`compile.py`, `backends.py`). The `process_group_dict` setup that was copy-pasted across several files is unified into `build_process_group_dict()`, and eagle/medusa/sampler/rejection_sampler now go through this shared entry point. Deleted `torch_compile_backend.py`, `_torch_dynamo_compat.py`, and `_torch_accelerator_compat.py`.

#### Clean up the attention backend/kernels

- Split the custom ops out of `flash_attention.py` into `v1/attention/ops/*_naive.py` front-ends, and moved `triton_kernels/*` into `v1/attention/ops/triton_*`. `flash_attention.py` now keeps only the backend layer.
- Removed unused fields and cleaned up: trimmed the GPU-runner-only fields from `RBLNFlashAttentionMetadata`, cutting it from 22 fields to 10.

#### Align quantization with vLLM's native registration interfaces

- Unless patching is truly unavoidable, use **vLLM's native registration interface** instead of monkey-patching (centralized in `patches/oot.py`):
  - `base_cls.register_oot(rbln_cls)` to register `FusedMoE` / `UnquantizedFusedMoEMethod` / `VocabParallelEmbedding` / `ParallelLMHead`.
  - Kernel selection now inserts the RBLN kernels into `linear._POSSIBLE_FP8_BLOCK_KERNELS[PlatformEnum.OOT]` / `_POSSIBLE_KERNELS[PlatformEnum.OOT]` instead of patching the selection function.
- Reduced custom implementations: removed `Fp8LinearMethod` / block-FP8 op from `fp8.py` and moved them into `kernel/linear/block_fp8.py`, simplified `mxfp4.py` to inherit from upstream `GptOssMxfp4MoEMethod`, and moved the `mixed_precision` kernel into `kernels/linear/`.

#### Other cleanup

- Consolidated the rejection sampler into `rbln_rejection_sampler.py` (Torch/RBLN strategy impls) and deleted `cpu_rejection_sampler.py`.
- Removed dead code (`logits_processor.py`).
- Reorganized `rbln_envs.py` -> `envs.py`, and simplified `logger.py` to reuse upstream `vllm.logger`.

### Behavior Impact

- **No behavior change**. Runtime behavior is the same as `dev`.
- Because patches are applied and verified through the registry, the old dependence on import order is gone.
- Some vLLM interfaces (from the GPU runner) are not carried over yet, so when adding features later, check whether that interface needs to be supported first.

### Not Yet Reflected

Features

- [ ] DeepSeek V3
  - [x] W8A16 FP8
  - [ ] W8A8 FP8
  - [x] MTP
- [ ] MiniMax FP8 W8A8 

Dev PRs

- [ ] #624
- [ ] #656
- [ ] #681
- [ ] #684
- [ ] #690
- [ ] #705
- [x] #731
- [ ] #754
- [ ] #758
- [x] #761
- [x] #765
- [x] #766
- [x] #772
- [x] #769
- [x] #695 
- [x] #643
- [x] #596
- [x] #749 
- [ ] #759
- [ ] #775
- [ ] #778
- [ ] #780 
- [x] #781 

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 Testing & Follow-up

Tests were rewritten mostly around the model runner, but coverage is still thin overall. There are no unit tests for attention or quantization, and some existing tests were removed due to the structural changes. As a follow-up, we're going to stronger feature tests and regression tests.

---

---

### 📋 Checklist

---

#### Motivation

`vllm-rbln` is a vLLM plugin that registers the RBLN platform, models, ops, and patches. Its vLLM-model mode has grown by keeping much of upstream vLLM's model / worker / runner structure and wiring RBLN-specific behavior in through patches, forks, overrides, and helpers.

The core issue is not that there is a lot of RBLN-specific code. It is that **RBLN's execution model does not match the assumptions behind the upstream GPU runner**. The GPU runner assumes a GPU runtime, CUDA graphs, and fairly dynamic runtime paths. The RBLN backend, on the other hand, works under compiler-driven constraints: static shapes, limited mixed batching, padding/bucketing, and so on. The harder we try to keep the GPU runner's method structure and control flow, the more RBLN-specific logic gets scattered as conditionals inside upstream methods. As a result, it becomes hard to tell which code is upstream reuse vs. a patch vs. a fork, RBLN-specific policy is hard to test on its own, the points that actually need review during an upstream sync are not visible, and it is hard to separate old workarounds from code that is still needed.

---


rebel-jonghewk:
@rebel-cjlee Could you merge dev into this branch to resolve the conflicts? It'd make it easier for me to review just the actual changes.

rebel-cjlee:
@rebel-jonghewk Thanks for checking.
This PR has massive changes everywhere in `vllm-rbln`. A standard merge to `dev` is too difficult for this. Instead, we need to apply the changes directly to the branch (for example, by using a rebase and a force-push).
Reviewing the diffs will be really hard, so you don't need to check every single line. It would be great if you could just review the overall flow and logic around the RBLN model runner.

PR #653, opened 2026-06-08
fix(cache): include toolchain versions in optimum compiled model cache key

rebel-eunji:
### 🚀 Summary of Changes

`_generate_model_path_name` computes the cache key (directory name under compiled_models/) for a compiled model. Until now it hashed only user-provided parameters (model, batch size, block size, max len, TP size, rbln_config), but not the versions of the toolchain that actually produces the binary.

This meant that after upgrading `rebel-compiler` or `optimum-rbln`, the same key could resolve to a stale cache entry compiled by a different toolchain — a silent cache collision. This was already flagged by a FIXME in the code.

This PR folds the installed `rebel-compiler` / `optimum-rbln` versions into the hash, so a toolchain upgrade naturally invalidates the cache and triggers a recompile.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/utils/optimum/converter/dispatch.py
+++ b/vllm_rbln/utils/optimum/converter/dispatch.py
@@ -15,6 +15,7 @@
 import hashlib
 import json
 import os
+from importlib.metadata import PackageNotFoundError, version
 from typing import TYPE_CHECKING
 
 import vllm_rbln.rbln_envs as envs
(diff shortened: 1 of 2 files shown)


vllm_rbln/utils/optimum/converter/dispatch.py
```diff
@@ -53,6 +54,31 @@ def _strip_runtime_only_keys(obj: dict) -> dict:
     return obj
 
 
+# Packages whose version changes the compiled binary, so a version bump must
+# invalidate the cache to avoid loading an artifact built by a different
+# toolchain.
```
rebel-jonghewk, line 59: Why is recompilation necessary here? Different versions may still be compatible.

PR #662, opened 2026-06-09
fix(scheduler): re-emit KV block dropped by mixed-batching eviction

rebel-ykchoi:
#### Summary

  Fix a `RBLNScheduler` bug where a KV block allocated to a decode request is
  permanently lost to the model runner when that request is evicted by the
  "disable mixed batching" path, leaving a stale block-id `0` in its block
  table and corrupting decode output.

  #### Root cause

  When a new prefill request is scheduled, `RBLNScheduler`'s mixed-batching
  eviction kicks out every running (decode) request already scheduled in that
  step, popping their entries from `req_to_new_blocks`. A block that
  `allocate_slots` just granted to an evicted request is already committed in
  the KV-cache coordinator but never reaches the runner. On the next step
  `allocate_slots` returns an empty delta (the coordinator already holds the
  block), so the block id is never communicated — the runner's block table
  keeps a stale `0`.

  This only corrupts **block-aligned prompts** (prompt length an exact
  multiple of `block_size`): their second block is needed on the very first
  decode step — exactly the prefill→decode transition step the eviction
  targets — so it is dropped. Non-aligned prompts allocate the second block
  one step later on a pure decode step and are unaffected.

  #### Impact

  Observed on `gpt-oss-120b` (hybrid Full + SWA), 64 prompts of length 1024
  (= block_size): only **2 of 64** requests kept a correct full-attention
  block table; the other 62 read/wrote generated-token KV at block `0`,
  producing incorrect outputs.

  #### Fix

  - Stash the dropped delta in `self._stranded_new_blocks` and re-emit it
    (prepended, preserving the runner's append order) on the request's next
    scheduled step via `KVCacheBlocks.__add__`.
  - Coordinator state is left untouched — only the runner's view is re-synced.
  - `_free_request` drops any stash entry so an aborted/finished request
    cannot leak.
  - The zeroing path (`new_block_ids_to_zero`) is unaffected: it is gated on
    `needs_kv_cache_zeroing == has_mamba_layers`, which is `False` for
    non-Mamba models.

  #### Verification

  - Added regression tests: stash + re-emit at a block boundary, and on-finish
    cleanup. Both fail without the fix and pass with it.
  - Device rerun (`gpt-oss-120b`, 1024-token prompts): full-attention block
    table now grows to 2 blocks for all 64 requests across the entire decode
    phase (previously only 2/64); zero requests stuck at 1 block during decode.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/torch_compile/unit/v1/core/test_scheduler.py
+++ b/tests/torch_compile/unit/v1/core/test_scheduler.py
@@ -127,6 +127,108 @@ def test_new_prefill_uses_full_budget_when_decode_running():
     assert output.num_scheduled_tokens[req_b.request_id] == max_num_batched_tokens
 
 
+def test_evicted_decode_block_is_restashed_and_reemitted():
+    """Regression: a decode request whose next KV block is allocated exactly
+    on the prefill->decode transition step, and which is then evicted by the
+    "disable mixed batching" path, must not lose that block.
+
+    The block is already committed in the coordinator, so the next step's
+    allocate_slots returns an empty delta — without re-emitting the stashed
+    delta the runner's block table would keep a stale block-id 0. This is the
+    failure mode observed for block-aligned prompts (prompt length an exact
+    multiple of block_size), whose second block is needed on the very first
+    decode step that the eviction targets.
+    """
+    block_size = 16
+    scheduler = create_scheduler(
+        max_num_batched_tokens=128,
+        max_num_seqs=4,
+        block_size=block_size,
+        num_blocks=10000,
+    )
+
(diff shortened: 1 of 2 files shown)


rebel-wonsubkim (approved):
LGTM

PR #677, opened 2026-06-15
perf(reorder): copy only the valid token range in InputBatch.swap_states

rebel-jongwonlee:
## Summary
Reduce the per-swap copy cost of `_may_reorder_batch` (the `VLLM_RBLN_SORT_BATCH` length-sort) by copying only the **valid token range** of `token_ids_cpu` / `is_token_ids` instead of the full `max_model_len`-wide row.

## Context — why the sort can't just be turned off
`VLLM_RBLN_SORT_BATCH` (coupled with `VLLM_RBLN_BATCH_ATTN_OPT`) is **required** by rebel's batched decode attention kernel: per KV partition `p` the kernel attends only the first `valid_batch[p]` sequences (count of seqs with `len > p*partition_len`) as the dynamic batch size, which selects the correct sequences **only if the batch is sorted by descending length**. On gpt-oss-120b the alternative configs don't work (`opt=0/sort=0` fails to compile; `opt=1/sort=0` is silently wrong on multi-partition decode). So the sort is mandatory and the right optimization is to make each swap cheaper.

## Change
Monkey-patch `InputBatch.swap_states` (no base-vLLM edit, gated by `VLLM_RBLN_FAST_SWAP`, default on): wrap the upstream method with **narrowed views** of `token_ids_cpu`/`is_token_ids` (`[:, :n]`, `n` = max valid length of the two rows via `num_tokens_no_spec`), so the original logic copies only `:n` columns (~1k) instead of `max_model_len` (131072 int32 ≈ 512KB/row). All other field swaps run unchanged; the full arrays are restored in `finally`. Robust to upstream changes (delegates to the original).

## Validation (gpt-oss-120b PDD, rebel 0.10.5.dev269)
Reduces the decode **overhead** component (E2E − d32 − sample) consistently with `VLLM_RBLN_FAST_SWAP=1` vs `0`: 1k/144 3.42→4.02 (−0.60ms), 1k/192 3.09→3.72 (−0.63ms), 8k/1k 2.83→2.88 (−0.05, less turnover). Correctness preserved (0 failed requests across MIX/PDD × {1k144,1k192,8k1k}). Toggle off with `VLLM_RBLN_FAST_SWAP=0`.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -155,6 +155,45 @@ if TYPE_CHECKING:
 logger = init_logger(__name__)
 
 
+# [OPT] Monkey-patch InputBatch.swap_states to copy only the valid token range.
+# token_ids_cpu/is_token_ids are [max_num_reqs, max_model_len]; a full-row swap
+# copies max_model_len(=131072) int32 ≈ 512KB/row. Under PDD decode request
+# turnover, _may_reorder_batch issues many swaps/step → ~0.85ms. We wrap the
+# upstream swap_states with NARROWED VIEWS of those two arrays (sharing memory),
+# so the original logic copies only :n columns (n = max valid token count of the
+# two rows) — all other field swaps run unchanged. No base-vLLM edit, robust to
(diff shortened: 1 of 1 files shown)


rebel-jongwonlee:
closed by #683

PR #685, opened 2026-06-17
refactor(mm): build multimodal forward inputs (embeds + MRoPE positions) at the runner level

rebel-eunji:
## Summary

Move multimodal **forward-input assembly** out of each model's `forward` and up to the **runner level**, mirroring upstream vLLM where `GPUModelRunner` produces `inputs_embeds` (and, for MRoPE models, positions) and hands them to the model.

This builds on `feature/mm-refactoring-2` (which introduced `RBLNOptimumMultimodalMixin` + shared `embed_multimodal` / `embed_input_ids`). Here we wire those into the runner, make the per-request MRoPE delta state **runner-owned**, and fold Qwen-VL into the same path.

Previously each model embedded inline in its prefill branch, and Qwen-VL additionally owned its own `rope_deltas` dict and computed decode positions inside `forward`. Now the runner assembles inputs via two **pure** model hooks and passes them to the model through `ModelInputForRBLN`.

## Design

Two pure hooks on `RBLNOptimumMultimodalMixin` (overridable per model):

- `build_prefill_forward_inputs(model_input) -> (inputs_embeds, position_embed | None, rope_delta | None)` — prefill. Default: `embed_multimodal` + `embed_input_ids` (no MRoPE). Qwen-VL overrides to also return the MRoPE `position_embed` and the rope delta.
- `compute_decode_position_embed(model_input, mrope_position_deltas) -> Tensor | None` — decode. Default `None` (non-MRoPE models pass `input_ids` through). Qwen-VL overrides.

The hooks are **pure**: they *return* the per-request MRoPE delta — they never mutate runner state. The **runner owns the state**: `RBLNOptimumModelRunner.mrope_position_deltas` (`request_id → delta`) mirrors upstream `CachedRequestState.mrope_position_delta`. It is populated at prefill, consumed at decode, and cleaned up on finished requests — all by the runner.

> The normal-path prefill hook (`build_prefill_forward_inputs`) is distinct from the EC-consumer-only `build_prefill_inputs_from_cache` (see below); the names are kept separate on purpose.

## Changes

- **`base.py`** — `ModelInputForRBLN` gains `inputs_embeds` and `position_embed` fields (runner-produced, model-consumed).
- **`optimum_model_runner.py`** — add runner-owned `mrope_position_deltas`; add `_build_forward_inputs()` that dispatches prefill/decode, stores the returned delta, attaches `inputs_embeds` / `position_embed`, and cleans up finished requests. Called before `self.model(model_input)` on the normal forward path (EC producer/consumer paths unchanged). `_build_forward_inputs()` runs *before* the `rebel.capture_reports()` context so the captured perf reports cover only the model forward, not the runner-level embed/position assembly.
- **`model_base.py`** — the two default hooks on `RBLNOptimumMultimodalMixin`.
- **`qwen_vl.py`** — override both hooks; drop the model-owned `self.rope_deltas` dict and the inline prefill / `_preprocess_embeds` logic from `forward`, which now just consumes `model_input.inputs_embeds` / `position_embed`.
- **Model forwards (6)** — `gemma3`, `paligemma`, `idefics3`, `llava`, `blip2`, `llava_next`: consume `model_input.inputs_embeds` instead of embedding inline.
- **`ec_disagg_helpers.py`** — the EC consumer path writes the rope delta into the runner-owned `mrope_position_deltas` (was the model's own dict).

## Naming (EC-path clarity)

The encoder-cache (EC) disaggregation hooks were renamed so their names match what they do at the runner level — they had `decoder` / a too-generic name even though they run on the EC consumer's **prefill** step:

| before | after | why |
| --- | --- | --- |
| `_run_decoder_with_cached_encoder` | `_run_prefill_with_cached_encoder` | It is the EC consumer's prefill step; `decoder` only referred to optimum-rbln's internal `prefill_decoder` runtime. |
| `build_prefill_inputs` | `build_prefill_inputs_from_cache` | EC-consumer-only entry point; disambiguates from the normal-path `build_prefill_forward_inputs`. |

`build_prefill_inputs_from_cache` is invoked **only** on the EC consumer prefill path (`_run_prefill_with_cached_encoder`); the normal path never reaches it. (EC itself is currently gated to `RBLNQwen3VLForConditionalGeneration`, so the `model_base` default and the Gemma3 `NotImplementedError` override are unreachable today.)

## Fix: MRO ordering

`RBLNOptimumMultimodalMixin` is now ordered **before** `RBLNOptimumDecoderMixin` in every MM model class. Previously the decoder mixin's `VllmModel.embed_input_ids(input_ids)` (a 2-arg Protocol stub) preceded the mixin's rich `embed_input_ids(input_ids, multimodal_embeddings, *, is_multimodal)` in the MRO, so the runner-level embed call raised:

```
TypeError: embed_input_ids() takes 2 positional arguments but 3 were given
```

`embed_input_ids` is the only method shared between the two mixin chains, so only its resolution changes (`forward` stays per-model, `compute_logits` / `preprocess_for_decoder` stay on the decoder mixin). Gemma3 additionally drops a redundant explicit `VllmModelForTextGeneration` base (already inherited via the decoder mixin) — MRO and `issubclass` checks are unchanged.

## Notes

- **Qwen-VL is now included** in the runner-level path — the earlier `merges_embeds_in_runner` opt-out flag has been removed.
- Qwen-VL decode `inputs_embeds` are still embedded inside `forward` (they depend on the padded `input_ids` produced by `preprocess_for_decoder`); only the positions (`position_embed`) are precomputed at the runner.
- EC producer/consumer encoder paths are otherwise unchanged; only MRoPE delta-state ownership moved to the runner.

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/base.py
+++ b/vllm_rbln/model_executor/models/optimum/base.py
@@ -35,13 +35,7 @@ class ModelInputForRBLN:
     cached_lengths: list[int] = field(default_factory=list)  # for prefix caching
     multi_modal_kwargs: BatchedTensorInputs | None = None
     dummy_block: int | None = None  # for prefix caching
-    # Prefill input embeddings, computed at the runner level for multimodal
-    # models (embed_multimodal + embed_input_ids) and consumed by the model
-    # forward. Mirrors upstream vLLM where the runner produces inputs_embeds.
     inputs_embeds: torch.Tensor | None = None
-    # MRoPE position embeddings (cos/sin), computed at the runner level for
-    # models that use multimodal rotary positions (e.g. Qwen-VL) and consumed
-    # by the model forward. None for models without MRoPE.
     position_embed: torch.Tensor | None = None
 
 
--- a/vllm_rbln/model_executor/models/optimum/base.py
+++ b/vllm_rbln/model_executor/models/optimum/base.py
@@ -35,6 +35,10 @@ class ModelInputForRBLN:
     cached_lengths: list[int] = field(default_factory=list)  # for prefix caching
     multi_modal_kwargs: BatchedTensorInputs | None = None
     dummy_block: int | None = None  # for prefix caching
+    # Prefill input embeddings, computed at the runner level for multimodal
+    # models (embed_multimodal + embed_input_ids) and consumed by the model
+    # forward. Mirrors upstream vLLM where the runner produces inputs_embeds.
+    inputs_embeds: torch.Tensor | None = None
 
 
 version_error = RuntimeError(
--- a/vllm_rbln/model_executor/models/optimum/base.py
+++ b/vllm_rbln/model_executor/models/optimum/base.py
@@ -39,6 +39,10 @@ class ModelInputForRBLN:
     # models (embed_multimodal + embed_input_ids) and consumed by the model
     # forward. Mirrors upstream vLLM where the runner produces inputs_embeds.
     inputs_embeds: torch.Tensor | None = None
+    # MRoPE position embeddings (cos/sin), computed at the runner level for
+    # models that use multimodal rotary positions (e.g. Qwen-VL) and consumed
+    # by the model forward. None for models without MRoPE.
+    position_embed: torch.Tensor | None = None
 
 
 version_error = RuntimeError(
--- a/vllm_rbln/model_executor/models/optimum/qwen_vl.py
+++ b/vllm_rbln/model_executor/models/optimum/qwen_vl.py
@@ -313,6 +313,12 @@ class RBLNOptimumQwenVLForConditionalGeneration(
         request_nums = input_ids.shape[0]
         is_prompt = model_input.is_prompt
 
+        # FIXME This should be removed in the future
+        # by moving the padding logic into model runner.
+        assert len(model_input.running_requests_ids) == request_nums, (
+            f"The number of running requests is {len(model_input.running_requests_ids)}, "
+            f"but the shape of input_ids is {input_ids.shape}")
+        
         kwargs = self.preprocess_for_decoder(
             is_prompt, block_tables, input_ids, cache_position
         )
--- a/vllm_rbln/model_executor/models/optimum/qwen_vl.py
+++ b/vllm_rbln/model_executor/models/optimum/qwen_vl.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from abc import ABC, abstractmethod
+from dataclasses import replace
 from typing import Any
 
 import torch
@@ -31,6 +32,8 @@ from vllm.model_executor.models.qwen2_vl import (
     Qwen2VLVideoPixelInputs,
 )
 
+from vllm_rbln.utils.optimum.bucket import select_bucket_size
+
 from .base import ModelInputForRBLN
 from .model_base import (
     RBLNOptimumDecoderMixin,
@@ -49,11 +52,6 @@ class RBLNOptimumQwenVLForConditionalGeneration(
     Automatically detects model type based on the model configuration.
     """
 
-    # Qwen-VL prefill uses a custom preprocess_prefill (rope_deltas /
-    # position embeddings) rather than the embed_input_ids merge pattern,
-    # so it embeds inside its own forward instead of at the runner level.
-    merges_embeds_in_runner: bool = False
-
     @classmethod
     def get_placeholder_str(cls, modality: str, i: int) -> str | None:
         if modality.startswith("image"):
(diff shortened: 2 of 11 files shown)


vllm_rbln/model_executor/models/optimum/qwen_vl.py
```diff
@@ -238,108 +239,110 @@ def _create_video_embedding_inputs(self, video_embeds, video_grid_thw) -> Any:
...
+        """
+        cache_position = model_input.input_positions
+        running_requests_ids = model_input.running_requests_ids
+        # int32 mirrors the cache_position dtype the prior decode path used
+        # (cast in preprocess_for_decoder before computing position embeds).
+        cache_position = cache_position.to(torch.int32)
+        padded_batch_size = self.decoder_batch_size
+        if self.use_multiple_decoder:
+            padded_batch_size = select_bucket_size(
+                len(running_requests_ids), self.decoder_batch_sizes
```
rebel-yskim, line 289: **[Should fix]** `padded_batch_size` is now recomputed here independently from `preprocess_for_decoder`, and the two use different inputs:

- here: `len(running_requests_ids)`
- `preprocess_for_decoder` (`model_base.py`): `request_nums = input_ids.shape[0]`

Before this refactor there was a single source — `preprocess_for_decoder` computed `padded_batch_size` and passed it into `_preprocess_embeds`. Now position computation runs at the runner level (before `forward`), so it can't reuse that result and re-derives the bucket. The `position_embed` produced here is padded to this value, while the actual decoder is selected in `forward` from the `padded_batch_size` that `preprocess_for_decoder` returns. If the two ever diverge, the `position_embed` batch dim won't match the decoder. The equality of `len(running_requests_ids)` and `input_ids.shape[0]` at decode is an implicit invariant. Please either share one bucket-selection helper between both sites, or assert `len(running_requests_ids) == input_ids.shape[0]` to make the assumption explicit.

> yskim-review-bot

vllm_rbln/model_executor/models/optimum/qwen_vl.py
```diff
@@ -238,108 +239,110 @@ def _create_video_embedding_inputs(self, video_embeds, video_grid_thw) -> Any:
...
+        self,
+        model_input: ModelInputForRBLN,
+        mrope_position_deltas: dict[str, float],
+    ) -> torch.Tensor:
+        """Decode-step MRoPE: advance each request's position from its stored
+        delta (``cache_position + mrope_position_delta``) and return the padded
+        position embeddings (cos/sin). Mirrors upstream vLLM's
+        ``get_next_input_positions_tensor``.
+        """
+        cache_position = model_input.input_positions
```
rebel-yskim, line 281: **[Nit]** `cache_position` is now read from `model_input.input_positions`, i.e. the value *before* `preprocess_for_decoder`. The old `_preprocess_embeds` received the post-`preprocess_for_decoder` (padded) `cache_position`. For the real request indices (`b_id < len(running_requests_ids)`) the values match since padding only appends at the end, and the `.to(torch.int32)` cast is preserved — so this looks equivalent. It does rely on `input_positions` row order always matching `running_requests_ids`; worth a quick confirmation.

> yskim-review-bot

rebel-yskim (changes requested):

vllm_rbln/model_executor/models/optimum/base.py
```diff
@@ -35,6 +35,14 @@ class ModelInputForRBLN:
     cached_lengths: list[int] = field(default_factory=list)  # for prefix caching
     multi_modal_kwargs: BatchedTensorInputs | None = None
     dummy_block: int | None = None  # for prefix caching
+    # Prefill input embeddings, computed at the runner level for multimodal
```
rebel-yskim: nit, remove verbose comments?

vllm_rbln/model_executor/models/optimum/qwen_vl.py
rebel-eunji: That's a good point. The `preprocess_for_decoder` must be refactored for clean code. So I just add assert for now.

rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/685/changes/b84a6ac4713b742f158971a1f13b1c47981c92fa

vllm_rbln/model_executor/models/optimum/qwen_vl.py
rebel-eunji: Yes. `input_positions` and `running_request_ids` are generated together in `_prepare_decode` method.
https://github.com/RBLN-SW/vllm-rbln/blob/refactor/multimodal-embed-call/vllm_rbln/v1/worker/optimum_model_runner.py#L708

rebel-jonghewk (approved):

rebel-yskim (approved):
LGTM

rebel-eunji:
TODO
## Refactoring Qwen VL.
Currently...
- embed_input_ids: not called at all
- embed_multimodal: called only for ec producer

## preprocess_for_decoder
- split prefill and decoder

PR #691, opened 2026-06-18
other(ci): pin fastapi below 0.137 for OpenAI entrypoint tests

rebel-eunji:
## Problem

`pytest tests/entrypoints/openai` (the `openai` matrix in the `pytest-arc` job) crashes on the first request to the spun-up OpenAI server:

```
AttributeError: '_IncludedRouter' object has no attribute 'path'
  prometheus_fastapi_instrumentator/routing.py, line 55, in _get_route_name
    route_name = route.path
```

## Root cause

- **fastapi 0.137.0** changed `include_router`: instead of flattening child routes into `app.routes` as `APIRoute` (which have `.path`), it now keeps them as `_IncludedRouter` objects — a `BaseRoute` subclass with **no `.path`**.
- **prometheus-fastapi-instrumentator** (≤8.0.0) iterates `app.routes` in its middleware and reads `route.path` for every request → `AttributeError`. This fires before any route handler, so any request to the server fails.
- vLLM pins `fastapi[standard]>=0.115.0` with **no upper bound**, so a fresh CI install resolves to the newest fastapi (0.137.x). It is purely a dependency-resolution/timing issue, not specific to vllm-rbln.

Verified: fastapi 0.136.3 still flattens routes into `APIRoute` and the instrumentator resolves route names fine; 0.137.x does not.

## Fix

Add a pin step in the `pytest-arc` job, **after** the wheel install (which re-resolves fastapi back to latest) and before `Run pytest`:

```yaml
- name: Pin fastapi below 0.137 (instrumentator compat)
  if: steps.should_skip.outputs.skip != 'true'
  run: |
    uv pip install --system "fastapi[standard]<0.137"
```

Scope notes:
- Only the `pytest-arc` job needs it — the other `pytest` job (core/worker/optimum_compile) doesn't launch the OpenAI server.
- `<0.137` still satisfies vLLM's `fastapi[standard]>=0.115.0` constraint.
- Project dependencies (`pyproject.toml`) are intentionally **not** touched — this is a CI-only pin until the instrumentator handles the new route type.

### DIFF
--- a/.github/workflows/rbln_vllm-rbln_pytest.yaml
+++ b/.github/workflows/rbln_vllm-rbln_pytest.yaml
@@ -357,7 +357,7 @@ jobs:
         run: |
           case "${{ matrix.test_type.name }}" in
             "openai")
-              pytest tests/entrypoints/openai -vv --durations 0
+              pytest tests/entrypoints/openai -vv --durations 0 --timeout=30
               ;;
             "llm")
               pytest tests/entrypoints/llm -vv --durations 0
--- a/.github/workflows/rbln_vllm-rbln_pytest.yaml
+++ b/.github/workflows/rbln_vllm-rbln_pytest.yaml
@@ -357,7 +357,7 @@ jobs:
         run: |
           case "${{ matrix.test_type.name }}" in
             "openai")
-              pytest tests/entrypoints/openai -vv --durations 0 --timeout=30
+              pytest tests/entrypoints/openai -vv --durations 0 --timeout=600
               ;;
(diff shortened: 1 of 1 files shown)


rebel-jonghewk (commented):
@rebel-eunji is there any timeout if this fails?

rebel-eunji:
@rebel-jonghewk
I've just added a timeout option (10 min) to the OpenAI test. The other pytests usually fail right away anyway.

rebel-jonghewk (approved):

PR #692, opened 2026-06-18
other(ci): pin fastapi below 0.137 for OpenAI entrypoint tests (dev)

rebel-eunji:
## Problem
It is not related with bumpup. It is related with recently released fastapi.
Same issue as #691 (which landed on `dev-0.22.0`): `pytest tests/entrypoints/openai` (the `openai` matrix in `pytest-arc`) crashes on the first request to the OpenAI server:

```
AttributeError: '_IncludedRouter' object has no attribute 'path'
  prometheus_fastapi_instrumentator/routing.py, line 55, in _get_route_name
    route_name = route.path
```

## Root cause

- **fastapi 0.137.0** (released 2026-06-14) changed `include_router`: child routes are kept as `_IncludedRouter` (a `BaseRoute` with no `.path`) instead of being flattened into `app.routes` as `APIRoute`.
- **prometheus-fastapi-instrumentator** (≤8.0.0) reads `route.path` for every request in its middleware → `AttributeError` before any handler runs.
- vLLM pins `fastapi[standard]>=0.115.0` with no upper bound, so a fresh CI install resolves to the newest fastapi (0.137.x). Dependency-resolution/timing issue, not specific to vllm-rbln.

## Fix

CI-only pin; `pyproject.toml` is intentionally untouched.


rebel-kblee (approved):
👍

PR #597, opened 2026-05-12
fix(whisper): remove workaround INVALID_TOKEN and add missing feature

rebel-eunji:
### 🚀 Summary of Changes

1. Fix workaround INVALID_TOKEN
    a. For now, we've used INVALID_TOKEN to identify the 1st step of decoder.
    b. But it triggers an error related to the shape of logits is [batch_size, vocab_size + 100])
    c. So we removed the workaround by running 1st decoder step in prefill step.
    d. For that, optimum-rbln always allocates one more block to padding the block table.
2. Allow forced_decoder_ids
    a. Whisper model has a feature to instruct the task (e.g. transcribe, translate.. ) and language (e.g. fr, en.. )-- `forced_decoder_ids`
    b. From now on, it is enabled by running the tokens one by one.
3. make script clearer
    a.  Currently, run_whisper.py code frequently causes Python GIL error by importing load_dataset globally . So I've made import load_dataset in the func module.
  

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to https://github.com/RBLN-SW/optimum-rbln/pull/560

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [x] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [x] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
5. Verify output: `...`
6. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #694, opened 2026-06-18
refactor(paligemma): move dataset/tokenizer access out of async path

rebel-eunji:
## What

Refactors `examples/optimum/run_paligemma.py` to prepare prompts and load the tokenizer **synchronously in `entry_point`** — before `asyncio.run` — instead of inside the async `main()`.

## Why

Accessing the `datasets` library (`load_dataset` + iteration) and `AutoTokenizer.from_pretrained` from within an async function caused GIL contention. This applies the same fix already landed for whisper in #597.

## Changes

- Move `generate_prompts(...)` call from async `main()` into sync `entry_point`; pass the prepared `inputs` list down.
- Make `from datasets import load_dataset` a local import inside `generate_prompts`.
- Load the tokenizer in `entry_point` and pass only `eos_token_id` into `main`/`generate` (no tokenizer object in the async path).

No behavioral change to the generated output.

### DIFF
--- a/examples/optimum/run_paligemma.py
+++ b/examples/optimum/run_paligemma.py
@@ -15,14 +15,14 @@
 import asyncio
 
 import fire
-from datasets import load_dataset
 from transformers import AutoTokenizer
 from vllm import AsyncEngineArgs, AsyncLLMEngine, SamplingParams
 
 
 def generate_prompts(batch_size: int):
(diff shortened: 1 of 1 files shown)


rebel-kblee (approved):
lgtm !

rebel-seinpark (approved):
👍

PR #699, opened 2026-06-19
fix(deps): pin fastapi below 0.137 in pyproject instead of CI

rebel-eunji:
## Summary
- Same issue in upstream vllm https://github.com/vllm-project/vllm/pull/45594 because of new release of fastapi.
- Move the `fastapi[standard]<0.137` pin out of the pytest CI workflow and into the project `dependencies` in `pyproject.toml`, so the constraint applies to every install (not just CI).
- Remove the now-redundant "Pin fastapi below 0.137" step from `.github/workflows/rbln_vllm-rbln_pytest.yaml`.

## Why
`fastapi>=0.137` keeps included routers as `_IncludedRouter` (no `.path`), which breaks prometheus-fastapi-instrumentator's route-name middleware and crashes the OpenAI server on every request. Keeping the pin in `pyproject.toml` ensures every environment resolves to a compatible fastapi until the instrumentator handles the new route type.

## Note
CI installs the wheel with `--force-reinstall --no-cache-dir ... --index-strategy unsafe-best-match`, so the resolver should now honor this constraint at install time. Worth confirming a CI run actually resolves fastapi <0.137.


rebel-jinhwan (approved):
LGTM 👍

rebel-jonghewk (approved):

PR #713, opened 2026-06-24
feature(pdd): unified RBLN NIXL kv connector for vLLM 0.22

rebel-ykchoi:
#### Purpose

  Add a single, unified RBLN NIXL KV connector for Prefill/Decode disaggregation on
  vLLM **0.22**, serving **both** transfer paths from one worker, branching internally
  on `kv_transfer_config.kv_buffer_device`:

  - **`cpu` (host-bounce)** — page-aligned host staging, RDMA over DRAM (`ibv_reg_mr`).
  - **`rbln` (D2D)** — direct device-memory RDMA (`ibv_reg_dmabuf_mr`) on the buffer
    exported by the `nixl-rbln` adapter; no host staging.

  Both paths share the same RBLN backend / RDMA NICs and require
  `VLLM_RBLN_USE_DEVICE_TENSOR=1`; they differ only in the registered memory segment
  (DRAM_SEG vs VRAM_SEG). When `nixl-rbln` is not installed, the connector **falls back
  to upstream NIXL (UCX)** on the host-bounce path and rejects D2D with an actionable
  error.

  Beyond porting the connector to the 0.22 NIXL API, this makes the connector path
  **compile-clean** and lets the **same compiled binary run with the connector on or
  off**, and adds **MultiConnector** support so NIXL can run alongside other connectors
  (e.g. LMCache).

  #### What's included

  - **Unified connector** (`distributed/kv_transfer/.../rbln_nixl/` package —
    `connector` / `scheduler` / `worker`): one `RblnNixlConnectorWorker` for host-bounce
    + D2D; deferred D2D registration (`finalize_kv_cache_registration` runs after warm-up
    materializes the KV backing memory); logical `block_size` pinning; chiplet sharding
    delegated to `nixl-rbln`.
  - **Upstream-NIXL fallback** when `nixl-rbln` is absent (UCX/DRAM host-bounce).
  - **Hybrid Full + SWA** KV cache support. Opt-in `VLLM_RBLN_NIXL_SWA_VIEW_OPT=1`
    publishes a second `sliding_window`-sized descriptor range at the same NIXL base
    addresses, trimming SWA groups' RDMA payload to `1/ratio` (host h2d/d2h still moves
    the full block; only the wire payload shrinks).
  - **vLLM 0.22 NIXL API adaptation**: register the remote engine in `TransferTopology`
    before the handshake; drop the removed base `_block_size`/`_tp_size` dicts.
  - **Attention KV as a graph input** (`model_executor/layers/attention/attention.py`):
    override `get_attention_context` to return no KV cache so the connector's per-layer
    wrapper never reads the layer's embedded cache and bakes it into the compiled graph;
    the attention op resolves its KV cache from attention metadata instead. Unblocks
    connector-path compilation and keeps the connector/no-connector graphs identical
    (same binary).
  - **MultiConnector support** (`distributed/kv_transfer/.../utils.py`): the RBLN-specific
    post-warm-up registration hook is not forwarded by `MultiConnector`; flatten the
    connector tree and finalize every child that implements it, via a runtime-checkable
    protocol — enables LMCache + NIXL together.
  - **Runner** (`rbln_model_runner.py`): pick one Full-attention canonical layer per HMA
    pool for `mark_static_address` and connector registration, iterating by layer index
    (deterministic, `PYTHONHASHSEED`-independent); contiguous K/V-split copy path for the
    host-buffer DMA fast path.
  - **SWA manager** (`v1/kv_cache.py`): `allocate_new_computed_blocks` override so the
    D-side P/D receive path allocates a single block per request (matches the P-side and
    RBLN's single-block in-place SWA ring buffer).
  - **Platform** (`platform.py`): allow `rbln` ↔ `cpu` for NIXL supported devices.
  - **Env** (`rbln_envs.py`): add `VLLM_RBLN_NIXL_SWA_VIEW_OPT`.

  #### Affected modules

  KV connector (distributed/kv_transfer) · Worker · Runner · Attn · Platform · Env

  #### Type of change

  Feature

  #### Not supported (by design)

  Pure-SWA single-group with `sliding_window < block_size` **under a KV connector**: the
  canonical-layer fallback would pick the SWA layer (kernel granularity), whose
  `cache.shape[0]` mismatches `num_blocks`. Non-disaggregated serving is unaffected.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

  - **Unit (CPU-runnable, no NPU / no nixl-rbln required):**
      ```bash
      pytest tests/torch_compile/unit/v1/core/test_pd_disaggregation.py \
             tests/torch_compile/unit/v1/worker/test_rbln_model_runner_kv_cache.py \
             tests/torch_compile/unit/v1/core/test_kv_connector_utils.py \
             tests/torch_compile/unit/model_executor/layers/attention/test_attention.py -vv
      ```
      Expected: **all pass** (62 + 23 + 12 + 9 cases). Tests stub `nixl_rbln` via
      `sys.modules`, so they pass with or without the adapter installed.

  - **E2E (needs NPU + nixl-rbln):** run a P/D pair with `VLLM_RBLN_USE_DEVICE_TENSOR=1`;
      `kv_buffer_device="rbln"` for D2D or `"cpu"` for host-bounce. For gpt-oss-120b SWA,
      set `VLLM_RBLN_NIXL_SWA_VIEW_OPT=1`.

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
```diff
@@ -0,0 +1,667 @@
...
+        assert not self._has_mamba, "RBLN NIXL connector does not support Mamba layers."
+
+        block_size_ratio = self.block_size // block_size
+        local_base_addresses = self.kv_caches_base_addr[self.engine_id][self.tp_rank]
+        num_blocks = self.num_blocks * block_size_ratio
+        blocks_data: list[tuple[int, int, int]] = []
+
+        # Two passes when SWA is present: Full descs first, then SWA descs
+        # at the same base addresses but `sliding_window`-sized.
+        length_divisors = [1] if self._sw_ratio is None else [1, self._sw_ratio]
```
rebel-yskim: At here,`self._sw_ratio` is always not None (L446 always guarantee)

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
```diff
@@ -0,0 +1,667 @@
...
+                self.src_xfer_handles_by_tp_ratio[tp_ratio].append(handle)
+
+        blocks_data: list[tuple[int, int, int]] = []
+        num_blocks = nixl_agent_meta.num_blocks
+
+        # Two passes when SWA is present: Full descs first, then SWA descs
+        # at the same base addresses (same `page_size` stride — the
+        # remote tensor's physical block stride is still Full-sized),
+        # shorter desc length.
+        length_divisors = [1] if self._sw_ratio is None else [1, self._sw_ratio]
```
rebel-yskim: Same comment as L463

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -5108,24 +5173,16 @@ def rbln_copy_kv_blocks(
...
-                    copy_fn = runtime._update_kv_cache
-                else:
-                    kv_caches = dst_kv_caches
-                    copy_fn = runtime._fetch_kv_cache
-
-                for idx in src_block_ids:
-                    for kv_name, kv_cache in kv_caches.items():
-                        block_size = kv_cache.shape[-2]
-                        copy_fn(kv_cache, idx, 0, block_size, kv_name)
+                for layer_name, dst_cache in dst_kv_caches.items():
```
rebel-yskim, line 5180: is it an intended behavior that it does not use `dst_block_ids`? (only used in assertion & if statement)

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
```diff
@@ -0,0 +1,667 @@
...
+            )
+
+        return remote_agent_name
+
+    def _compute_desc_ids(
+        self,
+        block_ids: BlockIds,
+        dst_num_blocks: int,
+        block_size_ratio: float | None,
+        physical_blocks_per_logical: int,
```
rebel-yskim, line 633: The None path forwards it via super()._compute_desc_ids(..., physical_blocks_per_logical), but the SWA branch's desc formula (region_ids * num_blocks + group_arr + offset) never uses it. This is correct because the connector pins _physical_blocks_per_logical_kv_block = 1, but accepting an argument and ignoring it is non-obvious. Suggest a one-line comment stating "the SWA desc formula assumes physical_blocks_per_logical == 1", or an explicit assert physical_blocks_per_logical == 1.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
```diff
@@ -0,0 +1,667 @@
...
+        self._validate_remote_agent_handshake(nixl_agent_meta, remote_tp_size)
+
+        tp_ratio = self.transfer_topo.tp_ratio(remote_tp_size)
+        indexes_into_remote = (
+            not self.transfer_topo.is_kv_replicated(engine_id) and tp_ratio > 0
+        )
+
+        # Heterogeneous TP path (P TP > D TP): logically split own regions
+        # into |tp_ratio| chunks. Mirrors upstream; preserved verbatim
+        # because RBLN may run heterogeneous TP in the future.
```
rebel-yskim: The comment says "preserved verbatim because RBLN may run heterogeneous TP in the future." RBLN currently runs homogeneous TP, so these ~18 lines are a branch that is never executed or tested (and the surrounding asserts like assert not self.use_mla narrow it further). Mirroring the upstream override's structure is understandable (trade-off with the next nit), but if hetero-TP is genuinely out of scope, an explicit assert tp_ratio >= 0 is safer than carrying an unverified branch. Author's call.

rebel-yskim (changes requested):

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
rebel-ykchoi: drop redundant _sw_ratio None-checks

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
rebel-ykchoi: drop redundant _sw_ratio None-checks

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: comment why rbln_copy_kv_blocks keeps the fixed CopyBlocksOp signature.
P/D transfers use identical block ids on both sides, which we assert (src_block_ids == dst_block_ids).

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
rebel-ykchoi: assert physical_blocks_per_logical == 1 in the SWA desc path

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
rebel-ykchoi: replace the untested heterogeneous-TP (remote>local) branch with assert tp_ratio >= 0

rebel-yskim (approved):
LGTM.
Could you please explain briefly that why we need `allocate_new_computed_blocks`? (I'm not familiar with this context)

rebel-ykchoi:
> LGTM. Could you please explain briefly that why we need `allocate_new_computed_blocks`? (I'm not familiar with this context)

@rebel-yskim 

## Background: upstream and RBLN use different SWA block allocation rules

- upstream SWA: allocates multiple blocks proportional to sequence length (cdiv(num_tokens, block_size)).
- RBLN SWA kernel: uses exactly one physical block per request and slides the KV in place within it (in-place ring buffer).

Therefore RBLN must enforce "SWA always uses 1 block" at the KV cache manager level, which requires an override.

## Problem: there are two block allocation paths

The scheduler allocates blocks through different methods depending on the situation.

| Situation | Allocation path | Behavior |
|---|---|---|
| Mix / Prefill (incl. P/D prefill) | allocate_new_blocks (tokens to be newly computed) | RBLN override applied, 1 block (correct) |
| P/D Decode | allocate_new_computed_blocks (externally received tokens) | without override, upstream cdiv, multiple blocks (incorrect) |

- Mix execution: prefill only goes through allocate_new_blocks, so the existing override correctly allocates 1 block.
- P/D disaggregation:
  - The prefill side likewise goes through allocate_new_blocks and correctly allocates 1 block.
  - The decode side receives KV transferred from the prefill node. That amount is recorded as num_external_computed_tokens, so it takes the allocate_new_computed_blocks path. Because this method was not overridden, it allocated multiple blocks per the upstream rule.
  - Result: the prefill node sends 1 block while the decode node allocates multiple blocks, so the block counts mismatch and the KV transfer breaks.

## Fix

Override allocate_new_computed_blocks as well, so the externally received portion also allocates only 1 block. Both P and D then agree on 1 block.

## One-line summary

SWA needs an override because it uses "1 block per request"; Mix was fine since prefill only goes through allocate_new_blocks, but the P/D decode side takes the allocate_new_computed_blocks path due to externally received tokens, and since that path follows the upstream rule, it needed its own separate 1-block override.

PR #731, opened 2026-06-26
refactor(metrics): merge model + sampler into a single MODEL+SAMPLER tracker

rebel-jonghewk:
### 🚀 Summary of Changes

Report model and sampler timings as a single combined `MODEL+SAMPLER` tracker instead of two separate `MODEL` / `SAMPLER` blocks. Per step the model time and sampler time are summed into one measurement, halving the metrics output and reflecting the real model+sampler compute time.

**1. Combined tracker:** Replace the separate `MODEL` / `SAMPLER` trackers with one `MODEL+SAMPLER` tracker. Model and sampler run 1:1 per step, so their timings are summed into a single per-step measurement.

**2. StepReport (metrics.py):** New `StepReport` dataclass — `from_reports()` extracts timings (host/device/ccl/prepare) from `capture_reports`, `merged_with()` sums model + sampler timings (None-safe), `record()` writes into a tracker. `collect_metrics` is reimplemented on top of it, so the E2E tracker and the optimum runner keep identical behavior.

**3. rbln_model_runner:** `execute_model` stashes the model `StepReport` in `_pending_model_report` instead of recording immediately; `_sample` builds the sampler `StepReport` and records `model.merged_with(sampler)` once. `_flush_pending_model_report()` records the model report alone on paths that skip sampling — intermediate chunked prefill (`logits.shape[0] == 0`), pooling models, and PP non-last rank — so no model timing is lost.

**4. rbln_worker:** Drop the separate `SAMPLER` print at shutdown; only `MODEL+SAMPLER` and `E2E` are printed.

**5. Scope:** The optimum runner/worker are unchanged (still separate `MODEL` / `SAMPLER`). The shared `collect_metrics` signature/behavior is preserved.

---

* N/A

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run any rbln model with metrics enabled: `VLLM_RBLN_METRICS=1 <serve/inference cmd>`
2. On shutdown, verify a single `FINAL PERFORMANCE STATISTICS [MODEL+SAMPLER]` block is printed (plus `[E2E]`) — no separate `[SAMPLER]` block.
3. Edge cases: pooling model, chunked prefill, and PP non-last rank still record model timing (no lost steps).

---

**Before** (two blocks):
```
FINAL PERFORMANCE STATISTICS [MODEL]
FINAL PERFORMANCE STATISTICS [SAMPLER]
```

**After** (one combined block; values are model + sampler per step):
```
FINAL PERFORMANCE STATISTICS [MODEL+SAMPLER]
FINAL PERFORMANCE STATISTICS [E2E]
```

---

### 📋 Checklist

---

### 💬 Notes

The optimum runner is intentionally left untouched in this PR. The E2E tracker stays separate (it measures whole-step time, which differs in meaning from model+sampler).

### DIFF
--- a/vllm_rbln/v1/worker/metrics.py
+++ b/vllm_rbln/v1/worker/metrics.py
@@ -14,7 +14,7 @@
 
 import os
 from collections import defaultdict
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from typing import TypeVar
 
 import vllm_rbln.rbln_envs as envs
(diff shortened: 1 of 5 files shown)


vllm_rbln/v1/worker/metrics.py
```diff
@@ -321,6 +321,93 @@ def print_final_stats(self):
...
+            performance_tracker.record_decode(
+                self.latency,
+                self.token_count,
+                host_time=self.host_time,
+                device_time=self.device_time,
+                ccl_time=self.ccl_time,
+                prepare_time=self.prepare_time,
+                padded_decode=self.padded_decode,
+                request_ids=self.request_ids,
+            )
```
rebel-eunji: I think it'd be better for record to be a method on the tracker rather than on the report — tracker.record(report) instead of report.record(tracker).

```
# StepReport stays a pure value; PerformanceTracker gains:
def record(self, report: StepReport) -> None:
    if report.is_prefill:
        self.record_prefill(report.latency, report.token_count, ...)
    else:
        self.record_decode(report.latency, report.token_count, ..., padded_decode=report.padded_decode)
```

vllm_rbln/v1/worker/metrics.py
```diff
@@ -321,6 +321,93 @@ def print_final_stats(self):
...
+    token_count: int = 0
+    host_time: int | None = None
+    device_time: int | None = None
+    ccl_time: int | None = None
+    prepare_time: int | None = None
+    is_prefill: bool = False
+    padded_decode: bool = False
+    request_ids: list[str] | None = None
+
+    @classmethod
```
rebel-eunji, line 365: Could `**meta` be spelled out as explicit keyword args? Right now the signature hides which fields are expected — you have to read both call sites to learn it's `token_count` / `is_prefill` / `padded_decode` / `request_ids` — and it gives up type checking on them. Explicit params make the constructor self-documenting at no real cost:

```python
@classmethod
def from_reports(cls, start_time, end_time, reports, *,
                 token_count: int = 0, is_prefill: bool = False,
                 padded_decode: bool = False,
                 request_ids: list[str] | None = None) -> "StepReport":
```

rebel-eunji (approved):
lgtm 👍

PR #719, opened 2026-06-26
feature: always use device tensors

rebel-jonghewk:
### 🚀 Summary of Changes

1. remove `VLLM_RBLN_USE_DEVICE_TENSOR` env var.
2. remove and refactor if-else branch for `VLLM_RBLN_USE_DEVICE_TENSOR`
3. remove `has_torch_rbln` flag and check torch-rbln is installed 
4. remove compile_context
5. change the default value of `VLLM_RBLN_AUTO_PORT` to True

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
4. Verify output: `...`
6. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #737, opened 2026-06-29
refactor(attn): remove unused attn metadata fields and add comment

rebel-eunji:
Follow-up to #719.

## What

Removes `query_start_loc` and 10 other fields from `RBLNFlashAttentionMetadata` that are set in `build()` but never read anywhere:
`num_actual_tokens`, `max_query_len`, `max_seq_len`, `use_cascade`, `common_prefix_len`, `cu_prefix_query_lens`, `prefix_kv_lens`, `suffix_kv_lens`, `scheduler_metadata`, `prefix_scheduler_metadata`.

Also drops the now-unused local computations that fed them.

## Why

Static audit of all consumers (`RBLNFlashAttentionImpl.forward`, `attention.py`, the runner) shows these fields are write-only. The 6 cascade/scheduler fields are vestigial copies of vLLM's `FlashAttentionMetadata` and are always `None`/`False` here; the others hold real values that nothing reads.

`slot_mapping` was also found write-only but is **kept** pending a runtime check (KV-cache correctness).

### DIFF
--- a/vllm_rbln/v1/attention/backends/flash_attention.py
+++ b/vllm_rbln/v1/attention/backends/flash_attention.py
@@ -997,24 +997,9 @@ class RBLNAttentionBackend(AttentionBackend):
 
 @dataclass
 class RBLNFlashAttentionMetadata:
-    num_actual_tokens: int  # Number of tokens excluding padding.
-    max_query_len: int
-    query_start_loc: torch.Tensor
-    max_seq_len: int
     seq_lens: torch.Tensor
     block_tables: torch.Tensor
     slot_mapping: torch.Tensor
-
-    # For cascade attention.
-    use_cascade: bool | None
-    common_prefix_len: int | None
-    cu_prefix_query_lens: torch.Tensor | None
-    prefix_kv_lens: torch.Tensor | None
-    suffix_kv_lens: torch.Tensor | None
-
-    # Optional aot scheduling
-    scheduler_metadata: torch.Tensor | None = None
-    prefix_scheduler_metadata: torch.Tensor | None = None
 
     # To distinguish prefill and decode
     is_prefill: bool = True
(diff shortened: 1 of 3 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
rebel-eunji, line 3530: @rebel-ykchoi 
Is slot_mapping required? I think slot_mapping is not used anymore.

rebel-ykchoi:
LGTM on removing the 11 write-only RBLNFlashAttentionMetadata fields

rebel-ykchoi:
tests/torch_compile/unit/test_forward_context.py still sets num_actual_tokens on a spec=RBLNFlashAttentionMetadata mock — this attribute no longer exists and may break the fixture. Please update before merge.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Agree it's write-only on RBLNFlashAttentionMetadata today; attention reads KV via block_tables/seq_lens, and `forward_context.slot_mapping` is also discarded in `_rbln_unified_attention_with_output`.

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -1170,11 +1151,10 @@ def build(
...
 
-        cu_prefix_query_lens = None
-        prefix_kv_lens = None
-        suffix_kv_lens = None
-        prefix_scheduler_metadata = None
-
+        # The length of the partition equals the block size.
+        partition_len = self.block_size
+        # num_partition is derived from max_model_len (not hardware block count)
+        # to ensure seq_idx/seq_lens dimensions stay within block_table bounds.
```
rebel-eunji:
```suggestion
```

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -1235,20 +1211,9 @@ def build(
         )
 
         attn_metadata = RBLNFlashAttentionMetadata(
-            num_actual_tokens=num_actual_tokens,
-            max_query_len=max_query_len,
-            query_start_loc=query_start_loc,
-            max_seq_len=query_max_seq_len,
-            seq_lens=seq_idx,
+            seq_lens=seq_lens,
```
rebel-eunji:
```suggestion
            seq_lens=seq_idx,
```

PR #740, opened 2026-06-30
fix(deps): add torch-rbln dependency

rebel-jonghewk:
### 🚀 Summary of Changes

Explicitly adds `torch-rbln` to the dependency list of the `vllm-rbln` project.

**Rationale:** Under the current architecture, the `torch-rbln` package is a hard requirement in any environment that uses VLLM_RBLN_USE_DEVICE_TENSOR=1. When this dependency is missing, it either raises a runtime error or forces the user to install it manually.
Declaring it in `pyproject.toml` automates installation and removes that friction.

- Resolves #
- Related to #
 -  https://github.com/rebellions-sw/vllm-rbln-internal/issues/127

---

### ✅ Type of Change

- [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Create a fresh environment without `torch-rbln` installed.
2. Install `vllm-rbln` and confirm `torch-rbln` is pulled in automatically.

---

### 📋 Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
     "torchvision",
     "torchaudio",
     "torchcodec",
-    "torch-rbln==0.2.2",
+    "torch-rbln~=0.2.2.dev0",
     "optimum-rbln>=0.11.1a1",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dependencies = [
     "torchvision",
     "torchaudio",
     "torchcodec",
+    "torch-rbln>=0.2.2",
     "optimum-rbln>=0.11.0",
 ]
 


rebel-eunji (approved):

rebel-seinpark (approved):

rebel-jinhwan (approved):

rebel-minhopark:
In the requirements of vllm (https://github.com/vllm-project/vllm/blob/main/requirements/cuda.txt), the dependency constraint is pinned tightly (e.g., ==2.x.x) instead of using >=. I believe this is due to changes in compilation behavior across PyTorch version updates.

Currently, if we only have `torch-rbln==0.2.2` deployed, running `uv pip install vllm-rbln==x.y.z` will resolve to 0.2.2. However, once we release `torch-rbln==0.2.3`, the same `uv pip install vllm-rbln` command might automatically pull 0.2.3, which could introduce breaking changes for that specific vllm-rbln version.

Does torch-rbln guarantee backward compatibility between patch/minor versions? Or should we strictly bound (ex: `torch-rbln~=0.2.2 or torch-rbln~=0.2.2.dev0` for dev package) or pin the torch-rbln version within vllm-rbln, similar to how upstream vLLM handles PyTorch?

Thank you!

rebel-jonghewk:
> In the requirements of vllm (https://github.com/vllm-project/vllm/blob/main/requirements/cuda.txt), the dependency constraint is pinned tightly (e.g., ==2.x.x) instead of using >=. I believe this is due to changes in compilation behavior across PyTorch version updates.
> 
> Currently, if we only have `torch-rbln==0.2.2` deployed, running `uv pip install vllm-rbln==x.y.z` will resolve to 0.2.2. However, once we release `torch-rbln==0.2.3`, the same `uv pip install vllm-rbln` command might automatically pull 0.2.3, which could introduce breaking changes for that specific vllm-rbln version.
> 
> Does torch-rbln guarantee backward compatibility between patch/minor versions? Or should we strictly bound (ex: `torch-rbln~=0.2.2 or torch-rbln~=0.2.2.dev0` for dev package) or pin the torch-rbln version within vllm-rbln, similar to how upstream vLLM handles PyTorch?
> 
> Thank you!

Thanks for raising this — it's a fair point.

It somewhat depends on the situation, and to be clear there's no formal backward-compatibility guarantee across torch-rbln patch/minor versions at the moment. That said, in practice torch-rbln and vllm-rbln have almost always been version-bumped together, and each vllm-rbln release has effectively been validated against its corresponding torch-rbln. So I think pinning with `==` is reasonable and reflects how the two have actually evolved so far.

Thanks!

rebel-minhopark (approved):
Just wondering, do you plan to stick with torch-rbln==0.2.2 until the next release, or is there any chance you'll need to use the nightly built dev versions?

LGTM, anyway!

rebel-jonghewk:
> 

> Just wondering, do you plan to stick with torch-rbln==0.2.2 until the next release, or is there any chance you'll need to use the nightly built dev versions?
> 
> LGTM, anyway!

It depends. I'll notify if there needs changes.

rebel-jinhwan:
`lmcache-rbln` ships with `vllm-rbln`, and its nightly-dev validation sometimes requires the latest torch-rbln/rebel-compiler dev packages. So version bounds look like the better approach `(e.g. torch-rbln~=0.2.2, or torch-rbln~=0.2.2.dev0 for dev package)` for me.

rebel-jonghewk:
@rebel-jinhwan @rebel-minhopark Is it okay to merge?

rebel-jinhwan:
@rebel-jonghewk yes!

PR #741, opened 2026-07-01
fix(test): raise near-tie logprob-gap tolerance to 0.3

rebel-jinhwan:
## What

`test_eagle_matches_base_generation[eagle3-32-8]` and `[eagle3-32-16]` fail with:

```
pos 11: spec=220 (' ') base_argmax=911 (' about') gap=0.2500
```

This is a genuine **near-tie** — `"mass of 1.9"` vs `"mass of about 1.9"`, both fluent — not a quality regression.

## Why the old bound was wrong

The previous tolerance (`DEFAULT_MAX_LOGPROB_GAP = 0.2`) assumed the largest bf16 near-tie argmax flip is `1 ULP = 0.125`. That only holds for argmax logit magnitudes in `[16,32)`. bf16 ULP scales with magnitude:

| magnitude | 1 ULP |
|-----------|-------|
| `[16,32)` | 0.125 |
| `[32,64)` | 0.250 |

`eagle3` produces larger-magnitude logits, where a single ULP is already `0.25` — so one legitimate ULP flip exceeds `0.2` and fails spuriously.

## Change

- `DEFAULT_MAX_LOGPROB_GAP`: `0.2 → 0.3`. Sits above the observed `0.25` (1 ULP at `[32,64)`) with f32-jitter margin, still well below the next ULP step (`0.375`) and far below the `>=~1 nat` gap a real verify-path bug would show.
- Make failing runs debuggable:
  - test prints the spec-decoded `output` text + `token_ids` per batch seq
  - `assert_spec_matches_base_within_noise` detokenizes `spec_text` / `base_argmax_text` and each offender token (`spec=220 (' ')` etc.)
- `conftest.py`: enable `VLLM_RBLN_BATCH_ATTN_OPT` for the spec_decode env.

### DIFF
--- a/tests/torch_compile/e2e/v1/spec_decode/conftest.py
+++ b/tests/torch_compile/e2e/v1/spec_decode/conftest.py
@@ -17,6 +17,7 @@ import pytest
 
 @pytest.fixture(scope="module", autouse=True)
 def common_specdec_env(monkeypatch_module):
+    monkeypatch_module.setenv("VLLM_RBLN_BATCH_ATTN_OPT", "1")
     monkeypatch_module.setenv("VLLM_RBLN_COMPILE_STRICT_MODE", "1")
     monkeypatch_module.setenv("VLLM_DISABLE_COMPILE_CACHE", "1")
     monkeypatch_module.setenv("VLLM_RBLN_ENABLE_WARM_UP", "1")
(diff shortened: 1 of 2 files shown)


rebel-clee (approved):
LGTM

PR #742, opened 2026-07-01
fix(model): force parameters/buffers contiguous at load time

rebel-jongho:
### 🚀 Summary of Changes

Make every model parameter and buffer contiguous at load time so that non-contiguous CPU tensors are never handed to the compile path, keeping the compiler from holding references to non-contiguous tensors.

Some models load a sliced view straight into a parameter — for example `gpt-oss-120b`'s MoE expert bias — which leaves that parameter non-contiguous. The loop runs in `RBLNModelRunner.load_model()` right after `self.get_model().eval()`, before `torch.compile` and where both the from-scratch load and the inplace reload converge. Doing it once at load time keeps it independent of any specific model.

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Compile a model whose weights include a non-contiguous parameter or buffer, e.g. `gpt-oss-120b`, through the vLLM Dynamo path.
2. Confirm the model loads and compiles.

---

### 📋 Checklist

---

### 💬 Notes

`p.set_(p.contiguous())` swaps only the underlying storage while keeping the Parameter object identity, so module attribute references stay valid. The optimum-rbln path (`optimum_model_runner.py`) loads already-compiled artifacts and is out of scope here.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -4196,6 +4196,13 @@ class RBLNModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
                 model_loader.load_weights(self.model, model_config=self.model_config)
 
         self.model = self.get_model().eval()
+        # weight-free compile hard-errors on non-contiguous CPU weights.
+        for p in self.model.parameters():
+            if not p.is_contiguous():
+                p.set_(p.contiguous())
+        for b in self.model.buffers():
+            if not b.is_contiguous():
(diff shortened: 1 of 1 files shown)


rebel-eunji:
It would be better to apply this logic during `warm_up_model` as well, for consistency and to avoid potential recompilation.

rebel-jongho:
Good point — the compile is actually triggered during warm-up, so I extracted the fixup into `_make_weights_contiguous()` and now call it from both `load_model` and `warm_up_model` (09822875).

rebel-eunji (approved):

PR #744, opened 2026-07-01
perf(worker): vectorize VLLM_RBLN_SORT_BATCH batch reorder

rebel-ykchoi:
### 🚀 Summary of Changes

#### Summary

On the native path (`VLLM_RBLN_USE_VLLM_MODEL=1`) with `VLLM_RBLN_SORT_BATCH=1`,
`RBLNModelRunner` re-sorts the persistent `InputBatch` by descending
`num_tokens_no_spec` every decode step so the attention backend sees a
length-ordered batch. This ran as N-1 `InputBatch.swap_states()` calls; in
steady-state decode every completion permutes the batch into a near-full
rotation, so it was a recurring per-step spike dominated by Python overhead
(`swap_states` tuple-swaps ~20 fields and copies a full `max_model_len` row).

This PR reframes the reorder as a single vectorized operation:

- **Skip when already sorted** — an O(n) non-increasing check; a stable argsort
  is identity in steady-state decode, so the argsort + reindex is skipped
  entirely on those steps.
- **Single permutation pass** applies the sort order to every per-request field
  at once instead of N-1 pairwise swaps, emitting the equivalent pairwise move
  records so logits processors realign (skipped for pooling models, matching
  `swap_states`).
- **Narrow the token-matrix reindex** to the valid token columns
  (`num_tokens + spec`) rather than the full `max_model_len` width.

No behavior change: the sorted result is identical to the `swap_states` loop.
It also fixes a latent upstream bug — `swap_states` permutes
`allowed_token_ids_mask` with an aliasing row tuple-swap, whereas the
vectorized gather is correct.

#### Structure

Mirrors upstream's reorder hook, which is a runner-method that delegates to a
free function:

- `_may_reorder_batch` stays the `RBLNModelRunner` override of
  `GPUModelRunner._may_reorder_batch` — the env / kv-group gate, the sort, and
  the move-record emission.
- The permutation itself is a free function `reorder_input_batch(input_batch,
  perm)` in `vllm_rbln/v1/worker/utils.py`, parallel to upstream
  `reorder_batch_to_split_decodes_and_prefills(input_batch, ...)`. It uses no
  runner state, so it is unit-tested directly on a real `InputBatch`.

#### Performance

The gain is concentrated in the short-output, large-batch regime: short output
sequences finish often, churning batch composition so reorders fire frequently,
and a large batch makes each reorder an O(N) swap chain — exactly what the
single vectorized pass collapses. With long outputs / stable batches the O(n)
already-sorted check skips the work in both old and new, so the benefit there
is small.

Measured on Qwen3-30B-A3B decode (P/D decode side, batch N≤32, ISL ~1k /
OSL ~100):

| metric | before | after | change |
|---|---|---|---|
| reorder overhead / step | ~234 µs | ~82 µs | −65% |
| worst-case event | 4.7 ms | 0.78 ms | −83% |

#### Compatibility

`reorder_input_batch` mirrors upstream `InputBatch.swap_states`' field set
(verified against vLLM 0.22.0) and **must be kept in sync on vLLM bumps**.
`swap_states`' field set is unchanged (req bookkeeping, per-request scalars,
`token_ids_cpu`/`is_token_ids`, `req_prompt_embeds`, block-table rows,
`request_lora_mapping`, sampling params, generators/`bad_words_token_ids`,
`allowed_token_ids_mask`); `BlockTable.swap_row` still touches only
`num_blocks_per_row` + `block_table.np`; `MoveDirectionality` is still exported
from `vllm.v1.sample.logits_processor`.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

- `test_reorder_matches_swap_states` guards equivalence over random
  permutations and every per-request field.
- Coverage for the valid-width narrowing, the `allowed_token_ids_mask` gather,
  and the pooling-model early return (no move records emitted).
- `pytest tests/torch_compile/unit/v1/worker/test_rbln_model_runner.py -k reorder` — 8 passed.

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/torch_compile/unit/v1/worker/test_rbln_model_runner.py
+++ b/tests/torch_compile/unit/v1/worker/test_rbln_model_runner.py
@@ -23,6 +23,7 @@ import numpy as np
 import pytest
 import torch
 from vllm.v1.outputs import ModelRunnerOutput
+from vllm.v1.sample.logits_processor import MoveDirectionality
 from vllm.v1.worker.gpu_input_batch import InputBatch
 
 import vllm_rbln.v1.worker.rbln_model_runner as model_runner_module
@@ -32,6 +33,7 @@ from vllm_rbln.v1.worker.rbln_model_runner import (
     ExecuteModelState,
     RBLNModelRunner,
 )
+from vllm_rbln.v1.worker.utils import reorder_input_batch
 
 # ===========================================================================
 # Helpers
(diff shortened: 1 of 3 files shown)


rebel-jongwonlee (approved):
LGTM

rebel-eunji (approved):
Great Idea! LGTM 👍

PR #749, opened 2026-07-02
feature: declare rebel-compiler dependency and add uv.lock (Part of #738)

### LINKED ISSUE #738: [RFC] uv-based rebel-compiler Dependency Management
## 1. Summary

This RFC proposes to **explicitly declare** `rebel-compiler` (and the internal package inter-dependencies) that `torch-rbln` / `lmcache-rbln` / `nixl-rbln` / `vllm-rbln` depend on, and to manage environments with `pyproject.toml` (spec) + `uv.lock` (full-graph pin).

- **Each repo declares only its direct dependencies** as range specs; transitive deps are not re-declared. The full graph is pinned by `uv.lock` (§4.2).
- **dev / main**: `~=0.11.0.dev0` (dev marker → prerelease enabled for internal packages only, tracks nightly). **release tag**: `==0.12.0` exact pin → consumer determinism. **next dev**: re-open with `~=0.12.1.dev0`.
- Both the UDC Nexus nightly index and the external index stay registered with `unsafe-best-match` (no index switching). Exact reproduction is handled by `uv.lock` at every stage. Spec transitions are mechanized via `replace_depends.py` (§4.4).

Key distinction: **lock files are for reproducing this repo's dev/CI environment and are NOT shipped to consumers** who install the published package. The only contract delivered to consumers is the dependency spec in `pyproject.toml` (→ wheel METADATA). The behavioral difference between these two is the heart of this RFC.

---

## 2. Motivation — Problems with implicit dependency management

`rebel-compiler` is currently not declared as a dependency; developers/CI install it manually. The resulting costs:

1. **Non-reproducible environments.** Nowhere in the code is it recorded "which `rebel-compiler` version is this `torch-rbln`/`vllm-rbln` compatible with". Different versions get installed per developer / per CI machine, so "works on my machine" happens constantly.

2. **Compatibility breaks surface only at runtime.** When `rebel-compiler`'s ABI/API changes, there's no dependency spec to block it at install/resolve time. It blows up while `vllm-rbln` is running.

3. **CI flakiness / non-reproducibility.** Nightly changes daily, and with no version pin, a build that passed yesterday fails today. You cannot tell whether the failure is from a code change or compiler nightly drift.

4. **Manual-install dependence.** Missing `pip install rebel-compiler` / wrong version / install-order dependence → higher setup and onboarding cost.

5. **Transitive dependency drift.** Without a lock, even the sub-dependencies `rebel-compiler` pulls in shift on every install.

6. **No traceability / auditability.** There's no record of which `rebel-compiler` version a given `vllm-rbln` build was bound to, making regression bisect, SBOM, and release notes impossible.

**Bottom line: not pinning dependencies in code loses build-wide reproducibility and CI trust.**

### 2.1 Measured against the actual codebase — the runtime dependency gap

Verified against the real repos, the biggest problem is **the gap of importing at runtime without declaring a runtime dependency**.

**torch-rbln** (partially managed):
- `rebel-compiler` is **absent** from `[project.dependencies]` (runtime).
- It IS in `[build-system].requires` and `[dependency-groups].build` as `rebel-compiler>=0.10.2,<0.20.0` (build/dev only).
- Yet `torch_rbln/__init__.py`, `_internal/*` **import `rebel_compiler` at runtime**.
- `[tool.uv.sources]` (`rebel-compiler = {index="rbln"}`), `uv.lock`, and `constraints-build-dev.txt` (pinning `rebel-compiler==0.10.3.dev27+g79bf6892`) do exist.
- → Build/CI work, but **a consumer who receives the published wheel does not automatically get `rebel-compiler`.** The runtime import raises `ImportError`, or relies on whatever version the consumer manually installed.

**vllm-rbln** (barely managed):
- `[project.dependencies]` has **neither `rebel-compiler` nor `torch-rbln`** (only `optimum-rbln`, `vllm==0.18.0+cpu`, etc.).
- **No `rbln` index** defined, **no `uv.lock`**.
- Yet `vllm_rbln/platform.py`, `v1/worker/*` etc. **import `rebel_compiler` at runtime**.
- → Spec, index, and lock are all absent. Fully implicit dependency.

**Where the version binding actually lives (outside the repo):**
- torch-rbln: `constraints-build-dev.txt` + fsw-integration `torch-rbln-resolve-versions.sh`.
- vllm-rbln: the **resolve → snapshot → replay** flow below + fsw-integration CI **runtime-injecting** via `uv add` / `uv pip install` / Dockerfile.
- prod pin: fsw-integration `job-mlfi-publish-pair.yaml` pins the triple into the `rblness` repo.

**vllm-rbln resolve flow (floating → snapshot → replay):**

1. **resolve (the decision):** `on-vllm-rbln-pr.yaml` / `on-vllm-rbln-nightly.yaml` → `scripts/resolve_rebel_compiler.py` picks the **latest from the Nexus `pypi-compiler-nightly-dev` simple index, floating** (prereleases included; highest version by PEP 440 ordering with `+local` ignored).
2. **snapshot (the record):** when the nightly build/test **succeeds**, `on-vllm-rbln-nightly.yaml` (L273~) **writes back** the result into `shared-ci-configs/environments/vllm_rbln-nightly.yaml` (`sed -i` updates `vllm_rbln:`/`rebel_compiler:`, then commits). So the env yaml is not an input but the **last-green snapshot (an output)**.
3. **replay (consumption):** other flows (`on-vllm-rbln-guidellm`, `flow-vllm-rbln-build-from-env`, executor, …) **read** this env yaml to reproduce the known-good combination.

The core problem is not "two conflicting sources of truth" but that **the real decision-maker is a single floating resolve**:
- The decision happens at CI runtime with no pin/lock, so running a PR twice a day apart can pick different rebel-compiler versions — the exact mechanism behind §2's "CI flaky / non-reproducible".
- That decision **is recorded nowhere in the vllm-rbln repo or lock**; only as an after-the-fact snapshot in a separate repo (shared-ci-configs). Source-code changes and dependency changes are not tied to the same commit / the same review.

In short, the **decision** of "which rebel-compiler is this vllm-rbln/torch-rbln compatible with" lives in a CI-runtime floating resolve rather than the package's own metadata, and the **record** survives only as an external-repo snapshot. Wheel consumers (outside CI) cannot see this information. This is the core gap this RFC fills.

> Note: the `rblness` repo already implements 3-way sync of `pyproject.toml` + `uv.lock` + `pylock.toml` (enforced by `check-pylock-sync.sh`) → it is the **internal reference** for the pattern this RFC proposes. This is not a new design but an extension of that pattern into torch-rbln/vllm-rbln proper.

### 2.2 Field symptoms — who is inconvenienced, and how, right now

How the structural problems above manifest day to day. All stem from the same root (dependencies scattered across CI/external repos instead of metadata/lock + floating resolve).

**Developers (local development):**
- Right after `clone`, `import rebel_compiler` → `ImportError`. **Where to get it and which version is not written in the repo** (no runtime `[project.dependencies]`) → you learn it verbally/from docs and install manually. No one-line `uv sync` setup.
- Locally a hand-installed arbitrary version, in CI a floating-resolve + Docker injection → **environment mismatch.** "Works locally but breaks in CI" (or vice versa) happens structurally.
- To answer "which compiler must my branch align with right now?" you must **cross-check** `constraints-build-dev.txt` / `shared-ci-configs` env yaml / CI logs. No single source.

**CI (PR / nightly):**
- **Re-running a PR without changing a single line of code** can change the result, because floating resolve picks a different nightly → you cannot tell whether red is your change or compiler drift. "Retry" becomes a debugging tool — an anomaly.
- A nightly green means "the latest combination at that instant passed", **not a guarantee that yesterday's green reproduces today.**
- On regression, you **cannot bisect which compiler nightly broke it** — the used version isn't pinned in the commit/lock, so you reverse-engineer CI logs.

**Bumping versions (manual toil):**
- Raising one compiler lower bound requires a human to **manually sync** scattered spots across 4 repos + `constraints-build-dev.txt` + env yaml + `replace_depends.py` + Dockerfile. Missing one → a silent mismatch → back to the CI/local symptoms above.

**Cloning rebel-compiler source / editable install (when co-developing the compiler):**
- **Version unidentifiable.** A source build makes setuptools-scm produce a local segment like `0.11.1.dev5+g{hash}` or `+dirty`. PEP 440 comparison ignores the local segment, so it doesn't match the spec/lock, and **even if it differs from the Nexus version the lock pinned, it is not tracked.**
- **Lock divergence.** An editable/path install bypasses `uv sync`'s lock management → only that developer runs different code. Team uniformity/reproducibility breaks even harder (the ultimate "works on my machine").
- **Header↔runtime ABI mismatch.** torch-rbln builds against `third_party/rebel_compiler` headers. If a locally built compiler (different commit/toolchain/build options) and the header version diverge, you get subtle runtime bugs.
- **Invisible upstream dependency.** Working on torch/vllm against an unmerged/unreleased rebel-compiler branch means the upstream work breaks unless that compiler lands. This upstream dependency appears nowhere in the code.
- **`file://` pin leakage risk (real).** torch-rbln's pyproject already allows `file://` direct references (pinned by `replace_depends.py`, noted in a pyproject comment). If a local path is accidentally committed, **other developers'/CI builds depend on that developer's local path** and break.

**Release / audit:**
- The shipped wheel **has no record in its metadata** of which compiler it was bound to → `ImportError` in customer environments; writing SBOMs/release notes requires reverse-engineering CI logs (§2-6).

> Summary: the reason it "works" today is not that the system is robust but that **a human patches it manually every time.** That manual patching, cross-checking, and retry-debugging is the real cost this RFC removes.

---

## 3. Role of each file

| File | Role | In the published wheel? | Delivered to consumers? |
|---|---|---|---|
| `pyproject.toml` `[project.dependencies]` | Abstract dependency **spec** (`==`, `~=`, …). Frozen into wheel METADATA (`Requires-Dist`) at build time | **Yes (as METADATA)** | **Yes** — the only contract the consumer's resolver reads |
| `uv.lock` | uv-native lockfile. Pins the full dependency graph to exact versions+hashes. Used by `uv sync`/`uv lock` | No | No (repo-internal dev/CI only) |
| `pylock.toml` | PEP 751 standard lockfile. Tool-agnostic. Produced by `uv export -o pylock.toml`; installable by any PEP 751 tool | No | Optional (CI / cross-tool sharing) |

Core principles:
- **Lock files (`uv.lock`/`pylock.toml`) are not included in the wheel.** They therefore do not determine the consumer's install result; they only reproduce this repo's dev/CI environment.
- **The only thing delivered to consumers is the `[project.dependencies]` spec.** dev/main use `~=...dev0` (range); **only release tags use `==` (exact)** → the release wheel is deterministic for consumers too. Internal reproduction is handled by `uv.lock` at every stage (§4.4).

### uv.lock vs pylock.toml
- `uv.lock`: uv-native. More expressive (dependency graph, markers, …) and the single source of truth for `uv sync`. **Recommended as this repo's source of truth.**
- `pylock.toml`: PEP 751 standard exchange format. Export with `uv export -o pylock.toml`; install with `uv pip install -r pylock.toml` / `uv pip sync pylock.toml`. Some uv features can't be expressed, so it cannot replace `uv.lock` → use only as an **export artifact** (e.g. to hand to a CI step using a different tool).

Sources: [uv export](https://docs.astral.sh/uv/concepts/projects/export/), [PEP 751](https://peps.python.org/pep-0751/), [uv #12584](https://github.com/astral-sh/uv/issues/12584)

---

### 4.1 Dependency graph

```mermaid
graph TD
    vllm[vllm-rbln]
    lmcache[lmcache-rbln]
    nixl[nixl-rbln]
    torch[torch-rbln]
    rc[rebel-compiler]

    vllm --> lmcache
    vllm --> torch
    vllm --> nixl
    vllm --> rc
    lmcache --> torch
    lmcache --> rc
    nixl --> rc
    nixl --> torch
    torch --> rc

    classDef leaf fill:#ffe6e6,stroke:#cc0000;
    class rc leaf;
```

Arrow = "depends on" (`A --> B`: A uses B). `rebel-compiler` (red) is the bottom leaf; `vllm-rbln` is the top consumer. All internal packages are published by nightly CI to the **UDC Nexus index** (the dev/main baseline index = Nexus).

### 4.2 Principle: declare direct deps only, lock the whole graph

- **`[project.dependencies]` declares only the packages each repo directly imports/uses**, as range specs.
- **Transitive (indirect) deps are not re-declared.** The resolver pulls them by following the direct deps' METADATA (§5.2).
- **`uv.lock` pins the whole graph (direct + transitive) to exact versions+hashes** → this is what actually unifies the dev environment.

| repo | declared in `[project.dependencies]` (direct) | pinned by `uv.lock` (direct + transitive) |
|---|---|---|
| `torch-rbln` | `rebel-compiler` | rebel-compiler |
| `lmcache-rbln` | `torch-rbln`, `rebel-compiler` | torch-rbln, rebel-compiler |
| `nixl-rbln` | `rebel-compiler`, `torch-rbln` | rebel-compiler, torch-rbln |
| `vllm-rbln` | `lmcache-rbln`, `torch-rbln`, `nixl-rbln`, `rebel-compiler` | lmcache-rbln, torch-rbln, nixl-rbln, rebel-compiler |

> **`vllm-rbln` declares `rebel-compiler` directly too.** Since `vllm_rbln/platform.py`, `v1/worker/*` etc. import `rebel_compiler` directly at runtime (§2.1), put it **directly** in `[project.dependencies]` rather than relying on transitivity. The "if you import it, declare it" rule — transitive availability can disappear if an upstream package (torch-rbln/lmcache-rbln/nixl-rbln) changes its spec, so it's not trustworthy.

### 4.3 Version bump policy (dev branch)

Core: **rarely change the spec; track nightly day to day via `uv.lock` only.**

1. **Routine (most of the time):** don't touch `pyproject.toml`.
   ```bash
   uv lock --upgrade-package rebel-compiler --upgrade-package torch-rbln ...
   ```
   → Absorb the latest nightly into the lock, leaving the spec as is. Commit the lock to sync the whole team.

2. **Spec (lower-bound) change (exceptional):** **only when runtime API compatibility breaks.**
   e.g. `torch-rbln` starts using a `rebel-compiler` runtime API added in a specific nightly (`0.11.1.dev5`) → raise the lower bound to that nightly:
   ```toml
   # before (dev baseline — dev marker tracks nightly + enables per-package prerelease)
   dependencies = ["rebel-compiler~=0.11.0.dev0"]  # >=0.11.0.dev0, <0.12.0
   # after — only when 0.11.1.dev5's new API becomes required
   dependencies = ["rebel-compiler~=0.11.1.dev5"]  # >=0.11.1.dev5, <0.12.0
   ```
   Change both `pyproject.toml` + `uv.lock` only in this case. Otherwise lock-only.

   > **The dev baseline must carry a dev marker.** `~=0.11.0` (no marker) does not enable prereleases under `if-necessary-or-explicit`, so it **won't pick nightlies**. A marker like `~=0.11.0.dev0` is needed to pick nightlies, and that prerelease allowance is scoped to the single `rebel-compiler` package. (Even `0.11.1dev` without the dot/number normalizes to `0.11.1.dev0` per PEP 440, but write the canonical `~=0.11.1.dev0`. Only the local `+g..` segment is forbidden in `~=`.)

3. **rebel-compiler source co-development (local checkout):** when you must use a local build instead of the Nexus wheel, do it **as a one-off override in your personal venv only**. Do not commit.
   ```bash
   uv pip install -e ../rebel-compiler   # current venv only; lock/pyproject unchanged
   ```
   - **Do not commit** a `{ path = ... }` / `file://` pin in `pyproject.toml`'s `[tool.uv.sources]` (§2.2's file:// leakage risk). If truly needed, declare it as a uv workspace and put it through review.
   - CI always runs `uv sync --locked` (§6) so this override cannot leak in → the source checkout becomes an **explicit, one-off override rather than an invisible default**. When done, return to the lock-based state with `uv sync`.

   > **Caution — no local segment (`+g8d9ab0a`) in a spec:** PEP 440 **does not permit** local versions (`+...`) in `~=` (or any prefix match) (`~=0.11.1.dev5+g8d9ab0a` is invalid). The dev suffix is also ignored in the prefix decision, so `~=0.11.1.dev5` = `>=0.11.1.dev5, ==0.11.*`. The spec only expresses "this nightly **or above**"; **pinning to exactly that `+g8d9ab0a` nightly is `uv.lock`'s job** (`uv lock --upgrade-package rebel-compiler` writes `0.11.1.dev5+g8d9ab0a` into the lock).

> **Caution (nightly = pre-release):** rebel-compiler nightly/pre versions arrive in these PEP 440 forms:
> - `.devN` — `0.10.4.dev99+g2b21b01b`. `.dev{distance from last tag}+g{commit hash}` (setuptools-scm).
> - `aN` — alpha (e.g. `0.10.3a6.dev7+g35de6908e`), `rcN` — release candidate.
>
> All are pre-releases, so range specs alone **won't pick them by default** (uv/pip exclude pre-releases by default).
>
> **Do not open prereleases globally with `"allow"` — allow them for internally managed packages only.** uv has no per-package prerelease setting; **if a specifier contains a prerelease marker, prereleases are allowed for that package only** (PEP 440 behavior). So:
> ```toml
> [tool.uv]
> prerelease = "if-necessary-or-explicit"   # no global "allow" — torch-rbln's current setting
> ```
> ```toml
> # put a dev marker only on internal packages to enable prerelease per-package
> dependencies = [
>   "rebel-compiler>=0.11.0.dev,<0.12",   # dev marker → prerelease for this package only
>   "some-public-dep>=2.0",               # no marker → stable only (unaffected)
> ]
> ```
> Global `"allow"` is dangerous because it lets **every** dependency pull prereleases. `if-necessary-or-explicit` + a dev marker on internal-package specs confines prereleases to the intended internal packages such as rebel-compiler.
>
> Since the `+g{hash}` local segment is not used for comparison/ordering, exact reproduction is `uv.lock`'s job, not the spec's — which is why committing the lock is mandatory in dev.

### 4.4 dev → main → release transition

**Index is NOT switched per stage.** The UDC Nexus nightly index and the external (release) index are **both registered**, and `unsafe-best-match` picks the best version satisfying the spec from either. What changes across stages is **the spec only** (dev `~=...dev0` ↔ release `==`); the index configuration is fixed. Which index's version is actually selected is decided by the spec (operator + prerelease marker) — dev picks nightly, release `==` picks the external stable.

| Stage | spec | index config (fixed) | what gets selected | lock |
|---|---|---|---|---|
| **dev branch** | `~=0.11.0.dev0` (dev marker → nightly·prerelease) | Nexus nightly + external both registered, best-match | latest nightly `.devN` | track via `uv lock --upgrade-package`, commit |
| **main branch** | same (`~=...dev0`) | same | the dev-validated version | freeze the dev lock as-is |
| **release tag** | **`==0.12.0` exact pin** | same (both registered) | the external stable `0.12.0` | freeze the release version |

- **Only when cutting a release tag do you `==`-pin.** `==0.12.0` has no prerelease marker, so it won't pick nightlies and fetches exactly `0.12.0` from whichever registered index (external) has it. The wheel's `Requires-Dist: rebel-compiler==0.12.0` → **deterministic for external consumers too**.
- **At the start of the next dev phase, re-open the dev branch with `~=x.y.z.dev0`** (re-attach the marker → resume nightly tracking).
- Automate with `torch-rbln`'s `tools/replace_depends.py`: at release, `--target "rebel-compiler==0.12.0"`; at the next dev start, `--target "rebel-compiler~=0.12.1.dev0"` (or the relevant next line). `--set-index` is unnecessary (both are always registered).

> **Spec transition summary:** `~=0.11.0.dev0` (dev/main) → `==0.12.0` (release tag) → `~=0.12.1.dev0` (next dev). Release nails it with `==` to give consumers determinism; dev re-opens nightly with `~=...dev0`. Since the operator toggles `~=`↔`==`, mechanize the transition via `replace_depends.py` (no manual editing). Exact version pinning is handled in parallel by `uv.lock` at every stage.

#### Index configuration (best-match)

```toml
[tool.uv]
index-strategy = "unsafe-best-match"   # search all registered indexes for the best version satisfying the spec

[[tool.uv.index]]
name = "rbln-nightly"                   # UDC Nexus (dev nightly)
url  = "https://nexus.internal.../pypi-compiler-nightly-dev/simple"

[[tool.uv.index]]
name = "rbln-release"                   # external release index
url  = "https://.../release/simple"
```

- `unsafe-best-match`: looks at both indexes and picks the highest version satisfying the spec → you can keep nightly/release registered simultaneously and control selection purely via spec + prerelease policy.
- **Security trade-off (must be understood):** best-match has a **wider dependency-confusion attack surface** than `first-index` (torch-rbln's current, secure setting). In particular, if the default public `pypi` is also registered, a same-named package appearing on public PyPI could be selected. Mitigations:
  - Assume internal packages like `rebel-compiler` do not exist on public PyPI, but consider a config that **excludes the public index from best-match search** (e.g. keep only the two internal indexes non-explicit, and mark the public one `explicit = true` so it's used only on direct request).
  - Because rebel-compiler must come from one of the two (nightly/release) indexes, a single-index `[tool.uv.sources]` explicit pin is unusable (a pin names exactly one index). So the residual confusion risk from relying on best-match must be explicitly accepted/controlled.

> **Consumer determinism note:** even at release, `==` ships in `Requires-Dist`, so external consumers get exactly that version. For dev/main wheels the spec is `~=` (a range), so consumers there get the latest within range. This model deliberately assigns exact reproduction to `uv.lock` (internal CI/dev) + curated index contents. If byte-level determinism is needed even for arbitrary external consumers, ship a pinned export via `uv export` alongside that distribution channel.

---

### 5.1 `uv sync` logic during development (this repo's dev environment)

On `uv sync`:

1. Read `pyproject.toml` (`[project.dependencies]`, dependency-groups, index config) and `uv.lock`.
2. **Consistency check**: if `uv.lock` matches `pyproject.toml` and is current → **skip resolution** and install exactly the versions recorded in the lock.
3. If the lock is stale (mismatch) → re-resolve, update the lock, then install. Except:
   - `--frozen`: **error** if the lock is missing or stale; don't touch the lock.
   - `--locked`: **assert** the lock is current (fail otherwise). Recommended for CI.
4. Install the **whole graph (incl. transitive)** at the lock's exact versions+hashes into `.venv`. The project itself is installed editable.
5. Package locations are found via the index config (`[[tool.uv.index]]`, env) and the source recorded in the lock is reused.

Key point: **in dev, `uv.lock` is the final arbiter.** The spec (`~=`) only sets "what range is allowed"; the version actually installed is nailed by the lock. → Everyone on the team gets the same environment.

### 5.2 Consumer install logic after publishing (`pip install` / `uv pip install`)

When a consumer runs `pip install vllm-rbln`:

1. The resolver reads `vllm-rbln`'s wheel/sdist **METADATA (`Requires-Dist`)**, frozen from `[project.dependencies]` at build time.
2. It follows recursively: `Requires-Dist: torch-rbln <spec>` → torch-rbln METADATA → `Requires-Dist: rebel-compiler <spec>`.
3. It selects and installs the **latest compatible version** satisfying all specs simultaneously.
4. **This repo's `uv.lock` / `pylock.toml` / `tool.uv.sources` are NOT delivered to the consumer.** The consumer's environment (or the consumer's own lock) decides exact versions.
5. So the consumer must **configure the index that hosts `rebel-compiler` in their own environment** (index info is not in the wheel METADATA).

Key point: **at distribution time, only the spec is the contract** (lock is not delivered).
- **dev/main** wheels: `~=0.11.0.dev0` (range) → consumers get the latest within range. For internal integration.
- **release tag** wheels: `==0.12.0` (exact) → consumers get exactly `0.12.0`. **Deterministic.**
- Byte-level reproduction for internal CI/dev is handled in parallel by `uv.lock` at every stage.

### 5.3 Side-by-side summary

| | Development (`uv sync`) | Consumer install after publish |
|---|---|---|
| Who decides the version | this repo's `uv.lock` | consumer env + each package's METADATA spec |
| Files used | `pyproject.toml` + `uv.lock` | wheel METADATA (`Requires-Dist`) only |
| Lock delivered? | repo-internal only | not delivered |
| Index config | this repo's uv settings | consumer configures their own |

---

### Developers (dev branch)
1. **Routine nightly tracking:** don't touch `pyproject.toml`; `uv lock --upgrade-package rebel-compiler [...]` → commit `uv.lock`. (§4.3)
2. **Change the spec only when runtime API compatibility breaks:** raise the lower bound and commit both `pyproject.toml` + `uv.lock`.
3. Sync the environment with `uv sync`. No manual `pip install`.

### CI
- Install lock-as-is with `uv sync --locked` (or `--frozen`). If the lock and `pyproject.toml` diverge, fail the build to catch drift early. → Replaces the current floating resolve (`resolve_rebel_compiler.py`).
- On a release tag, keep `uv.lock` (and a `pylock.toml` export if needed) as a release artifact → traceability of the build–compiler version binding.

### Stage policy summary

| Stage | spec | index | lock role |
|---|---|---|---|
| `dev` branch | `~=0.11.0.dev0` (dev marker on internal packages only; global `if-necessary-or-explicit`) | nightly + external both, best-match | track nightly via `--upgrade-package`, unify the team |
| `main` branch | same (`~=...dev0`) | same | freeze the dev-validated lock, integration reproduction |
| release tag | `==0.12.0` exact pin (`replace_depends.py`) | same (both registered) | freeze the release version, consumer determinism |
| (next dev) | re-open with `~=0.12.1.dev0` | same | resume nightly tracking |

---

## 7. Alternatives

- **Keep current state (implicit manual install):** all the costs of §2. **Not recommended**.
- **Keep `~=` at release too (lower-bound bump + drop the marker):** avoids the dev↔release operator toggle, but cannot give external consumers wheel-level determinism. This model instead **nails release with `==`** for consumer determinism and re-opens dev with `~=...dev0`. The `~=`↔`==` transition is mechanized via `replace_depends.py`, so there's no manual-edit burden.
- **Don't commit the lock:** loses environment unification / CI reproducibility. Conflicts with the dev branch's purpose.

rebel-minhopark:
### 🚀 Summary of Changes

RFC #738 Phase 1 — declare `rebel-compiler` as a project dependency and lock the full resolution.

**What this PR does:**
- Adds `rebel-compiler` dependency spec to `pyproject.toml` with 3 internal index sources (Nexus nightly, Nexus group, internal PyPI)
- Commits `uv.lock` with fully resolved, pinned dependency graph
- Documents the uv-based developer setup in `README.md`

**Locked `rebel-compiler` version:** `0.11.1.dev160+g7bd15fb8`

**Verification:** Ran `uv sync --locked` inside a clean Ubuntu 22.04 Docker container; `import rebel` succeeded with no errors.

**Known constraints:**
- `rebel-compiler` nightly wheels are cp312-only; Python 3.12 is required for internal dev setup
- Most third-party packages resolve through the Nexus group index (`unsafe-best-match` — intentional design decision, see RFC §4.4)
- Installing without `--locked` is outside the defined contract

**CI impact:** No changes to existing CI workflows. Floating resolve continues to install versions within the declared spec range. Lock-based CI switchover is scheduled for Phase 2 (fsw-integration).

---

* Part of #738

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

1. Set internal index credentials in your shell (see `README.md` — Development setup)
2. Run `uv sync --locked` from repo root on Linux x86_64 + Python 3.12
3. Verify: `python -c "import rebel; print(rebel.__version__)"` prints `0.11.1.dev160+g7bd15fb8`

---

### 📋 Checklist

---

### 💬 Notes

Phase 1 only — `uv.lock` is committed and the dependency is declared. CI still uses floating resolve; Phase 2 will switch PR/nightly workflows to `uv sync --locked`. Do not close #738 until Phase 2 merges.

## Verification (lock-based CI, fsw-integration#790)

The uv.lock-based CI flow was verified end-to-end against this PR's head (`8b66056a`) using fsw-integration#790 — resolve (rebel-compiler `0.11.1.dev160+g7bd15fb8` + torch-rbln `0.2.3.dev36+g9e7b752` from this repo's committed uv.lock) → wheel build → runtime image (`uv sync --locked`) → NPU unit pytest, **all green**: [verification comment](https://github.com/rebellions-sw/fsw-integration/pull/790#issuecomment-4874608419) ([run 28649103949](https://github.com/rebellions-sw/fsw-integration/actions/runs/28649103949)).

> The remaining red `vLLM-RBLN PR CI` check on this PR runs fsw-integration's **old (dev-branch) flow**, which cannot handle the new wheel metadata — it turns green once fsw-integration#790 merges (merge order: this PR first, then #790).

**Real-world validation (2026-07-06):** a rebel-compiler nightly regression (`dev208+`) turned every floating-version PR CI red, while this PR stayed green on the lock-pinned `dev160` after `f1c6beb0` made the lock the default — see the [case study comment](https://github.com/RBLN-SW/vllm-rbln/pull/749#issuecomment-4888497590).


rebel-minhopark:
## Case study: uv.lock shielded this PR's CI from a compiler nightly regression (2026-07-06)

Since 07-04, rebel-compiler nightly `0.11.1.dev208+` carries a regression (the engine core dies without a traceback during computation-graph compilation; `tests/entrypoints/llm/test_default_max_num_seqs.py` and friends), so **every PR CI using a floating compiler version is red**:

| Branch | How rebel-compiler is decided | Result |
|---|---|---|
| [feat/mm-pf](https://github.com/RBLN-SW/vllm-rbln/actions/runs/28731170518) and all other PRs | floating (artifact pin `dev208`/`dev209`) | ❌ red |
| This PR, before removing the fallback ([run](https://github.com/RBLN-SW/vllm-rbln/actions/runs/28761187367)) | trigger fallback overrode the lock (`dev209`) | ❌ red |
| This PR, after removing the fallback ([run](https://github.com/RBLN-SW/vllm-rbln/actions/runs/28762233752)) | **uv.lock pin (`0.11.1.dev160+g7bd15fb8`)** | ✅ **all green** |

`f1c6beb0` removed the trigger's get-compiler-version fallback so the lock is the default — CI went green immediately. This is a live demonstration of exactly what RFC #738 §2 describes ("floating resolve breaks CI trust / retry becomes a debugging tool"), and of what the lock buys us:

- A broken nightly cannot affect this PR's CI unless someone explicitly bumps the lock (isolation).
- The `dev208+` compiler regression itself needs a separate report — other PRs are still blocked by it.

rebel-jinhwan (approved):

rebel-cjlee (approved):

rebel-minhopark:
## rebel-compiler lock pin history (for reviewers)

The locked `rebel-compiler` version moved a few times in this PR — each bump has a concrete reason:

| Lock pin | Commit | Why |
|---|---|---|
| `0.11.1.dev160+g7bd15fb8` | `84c2098f` | Initial lock (latest nightly at lock time; stayed green through the `dev208+` nightly regression that broke all floating-version PRs) |
| `0.11.1.dev224+gb0b1b8d9` | `819d715f` | Routine bump to a newer validated nightly |
| `0.11.1.dev304+g6948edb6.prod` | `c6f3637c` | **Required by torch-rbln**: dev brought `torch-rbln~=0.3.0rc0` (#765), and torch-rbln `0.3.0rc1` is built against rebel-compiler `0.11.1.dev304+g6948edb6.prod` per its [`constraints-build-dev.txt`](https://github.com/RBLN-SW/torch-rbln/blob/dev/constraints-build-dev.txt). torch-rbln does not (yet) declare rebel-compiler as a runtime dependency, so the pairing must be kept in sync manually in the lock. |

> Interim rule until this is structural: when bumping `torch-rbln`, check its `constraints-build-dev.txt` and bump `rebel-compiler` to the matching version in the same lock update.
>
> Planned follow-up removes the manual pairing: torch-rbln will declare rebel-compiler in an optional extra, and vllm-rbln's `runtime` extra will reference `torch-rbln[runtime]`, so the resolver enforces the pairing automatically.

pyproject.toml
```diff
@@ -143,6 +156,18 @@ name = "pytorch-cpu"
...
+url = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/simple/"
+
+[[tool.uv.index]]
+name = "rebellions"
+url = "https://pypi.rebellions.in/simple/"
+
+[[tool.uv.index]]
+name = "rbln-release"
+url = "https://pypi.rbln.ai/simple/"
+
```
rebel-thkim, line 170: [uv.index] is defined, but it isn't being used in the [tool.uv.source]. Where is uv.index used?

rebel-thkim:
It seems the meeting’s content has been well reflected, and I don’t think there will be any issues from the SDK team’s perspective. 👍 

One thing I'm wondering is whether "rbln-release" needs to be specified in `uv.index`; since we've already defined `uv.index` for internal pypi url, shouldn't it be used in `uv.source` instead?

rebel-thkim:
Since these changes need to be reflected in our CI system as well, could you please hold off on the merge until the SDK team’s CI is ready?

pyproject.toml
rebel-minhopark: Hello @rebel-thkim? Thank you for your review!

Here is the guide how to use `tool.uv.source` and `tool.uv.index` https://docs.astral.sh/uv/concepts/indexes/#pinning-a-package-to-an-index

I use a index strategy as a `unsafe-best-match` which means if developer wants to use `rebel-compiler==0.11.1.devN+g{commit_hash}`, `uv` finds that package from all indices.

And I think if we pin a rebel-compiler as a official version (ex : `0.11.1`), uv can find a package only in the `rbln-release`

rebel-seinpark (approved):
LGTM!

.github/workflows/rbln_optimum_ci.yaml
```diff
@@ -88,62 +84,55 @@ jobs:
...
           --max_seq_len 32768
           --num_input_prompt $((2 * ${{ env.NUM_INPUT_PROMPT }}))
           --model ${{ env.REBEL_VLLM_PRE_COMPILED_DIR }}/qwen3-reranker-0_6b-40k_kv4k_batch2
 
       - name: Run gemma2 model (hybrid attention) (V1)
         run: >
-          python3 examples/optimum/decoder_models/basic_offline.py
-          --model ${{ env.REBEL_VLLM_PRE_COMPILED_DIR }}/gemma-2-2b-8k_batch4
+          uv run --no-sync examples/optimum/decoder_models/basic_offline.py
+          --model_id ${{ env.REBEL_VLLM_PRE_COMPILED_DIR }}/gemma-2-2b-8k_batch4
```
rebel-seinpark:
```suggestion
          --model ${{ env.REBEL_VLLM_PRE_COMPILED_DIR }}/gemma-2-2b-8k_batch4
```

PR #753, opened 2026-07-05
feature: Support automatic prefix caching for multimodal models

rebel-eunji:
### 🚀 Summary of Changes

1. Support for APC
This PR enables APC (Automatic Prefix Caching) for multimodal models that use simple full attention.

2. Skip generating multimodal embeddings on prefix cache hits
When a prefix cache hit occurs, multimodal embedding generation is skipped, avoiding redundant computation.

3. Fix the logic of validating the model is using sliding window
The configuration of multimodal models usually have sliding window related parameters in the nested text_config
- https://huggingface.co/google/gemma-3-4b-it/blob/main/config.json
- https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B/blob/main/config.json

4. Refactor runner and Qwen-VL codes
- qwen_vl.py file was so messy because there are APC, EC, and special handling for multimodal(because of deepstack). So I've split the logics of qwen3_vl to qwen3_vl.py
- I reordered the methods of runner and qwen_vl codes to know the workflow easily

#### Qwen-VL family: model flow

**Class hierarchy**

```mermaid
classDiagram
    class RBLNOptimumQwenVLForConditionalGeneration {
      <<abstract>>
      owns shared prefill/decode flow
    }
    RBLNOptimumQwenVLForConditionalGeneration <|-- RBLNOptimumQwen2VLForConditionalGeneration
    RBLNOptimumQwenVLForConditionalGeneration <|-- RBLNOptimumQwen2_5_VLForConditionalGeneration
    RBLNOptimumQwen2_5_VLForConditionalGeneration <|-- RBLNOptimumQwen3VLForConditionalGeneration
    RBLNOptimumQwen3VLForConditionalGeneration <|-- RBLNOptimumQwen3VLMoeForConditionalGeneration
```

The base owns the whole prefill/decode flow; subclasses only fill in per-model hooks (`_create_*`, `_add_model_specific_args`) and Qwen3's deepstack.

**Prefill flow**

`build_prefill_forward_inputs` (owned by the base mixin) dispatches on whether `partial_prefix` is set; the EC consumer enters through `build_prefill_inputs_from_cache`. All four entry points converge on the same building blocks.

```mermaid
flowchart TD
    RUN["runner: execute_model"] --> K{"prefill path?"}
    K -->|"non-EC"| BPF["build_prefill_forward_inputs<br/>(base dispatch)"]
    K -->|"EC consumer"| BPC["build_prefill_inputs_from_cache"]

    BPF -->|"no hit"| FULL["_build_full_prefill_forward_inputs"]
    BPF -->|"partial hit"| PART["_build_partial_prefill_forward_inputs"]
    BPC -->|"no hit"| ECF["(full branch)"]
    BPC -->|"partial hit"| ECP["_build_partial_prefill_inputs_from_cache"]

    FULL --> EMB["embed_multimodal<br/>(vision encoder)"]
    PART --> EMB
    ECF --> C2M["_cache_to_mm<br/>(encoder cache)"]
    ECP --> C2M

    EMB --> MM["whole-prompt mm dict<br/>{embeds, grid_thw[, deepstack]}"]
    C2M --> MM
    MM -->|"partial hit only"| SLICE["_build_partial_mm_embeds<br/>(tail slice)"]
    MM --> SCAT["embed_input_ids<br/>(scatter over placeholders)"]
    SLICE --> SCAT
    SCAT --> POS["_build_prefill_position_embed<br/>MRoPE over full prompt, slice [num_cached:]"]
    POS --> OUT["prefill decoder inputs"]
```

| Path | Entry point | Feature source | Tail slice | Positions |
|---|---|---|---|---|
| non-EC, full | `_build_full_prefill_forward_inputs` | `embed_multimodal` (encoder) | — | `_build_prefill_position_embed` |
| non-EC, partial hit | `_build_partial_prefill_forward_inputs` | `embed_multimodal` | `_build_partial_mm_embeds` | `_build_prefill_position_embed` |
| EC, full | `build_prefill_inputs_from_cache` | `_cache_to_mm` (cache) | — | `_build_prefill_position_embed` |
| EC, partial hit | `_build_partial_prefill_inputs_from_cache` | `_cache_to_mm` | `_build_partial_mm_embeds` | `_build_prefill_position_embed` |

`_slice_to_tail` / `_mm_feature_counts` are the slice/count helpers; `_build_prefill_position_embed` delegates to `_compute_mrope_position` (runs `get_rope_index` with the encoder skipped).

**EC (encoder-cache disaggregation)**

- **Producer** (`_run_encoder_and_save`, runner): encodes via `model.embed_multimodal` and caches the result.
- **Consumer**: has no vision-encoder runtime; gathers the cached embeds through `_cache_to_mm` and reuses the flow above.
- **Enabled for**: `RBLNQwen3VLForConditionalGeneration` only (`ec_enabled_model` gate). For the other Qwen models, `embed_multimodal` is reachable solely as the EC producer.

**Decode**

The decode branch of `forward` calls `compute_decode_position_embed`, which advances each request's position from its stored `rope_deltas`, then runs the decoder.

**Per-model differences**

| | `_create_*` input types | Temporal MRoPE (`second_per_grid_ts`) | Deepstack |
|---|---|---|---|
| Qwen2-VL | `Qwen2VL*` | unused (`_add_model_specific_args` no-op) | none |
| Qwen2.5-VL | `Qwen2_5_VL*` | **used** — forwards `second_per_grid_ts × tokens_per_second` into `get_rope_index` | none |
| Qwen3-VL | reuses `Qwen2_5_VL*` | **unused** — removed by design, replaced by timestamp tokens (no-op) | **yes** |

- **Qwen3 deepstack**: the vision encoder also emits per-layer side outputs (deepstack). `_process_*` / `_cache_to_mm` carry `deepstack_*_embeds` in the mm dict, `_build_partial_mm_embeds` tail-slices them with the **same boundaries** as the base features, and `_pack_deepstack_from_mm` (with `_pack_partial_deepstack` / `_extract_cached_deepstack`) packs them for the prefill decoder.
- **Partial prefix-cache hit**: the boundary may fall inside an image. Because each multimodal item shifts every later token's position, MRoPE is always computed over the **full prompt and then sliced** to `[num_cached:]` (`_build_prefill_position_embed`), and only the tail of each item's features is re-scattered (`_build_partial_mm_embeds`). The cached prefix's KV is copied and reused via `reuse_prefix_cached_kv`.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
6. Verify output: `...`
7. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-eunji:
@rebel-yskim 
Prefix caching in EC is now also supported in Qwen3-VL, so the code has changed significantly.

vllm_rbln/platform.py
```diff
@@ -411,17 +410,43 @@ def _disable_prefix_caching(cls, vllm_config: VllmConfig, reason: str) -> None:
...
+        sub-config (multimodal composites nest it), honors a
+        ``use_sliding_window=False`` opt-out, and treats a sliding ``layer_types``
+        entry as sliding. Errs toward True (disabling prefix caching is safe).
+        """
+        config = (
+            hf_config.get_text_config()
+            if hasattr(hf_config, "get_text_config")
+            else hf_config
+        )
+        if getattr(config, "sliding_window", None) is not None and getattr(
```
rebel-yskim, line 425: **[Should fix]** The `use_sliding_window=False` opt-out only guards this `sliding_window`-based branch, not the `layer_types` branch right below it. A model that has `sliding_window` set but explicitly opts out with `use_sliding_window=False`, while still carrying `sliding`-typed entries in `layer_types`, would have prefix caching disabled despite the opt-out. In practice HF usually derives `layer_types` as all `full_attention` when `use_sliding_window=False`, so this errs toward the safe (over-disable) direction rather than a correctness bug — but it can silently withhold the optimization this PR is trying to enable. Is that intended? If the opt-out should be honored consistently, consider hoisting the `use_sliding_window` check into a shared guard covering both branches.

> yskim-review-bot

examples/optimum/prefix_caching/prefix_caching_mm_offline.py
```diff
@@ -0,0 +1,165 @@
...
+
+prompts = [
+    "What is the largest mammal in the world?",
+    "Who developed the theory of relativity?",
+    "Where is the Great Wall of China located?",
+    "Where does the process of photosynthesis occur?",
+    "What does the Pythagorean theorem state?",
+    "What is the chemical symbol for gold?",
+]
+os.environ["VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK"] = "4"
```
rebel-yskim, line 58: **[Nit]** `VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK` is set here again after it was already set at line 21 (before `from vllm import ...`). This second assignment runs after vllm is imported, so it is effectively a no-op and can be removed.

> yskim-review-bot

examples/optimum/prefix_caching/prefix_caching_mm_offline.py
```diff
@@ -0,0 +1,165 @@
...
+    # Create an LLM with prefix caching enabled.
+    prefix_cached_llm = LLM(
+        model=MODEL,
+        block_size=16384,
+        max_num_seqs=4,
+        enable_prefix_caching=True,
+    )
+
+    # Warmup so that the shared prompt's KV cache is computed.
+    # prefix_cached_llm.generate(generating_prompts[0], sampling_params)
```
rebel-yskim, line 133: **[Nit]** Leftover commented-out code (this warmup `generate` call, and `# cleanup_dist_env_and_memory()` earlier at line 122). Consider dropping these before merge to keep the example clean.

> yskim-review-bot

vllm_rbln/model_executor/models/optimum/qwen2_vl.py
```diff
@@ -44,8 +45,44 @@
...
+        "video",
+        "video_grid_thw",
+        "pixel_values_videos",
+        "video_embeds",
+        "video_token_id",
+    ),
+)
+
+
+def iter_modalities(
```
rebel-yskim, line 73: **[Nit]** `iter_modalities()` is used only once (in `_compute_mrope_position`) and just yields a fixed two-element sequence. Inlining it as `for spec, mm_input in zip(MODALITIES, (image_input, video_input)):` removes a layer of indirection. (The `ModalitySpec`/`MODALITIES` abstraction itself is justified by its many call sites — only this single-use helper is the nit.)

> yskim-review-bot

rebel-yskim (changes requested):

PR #755, opened 2026-07-06
fix(scheduler): support partial remote-KV match in RBLNScheduler

rebel-jinhwan:
## Problem

`RBLNScheduler.schedule()` assumed a request promoted from `WAITING_FOR_REMOTE_KVS` is decode-ready (`num_new_tokens == 1`) — true only for a **full** remote match. LMCache prefix caching returns **partial** matches (e.g. 2048 of 2388 tokens), leaving a local prefill remainder, so the promotion-path assert fires:

```
AssertionError: promoted remote-KV request ... is still in prefill (num_new_tokens=340);
  a partial remote KV match leaves a local prefill remainder that cannot mix into the decode batch.
```

The DP rank dies → the per-step gloo DP all-reduce breaks → other ranks cascade with `Connection closed by peer` → `EngineDeadError`, serve down. With chunk_size=1024 a partial match is the norm, so this fires on essentially every cache hit.

## Fix

Route waiting requests by **`is_prefill(request)` alone**, dropping the `promoted_from_waiting_for_remote_kvs` gate (now unused, removed):

- **decode-ready** (`not is_prefill`, `num_new_tokens == 1`) → join the decode batch (regardless of origin);
- **prefill** (`is_prefill`) → run as a lone prefill; deferred to the next step if a decode was already scheduled this step.

A partial match is `is_prefill`, so it takes the lone-prefill path and reaches decode via the running loop on a later step — exactly like a normal chunked prefill. Scoped to `rbln_scheduler.py`; no runner / connector changes.

### Waiting-queue scheduling: before / after

**Before** — gated on `promoted_from_waiting_for_remote_kvs`; a partial match hit the assert:

```mermaid
flowchart TD
    A[waiting request] --> B{promoted from<br/>WAITING_FOR_REMOTE_KVS?}
    B -->|yes| C{is_prefill?}
    C -->|"no · FULL match"| DEC[join decode batch]
    C -->|"yes · PARTIAL match"| X([AssertionError →<br/>DP rank dies → serve down])
    B -->|no| G{decode already<br/>scheduled this step?}
    G -->|yes| DEFER[defer]
    G -->|no| LP[lone prefill · evict decodes]
    style X fill:#f7d7d7,stroke:#c22
    style DEC fill:#d7e7f7,stroke:#26a
    style LP fill:#d7f7d7,stroke:#2a2
```

**After** — gated purely on `is_prefill`; no crash, origin no longer matters:

```mermaid
flowchart TD
    A[waiting request] --> D{is_prefill?}
    D -->|"no · decode-ready<br/>(any origin)"| DEC[join decode batch<br/>+ spec-slide]
    D -->|yes| G{decode already<br/>scheduled this step?}
    G -->|yes| DEFER[defer to next step]
    G -->|no| LP[lone prefill · runs alone<br/>→ decode via running loop later]
    style DEC fill:#d7e7f7,stroke:#26a
    style LP fill:#d7f7d7,stroke:#2a2
```

The two-level `promoted?` → `is_prefill?` decision (with a crash leaf for partial matches) collapses into a single `is_prefill?` decision: the crash leaf is gone, a partial match flows to lone prefill, and a non-promoted decode-ready request reaches the decode batch.

## Tests

`tests/torch_compile/unit/v1/core/test_pd_disaggregation.py` — partial match scheduled as a lone prefill (**red on unpatched: reproduces the assert; green after**), prefill→decode transition, spec-decode variant; new `test_decode_ready_sync_match_joins_decode_batch`; `test_eviction_drops_evicted_decode_slide_distance` retargeted to a prefill evictor (`num_new > 1`).

167 scheduler unit tests pass; `ruff check` + `ruff format` clean.

## Remaining before merge (hardware)

Verify a non-promoted (local/sync-matched) `NewRequestData` is safe in a decode batch — i.e. its KV is resident in the allocated blocks. On the DP4 + LMCache serve, send a >1024-token prompt twice → expect no gloo/DP crash, `external_prefix_cache_hits` climbs, coherent second response.

### DIFF
--- a/tests/torch_compile/unit/v1/core/test_pd_disaggregation.py
+++ b/tests/torch_compile/unit/v1/core/test_pd_disaggregation.py
@@ -410,33 +410,123 @@ class TestPDDisaggregationScheduler:
         pad_window_sum = sum(ns[r] + slide.get(r, 0) for r in ns)
         assert num_input_tokens == pad_window_sum
 
-    def test_promoted_partial_kv_match_is_rejected(self):
-        """A promoted remote-KV request must be a genuine single-token decode:
-        its prompt KV was prefilled remotely, so num_new_tokens == 1. A partial
-        remote match (num_tokens > matched + 1) would leave a local prefill
-        remainder that cannot mix into the decode batch -- the scheduler asserts
-        loudly rather than emit a corrupt (non-uniform) decode batch.
-        """
-        import pytest
+    def test_promoted_partial_kv_match_scheduled_as_prefill(self):
+        """A PARTIAL remote-KV match (num_tokens > matched + 1) leaves a local
+        prefill remainder. The promoted request must be scheduled as a lone
+        prefill of that remainder -- NOT rejected, and NOT forced into the
+        decode batch.
 
-        matched = _BLOCK_SIZE  # 16
-        # num_tokens = matched + 5 -> num_new_tokens == 5 (> 1): partial match.
-        remote = _create_pd_request(matched + 5, "remote")
+        REGRESSION for the DP4 + LMCache serve crash: the old code asserted
+        `num_new_tokens == 1` in the promotion path and killed the DP rank on
+        every partial prefix-cache hit (gloo cascade -> whole serve dies). With
+        chunk-granular prefix caching a partial match is the norm, so this is
+        the common case, not an edge case.
+        """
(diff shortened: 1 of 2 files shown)


rebel-ykchoi (approved):
lgtm

PR #684, opened 2026-06-17
model(minimax w8a8): Minimax w8a8

rebel-jaehob:
@rebel-thkim , minimax_w8a8 model works well with the rebel_compiler so we need to prepare the last few steps for shipping.

this is the first time doing vllm-rbln related work for me, so I will need a lot of help from you. Thanks in advance.

I've listed some issues that need to be resolved:

1. there seems to be a pre-commit ci stage that I can't resolve on my local?
2. in the moe code in fp8.py file, we fixed the output dtype to be bfloat16, can we change this to reflect the actual dtype of the input tensor?

PR #757, opened 2026-07-07
fix(quant): restore MoE fp8 custom op registrations

rebel-thkim:
## Summary

PR #684 removed the Python `@torch.library.custom_op` registration of `custom_moe_glu_group_dequantize` and, in the same change, added calls to a new `custom_moe_glu_w8a8` op **without registering it**. Neither op is registered anywhere else in the tree (or in an installed package), so the block-fp8 MoE path (`Fp8MoEMethod.apply`) fails with an `AttributeError` on `torch.ops.rbln_custom_ops.*` for **both** the W8A8 (default) and W8A16 branches.

This restores both custom op registrations in `vllm_rbln/model_executor/layers/quantization/fp8.py`.

## Changes

- **`custom_moe_glu_group_dequantize`** (op + `register_fake`) — restored verbatim from the pre-#684 version. Used by the W8A16 path (`VLLM_RBLN_USE_W8A16=True`).
- **`custom_moe_glu_w8a8`** (op + `register_fake`) — new registration matching the call site added in #684:
  - Activation is handed in as fp8 plus a per-(token, K-block) `hidden_states_scale`, dequantized back to `compute_dtype`.
  - The GLU intermediate is fp8-requantized before the down projection (true W8A8 semantics).
  - `compute_dtype` sets the output dtype (fp8 input can't infer it) and is a torch-level-only arg, ignored by the compiler frontend.

The custom op bodies are the eager PyTorch reference implementations; on device they are replaced by the RBLN custom kernels.

## Verification

- Both ops register on `torch.ops.rbln_custom_ops` after import.
- `custom_moe_glu_w8a8` eager run produces correct shape/dtype; W8A8 output tracks the W8A16 reference within fp8 quantization error.
- `register_fake` (meta) path returns correct shape/dtype/device.
- `ruff check` / `ruff format` clean.

> Note: the on-device kernel path was not exercised here — this environment lacks `librbln.so` (rebel-compiler), so only the CPU eager reference and meta/fake paths were validated.

### DIFF
--- a/vllm_rbln/model_executor/layers/quantization/fp8.py
+++ b/vllm_rbln/model_executor/layers/quantization/fp8.py
@@ -454,6 +454,358 @@ class Fp8LinearMethod(LinearMethodBase):
             return torch.nn.functional.linear(x, weight_bf16.t(), bias)
 
 
+@torch.library.custom_op(
+    "rbln_custom_ops::custom_moe_glu_group_dequantize",
+    mutates_args=(),
+)
+def custom_moe_glu_group_dequantize(
+    hidden_states: torch.Tensor,
+    gate_proj_weight: torch.Tensor,
+    gate_proj_scale: torch.Tensor,
+    up_proj_weight: torch.Tensor,
+    up_proj_scale: torch.Tensor,
+    down_proj_weight: torch.Tensor,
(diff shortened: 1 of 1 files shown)


rebel-kblee (approved):
👍

PR #766, opened 2026-07-09
fix: honor RBLN_FORCE_NPU_NAME (renamed from RBLN_TARGET_SOC)

rebel-thkim:
### Problem
rebel-compiler renamed the target-NPU env var `RBLN_TARGET_SOC` → `RBLN_FORCE_NPU_NAME`, and the compiler CI now sets the new name. vllm-rbln still referenced the old name in several places:

- `platform.py` `get_device_name()`: on a CPU-only compile host (no NPU) it fell back to `RBLN_TARGET_SOC` only, so the renamed var was missed and the device-less compile raised `RuntimeError`.
- `examples/cpu_only_compile.py` (FSW example): set/documented `RBLN_TARGET_SOC`.
- `tests/torch_compile/unit/test_compile_only.py`: asserted the error mentions `RBLN_TARGET_SOC`.
- `rbln_envs.py`: comment referenced `RBLN_TARGET_SOC`.

### Solution
Honor `RBLN_FORCE_NPU_NAME` (current name) across vllm-rbln, keeping `RBLN_TARGET_SOC` as a back-compat fallback where it's read:

- `get_device_name()`: read `RBLN_FORCE_NPU_NAME` first, then legacy `RBLN_TARGET_SOC`; error message now points at `RBLN_FORCE_NPU_NAME` / "target NPU".
- Example: use `RBLN_FORCE_NPU_NAME` in docs, default, help, and the env it sets.
- Test: assert on `RBLN_FORCE_NPU_NAME`; clear both env vars so the raise path is deterministic.
- Comment updated.

The NPU-attached path (`rebel.get_npu_name()` returns a value) is unaffected.

### Verification
- `py_compile` OK on all changed files.
- NPU host: `rebel.get_npu_name()` takes precedence → unchanged.
- CPU compile host + `RBLN_FORCE_NPU_NAME=RBLN-CA25`: resolves the target; legacy `RBLN_TARGET_SOC` still works.

Related: minimax-m2_7 CR-13 compile-only cases in rebel_compiler's REBEL vLLM-Dynamo compile-only PR.

## Reviewer
* **Primary**: (TBD)
* **Secondary**: (code owner)

### DIFF
--- a/examples/cpu_only_compile.py
+++ b/examples/cpu_only_compile.py
@@ -34,9 +34,9 @@ torch.compile pipeline). Three things follow from that:
   (the optimum-rbln path) makes ``VLLM_RBLN_COMPILE_ONLY=1`` raise an error.
 * The compile cache must stay enabled (this is where the artifacts land), so
   ``VLLM_DISABLE_COMPILE_CACHE=1`` is rejected in this mode.
-* Without an NPU, ``rebel.get_npu_name()`` returns ``None`` and the target SOC
-  can no longer be auto-detected, so you must set ``RBLN_TARGET_SOC`` (e.g.
-  ``RBLN-CA25``) to tell the compiler what to target.
+* Without an NPU, ``rebel.get_npu_name()`` returns ``None`` and the target NPU
+  can no longer be auto-detected, so you must set ``RBLN_FORCE_NPU_NAME``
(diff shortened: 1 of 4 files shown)


rebel-eunji (approved):
lgtm +1

rebel-seinpark (approved):

PR #769, opened 2026-07-10
refactor(metrics): report latency percentiles instead of outlier-removed mean

rebel-daeyang:
### 🚀 Summary of Changes

Latency was summarized as a single mean after dropping the one sample farthest from the mean. This misfires whenever a phase's per-step latencies fall into distinct groups instead of clustering around one value.

Example — chunked prefill: a prompt is processed in chunks and only the last chunk produces a token (and runs the sampler), so the per-step latencies are bimodal — fast intermediate chunks and slow final chunks. Dropping the single farthest-from-mean sample then throws away a legitimate fast or slow step, so the reported average is unstable: for the same workload it swung between ~49 ms and ~67 ms depending on which sample happened to be dropped.

This PR follows upstream vLLM's benchmark convention (report mean + percentiles over the raw data) and cleans up the surrounding metrics reporting:

- **Drop the single-outlier removal.** Report `mean | p50 | p90 | p99 | max` per phase, so one-off spikes show up in the tail instead of skewing/silently dropping a sample.
- **Fix throughput.** It is now `total_tokens / total_time`. Previously outliers were removed from the latency and token-count lists independently, which could drop different steps from each list and misalign the ratio.
- **Add a per-request prefill summary** (sum of a request's chunk latencies) — a stable "time to prefill a request" number that does not depend on chunking.
- Update unit tests accordingly.

No engine/runtime behavior changes — this only affects reported metrics.

### 🔎 Example output (prefill)

Before:
```
PREFILL METRICS:
  Average latency: 49.07 ms   # single outlier-removed mean (unstable: ~49–67 ms)
```

After:
```
PREFILL METRICS:
  Latency (ms): mean 58.53 | p50 66.77 | p90 109.01 | p99 110.29 | max 110.43
  Average throughput: 8747.39 tokens/sec
  Per-request prefill (ms): mean 117.06 | p50 111.03 | max 136.47 (4 requests)
```

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

### 🧪 How to Test

1. `pytest tests/torch_compile/unit/v1/worker/test_metrics.py` → passes.
2. Run inference with `VLLM_RBLN_METRICS=1` and inspect the `FINAL PERFORMANCE STATISTICS` block: each phase shows `Latency (ms): mean | p50 | p90 | p99 | max`, and prefill shows `Per-request prefill (ms): ...`.

### 📋 Checklist

### DIFF
--- a/vllm_rbln/v1/worker/metrics.py
+++ b/vllm_rbln/v1/worker/metrics.py
@@ -15,14 +15,13 @@
 import os
 from collections import defaultdict
 from dataclasses import dataclass, field, replace
-from typing import TypeVar
+
+import numpy as np
 
 import vllm_rbln.rbln_envs as envs
 from vllm_rbln.logger import init_logger, make_file_handler
 
 logger = init_logger(__name__)
-
-T = TypeVar("T", int, float)
 
 _metrics_file_attached = False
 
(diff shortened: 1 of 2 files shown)


vllm_rbln/v1/worker/metrics.py
```diff
@@ -79,92 +78,89 @@ def add_measurement(
...
+                "  Average throughput: %.2f tokens/sec", self.get_avg_throughput()
+            )
+        # Timing breakdown is only accurate in synchronous mode.
+        has_breakdown = bool(
+            self.host_times
+            or self.device_times
+            or self.ccl_times
+            or self.prepare_times
+        )
+        if has_breakdown and not envs.VLLM_RBLN_FORCE_SYNC:
```
rebel-jonghewk: I think this is a good idea, but how about adding this logic to rebel-compiler instead of here? We could also skip emitting the info in the async case, since sync/async can be used in various situations (e.g., async scheduling).

rebel-daeyang: Good point — removed the env-based gating from this PR; will handle it in rebel-compiler as you suggested.

rebel-jonghewk (approved):

rebel-eunji (commented):
lgtm 👍

rebel-eunji (approved):

PR #777, opened 2026-07-15
feature(pp): make pipeline parallelism first-class on the vLLM-native decode path

rebel-ykchoi:
### 🚀 Summary of Changes

#### Summary

Pipeline parallelism (`pipeline_parallel_size > 1`) on the vLLM-native path
(`VLLM_RBLN_USE_VLLM_MODEL=1`) worked for plain decode but broke down the moment
decode batching, speculative decoding, or disaggregated serving entered the
picture: a step's decode batch could overflow the compiled per-stage bucket,
spec-decode misclassified the prefill/decode phase inconsistently across ranks,
and the KV connector saved only part of the model's layers. This PR reorganizes
the scheduler/runner so PP is a first-class, correct, and efficient mode across
all three.

The work is one story told in four steps: give the scheduler a single PP-aware
decode-admission budget, make speculative decoding rank-consistent on top of it,
consolidate the prefill/decode phase authority onto the scheduler (removing the
now-dead per-rank classifier), and turn the batch-balancing behavior on by
default for PP.

##### 1. PP-aware decode-batch admission (scheduler)
Under PP the engine keeps one decode microbatch in flight per stage, so a single
step's decode batch must fit the compiled per-stage bucket
(`max_num_seqs // pp_size`). Decode requests are admitted from two places — the
running queue and, in disaggregated serving, requests whose KV transfer just
completed — so admission is centralized into **one per-step budget** shared by
both loops (`DecodeAdmissionController` / `DecodeBatchBudget` in
`v1/core/utils.py`):
- Hard cap = the compiled bucket size on every decode that joins the batch
  (exceeding it leaves the runner without a matching bucket → crash).
- Optional **demand spreading** across stages (`ceil(demand / pp_size)`) so no
  stage idles after a pipeline drain.
- Requests that become decode-ready outside the demand snapshot (full local
  prefix-cache hit, resume-after-eviction) are gated by the hard cap only.
- The admitted count stays consistent when the batch shrinks (cleared on
  whole-batch eviction, decremented on preemption of an admitted decode).

`pp_size == 1` degenerates to `max_num_seqs` — unchanged.

##### 2. Speculative decoding under PP (scheduler + runner)
Enables spec decode (ngram, EAGLE, MTP, …) together with `pp_size > 1`. RBLN runs
synchronous scheduling with `pp_size` microbatches in flight, so a verify step —
whose query spans `num_spec_tokens + 1` positions and whose anchor/accepted
tokens are produced only on the last PP rank — must stay bit-consistent on every
non-last rank one step later. Three mechanisms:
- **Uniform per-query-length bookkeeping** — the prefill/decode phase is stamped
  once on `RBLNSchedulerOutput` by the scheduler and read by every rank (never
  re-derived per-rank), PP inter-stage tensors are sized per `(bucket, query_len)`,
  and non-last ranks write propagated tokens from `num_tokens_no_spec`.
- **Anchor-reconciled scheduling** — `should_defer_spec_step` admits a verify only
  when its base token is reconciled, deciding on the base count (not the
  draft-inflated `num_new_tokens`). Gated on spec being enabled → non-spec decode
  is untouched.
- **Multi-accept token propagation** — the scheduler extends a lagging rank's
  `new_token_ids` payload back by `num_spec_tokens`; the runner writes by absolute
  position (idempotent over mis-speculated slots).

**KV-connector finalization is made PP-correct**: the deferred per-step finalize
(so draft-token KV is saved) runs in the sample path that only the last rank
reaches, so it is gated on the last rank — non-last ranks finalize in-context, so
every rank still saves its own layers' KV.

##### 3. Scheduler as the single phase authority (runner cleanup)
With `RBLNSchedulerOutput` now the unified type every step carries, the legacy
per-rank `is_prefills()` classifier (which diverged across PP ranks under spec
decode) is removed along with its test and a dangling comment. The defensive
`isinstance`/`getattr` fallbacks that hedged for a plain `SchedulerOutput` are
dropped: RBLN-only helpers are narrowed to `RBLNSchedulerOutput`, and the
base-override methods sit behind a single boundary assert that both guards the
invariant and narrows the type for direct field access.

##### 4. PP decode-batch balancing on by default
`VLLM_RBLN_PP_BALANCE_DECODE_BATCH` now defaults to **on**. The dynamic cap only
ever caps *lower* than the static one (never overflows the bucket), is floored at
1, matches the static cap at saturation, and is a **no-op for `pp_size == 1`** —
so non-PP configs are unaffected. Remains as an escape hatch: set `=0` to force
the static cap.

#### Testing

- **Unit** — new/extended coverage for the backfill/slide query-window math
  (`test_query_backfill.py`), the anchor-reconciled deferral predicate and
  multi-accept payload (`test_rbln_model_runner.py`), the PP decode-admission
  budget and demand-spread contrast (`test_scheduler.py`), and PP disaggregation
  (`test_pd_disaggregation.py`).
- `pre-commit run --hook-stage manual` clean (ruff, mypy 3.10–3.13, typos,
  license); full unit suites for runner/scheduler/pd green.

#### Risk / compatibility

- `pp_size == 1` and non-spec decode paths are untouched (predicate gated on spec;
  cap degenerates to `max_num_seqs`; balancing is a no-op).
- The default flip is confined to PP; the escape hatch preserves the old behavior.

#### Affected modules

Platform, Scheduler, Runner.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/v1/core/rbln_scheduler.py
+++ b/vllm_rbln/v1/core/rbln_scheduler.py
@@ -123,7 +123,11 @@ from vllm.utils.hashing import get_hash_fn_by_name
 from vllm.v1.core.kv_cache_manager import KVCacheBlocks
 from vllm.v1.core.kv_cache_utils import init_none_hash
 from vllm.v1.core.sched.interface import PauseState
-from vllm.v1.core.sched.output import NewRequestData, SchedulerOutput
+from vllm.v1.core.sched.output import (
+    CachedRequestData,
+    NewRequestData,
+    SchedulerOutput,
+)
 from vllm.v1.core.sched.request_queue import SchedulingPolicy, create_request_queue
 from vllm.v1.core.sched.scheduler import Scheduler
 from vllm.v1.engine import EngineCoreEventType, EngineCoreOutputs
@@ -138,7 +142,11 @@ from vllm_rbln.v1.core.rbln_kv_cache_manager import (
     RBLNKVCacheManager,
     SubBlockMatch,
 )
-from vllm_rbln.v1.core.utils import DecodeAdmissionController
+from vllm_rbln.v1.core.utils import (
+    DecodeAdmissionController,
+    num_base_tokens,
+    should_defer_spec_step,
+)
 
 logger = init_logger(__name__)
 
@@ -172,6 +180,15 @@ class RBLNSchedulerOutput(SchedulerOutput):
     # positions' logits are discarded; past KV gets idempotently
     # re-written. Empty / missing key => slide_distance is 0 (no slide).
     spec_decode_slide_distance: dict[str, int] = field(default_factory=dict)
+    # NOTE(RBLN): whether this step runs the prefill graph (True) or the decode
+    # graph (False). RBLN has no mixed batching, so a step is uniformly one or
+    # the other. Stamped by the scheduler (a single authority) from the
+    # authoritative Request state, so every PP rank reads an identical value.
+    # The runner MUST use this instead of re-deriving the phase from its own
+    # per-rank input_batch: num_tokens_no_spec is maintained on different paths
+    # on the last PP rank (sampling) vs other ranks (scheduler output), so a
+    # runner-side classification diverges across ranks under spec-decode + PP.
+    is_prefill_step: bool = False
 
 
 def is_prefill(request: Request) -> bool:
@@ -416,6 +433,17 @@ class RBLNScheduler(Scheduler):
                 + request.num_output_placeholders
                 - request.num_computed_tokens
             )
+            # NOTE(RBLN): Under sync-scheduling PP + spec decode, defer a step
+            # with no reconciled base (anchor) token yet -- num_computed_tokens is
+            # optimistically advanced by the drafts, so the RAW num_new_tokens can
+            # still be draft-inflated (drafts held) or negative (post-verify
+            # overshoot). See should_defer_spec_step. Non-spec decode is untouched
+            # (its no-new-tokens case is the plain `== 0` defer below).
+            if should_defer_spec_step(
+                self.num_spec_tokens, request.spec_token_ids, num_new_tokens
+            ):
+                req_index += 1
+                continue
             if 0 < self.scheduler_config.long_prefill_token_threshold < num_new_tokens:
                 num_new_tokens = self.scheduler_config.long_prefill_token_threshold
             num_new_tokens = min(num_new_tokens, token_budget)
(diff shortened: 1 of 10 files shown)


vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -288,6 +321,21 @@ def _decide_spec_slide(self, request: Request, new_n: int) -> tuple[int, bool]:
...
+        = running decodes + remote-KV requests ready to be admitted (transfer
+        complete, awaiting promotion). Including the ready remote-KV gives the
+        dynamic cap headroom to ramp the decode batch on the P/D-disaggregated
+        decode side; running-only would stall the ramp. Demand is invariant
+        under admission (promoting a ready remote-KV moves it from ready ->
+        running), so this snapshot is exact even as the running/ready split
+        shifts during the waiting loop. Evaluated only when PP balancing is on.
+        """
+        num_running_decodes = sum(1 for r in self.running if not is_prefill(r))
+        num_ready_remote_kv = len(self.finished_recving_kv_req_ids)
```
rebel-jinhwan, line 336: LMCache MP Connector return `load_kv_async=True` even for partial prefix hit. Those requests are still prefill request.

rebel-ykchoi: This is intentional. `_decode_demand` is a deliberately *inclusive* count of
requests that may become decodes (running decodes + ready remote-KV), used only
to size the PP-balancing **soft/spreading cap** `ceil(demand / pp_size)`. It does
not — and need not — resolve prefill-vs-decode at count time; that determination
is deferred to the admission loop below, per request:

- The soft cap gates only remote-KV promotions (`apply_soft_cap =
  request.status == WAITING_FOR_REMOTE_KVS`). A gated request stays in
  `WAITING_FOR_REMOTE_KVS` and is re-evaluated next step — it is not flipped and
  mis-routed through local prefill (see the comment at the admission site).
- So if such a finished-recving request is in fact a prefill (e.g. the LMCache
  MP partial-hit `load_kv_async=True` case you point out), it is handled as a
  prefill downstream. RBLN has no mixed batching, so a prefill step evicts the
  decode batch and the per-step decode budget is reset — the transient
  over-inclusion in the demand snapshot doesn't leak into the decode accounting.
- The **hard cap** (the compiled decode bucket, `max_num_seqs // pp_size`),
  enforced unconditionally by `can_admit()`, is the correctness limit and is
  independent of `_decode_demand`. An over-count can only make the soft cap
  slightly more permissive (weaker spreading for that step) — never breach the
  bucket. And it self-corrects: once the request enters `running`, the next
  step's snapshot reflects reality.

So the partial-hit-prefill case is covered by design — the demand snapshot is an
approximate spreading input, and the actual routing/accounting is exact at
admission time.

PR #781, opened 2026-07-16
fix(model): make plain-attribute weights contiguous for weight-free compile

rebel-jongho:
### 🚀 Summary of Changes

Weight-free compile hard-errors on non-contiguous CPU tensors (rebel-compiler). `_make_weights_contiguous()` (added in #742) only walked `self.model.parameters()` and `self.model.buffers()`, so it covered `PARAMETER` and `BUFFER` inputs but missed a third kind.

`torch.export` lifts three kinds of stored tensors as graph inputs that the compiler treats as weights and routes through the same non-contiguous check: `PARAMETER`, `BUFFER`, and `CONSTANT_TENSOR`. **Plain tensor attributes on submodules are lifted as `CONSTANT_TENSOR`**, and `parameters()`/`buffers()` never reach them. A concrete example is upstream MLA attention:

```python
# vllm/model_executor/layers/attention/mla_attention.py, process_weights_after_loading
self.W_UV = W_UV.transpose(0, 1)      # plain attr, non-contiguous
self.W_UK_T = W_UK.permute(1, 2, 0)   # plain attr, non-contiguous
```

These are neither parameters nor buffers, so #742 does not make them contiguous, and a weight-free CPU compile would hard-error on them.

This PR extends `_make_weights_contiguous()` to also walk each submodule's `__dict__` and make non-contiguous plain tensor attributes contiguous.

**Verified** (torch.export repro): before the walk, plain attributes export as non-contiguous `CONSTANT_TENSOR` (`contiguous=False`); after `setattr(module, name, value.contiguous())`, they export as contiguous `CONSTANT_TENSOR`.

#### Limitation (documented in the method docstring)

Only tensors held as **direct** module attributes are covered. Tensors that `torch.export` still lifts as `CONSTANT_TENSOR` but that are **not** direct attributes remain uncovered:

- tensors inside container attributes — `self.x = [t]`, `self.x = {"k": t}`, tuples/sets;
- tensors on non-`Module` holder objects — `self.cfg.t`, dataclasses, `SimpleNamespace`.

`self.model.modules()` only iterates `nn.Module` instances, and the `isinstance(value, torch.Tensor)` check skips containers/holder objects. Closing these fully from the consumer side requires a fragile recursive object-graph walk (and even that cannot replace a tensor held directly inside an immutable tuple). The structurally complete fix lives at the lifted-constant boundary (normalizing `ExportedProgram.constants` in the compiler), which is out of scope here. This PR covers the direct-attribute case, which is what MLA and similar layers use.

---

* Related to #742

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Load an MLA model (e.g. DeepSeek-V2/V3) whose `process_weights_after_loading` stores permuted weights as plain attributes (`self.W_UV`, `self.W_UK_T`).
2. Run a weight-free compile on CPU (no `VLLM_RBLN_USE_DEVICE_TENSOR`).
3. Before this change the compile raises `ValueError: weight-free mode requires contiguous CPU tensors ...`; after it, the plain-attribute weights are made contiguous and the compile proceeds.

Minimal export-level check:

```python
import torch
from torch.export.graph_signature import InputKind

class Inner(torch.nn.Module):
    def __init__(self):
        super().__init__()
        w = torch.randn(4, 8, 16)
        self.W_UV = w.transpose(0, 1)  # plain attr, non-contiguous
    def forward(self, x):
        return torch.bmm(x, self.W_UV)

m = Inner().eval()
# apply the same walk as _make_weights_contiguous
for module in m.modules():
    for name, value in list(module.__dict__.items()):
        if isinstance(value, torch.Tensor) and not value.is_contiguous():
            setattr(module, name, value.contiguous())

ep = torch.export.export(m, (torch.randn(8, 5, 4),))
for s in ep.graph_signature.input_specs:
    if s.kind == InputKind.CONSTANT_TENSOR:
        print(s.target, ep.constants[s.target].is_contiguous())  # -> True
```

---

### 📋 Checklist

---

### 💬 Notes

The hard-error only fires for CPU tensors; in device-tensor mode (`VLLM_RBLN_USE_DEVICE_TENSOR`) weights are offloaded to the rbln device and handled by the on-device contiguous path, so this walk is a no-op there (already-contiguous tensors are skipped).

#### Upstream context

The plain-attribute nature of MLA's `W_UV`/`W_UK_T` has already been flagged upstream, though for a different symptom — weight-reload / CUDA-graph storage identity in RL, not export-time contiguity:

- [vllm-project/vllm#48251](https://github.com/vllm-project/vllm/pull/48251) — "Preserve post-load tensors across weight reloads." Root cause: standard `MLAAttention` reassigns derived `W_UV`/`W_UK_T` views, so graphs captured before a reload keep referencing the old allocations. Proposed fix uses `replace_parameter(..., prefer_copy=True)`.
- [vllm-project/vllm#48312](https://github.com/vllm-project/vllm/issues/48312) — "[RFC] Weight Reload Correctness for RL," which tracks `W_UV`/`W_UK_T` (and other unregistered runtime-derived tensors across MoE/Marlin/Machete/FlashInfer) under a "storage identity" category.

Our concern (export lifting these as non-contiguous `CONSTANT_TENSOR`) is RBLN-specific and not covered by those. Forward-looking: `replace_parameter` wraps the result as an `nn.Parameter`, so if #48251 lands, MLA's `W_UV`/`W_UK_T` would become parameters and be covered by the existing `parameters()` walk. This `__dict__` walk then remains as a general net for any other plain-attribute derived weights.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -2344,7 +2344,7 @@ class RBLNModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
     def _make_weights_contiguous(self) -> None:
         """Force weights contiguous before weight-free compile, which hard-errors
         on non-contiguous CPU tensors. Covers parameters, buffers, and plain
-        tensor attributes (torch.export lifts the latter as CONSTANT_TENSOR)."""
+        tensor attributes."""
         for p in self.model.parameters():
             if not p.is_contiguous():
                 p.set_(p.contiguous())
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -2342,13 +2342,30 @@ class RBLNModelRunner(LoRAModelRunnerMixin, KVConnectorModelRunnerMixin):
         return self.model.compute_logits(hidden_states)
 
     def _make_weights_contiguous(self) -> None:
-        # weight-free compile hard-errors on non-contiguous CPU weights.
+        """Force model weights contiguous before weight-free compile, which
+        hard-errors on non-contiguous CPU tensors.
+
+        torch.export lifts three kinds of stored tensors as graph inputs that
+        the compiler treats as weights: PARAMETER, BUFFER, and CONSTANT_TENSOR.
+        parameters()/buffers() cover the first two; plain tensor attributes on
+        submodules (e.g. MLA's ``self.W_UV = W_UV.transpose(0, 1)``) are lifted
+        as CONSTANT_TENSOR and are reached here via each module's ``__dict__``.
+
+        Limitation: only tensors held as direct module attributes are covered.
+        Tensors inside container attributes (``self.x = [t]``, ``{"k": t}``) or
+        on non-Module holder objects (``self.cfg.t``) are also lifted as
+        CONSTANT_TENSOR but are not reachable this way and are still missed.
+        """
         for p in self.model.parameters():
             if not p.is_contiguous():
                 p.set_(p.contiguous())
         for b in self.model.buffers():
             if not b.is_contiguous():
                 b.set_(b.contiguous())
+        for module in self.model.modules():
+            for name, value in list(module.__dict__.items()):
+                if isinstance(value, torch.Tensor) and not value.is_contiguous():
+                    setattr(module, name, value.contiguous())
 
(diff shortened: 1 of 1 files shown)


rebel-jongho:
@rebel-eunji Could you recommend any reviewers for this PR?

rebel-jonghewk (approved):

rebel-eunji (approved):

PR #792, opened 2026-07-20
fix(spec-decode): avoid DP deadlock in MTP chunked prefill

rebel-eunik:
### 🚀 Summary of Changes

Fixes a hang when serving an MTP (eagle) model with **DP + EP** and **chunked prefill**.

`RBLNEagleProposer.propose()` early-returned on intermediate chunked-prefill steps, skipping the multi-token draft loop. Each loop iteration runs a draft forward that performs a DP/EP collective — skipping them on one rank while the peer rank runs the full count diverged the per-step collective count and deadlocked.

Fix: run the full loop so both ranks issue the same collectives, then zero the (already-discarded) result instead of returning early. Only reproduced with `num_speculative_tokens > 1` and a prompt long enough to be chunked.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Serve MTP model with DP + EP + chunked prefill and `num_speculative_tokens > 1`:
   ```bash
   vllm serve DeepSeek-V3 \
       --data-parallel-size 4 --enable-expert-parallel \
       --enable-chunked-prefill --max-num-batched-tokens 512 \
       --max-model-len 10240 --max-num-seqs 1 \
       --speculative-config '{"method": "mtp", "num_speculative_tokens": 3}'

2. Send request which has a prompt longer than `max_num_batched_tokens` so chunked prefill kicks in.

---
📸 Screenshots / Logs (if applicable)


---
📋 Checklist

---
💬 Notes



---


rebel-cjlee:
@rebel-eunik I originally added this condition to skip unnecessary computation during chunked prefill. Looking at it again, in the DP case every rank has to run the full computation anyway for communication. So, instead of building and returning a dummy result, we could just return the actual `draft_token_ids` we already computed and remove the `if intermediate_chunked_prefill` condition. What do you think?

rebel-eunik:
> @rebel-eunik I originally added this condition to skip unnecessary computation during chunked prefill. Looking at it again, in the DP case every rank has to run the full computation anyway for communication. So, instead of building and returning a dummy result, we could just return the actual `draft_token_ids` we already computed and remove the `if intermediate_chunked_prefill` condition. What do you think?

Updated as suggested.

rebel-cjlee (approved):

vllm_rbln/v1/spec_decode/eagle.py
rebel-eunik, line 128: @rebel-cjlee Will this work correctly in the MTP module that uses input embedding?

rebel-cjlee: I haven't tried using the MTP with `input_embeds` before. Is there a specific use case you have in mind where this would be needed?

rebel-eunik: My initial concern was regarding compatibility with models utilizing input text embeddings. However, it appears to be fine, as the DeepSeek MTP module contains its own logic to embed `input_ids`. Additionally, the upstream implementation only uses input_embeds if `mm_embed_inputs` is provided, applying `None` otherwise.

PR #795, opened 2026-07-20
model(qwen3_5): add Qwen3.5 to optimum model path

rebel-seinpark:
## Summary
Add the vision-language **Qwen3.5** (`Qwen3_5ForConditionalGeneration`) to the optimum model path, **including multi-request continuous-batching support** for its hybrid linear-attention state.

Qwen3.5 is a **hybrid decoder**: `linear_attention` (GatedDeltaNet — `conv_state` + `recurrent_state`) layers + gated `full_attention` (paged KV) layers. It is essentially **"Qwen3-VL without deepstack"**.

### 1. Model wiring
- **registry**: map `Qwen3_5ForConditionalGeneration → ("qwen3_5", "RBLNQwen3_5ForConditionalGeneration")`.
- **`qwen3_vl.py`**: `RBLNOptimumQwen3_5ForConditionalGeneration` **inherits `RBLNOptimumQwen3VLForConditionalGeneration`** — Qwen3.5 is Qwen3-VL minus deepstack, so it reuses the no-`second_per_grid_ts` video handling and only overrides `_build_prefill_params` back to the 3-tuple form (its vision encoder returns a single tensor, no deepstack), plus `_image_token_id` (top-level `image_token_id`).
- **`optimum/__init__.py`**: import the wrapper (auto-discovery picks it up).
- **compilation**: `get_param_qwen3_5` — **forces `flash_attn`** (the hybrid backbone's multi-window prefill is mis-lowered by the eager op; flash is correct).

### 2. Linear-attention state slotting (multi-request)
The GatedDeltaNet `conv_state`/`recurrent_state` is a **fixed `[max_num_seqs]` on-device cache indexed by batch ROW**, shared by prefill and decode. The compiled prefill graph declares `batch_idx` as a **required input** (which row to write).

- **`batch_idx` in prefill**: without it the runtime raises `KeyError: 'batch_idx'`. A hardcoded `0` is only correct for `max_num_seqs=1`; with >1 the rows collide and decode reads mismatched state.
- **`RBLNMambaStateIndexAllocator`**: pins each request to a **stable batch index** for its lifetime — allocate on prefill, free on finish. vLLM's `req_id_to_index` can't be used (its `condense()` reassigns rows), so we own the mapping. Mirrors vLLM's mamba `state_indices` (index allocation only; the cache is owned by optimum-rbln).
- **decode laid out by row**: each running request is placed at `row == batch_idx` (via `input_block_ids` + a slot-ordered `compute_decode_position_embed`), then logits are gathered back to running order. So a request always meets its own recurrent-state row regardless of vLLM's running order.
- **bucketing disabled**: the recurrent state is fixed-width, so `decoder_batch_sizes` is single (`= max_num_seqs`); decode always runs full width (empty rows are dummy-padded and dropped).

The `conv_state`/`recurrent_state` caches and their 0/1 control masks are handled **inside** optimum-rbln's `RBLNQwen3_5RuntimeModel`; this wrapper passes only standard kwargs + `batch_idx`.

### 3. Tests — `tests/v1/models/optimum/test_qwen3_5_state_allocator.py` (15, no hardware)
- **Allocator contract**: sequential/stable/idempotent allocation, `indices` order, lowest-free reuse, full lifecycle.
- **Edge cases**: unknown/double `free` no-ops, exhaustion raises, realloc-when-full, unallocated lookup raises.
- **Real decode-layout wiring** (drives the actual `forward` / `pad_decoder_items`, only the model mocked): scatter to `batch_idx` rows; permuted running order still pairs each request with its own row + gathers back to running order; finish→free→reuse then decode with mixed rows; single running request pads the freed row and drops its output.

## Design notes
- **Why not `req_id_to_index`**: vLLM condenses it on finish, but the recurrent state lives at a fixed address — the row must be stable per request.
- **Why row == `batch_idx`**: the state cache is indexed by batch row; we align the *input* to the fixed state layout (scatter), not the state to the input.
- **Prefill is single-request** (the runner forbids mixing prefill with decode), so prefill needs exactly one `batch_idx`.

## Verification
- `ruff` (pinned v0.14.0) + `mypy` (3.10–3.13) + `pytest` (15/15) pass.
- **Live import** (vLLM 0.22): `load_model` dispatch reaches the wrapper; MRO via Qwen3-VL; compile fn resolves to `get_param_qwen3_5` (flash).
- **Hardware smoke**: a full Qwen3.5-0.8B runs end-to-end (`max_num_seqs=1`) through compile → vision encoder → hybrid prefill/decode → logits.

## Follow-up
- **Multi-request accuracy on hardware**: the slot logic is unit-tested, but `max_num_seqs>1` end-to-end correctness (two concurrent prompts) still needs a hardware run.


examples/optimum/multimodal_models/qwen_vl_offline.py
```diff
@@ -191,7 +191,10 @@ def generate_prompts_wo_processing(batch_size: int, model: str):
 
 def main(
     num_input_prompt: int = 1,
-    # NOTE: This example supports Qwen2-VL, Qwen2.5-VL, and Qwen3-VL.
+    # NOTE: This example supports Qwen2-VL, Qwen2.5-VL, Qwen3-VL, and Qwen3.5.
+    # For Qwen3.5 (hybrid GatedDeltaNet backbone), flash attention is forced at
+    # compile time, so keep block_size / max_model_len >= 4096 (already the case
+    # below). Pass a Qwen3.5 checkpoint via `--model`.
```
rebel-seinpark:
```suggestion
    # below).
```

examples/optimum/multimodal_models/qwen_vl_offline.py
```diff
@@ -233,4 +233,4 @@ def main(
 
 
 if __name__ == "__main__":
-    fire.Fire(main)
+    fire.Fire(main)
```
rebel-seinpark:
```suggestion
    fire.Fire(main)

```

rebel-eunji:
It seems like there's an error... I think the PR needs this patch.
https://github.com/RBLN-SW/vllm-rbln/commit/33d3969fd149f32f7d0e512d76d7bc2a54e8c982

vllm_rbln/model_executor/models/optimum/qwen3_vl.py
```diff
@@ -343,3 +391,140 @@ class RBLNOptimumQwen3VLMoeForConditionalGeneration(
...
+        ``[max_num_seqs]`` on-device cache indexed by batch row. ``prefill`` writes
+        one row (``batch_idx``, a REQUIRED prefill graph input — omitting it raises
+        ``KeyError: 'batch_idx'`` in ``RBLNQwen3_5RuntimeModel._run``); ``decode``
+        reads/writes every row. So each request is pinned to a stable ``batch_idx``
+        (``RBLNMambaStateIndexAllocator``): prefill passes its ``batch_idx`` and
+        decode lays the batch out with the request at ``row == batch_idx`` (padding
+        the empty rows), then gathers logits back to running order.
+        """
+        # Reclaim batch indices from requests that finished since the last step.
+        self._state_index_allocator.free(model_input.finished_requests_ids)
```
rebel-eunji: This finished_requests_ids is already deprecated and not works.
I'll remove it clearly. You have to handle the finished requests in model runner level.
Recommendation
- Generate a subclass of attention of optimum_attention
- And just add sorting.

rebel-eunji (changes requested):

vllm_rbln/utils/optimum/converter/from_optimum.py
```diff
@@ -151,3 +153,15 @@ def update_num_blocks(vllm_config: VllmConfig, num_blocks: int) -> None:
     if vllm_config.cache_config.num_gpu_blocks_override is not None:
         vllm_config.cache_config.num_gpu_blocks_override = adjusted_num_blocks
     vllm_config.additional_config["num_blocks_synced"] = True
+
+def update_mamba_block_size(vllm_config: VllmConfig, params: "RBLNParams") -> None:
```
rebel-seinpark, line 172: @rebel-eunji Thanks!

rebel-eunji:
@rebel-seinpark Could you remove the Claude-generated comments? There are a lot of them.

vllm_rbln/model_executor/models/optimum/optimum_attention/optimum_attention_strategy.py
```diff
@@ -223,6 +223,18 @@ def preprocess(
         )
 
 
+class LinearAttentionStrategy(InnerAttentionStrategy):
+    def preprocess(  # type: ignore[override]
```
rebel-eunji, line 227: I will fix the return type ignore by refactoring it later.

vllm_rbln/model_executor/models/optimum/compilation/multimodal/qwen.py
```diff
@@ -55,6 +55,32 @@ def get_param_qwen2_vl(
...
 
 
+def get_param_qwen3_5(
+    batch_size: int,
+    max_model_len: int,
+    block_size: int,
+    num_devices: int,
+    memory_budget: float,
+    prefill_chunk_size: int | None = None,
+    gdn_chunk_size: int | None = None,
```
rebel-eunji: gdn_chunk_size is not passed in as a parameter.
https://github.com/RBLN-SW/vllm-rbln/blob/e9fbe9fce706791c20637c6de17c6cb7ed9d6215/vllm_rbln/model_executor/models/optimum/compilation/__init__.py#L262

rebel-eunji: I think it is better that gdn_chunk_size is not processed in param, additional_config["rbln_config"].

vllm_rbln/model_executor/models/optimum/qwen3_vl.py
```diff
@@ -343,3 +345,127 @@ class RBLNOptimumQwen3VLMoeForConditionalGeneration(
...
+            self.attention_manager.add(req_id, batch_idx)
+            kw = self.preprocess_for_decoder(
+                True, block_tables, input_ids, cache_position
+            )
+            prefill_kwargs = {
+                "inputs_embeds": model_input.inputs_embeds,
+                "position_embed": model_input.position_embed,
+                "block_tables": kw.pop("block_tables"),
+                "cache_position": kw.pop("cache_position"),
+                "batch_idx": batch_idx,
```
rebel-eunji, line 450: Until now, batch_idx of optimum is not used in vllm-rbln. batch_idx can triggers the confusion.
It must be refactored or renamed later to align with other models.

vllm_rbln/model_executor/models/optimum/compilation/multimodal/qwen.py
rebel-seinpark: https://github.com/RBLN-SW/vllm-rbln/pull/795/changes/e066b029db33123f2735cf018c039e1f10e313a3

vllm_rbln/model_executor/models/optimum/optimum_attention/optimum_attention_strategy.py
rebel-seinpark: Thanks, leaving the type: ignore for now

vllm_rbln/model_executor/models/optimum/qwen3_vl.py
rebel-seinpark: How about `local_table_id`?

rebel-eunji: Great!

rebel-seinpark: ~~Renamed the wrapper-side names to `local_table_id`/`local_table_ids` to match the kwargs used in vllm-rbln. Please check this [commit](https://github.com/RBLN-SW/vllm-rbln/pull/795/commits/36147d732e9c131ffdb3ac8eebecb6d2b19df246) once again.~~

I looked into aligning with the `local_table_id` naming the framework uses, but that name comes from the sliding-window models (exaone/gemma3), where it's the row into a local block table. Qwen3.5 has no local block table - `batch_idx` here is the row into the fixed [`max_num_seqs`] `GatedDeltaNet` recurrent/conv state cache, so importing `local_table_id` felt more misleading than helpful.

I decided to keep batch_idx because (1) it's optimum-rbln's own name for this graph input (modeling_qwen3_5 declares input_info as "`batch_idx`", and the architecture/runtime code all use it), and (2) since the prefill kwarg key must stay "`batch_idx`", keeping the local variable matching it avoids a translation layer (`"batch_idx": batch_idx`).

rebel-eunji (approved):
It works now. And I will refactor it in the future.

PR #798, opened 2026-07-20
fix(sampler): disable caching sampler

rebel-eunji:
### 🚀 Summary of Changes

1. Skip caching sampler
    - The common `compile` method triggers caching sampler.
2. Refactor the parameters of `compile`
    - It is better to use same parameters for greedy op and topk_topp_sampler.
3. Skip sampler warm-up in pooler models

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
4. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/v1/sample/rbln_sampler.py
+++ b/vllm_rbln/v1/sample/rbln_sampler.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections.abc import Callable
 from typing import Any
 
 import rebel
@@ -25,7 +26,7 @@ from vllm.v1.sample.sampler import Sampler as VLLMSampler
 import vllm_rbln.envs as envs
 from vllm_rbln.compilation import compile, create_compile_context
 from vllm_rbln.logger import init_logger
-from vllm_rbln.platform import HAS_TORCH_RBLN, USE_DEVICE_TENSOR
+from vllm_rbln.platform import USE_DEVICE_TENSOR
 from vllm_rbln.v1.sample.ops.logprobs import batched_count_greater_than
 from vllm_rbln.v1.sample.ops.penalties import (
     apply_all_penalties as rbln_apply_all_penalties,
@@ -59,6 +60,32 @@ def rbln_greedy_sample(logits: torch.Tensor) -> torch.Tensor:
     return torch.ops.rbln.argmax(logits)
 
 
+def compile_sampler(
+    op: Callable[..., torch.Tensor],
+    compile_context: rebel.CompileContext | None,
+) -> Callable[..., torch.Tensor]:
+    compile_context = (
+        compile_context
+        or create_compile_context(
+            use_global_ctx=True,
+        )
+        if not USE_DEVICE_TENSOR
+        else None
+    )
+    return compile(
+        op,
+        dynamic=False,
+        fullgraph=True,
+        compile_context=compile_context,
+        num_devices=1,
+        model_trace_method="export" if USE_DEVICE_TENSOR else "",
+        mode="strict" if envs.VLLM_RBLN_COMPILE_STRICT_MODE else "",
+        use_global_ctx=None if USE_DEVICE_TENSOR else True,
+        global_device_id=None if USE_DEVICE_TENSOR else 0,
+        use_cache=False,
+    )
+
+
 class RBLNTopKTopPSampler(nn.Module):
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -59,6 +60,32 @@ def rbln_greedy_sample(logits: torch.Tensor) -> torch.Tensor:
...
+        op,
+        dynamic=False,
+        fullgraph=True,
+        compile_context=compile_context,
+        num_devices=1,
+        model_trace_method="export" if USE_DEVICE_TENSOR else "",
+        mode="strict" if envs.VLLM_RBLN_COMPILE_STRICT_MODE else "",
+        use_global_ctx=None if USE_DEVICE_TENSOR else True,
+        global_device_id=None if USE_DEVICE_TENSOR else 0,
+        use_cache=False,
```
rebel-cjlee, line 87: This unconditionally disables the compile cache for **every** caller of `compile_sampler `. But this sampler is instantiated by both runners (`optimum_model_runner.py` and `rbln_model_runner.py`).

It looks like the caching problem only reproduces with `VLLM_RBLN_USE_VLLM_MODEL=0`. But, this change also drops caching on the `=1` path and forces a sampler recompile on every startup there for no reason.

Could we scope the disable to the affected path? e.g. take `use_cache` as an argument so the optimum runner passes `False` while the dynamo path keeps `True`, or gate it on `envs.VLLM_RBLN_USE_VLLM_MODEL` directly.

Also, could you add a short comment on **why** the cached sampler artifact is a problem on this path?
The commit just says 'tmp', so a short note (or a linked tracking issue) would help follow up on it later.

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -59,6 +60,32 @@ def rbln_greedy_sample(logits: torch.Tensor) -> torch.Tensor:
...
+    return compile(
+        op,
+        dynamic=False,
+        fullgraph=True,
+        compile_context=compile_context,
+        num_devices=1,
+        model_trace_method="export" if USE_DEVICE_TENSOR else "",
+        mode="strict" if envs.VLLM_RBLN_COMPILE_STRICT_MODE else "",
+        use_global_ctx=None if USE_DEVICE_TENSOR else True,
+        global_device_id=None if USE_DEVICE_TENSOR else 0,
```
rebel-cjlee: This drops the `HAS_TORCH_RBLN` gating in the previous code.
`torch-rbln` is currently gated to x86_64 only (`pyproject.toml:47`), so `HAS_TORCH_RBLN` is genuinely `False` on non-x86_64 environments. On those, the not `USE_DEVICE_TENSOR` branch now produces a different set of compile options instead of the previous `None`s. This behavior change looks unrelated to the caching fix.

It's also applied inconsistently: `rbln_rejection_sampler.py` and `rbln_model_runner.py` still use `HAS_TORCH_RBLN` (and `dev-legacy` branch relies on it as `has_torch_rbln`).

Could you explain why `HAS_TORCH_RBLN` was removed here? If it's intentional, I think it should be applied repo-wide in a separate PR rather than only in this one.

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -59,6 +60,32 @@ def rbln_greedy_sample(logits: torch.Tensor) -> torch.Tensor:
...
+    compile_context: rebel.CompileContext | None,
+) -> Callable[..., torch.Tensor]:
+    compile_context = (
+        compile_context
+        or create_compile_context(
+            use_global_ctx=True,
+        )
+        if not USE_DEVICE_TENSOR
+        else None
+    )
```
rebel-cjlee, line 74: The `compile_context` resolution logic is now duplicated in 3 places.

It looks like one of the goals of extracting `compile_sampler` was to reduce the duplicated `compile` code. But it was already duplicated in 2 places before this PR, and now it grew to 3. Since callers always pass an already-resolved context, the inner resolution logic is redundant.

`RBLNTopKTopPSampler` is only ever constructed inside `RBLNSampler`, so we could resolve the context once in `RBLNSampler.__init__` and just pass it down, letting both `RBLNTopKTopPSampler` and `compile_sampler` use the value as-is.

rebel-cjlee:
@rebel-eunji Looking at the CI results of another PR, the Optimum Test CI fails on RSD 1 but passes on RSD 4 and RSD 8. If this is really a sampler caching problem, could you explain why it only reproduces on RSD 1?

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: I thought the previous logic looks deprecated and has_torch_rbln is not required anymore. I'll re-apply this. I'll remove it with next PR (default to dtensor).

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/798/changes/0f32d67712aed808ec9b1be71a1d16d503b268cd

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/798/changes/0656ac884f3456cec7abb6c563993a11fd2acb77

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/798/changes/d0a29cbdbc99f02c51dc7ca5d5895c821c364654
Further details are omitted as they relate to a confidential compiler.

rebel-cjlee (approved):
Please make sure to share the details later.

PR #802, opened 2026-07-21
other(bump): bump vllm to 0.24.0 (optimum)

rebel-eunji:
### 🚀 Summary of Changes

## Changes
1. Freeing block logic changed.
In vllm-rbln, the inner→outer block mapping is dissolved only in `_evict_block`, which `can_allocate` triggers only when free outer blocks run short.

- 0.22: freed blocks (hashed or not) all went to the tail of vLLM's free queue, so a block was only reused under pool pressure. Inner-block reassignment therefore coincided with outer-block eviction, and reconciling the index at eviction time (remove_mapping) was sufficient.
- 0.24.0 (upstream #42656): unhashed freed blocks now go to the head of the queue and are reused first, even while free blocks remain. `can_allocate` still finds enough and never evicts, so the stale inner→outer mapping is never dissolved — leaving prefix cache-hit lookups to miss or resolve against a stale outer block.

0.22
```python
def free_blocks(self, ordered_blocks, prepend: bool = False):
    ...
    if prepend:
        self.free_block_queue.prepend_n(freed_blocks)   # HEAD (not used cases in RBLN)
    else:
        self.free_block_queue.append_n(freed_blocks)    # TAIL (default)
```

0.24
```python
def free_blocks(self, ordered_blocks):
    for block in ordered_blocks:
        block.ref_cnt -= 1
        if block.ref_cnt == 0 and not block.is_null:
            if block.block_hash is None:
                blocks_without_hash.append(block)
            else:
                blocks_with_hash.append(block)
    self.free_block_queue.prepend_n(blocks_without_hash)  # HEAD → reused first
    self.free_block_queue.append_n(blocks_with_hash)      # TAIL
```

2. The new parameter used in sampler -- use_fp64_gumbel
- This is not available in RBLN because we don't support fp64 logits in sampler as well as gumbel.
- So I'ved added assertion to validate use_fp64_gumbel is False.

3. pin_memory 
- Added a local _StreamPlaceholder (duck-typed, only .synchronize()) instead of importing it from torch_utils, to avoid the import-time circular import.

4. Add assertion of not used var -- VLLM_PREFIX_CACHE_RETENTION_INTERVAL, watermark

## Other changes
1. Changing kv_cache for type2 is just temporary patch for preventing global patch.
2. change vllm version from 0.22.0 to 0.24.0 for ci.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
5. Verify output: `...`
6. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-jonghewk:
@rebel-eunji can you summarize the changes for apply the new version?

rebel-eunji:
@rebel-jonghewk Done, summary's above.

rebel-jonghewk (approved):

vllm_rbln/platform.py
```diff
@@ -62,6 +62,17 @@ def bypass_backend(graph_module: torch.fx.GraphModule, example_inputs):
...
 
+class _StreamPlaceholder:
+    # Local copy of vllm.utils.torch_utils._StreamPlaceholder. That module
+    # computes `PIN_MEMORY = is_pin_memory_available()` at import time (vLLM
+    # 0.24), which resolves the platform and imports this module while
+    # torch_utils is still initializing. Importing the symbol from there at top
+    # level therefore hits a partially initialized module (circular import), so
+    # we duck-type our own here (only `.synchronize()` is required).
+    def __init__(self):
+        self.synchronize = lambda: None
```
rebel-cjlee: Just curious, I haven't run into this issue myself. In what cases does this usually happen?

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -134,6 +134,7 @@ class RBLNSampler(VLLMSampler):
     def __init__(
         self,
         logprobs_mode: LogprobsMode = "raw_logprobs",
+        use_fp64_gumbel: bool = False,
```
rebel-cjlee, line 137: I think its default value is False and we don't support this for now. So, we don't really need to introduce it. I'll clean it up as part of my current work.

rebel-cjlee (approved):

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: I just kept that for bump even though it is not used in rbln sampler because it overrides upstream vllm's sampler. overriding class emits the original class's parameter looks not good.

vllm_rbln/platform.py
rebel-eunji: I missed the current_stream is disappeared after refactoring. I'll remove it.

vllm_rbln/platform.py
```diff
@@ -62,6 +62,17 @@ def bypass_backend(graph_module: torch.fx.GraphModule, example_inputs):
...
+
+
```
rebel-eunji:
```suggestion
```

PR #805, opened 2026-07-21
feature: introduce mega cache

rebel-jonghewk:
### 🚀 Summary of Changes

Persist torch.compiler mega-cache so the rbln backend's `.rbln` blobs ship as a single atomic bundle per **(model, config-signature, local_rank)** under `VLLM_CACHE_ROOT/rbln`. Re-creation of #619 on the current `dev` (old branch was closed after `dev` was rewritten).

**1. Atomic bundle:** the per-file `cache_dir` scheme leaves partial artifacts on disk if compile fails mid-way (prefill ok, decode fails, ...); those orphan `.rbln` cache-hit and silently desync on the next run. Mega-cache writes `mega_cache.bin` **only after warm-up fully succeeds** — every program (prefill, decode, compute_logits, all buckets) must have compiled before the bundle becomes durable.

**2. Config-signature in the path:** a bundle keyed only by `(model, rank)` is shared by runs with different compile configs. torch's `save_cache_artifacts()` only serializes freshly compiled graphs and `load_cache_artifacts()` does not re-record loaded ones, so a stale bundle turns would-be-misses into un-re-recorded hits and the next `save()` writes a **partial** bundle. `config_signature = launch-stable vllm_config hash + vLLM worker-aligned compile env factors + rebel major.minor` keeps each config's snapshot isolated. Per-worker/runtime-only env vars are dropped so all TP/DP ranks across launches share one signature; the rank subdir isolates shards. Patch-level rebel bump stays cache-compatible; a minor bump invalidates.

**3. What's added:**
- **`v1/worker/mega_cache.py`:** `config_signature` / `bundle_path` / `load` / `save`; rebel import is lazy (module imports without a built rebel extension → unit-testable).
- **`v1/worker/rbln_model_runner.py`:** bracket `warmup_model` with `mega_cache.load`/`save`; `save` flushes disk-staged blobs before `save_cache_artifacts`.
- **`compilation/compiler.py`:** pass `mega_cache_only=True` alongside `cache_dir` so only the bundle is durable (no permanent per-file `.rbln`).
- **`tests/.../test_mega_cache.py`:** version parsing (patch-stable / minor-invalidates), bundle-path signature isolation, `config_signature` behavior.

* Re-creation of #619 on current `dev`
* Depends on rebellions-sw/rebel_compiler#11035 (`update(runtime): introduce mega cache`)

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

### 🧪 How to Test

1. Cold start: `vllm serve <model>` → only `~/.cache/vllm/rbln/<model>-<hash>/<sig>/rank0/mega_cache.bin` exists (no orphan `*.rbln`).
2. Warm restart (same config): rerun → `Loaded rbln mega-cache bundle from ...`, warm-up skips re-compile.
3. Config change: rerun with a different `VLLM_RBLN_DECODE_BATCH_BUCKET_STRATEGY` → new `<sig>/` dir; previous bundle preserved.
4. Failure injection: kill worker mid warm-up → bundle unchanged (no partial persistence).
5. TP=N: each rank writes its own `rank{N}/` subdir.

### 📋 Checklist


rebel-jonghewk:
Related: https://github.com/rebellions-sw/vllm-rbln-internal/issues/97

tests/torch_compile/unit/v1/worker/test_mega_cache.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-eunji: nit: 2026 copyright

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,206 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
```
rebel-eunji: nit: 2026

rebel-jonghewk:
Tested with ATOM dynamo CI: https://obedients.k8s.rebellions.in/orgs/main/pipelines/standalone-rebel-compiler-atom-vllm-dynamo-ci/builds/123

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,229 @@
...
+        return
+    from rebel.core import mega_cache as rbln_mega_cache
+
+    rbln_mega_cache.set_dir(cache_root())
+    path = bundle_path(model, sig)
+    if not os.path.isfile(path):
+        return
+    try:
+        with open(path, "rb") as src:
+            torch.compiler.load_cache_artifacts(src.read())
```
rebel-eunji: How about checking the return value of load_cache_artifacts?

```python
def load_cache_artifacts(serialized_artifacts: bytes) -> Optional["CacheInfo"]:
    from ._cache import CacheArtifactManager, CacheInfo

    artifacts = CacheArtifactManager.deserialize(serialized_artifacts)
    if artifacts is not None:
        return CacheArtifactManager.populate_caches(artifacts)
    return None
```

```suggestion
            info = torch.compiler.load_cache_artifacts(src.read())
        if info is None:
            logger.warning(
                "Ignored an unreadable rbln mega-cache bundle at %s; recompiling",
                path,
            )
        else:
            logger.info("Loaded rbln mega-cache bundle from %s", path)
```

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,229 @@
...
+    path = bundle_path(model, sig)
+    tmp_path = path + ".tmp"
+    try:
+        rbln_mega_cache.flush_to_bundle()
+        result = torch.compiler.save_cache_artifacts()
+        if result is None:
+            return
+        artifact_bytes, _ = result
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        with open(tmp_path, "wb") as dst:
```
rebel-eunji, line 196: To prevent the conflicts of multiple instances with same model(e.g. CI) in the same server, it is better to add pid to temp file name.

rebel-cjlee:
@rebel-jonghewk If this PR is now ready and the core logic is finalized, would it be okay if I add the corresponding tests to `tests/native`? 
(For reference, `tests/native` was first introduced in #887 and will soon replace `tests/torch_compile`.)

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,222 @@
...
+        "VLLM_RBLN_USE_MOE_TOKENS_MASK",
+        "VLLM_RBLN_DISPATCH_ALL2ALL",
+        "VLLM_RBLN_COMBINE_ALL2ALL",
+        "VLLM_RBLN_DECODE_BATCH_BUCKET_STRATEGY",
+        "VLLM_RBLN_DECODE_BATCH_BUCKET_MIN",
+        "VLLM_RBLN_DECODE_BATCH_BUCKET_STEP",
+        "VLLM_RBLN_DECODE_BATCH_BUCKET_LIMIT",
+        "VLLM_RBLN_DECODE_BATCH_BUCKET_MANUAL_BUCKETS",
+    }
+)
```
rebel-eunji, line 78: Keeping this list in sync looks fragile.
If we miss one new compile-affecting env var and we silently load a stale bundle.
So how about changing `RBLN_COMPILE_ENV` to `RBLN_NON_COMPILE_ENV` and   moving it to rbln_envs.py?

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,222 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""torch.compiler mega-cache bundle helpers for the rbln model runner.
+
+Persists/restores `torch.compiler.{save,load}_cache_artifacts()` bundles as a
+per-(model, config-signature, rank) file under VLLM_CACHE_ROOT.
+"""
+
+from __future__ import annotations
```
rebel-eunji, line 20: nit: this import looks unused
Can we drop it?

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,222 @@
...
+    pc = getattr(vllm_config, "parallel_config", None)
+    saved: dict = {}
+    if pc is not None and dataclasses.is_dataclass(pc):
+        for field in dataclasses.fields(pc):
+            if "port" in field.name.lower():
+                try:
+                    saved[field.name] = getattr(pc, field.name)
+                    object.__setattr__(pc, field.name, 0)
+                except Exception:  # pylint: disable=broad-exception-caught
+                    saved.pop(field.name, None)
```
rebel-eunji, line 108: Do we still need the field scan? `ParallelConfig.compute_hash()`
already excludes the port fields (https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/config/parallel.py#L746), with `_coord_store_port`
as the one exception.

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -0,0 +1,222 @@
...
+    cfg = _stable_compute_hash(vllm_config)
+    env = _compile_env_factors()
+    rebel_ver = _rebel_major_minor()
+    digest = hashlib.sha1("|".join([cfg, env, f"rebel={rebel_ver}"]).encode("utf-8"))
+    sig = digest.hexdigest()[:16]
+    logger.info(
+        "mega-cache config_signature=%s (cfg=%s env=%s rebel=%s)",
+        sig,
+        cfg[:8],
+        hashlib.sha1(env.encode("utf-8")).hexdigest()[:8],
```
rebel-eunji, line 129: `env` is already a hash. Is it intended?

vllm_rbln/v1/worker/mega_cache.py
rebel-jonghewk: My worry with inverting it to RBLN_NON_COMPILE_ENV is that the failure mode just flips: every rbln_envs var not on the denylist gets hashed into the cache key, so a newly added runtime-only var (metrics path, warm-up/port toggle, ...) that misses the denylist unintentionally becomes a recompile condition until someone adds it to the list.

vllm_rbln/v1/worker/mega_cache.py
rebel-jonghewk: Applied 👍

vllm_rbln/v1/worker/mega_cache.py
rebel-jonghewk: _coord_store_port is actually the case that bit us — since upstream's compute_hash() doesn't exclude it, a different coordinator port between runs invalidates the cache even though nothing compile-relevant changed. Ports are runtime-assigned and never affect the compiled artifact, so I'd prefer to keep the field scan and exclude *all* port-related fields uniformly, rather than depend on upstream's exclusion list staying complete. It also keeps us robust if upstream adds a new port field later without updating compute_hash().

vllm_rbln/v1/worker/mega_cache.py
rebel-jonghewk: Thanks. Fix applied.

vllm_rbln/v1/worker/mega_cache.py
rebel-jonghewk: Added both `RBLN_COMPILE_ENV` and `RBLN_NON_COMPILE_ENV` to `envs.py`, with a test asserting they exactly partition environment_variables. Any newly added env var now fails CI until it is explicitly classified into one of the two, so the list is forced to stay in sync. How about this?

rebel-eunji: Great Idea!

vllm_rbln/v1/worker/mega_cache.py
rebel-eunji: Agreed. Let's keep the scan.

rebel-eunji (approved):
LGTM 👍

rebel-cjlee (approved):
Looks good to me.
Just one thing: it seems like the ownership for the new feature will automatically be assigned to our team. Do we need to update the ownership?

rebel-cjlee:
+) I haven't gotten around to the tests in `tests/native` yet. I'll update them separately later.

tests/torch_compile/unit/v1/worker/test_mega_cache.py
```diff
@@ -0,0 +1,708 @@
...
+
+    def test_out_of_space_keeps_previous_bundle(self, bundle_env, monkeypatch):
+        mega_cache.save("m", "sig")
+        monkeypatch.setattr(
+            mega_cache.os, "replace", _raiser(OSError(errno.ENOSPC, "no space"))
+        )
+        bundle_env.artifact = b"second-bundle"
+        bundle_env.save_result = (bundle_env.artifact, object())
+        mega_cache.save("m", "sig")
+        monkeypatch.undo()
```
rebel-yskim: **[Nit]** `monkeypatch.undo()` also reverts the `bundle_env` fixture's patches.
  Pytest's `monkeypatch` is function-scoped and shared between the fixture and the
  test, so this call rolls back the `VLLM_CACHE_ROOT` → `tmp_path` redirection (and
  the `torch.compiler` / `rebel.core.mega_cache` stubs) along with the `os.replace`
  patch it is meant to undo. The assertions below only pass because
  `bundle_env.path` was captured as a string earlier — the moment one of them
  recomputes a path it will look at the real cache root instead of `tmp_path`.

  Fixture teardown already restores everything at the end of the test, so the
  simplest fix is to drop the `undo()` call. If a patch really has to be lifted
  mid-test, scope just that one:

  ```python
  with monkeypatch.context() as m:
      m.setattr(mega_cache.os, "replace", _raiser(OSError(errno.ENOSPC, "no space")))
      mega_cache.save("m", "sig")

  Same applies to the undo() calls in test_out_of_space_keeps_previous_bundle
  (L547) and test_fsync_failure_is_survivable (L566).

rebel-jonghewk:
> Looks good to me. Just one thing: it seems like the ownership for the new feature will automatically be assigned to our team. Do we need to update the ownership?

Yes that would be great.

tests/torch_compile/unit/v1/worker/test_mega_cache.py
rebel-jonghewk: Fix appiled.

rebel-jonghewk:
> +) I haven't gotten around to the tests in `tests/native` yet. I'll update them separately later.

Thanks. I'll merge this once the CI passes.

PR #807, opened 2026-07-21
model: support w8a8 linear & moe

rebel-kblee:
### 🚀 Summary of Changes

PR to cover that running with w8a8
- Quantize all fp8 block-quant Linear values where weight_block_size is present
- Quantize the gate/up/down project by MoE experts

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-cjlee:
Please fill the description about this change, how to test, or more information to understand it.

rebel-kblee:
> Please fill the description about this change, how to test, or more information to understand it.

I fill the description.
- how to test -> I think you could try it out using cr13 with minimax m2.7 model (maybe all configurations include spec dec). How would you prefer me to send you the testing details?

rebel-cjlee:
@rebel-kblee Since I don't currently have a device available, I'm wondering about the actual environment where this change can be run. Can it be executed with the compiler version in `uv.lock`? If not, it seems we need to explicitly state the compiler version or update `uv.lock`.

rebel-cjlee:
About how to test, I think what you mentioned here is sufficient:
> try it out using cr13 with minimax m2.7 model

vllm_rbln/envs.py
```diff
@@ -77,6 +77,22 @@
...
+    if value is not None:
+        return value.lower() in ("true", "1")
+
+    from vllm.platforms import current_platform
+
+    try:
+        device_name = current_platform.get_device_name()
+    except Exception:
+        device_name = ""
+    return "cr03" in device_name.lower()
```
rebel-cjlee: `get_use_w8a16()` is exposed differently from every other env var in this file, which breaks the uniform access pattern.

The convention here is that all env vars are registered in the `environment_variables` dict and accessed via `__getattr__` as `envs.VLLM_RBLN_XXX`. Even variables with non-trivial default-resolution logic (`use_auto_port`, `get_num_devices_per_local_rank`) still register their helper as the dict value (`"VLLM_RBLN_AUTO_PORT": use_auto_port`), so the access interface stays identical.

`get_use_w8a16()` isn't registered in the dict, so callers (`fp8.py`, `oot.py`) import and call the function directly. On top of that, the `TYPE_CHECKING` block declares `VLLM_RBLN_USE_W8A16: bool = False`, which makes it look accessible via `envs.VLLM_RBLN_USE_W8A16`. But, since it's not in the dict, that actually raises `AttributeError`.

Suggestion: register it in the dict just like `use_auto_port`:
```python
"VLLM_RBLN_USE_W8A16": get_use_w8a16,
```
Then callers can use `envs.VLLM_RBLN_USE_W8A16` consistently.

rebel-kblee: it is good suggestion. I applied it!

rebel-kblee:
@rebel-cjlee @rebel-jonghewk 
If we can implement this PR now, I think we’ll be able to test the w8a8 cases in the CI schedule. What would be the best way to proceed? If a review is possible, could you carry it out now?

@rebel-cjlee I have a question.
Regarding the refactoring, I’m not sure if I’ve understood correctly, but it seems you’ve changed the code to treat the method and kernel separately—is that correct?
If this is the case, unlike when we previously implemented custom methods for modelling, we’ll now be able to modify only the kernel (whilst the method follows the upstream implementation exactly). This suggests there may be differences compared to the graph we were previously tracking, so I’d like to ask if my understanding is correct.

vllm_rbln/envs.py
```diff
@@ -81,18 +81,31 @@
...
-        device_name = current_platform.get_device_name()
+        device_name = current_platform.get_device_name() or ""
     except Exception:
         device_name = ""
-    return "cr03" in device_name.lower()
+
+    if not device_name:
+        return True
+
+    _USE_W8A16 = "cr03" in device_name.lower()
```
rebel-cjlee: I'm a bit concerned about `get_use_w8a16()` deciding the default based on whether `"cr03"` is in the device name.

Previously the FP8 block linear was W8A16-only, so ATOM also ran W8A16 by default. With this change, ATOM (which doesn't match `"cr03"`) will silently default to W8A8. Unless the user explicitly sets `VLLM_RBLN_USE_W8A16=1`, a path that used to work fine on ATOM changes silently. Since ATOM can also run W8A16, this looks like it could break backward compatibility.

Hardcoding the device name via substring matching is also fragile:
1. it requires adding a new string for every new device
2. it's sensitive to any format change in `get_device_name()`

Is there a better approach here? Or could we decide this purely from the `VLLM_RBLN_USE_W8A16` (or `W8A8`) setting alone, defaulting it to use W8A16? That way W8A16 stays the safe default everywhere and W8A8 becomes opt-in, which keeps the existing behavior on ATOM or CR03 working.

rebel-cjlee:
> Regarding the refactoring, I’m not sure if I’ve understood correctly, but it seems you’ve changed the code to treat the method and kernel separately—is that correct?
If this is the case, unlike when we previously implemented custom methods for modelling, we’ll now be able to modify only the kernel (whilst the method follows the upstream implementation exactly). This suggests there may be differences compared to the graph we were previously tracking, so I’d like to ask if my understanding is correct.

@rebel-kblee Yes, your understanding is correct. We felt that extending in a vLLM-friendly way was the natural direction, and the refactoring follows that intent: the method now follows the upstream implementation, and the RBLN-specific behavior is isolated in the kernel.

On the graph point you raised, you're right that this can differ from the graph we were tracking before, since the method is no longer fully custom. When I did the refactoring I checked that, apart from the kernel, our path lines up almost exactly with upstream. If there are any parts we intentionally need to differ, the door is open to override them explicitly, so we can reflect any such cases that way.

rebel-jonghewk:
@rebel-kblee do we need to update the uv.lock version for rebel-compiler?

vllm_rbln/envs.py
rebel-jonghewk: @rebel-cjlee I agree that matching on a substring of the device name is fragile, and we need a better approach.

That said, I think selecting the custom kernel based on the device is necessary, and I don't see this as fundamentally different from how GPUs handle it. Defaulting to `VLLM_RBLN_USE_W8A16` (or the alternative) would require setting an environment variable, which is exactly what I want to avoid—we already have too many environment variables to manage.

rebel-kblee: I’ve amended the content you left in the review; could you please check that it’s on the right track?

rebel-kblee:
> @rebel-kblee do we need to update the uv.lock version for rebel-compiler?

I check with updated version.

vllm_rbln/envs.py
```diff
@@ -77,6 +77,36 @@
...
+        device_name = current_platform.get_device_name() or ""
+    except Exception:
+        device_name = ""
+
+    if not device_name:
+        return True
+
+    _USE_W8A16 = device_name.strip().upper() not in _W8A8_CAPABLE_NPUS
+
+    return _USE_W8A16
```
rebel-cjlee:
```suggestion
    _USE_W8A16 = not device_name or device_name.strip().upper() not in _W8A8_CAPABLE_NPUS
    return _USE_W8A16
```

rebel-jonghewk (approved):
lgtm

rebel-cjlee (approved):

PR #811, opened 2026-07-22
feature(metrics): rework metrics_v2 performance reporting

rebel-cjlee:
### 🚀 Summary of Changes

Reworks worker-level performance reporting in `metrics_v2.py`. Building on #769 (latency percentiles + removal of single-outlier averaging) and #731 (model+sampler), this PR applies those changes to `metrics_v2.py` and overhauls how metrics are computed, labeled, logged, and persisted.

#### Statistics

- Replace the unstable "drop one outlier, then average" logic with `mean | p50 | p90 | p99` latency reporting, following #769.
- Drop token count & throughput — misleading at the worker layer (padded device work, padded sampler bucket, spec-decode proposed-vs-accepted tokens) and not aggregatable across ranks.

#### Model + sampler per step

- Remove the standalone SAMPLER section; report `PREFILL + SAMPLE` / `DECODE + SAMPLE`, split by the model step's phase.
- Stash each model step as a pending report and merge the following sampler timing into it.
- Replace `ProfileSection`-based `profile()` with `profile_model()` / `profile_sampler()`.

#### End-to-end latency

- Report `PREFILL E2E` / `DECODE E2E` (latency only).
- Measured from the model span start to the sampler span and, so it captures the overhead between the two spans that the summed metric misses - hence `E2E >= (model + sampler)`.

#### Multi-worker reporting

- Tag each report with a per-worker rank covering TP/PP/DP/EP, including only the axes whose degree > 1 (e.g. a TP=2 run shows `[runner | TP0]` / `[runner | TP1]`; a single worker shows no tag). Each axis is probed independently, so an uninitialized group (e.g. EP when not used) is simply omitted rather than dropping the whole tag.
- The tag is captured at context initialization (after distributed init), so it is stable and survives distributed-group teardown at shutdown.
- Emit the console report as a single atomic `logger.info`, so the log prefix appears once and multi-rank blocks stay intact instead of interleaving line-by-line across workers.

#### File output

- Redesign file output as per-work JSON via the new `VLLM_RBLN_METRICS_DIR` (a directory). Each worker writes `metrics_<rank>.json` (`metrics.json` when not distributed), with `name` / `rank` / per-section stats separated for easy cross-rank aggregation. Console stays human-readable; the file is machine-readable. This replaces the pid-suffixed text file for v2; `VLLM_RBLN_METRICS_FILE` is kept for v1 (`metrics.py`).

#### Cleanups

- Extract a module-level `_mean` helper, add `Metrics.to_dict()` for serialization, rename `avg_* -> mean_*`, and drive the percentile render loop from the returned dict (no hardcoded keys).

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Run a short inference with metrics enabled and inspect the output.

```bash
# Console only
VLLM_RBLN_METRICS=1 RBLN_RUNTIME_TIMER=1 python <offline inference script>

# Console + per-worker JSON files
VLLM_RBLN_METRICS=1 VLLM_RBLN_METRICS_DIR=/tmp/metrics RBLN_RUNTIME_TIMER=1 \
  python <offline inference script>
```

Check:
1. Console shows 4 sections per run - `PREFILL + SAMPLE`, `DECODE + SAMPLE`, `PREFILL E2E`, `DECODE E2E` - each with Mean / Median / P90 / P99 latency. With parallelism, the header carries the rank tag (e.g. `[runner | ]`).
2. For each phase, E2E latency >= (+ SAMPLE) latency (the gap is the inter-span overhead).
3. When `VLLM_RBLN_METRICS_DIR` is set, `/tmp/metrics/` contains one JSON file per worker (e.g. `metrics_tp0.json`, or `metrics.json` on a single worker) with the same statistics in structured form.

> Unit tests are not included in this PR and will be added in a follow-up.

Output sample:
```
==================================================
PERFORMANCE STATISTICS [runner | TP1]
==================================================
PREFILL + SAMPLE METRICS:
  Total call counts        : ..
  Mean latency (ms)        : ..
  Median latency (ms)      : ..
  P90 latency (ms)         : ..
  P99 latency (ms)         : ..
  Mean host time (us)      : ..
  Mean device time (us)    : ..
  Mean ccl time (us)       : ..
  Mean prepare time (us)   : ..
--------------------------------------------------
DECODE + SAMPLE METRICS:
  Total call counts        : ..
  Mean latency (ms)        : ..
  Median latency (ms)      : ..
  P90 latency (ms)         : ..
  P99 latency (ms)         : ..
  Mean host time (us)      : ..
  Mean device time (us)    : ..
  Mean ccl time (us)       : ..
  Mean prepare time (us)   : ..
--------------------------------------------------
PREFILL E2E METRICS:
  Total call counts        : ..
  Mean latency (ms)        : ..
  Median latency (ms)      : ..
  P90 latency (ms)         : ..
  P99 latency (ms)         : ..
--------------------------------------------------
DECODE E2E METRICS:
  Total call counts        : ..
  Mean latency (ms)        : ..
  Median latency (ms)      : ..
  P90 latency (ms)         : ..
  P99 latency (ms)         : ..
--------------------------------------------------
==================================================
```

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/v1/worker/metrics_v2.py
+++ b/vllm_rbln/v1/worker/metrics_v2.py
@@ -284,9 +284,15 @@ def _write_metrics_json(
     }
 
     try:
+        payload_str = json.dumps(payload, indent=2)
+    except (TypeError, ValueError) as e:
+        logger.warning("Failed to serialize metrics JSON: %s", e)
+        return
+
+    try:
         os.makedirs(envs.VLLM_RBLN_METRICS_DIR, exist_ok=True)
         with open(path, "w", encoding="utf-8") as f:
-            json.dump(payload, f, indent=2)
+            f.write(payload_str)
     except OSError as e:
         logger.warning("Failed to write metrics JSON to %s: %s", path, e)
 
--- a/vllm_rbln/v1/worker/metrics_v2.py
+++ b/vllm_rbln/v1/worker/metrics_v2.py
@@ -146,11 +146,15 @@ class _TimingSpan:
     def __exit__(self, *args):
         # Record time before closing capture_ctx to exclude any internal
         # synchronization overhead inside rebel from the measured latency.
-        latency = time.perf_counter() - self._start
+        end = time.perf_counter()
+        latency = end - self._start
         self._capture_ctx.__exit__(*args)
         host, device, ccl, prepare = _parse_reports(self._reports)
         self._ctx._submit(
-            _Sample(latency, host, device, ccl, prepare, self._phase), self._is_sampler
+            _Sample(latency, host, device, ccl, prepare, self._phase),
+            self._is_sampler,
+            self._start,
+            end,
         )
         return False
 
@@ -171,6 +175,8 @@ class _PerformanceContext:
         self.rank_tag = _rank_tag()
         self._metrics: dict[bool, Metrics] = defaultdict(Metrics)
         self._pending: _Sample | None = None
+        self._e2e_start: float | None = None
+        self._e2e: dict[bool, Metrics] = defaultdict(Metrics)
 
     def profile_model(self, is_prefill: bool) -> _TimingSpan:
         # A prior model step with no sampler (intermediate chunked prefill,
@@ -181,19 +187,26 @@ class _PerformanceContext:
     def profile_sampler(self) -> _TimingSpan:
         return _TimingSpan(self, phase=None, is_sampler=True)
 
-    def _submit(self, sample: _Sample, is_sampler: bool) -> None:
+    def _submit(
+        self, sample: _Sample, is_sampler: bool, start: float, end: float
+    ) -> None:
         if not is_sampler:
             self._pending = sample  # model: stash, wait for the sampler
+            self._e2e_start = start
             return
         if self._pending is None:
             return  # sampler with no preceding model step; ignore
         self._record(self._pending.merged(sample))
+        if self._e2e_start is not None:
+            self._e2e[bool(self._pending.phase)].record(end - self._e2e_start)
         self._pending = None
+        self._e2e_start = None
 
     def _flush_pending(self) -> None:
         if self._pending is not None:
             self._record(self._pending)
             self._pending = None
+        self._e2e_start = None
 
     def _record(self, s: _Sample) -> None:
         self._metrics[bool(s.phase)].record(
@@ -202,10 +215,11 @@ class _PerformanceContext:
 
     def print_stats(self) -> None:
         self._flush_pending()  # record the final step if it had no sampler
-        sections = {
-            ("PREFILL + SAMPLE" if phase else "DECODE + SAMPLE"): m
-            for phase, m in sorted(self._metrics.items(), key=lambda x: not x[0])
-        }
+        sections: dict[str, Metrics] = {}
+        for phase, m in sorted(self._metrics.items(), key=lambda x: not x[0]):
+            sections["PREFILL + SAMPLE" if phase else "DECODE + SAMPLE"] = m
+        for phase, m in sorted(self._e2e.items(), key=lambda x: not x[0]):
+            sections["PREFILL E2E" if phase else "DECODE E2E"] = m
         name = f"{self.name} | {self.rank_tag}" if self.rank_tag else self.name
         _report_metrics(name, sections)
         _write_metrics_json(self.name, self.rank_tag, sections)
--- a/vllm_rbln/v1/worker/metrics_v2.py
+++ b/vllm_rbln/v1/worker/metrics_v2.py
@@ -18,7 +18,6 @@ import time
 from collections import defaultdict
 from contextlib import nullcontext
 from dataclasses import dataclass, field
-from enum import Enum
 from typing import TypeVar
 
 import numpy as np
@@ -89,6 +88,31 @@ class Metrics:
         return stats
 
 
+@dataclass
+class _Sample:
+    """One span's timing, summable across model+sampler within a step."""
+
+    latency: float
+    host: int | None = None
+    device: int | None = None
+    ccl: int | None = None
+    prepare: int | None = None
+    phase: bool | None = None  # True=prefill, False=decode (set by the model span)
+
+    def merged(self, other: "_Sample") -> "_Sample":
+        def _add(a: int | None, b: int | None) -> int | None:
+            return None if a is None and b is None else (a or 0) + (b or 0)
+
+        return _Sample(
+            latency=self.latency + other.latency,
+            host=_add(self.host, other.host),
+            device=_add(self.device, other.device),
+            ccl=_add(self.ccl, other.ccl),
+            prepare=_add(self.prepare, other.prepare),
+            phase=self.phase,  # keep the model step's phase
+        )
+
+
 try:
     import rebel  # type: ignore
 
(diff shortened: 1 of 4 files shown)


rebel-cjlee:
@rebel-jongwonlee I took a look at your changes in #808, which are related to this PR. I'm just mentioning it for reference, but if needed, I'm totally fine with moving the dump timing in this PR.

rebel-jonghewk (commented):
Is folding padded decode into the regular DECODE + SAMPLE bucket intentional?

vllm_rbln/v1/worker/metrics_v2.py
```diff
@@ -143,74 +169,130 @@ def __exit__(self, *_):
...
-        with open(path, "a", encoding="utf-8") as f:
-            f.write("\n".join(lines))
-            f.write("\n")
+        os.makedirs(envs.VLLM_RBLN_METRICS_DIR, exist_ok=True)
+        with open(path, "w", encoding="utf-8") as f:
+            json.dump(payload, f, indent=2)
     except OSError as e:
-        logger.warning("Failed to write metrics to file %s: %s", path, e)
+        logger.warning("Failed to write metrics JSON to %s: %s", path, e)
+
```
rebel-eunji, line 298: Consider catching TypeError/ValueError too — json.dump can fail on non-serializable values (e.g. numpy scalars) and would leave a truncated file.

```python
try:
    data = json.dumps(payload, indent=2)
    os.makedirs(envs.VLLM_RBLN_METRICS_DIR, exist_ok=True)
    with open(path, "w", encoding="utf-8") as f:
        f.write(data)
except (OSError, TypeError, ValueError) as e:
    logger.warning("Failed to write metrics JSON to %s: %s", path, e)
```

rebel-cjlee:
> Is folding padded decode into the regular DECODE + SAMPLE bucket intentional?
@rebel-jonghewk yes, it is. Do we need to distinguish them?

rebel-jonghewk:
> > Is folding padded decode into the regular DECODE + SAMPLE bucket intentional?
> > @rebel-jonghewk yes, it is. Do we need to distinguish them?

If it's not too difficult, I'd appreciate it if these could be distinguished.

vllm_rbln/v1/worker/metrics_v2.py
rebel-cjlee: Fixed as below: separated serialization and writing processes.
https://github.com/RBLN-SW/vllm-rbln/blob/8566b64fba3215520ca3d7f35f0e96c8cccfbed3/vllm_rbln/v1/worker/metrics_v2.py#L286-L297

rebel-cjlee:
@rebel-jonghewk I looked into porting the `padded_decode` split, but I'd rather not carry over the legacy definition as-is. It isn't a clean signal.

The legacy flag is:
```
padded_decode = (
    not is_prefill_phase
    and num_padded_tokens is not None
    and num_padded_tokens != batch_bucket_size
)
```

The main issue is that it conflates speculative decode with padding: `num_padded_tokens` is a token count while `batch_bucket_size` is a batch count. Under spec decode `num_padded_tokens` scales with `num_spec_tokens + 1` (e.g. `batch_bucket_size * max_tokens_per_req` in the specialized MoE-decode path), so it's always `!= batch_bucket_size` and every spec-decode step gets flagged as "padded" regardless of actual padding. (It's also DP-only, but that part is easy to gate.)

I agree distinguishing them is worthwhile, but I'd just rather settle on a correct criterion in a follow-up than reproduce a definition we already know is imprecise. Does it sound OK?

rebel-jonghewk:
> @rebel-jonghewk I looked into porting the `padded_decode` split, but I'd rather not carry over the legacy definition as-is. It isn't a clean signal.
> 
> The legacy flag is:
> 
> ```
> padded_decode = (
>     not is_prefill_phase
>     and num_padded_tokens is not None
>     and num_padded_tokens != batch_bucket_size
> )
> ```
> 
> The main issue is that it conflates speculative decode with padding: `num_padded_tokens` is a token count while `batch_bucket_size` is a batch count. Under spec decode `num_padded_tokens` scales with `num_spec_tokens + 1` (e.g. `batch_bucket_size * max_tokens_per_req` in the specialized MoE-decode path), so it's always `!= batch_bucket_size` and every spec-decode step gets flagged as "padded" regardless of actual padding. (It's also DP-only, but that part is easy to gate.)
> 
> I agree distinguishing them is worthwhile, but I'd just rather settle on a correct criterion in a follow-up than reproduce a definition we already know is imprecise. Does it sound OK?

Agreed, let's not carry the legacy definition over. Let's leave the split out of this PR and settle the criterion in a follow-up PR. cc @rebel-ykchoi

rebel-jonghewk (approved):

vllm_rbln/v1/worker/metrics_v2.py
```diff
@@ -143,74 +169,136 @@ def __exit__(self, *_):
...
+    ) -> None:
+        if not is_sampler:
+            self._pending = sample  # model: stash, wait for the sampler
+            self._e2e_start = start
+            return
+        if self._pending is None:
+            return  # sampler with no preceding model step; ignore
+        self._record(self._pending.merged(sample))
+        if self._e2e_start is not None:
+            self._e2e[bool(self._pending.phase)].record(end - self._e2e_start)
```
rebel-daeyang, line 201: This E2E ends at the sampler span, so it excludes the `tolist()` in `_bookkeeping_sync (rbln_model_runner.py:1274)`, where the host sync actually lands in async mode. The `dev-legacy` branch E2E tracker spanned through bookkeeping. Is stopping at the sampler intentional, or should it extend past it to reflect the real per-step wall time? (cc @rebel-jonghewk )

rebel-cjlee: First, although it's a narrow case, the measured times seemed quite similar when compared to the legacy branch. While we could extend the context span, matching the exact span from legacy might be tricky to achieve with just a simple `contextmanager`.

I understand you're currently running measurements related to the inplace device tensor. If it's okay with you, how about we apply this PR, continue our review from here, and address this issue in a follow-up PR?

+) Does the host sync always occur exactly when `tolist()` is called? I heard previously that it varies sometimes.
(Ideally, the best approach would be to have an explicit sync call that we can invoke right at the end of timing span..)

rebel-daeyang: Okay, let's land it and fix the span in a follow-up.

My main concern is coverage: the current E2E skips preprocess and bookkeeping, so if torch-rbln overhead grows there, none of the sections would catch it. A full-step E2E like legacy would surface that, even if it's not trivial to capture with a plain contextmanager.

`tolist()` forces a device to host sync only when the tensor is still device-resident.
And we can use torch.rbln.synchronize() at the span end if needed.

rebel-daeyang (approved):

vllm_rbln/v1/worker/metrics_v2.py
rebel-cjlee: I though `torch.rbln.synchronize()` wasn't supported yet. In that case, it seems more appropriate to call it right before the span end time.

rebel-daeyang: It's supported now, but it only drains async memory transfers, not the model/sampler compute, so calling it before span end wouldn't force compute completion by itself. We'd need to extend the span through `tolist()`, or implement a proper sync runtime API.

PR #732, opened 2026-06-28
feature: run rejection sampler in device tensor

rebel-eunji:
### 🚀 Summary of Changes

1. make rejection sampler fully running in device tensor 
2. refactor the process of building compile option

## Issues
- add clamp func to remove negative numbers in `num_accepted`. (It's weird...)
-> It is resolved by forcing the dtype of target_probs to float16.
---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
4. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #816, opened 2026-07-22
feature: dtensor support in rejection sampler

rebel-kblee:
### 🚀 Summary of Changes

dtensor support when use rejection sampler

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -422,12 +420,10 @@ def rejection_sample(
...
             src_offset += n
 
-        # FIXME required for device tensor?
-        # cu_num_draft_tokens = cu_num_draft_tokens.to(device=cpu_device)
         if sampling_metadata.top_k is not None:
-            sampling_metadata.top_k = sampling_metadata.top_k.to(device=cpu_device)
+            sampling_metadata.top_k = sampling_metadata.top_k.to(device=op_device)
         if sampling_metadata.top_p is not None:
-            sampling_metadata.top_p = sampling_metadata.top_p.to(device=cpu_device)
+            sampling_metadata.top_p = sampling_metadata.top_p.to(device=op_device)
```
rebel-eunji: It doesn't matter to delete these if branch.
ref) https://github.com/RBLN-SW/vllm-rbln/pull/732

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -340,15 +341,12 @@ def rejection_sample(
...
-                "VLLM_RBLN_USE_DEVICE_TENSOR is enabled, but the RBLN rejection "
-                "sampler only supports CPU input tensors. Forcing rejection sampler "
-                "inputs to CPU."
-            )
         cpu_device = "cpu"
+        op_device = (
+            target_probs.device if USE_DEVICE_TENSOR else torch.device(cpu_device)
+        )
+        draft_token_ids = draft_token_ids.to(cpu_device)
+        target_probs = target_probs.to(cpu_device)
```
rebel-eunji: 1. The cpu_device variable is no longer needed once the if branch is removed.
2. The name op_device is a bit hard to understand—could we use something clearer?

rebel-eunji (approved):

rebel-jonghewk:
@rebel-cjlee can you also take alook?

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -574,7 +574,9 @@ def profile(self, is_start: bool = True, profile_prefix: str | None = None):
 
     def execute_dummy_batch(self) -> None:
         bucket_size = self.model_runner.bucketing_manager.find_decode_batch_bucket(1)
-        query_len = 1 + self.model_runner.num_spec_tokens
+        # qlen drops to 1 when some rank elected no-spec, else num_spec+1.
+        no_spec = self.model_runner._reduce_cross_dp_no_spec(False)
```
rebel-cjlee: Does this part work properly?
Not all ranks will call `execute_dummy_batch` at the same time (some ranks can run the real execution), so is it safe to include communication operations inside this call?

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -48,6 +48,10 @@ class RBLNSchedulerOutput(SchedulerOutput):
     prefix caching."""
 
     kv_cache_copy_ops: list[KVCacheCopyOp] = field(default_factory=list)
+    # NOTE(RBLN): Set True when at least one finally-scheduled decode request
+    # cannot reach full spec (num_spec+1) via in-block query backfill, so this
+    # rank forced its decode batch to qlen=1 (no-spec).
+    step_no_spec_required: bool = False
```
rebel-cjlee: This seems unnecessary. Since `num_scheduled_tokens` is set to 1 when True, we can just use that information instead of adding a new field to `RBLNSchedulerOutput`. Also, I think it would be better to avoid adding a new field to the scheduler unless absolutely necessary.
Please let me know if I missed anything.

rebel-cjlee:
Actually, apart from the rejection sampler, the other changes seem out of scope. It would be helpful to have a more detailed explanation for the other modifications (such as why they are necessary and what the intention behind them is).

vllm_rbln/v1/worker/rbln_worker.py
rebel-cjlee: On closer inspection, it seems `_reduce_cross_dp_no_spec` is called exactly once on all ranks, just at different call sites. It doesn't look like it will hang. Could you please confirm?

vllm_rbln/v1/core/rbln_scheduler.py
rebel-kblee: I think `num_scheduled_tokens == 1` isn't unique to the forced case. A normal spec-decode request that just gets 0 drafts this step (ngram miss, or a block-boundary clamp) also ends up at num_scheduled=1 without being the backfill-forced no-spec, and plain non-spec decode is 1 too — so we can't tell them apart from num_scheduled alone.
But, i agree on dropping the another field -> i will try another mechanism to access

rebel-kblee:
@rebel-wonsubkim @rebel-jinhwan  It seems to be working fine for me, but could you please review it and check whether I’ve missed anything or made any mistakes?

rebel-kblee:
> Actually, apart from the rejection sampler, the other changes seem out of scope. It would be helpful to have a more detailed explanation for the other modifications (such as why they are necessary and what the intention behind them is).

Sure, as the changes to the compiler need to be included as well, I’ll summarize them and post them again above.

vllm_rbln/v1/worker/rbln_worker.py
rebel-kblee: Confirmed. Every rank calls `_reduce_cross_dp_no_spec` exactly once per step — busy ranks from inside execute_model, idle ranks from inside `execute_dummy_batch`. The two call sites look different but together they cover all ranks, so the all_reduce always has all its participants and can't deadlock.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-cjlee: Oh, we need to distinguish between the forced no-spec and the plain no-spec?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2907,6 +3066,9 @@ def warmup_model(self) -> None:
                 for size in self.bucketing_manager.batch_buckets:
                     self._dummy_sampler_run(size)
 
+            # 4. rejection sampler warmup
+            self._warmup_sampler_decode_batches()
```
rebel-cjlee, line 2984: Could we revisit the approach here for `_warmup_sampler_decode_batches` + `_add_dummy_requests` / `_make_dummy_scheduler_outputs` / `_execute_dummy_requests`?

I understand the goal is to warm up the rejection sampler, but this implementation fabricates dummy `NewRequestData` and `RBLNSchedulerOutput` (exec/cleanup) and pushes them through the full `execute_model()` -> `sample_tokens()` path. This style of warmup is actually something I intentionally removed in an earlier refactor (I treated it as an anti-pattern because of the coupling/maintenance cost), so I'd be a little hesitant to reintroduce it.

I think `_dummy_sampler_run()` right above is a good reference: it warms up the sampler by building only a dummy `SamplingMetadata` and calling `self.sampler(...)` directly. In the same way, could we warm up the rejection sampler by constructing its required inputs directly?

If there's a reason a direct call can't cover this, let me know.

rebel-kblee:
https://github.com/rebellions-sw/rebel_compiler/pull/12070
https://github.com/rebellions-sw/rebel_compiler/pull/12527
Those pr need to support dtensor rejection sampler.
@rebel-cjlee There are likely some sections of code that were included to check what needs to be amended in the compiler; further work is planned in line with your review comments.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-kblee: Yes — I think we do need to distinguish the two, and it matters specifically under DP+EP.
All DP ranks run a single fixed-shape decode graph in lockstep, so they must agree on the per-step query length (num_spec+1 vs 1):
- plain no-spec (0 drafts / ngram miss, or a lone non-spec decode): the request's single real query token can be padded up to num_spec+1 by _determine_batch_padding, so it can still ride along in a spec-qlen batch. It should not drag other ranks down.
- forced no-spec (backfill would cross the current KV block boundary): the request physically cannot form a contiguous num_spec+1 window, so it cannot be padded up — it must run at qlen=1, and therefore every DP rank must drop to qlen=1 for that step.

That's why the cross-DP OR-reduce keys on the "forced" flag. If we keyed it on num_scheduled_tokens == 1 instead, it would also fire for plain/0-draft/non-spec decode, so a single rank's ngram miss would force the whole DP group off spec for that step — a real throughput regression, not just a style concern.

If that is not the case, please let me know.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-kblee: I’d temporarily included this code whilst debugging to find a compiler bug, but I’ve now fixed it. Would you mind checking it for me?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2849,6 +2888,71 @@ def model_tensors(model: torch.nn.Module):
...
+        max_decode_batch = self.bucketing_manager.decode_batch_buckets[-1]
+
+        for batch_size in range(1, max_decode_batch + 1):
+            num_tokens = batch_size * num_spec
+            num_draft_tokens = [num_spec] * batch_size
+            draft_token_ids = torch.zeros(
+                num_tokens, dtype=torch.int32, device=self.device
+            )
+            target_probs = torch.zeros(
+                num_tokens, vocab_size, dtype=torch.float32, device=self.device
```
rebel-eunji, line 2880: Is hardcoded target_probs dtype okay?

rebel-kblee: The float32 hardcoding is intentional. In serving, target_probs = target_logits.softmax(dim=-1, dtype=torch.float32) (in forward) always produces float32 regardless of the model dtype, and since the rejection op is compiled with dynamic=False, the warmup inputs must match the serving dtype exactly. Switching to the model dtype would make the warmup and serving graphs diverge, causing a recompile/error.

rebel-eunji: I got it. Thanks.

rebel-cjlee:
Are there any plans to add test codes for this? We are currently in the middle of renewing our tests, so it would be a huge help if you could write even a simple test for this change.

rebel-eunji (approved):

rebel-minkyu:
@rebel-kblee Apologies for a late review. In the original code, only MoE module is padded to max_qlen across ranks, with everything else running on each rank's own qlen. Is there a specific reason this has to be aligned via `_reduce_cross_dp_no_spec`? Even considering a backfill case, attention already gets its own qlen after slicing at the end of MoE. If the issue is recompilation, can't it be solved with warm-up instead? It'd be great to confirm the failure case of the original code, since this reduces how often speculation applies(in ngram or suffix), which has a large impact on throughput.

rebel-kblee:
> @rebel-kblee Apologies for a late review. In the original code, only MoE module is padded to max_qlen across ranks, with everything else running on each rank's own qlen. Is there a specific reason this has to be aligned via `_reduce_cross_dp_no_spec`? Even considering a backfill case, attention already gets its own qlen after slicing at the end of MoE. If the issue is recompilation, can't it be solved with warm-up instead? It'd be great to confirm the failure case of the original code, since this reduces how often speculation applies(in ngram or suffix), which has a large impact on throughput.

Thank you for the review. I see that this has already been added to the dev branch. I’ll rebase it and take your feedback into consideration. 👍

rebel-kblee:
@rebel-minkyu  I have a question. I'm not entirely sure if I understand this correctly, but if “drafter” refers to a device model (MTP, Eagle, etc.), would what you described be possible in that scenario? If it's a CPU model (ngram or suffix), it seems like it would work as you described, but if it's a device model, I think it might hang due to the DP collective.

rebel-minkyu:
> @rebel-minkyu I have a question. I'm not entirely sure if I understand this correctly, but if “drafter” refers to a device model (MTP, Eagle, etc.), would what you described be possible in that scenario? If it's a CPU model (ngram or suffix), it seems like it would work as you described, but if it's a device model, I think it might hang due to the DP collective.

I think even with a device drafter, when using PD disaggregation, a rank performing its first decode step after a KV receive may run without any draft tokens, resulting in a qlen mismatch across ranks.

Could you elaborate a bit more on what exactly causes the hang from the DP collective? My understanding is that on the shape side all ranks are padded to max_pad before collective, and on the call-count side the number of collectives is the same regardless of qlen.

rebel-cjlee (approved):

rebel-minkyu (approved):

PR #804, opened 2026-07-21
fix(model): make weights contiguous after MTP/drafter setup, before compile

rebel-cjlee:
### 🚀 Summary of Changes

- **Problem**: DeepSeek-V3 fails to compile with a non-contiguous CPU tensor error when MTP is enabled, while the same model without MTP compiles fine. Also, this logic is not applied to **a draft model**.
- **Root cause**: `RBLNModelRunner.load_model()` called `_make_weights_contiguous()` immediately after loading the target model, before the drafter/MTP model was loaded.
- **Fix**: Move the `_make_weights_contiguous()` call so it run after drafter/MTP load, before compilation. This guarantees that all post-load weight transformations, including the MTP path's absorbed MLA weights.

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [x] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

Run DeepSeek-V3 with MTP.

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [x] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [x] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #821, opened 2026-07-23
fix(spec-decode): enable DeepSeek MTP under device tensor

rebel-eunik:
### 🚀 Summary of Changes

Enables MTP speculative decoding (--speculative-config '{"method":"mtp",...}') with VLLM_RBLN_USE_DEVICE_TENSOR=1, which previously failed at warmup compilation and then at inference.

- new patch `vllm_rbln/patches/deepseek_mtp.py`overrides `DeepSeekMultiTokenPredictorLayer.forward` to drop the position-0 embedding mask. 
The RBLN graph optimizer rejects that position-conditional select with `RuntimeError: RBLNCompileError: Graph Optimization: [UNEXPECTED_GRAPH]` under device-tensor.
- Runtime fixes — align CPU/device tensors in the multi-token draft loop.
---

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Verified on DeepSeek-V3 with num_speculative_tokens=3

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/patches/__init__.py
+++ b/vllm_rbln/patches/__init__.py
@@ -26,6 +26,7 @@ from vllm_rbln.patches.registry import (
 # ruff: noqa: F401
 from . import (
     attention,
+    deepseek_mtp,
     deepseek_v2,
     fp8_moe_method,
     gpt_oss,
--- a/vllm_rbln/patches/deepseek_mtp.py
+++ b/vllm_rbln/patches/deepseek_mtp.py
@@ -1,0 +1,59 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+from vllm.model_executor.models.deepseek_mtp import DeepSeekMultiTokenPredictorLayer
+
+from vllm_rbln.patches import register_patch
+from vllm_rbln.platform import USE_DEVICE_TENSOR
+
+
+@register_patch(
+    target=(
+        "vllm.model_executor.models.deepseek_mtp."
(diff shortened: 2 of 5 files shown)


rebel-kblee:
> The RBLN graph optimizer rejects that position-conditional select with RuntimeError: RBLNCompileError: Graph Optimization: [UNEXPECTED_GRAPH] under device-tensor.

Could you perhaps tell me which version of the compiler you’re using?

rebel-eunik:
I confirmed this issue in `0.11.1.dev304+g6948edb6.prod`

rebel-cjlee:
@rebel-kblee Would you follow-up this issue? 
And, could we resolve this issue by using a temporal patch?

rebel-kblee:
As I had already completed the deepseekv3 + mtp and verified that it worked before the refactoring branch was merged, it should work. It seems likely to be the same as the issues I encountered when I built it; could you perhaps check it?
fix pr : https://github.com/rebellions-sw/rebel_compiler/pull/12098
check with `rebel_compiler==0.11.1.dev432+g176ccc60.prod`

rebel-cjlee:
@rebel-kblee After bumping up vLLM version to 0.24.0, I will check it with the compiler version as you mentioned. 
Or, is it okay to use a compiler version used in #804 ?

rebel-kblee:
> RBLNCompileError: Graph Optimization: [UNEXPECTED_GRAPH] under device-tensor.

The compiler version I shared is the one that allows you to verify the functionality despite this error. I think you should be able to verify it without any issues using the version I shared!

rebel-cjlee:
@rebel-kblee @rebel-eunik I checked DeepSeek-V3 with MTP after bumping vLLM to 0.24.0 (using compiler `0.11.1.dev432+g176ccc60.prod`), and it works well, no compile error.

@rebel-eunik It works well, but do we need any other changes besides the patch to remove implicit compilations?

rebel-kblee:
@rebel-cjlee Do you happen to know when this PR will be merged!?

rebel-cjlee:
@rebel-kblee I'm checking this PR and observed a compiler error without the patch applied in this PR.

rebel-cjlee:
I found that the root cause of the error is the rejection sampler. Since the rejection sampler currently only supports CPU and doesn't support warmup yet, compilation(warmup) passes without issues, but it fails at runtime when it gets compiled with CPU inputs.
Device tensor support for the rejection sampler is currently being addressed in https://github.com/RBLN-SW/vllm-rbln/pull/816.

rebel-cjlee (approved):

PR #810, opened 2026-07-22
feature(optimum): wire gpu_memory_utilization to compile-time KV-cache memory_budget

rebel-eunji:
## Problem
The KV-cache estimation can be so tight that no memory is left for later allocations, causing OOM while creating the cs and compiling the sampler.

## Solution
The estimation needs to leave a sufficient margin.

optimum-rbln added a memory_budget feature for exactly this (PR #637): it controls the proportion of total DRAM the KV-cache estimation may assume (e.g. 0.92 → 92% of total DRAM).

This PR maps that memory_budget to vLLM's gpu_memory_utilization, so a single knob bounds how much DRAM the compiled model reserves for the KV cache.

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #825, opened 2026-07-24
fix(optimum): drop memory_budget from pooling/encoder compile config

rebel-thkim:
## Problem
Compiling pooling / encoder-only models (e.g. `bge-*`) fails with:

```
ValueError: Unexpected arguments: dict_keys(['memory_budget'])
```

`memory_budget` (wired from `gpu_memory_utilization` in #810) is a
decoder-only knob. The pooling/encoder RBLN config classes do not accept
it, so passing it into `from_pretrained` aborts compilation. The
enc-dec (Whisper) path already omits it; only the pooling path still
forwarded it.

## Fix
Drop `memory_budget` from `_for_pooling`'s `rbln_config` (and from its
signature / call site). Decoder and multimodal paths are unchanged and
keep passing it.

## Verify
- `py_compile` on `compilation/__init__.py`.
- Pooling compile no longer builds a config containing `memory_budget`.

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/compilation/__init__.py
+++ b/vllm_rbln/model_executor/models/optimum/compilation/__init__.py
@@ -207,8 +207,6 @@ class RBLNCompileSpec:
         model_cls = getattr(optimum.rbln, model_cls_name)
         assert model_cls is not None
 
-        # NOTE: memory_budget is a decoder-only knob. Pooling/encoder RBLN
-        # configs reject it ("Unexpected arguments"), so it is omitted here.
         rbln_config: dict[str, Any] = {
             "num_devices": num_devices,
             "batch_size": batch_size,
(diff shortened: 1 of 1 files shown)


vllm_rbln/model_executor/models/optimum/compilation/__init__.py
```diff
@@ -203,17 +202,17 @@ def _for_pooling(
...
         max_model_len: int,
         num_devices: int,
-        memory_budget: float,
     ) -> "RBLNCompileSpec":
         _, model_cls_name = get_rbln_model_info(config)
         model_cls = getattr(optimum.rbln, model_cls_name)
         assert model_cls is not None
 
+        # NOTE: memory_budget is a decoder-only knob. Pooling/encoder RBLN
+        # configs reject it ("Unexpected arguments"), so it is omitted here.
```
rebel-eunji:
```suggestion
```
I think there's no need to commen that.

rebel-eunji (approved):

rebel-jonghewk (approved):

rebel-cjlee (approved):

PR #826, opened 2026-07-24
fix(optimum): pin gemma3 prefill_chunk_size to mm_tokens_per_image

rebel-thkim:
## Problem
Compiling `gemma3` multimodal models fails with:

```
NotImplementedError: Not implemented for different prefill chunk sizes
between text and image prefill.
```

optimum-rbln forces `image_prefill_chunk_size == mm_tokens_per_image`
and then requires the text `prefill_chunk_size` to equal it. The text
chunk is therefore dictated by the model, but vllm-rbln forwarded its
own resolved `prefill_chunk_size` (defaulting to 128) into the gemma3
language-model config, which diverged from `mm_tokens_per_image` (256)
and tripped the check.

## Fix
In `_for_multimodal`, for `gemma3` pin the text `prefill_chunk_size` to
`config.mm_tokens_per_image` before building the compile config, so text
and image prefill agree. Guarded by architecture; `gemma4` buckets image
prefill independently and is left unchanged.

## Verify
- `py_compile` on `compilation/__init__.py`.
- gemma3 compile now uses `prefill_chunk_size == mm_tokens_per_image`,
  matching the forced `image_prefill_chunk_size`.

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/compilation/__init__.py
+++ b/vllm_rbln/model_executor/models/optimum/compilation/__init__.py
@@ -243,6 +243,16 @@ class RBLNCompileSpec:
             )
         model_cls = getattr(optimum.rbln, model_cls_name)
         assert model_cls is not None
+        # gemma3 forces image_prefill_chunk_size == mm_tokens_per_image and then
+        # requires text prefill_chunk_size == image_prefill_chunk_size, else it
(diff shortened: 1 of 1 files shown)


rebel-thkim:
Superseded: the gemma3 prefill_chunk_size fix will be handled in vllm-optimum-executor (set prefill_chunk_size=256) instead of vllm-rbln.

PR #829, opened 2026-07-26
other(bump): bump vllm to 0.24.0

rebel-eunji:
### 🚀 Summary of Changes

It bumps vllm to 0.24.0

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/.github/workflows/rbln_trigger_fsw_merge_ci.yaml
+++ b/.github/workflows/rbln_trigger_fsw_merge_ci.yaml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         vllm_version:
-          - "0.22.0+cpu"
+          - "0.24.0+cpu"
     steps:
       - name: Dispatch to fsw-integration (vllm ${{ matrix.vllm_version }})
         uses: Wandalen/wretry.action@v3
--- a/.github/workflows/rbln_trigger_fsw_nightly_e2e.yaml
+++ b/.github/workflows/rbln_trigger_fsw_nightly_e2e.yaml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         vllm_version:
-          - "0.22.0+cpu"
+          - "0.24.0+cpu"
     steps:
       - name: Dispatch to fsw-integration (vllm ${{ matrix.vllm_version }})
         uses: Wandalen/wretry.action@v3
--- a/.github/workflows/rbln_trigger_fsw_pr_ci.yaml
+++ b/.github/workflows/rbln_trigger_fsw_pr_ci.yaml
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         vllm_version:
-          - "0.22.0+cpu"
+          - "0.24.0+cpu"
     steps:
       - name: Dispatch to fsw-integration (vllm ${{ matrix.vllm_version }})
         uses: Wandalen/wretry.action@v3
--- a/.github/workflows/rbln_trigger_on_pr.yaml
+++ b/.github/workflows/rbln_trigger_on_pr.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - dev
-      - dev-0.22.0
+      - dev-0.24.0
 
 jobs:
   check-skip-ci:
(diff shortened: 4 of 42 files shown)


rebel-kblee:
vLLM >= 0.24 resolves platform plugins during `import vllm` - so a module-scope import makes `import vllm_rbln` re-enter the plugin loader before `register` is bound, and vLLM silently falls back to CpuPlatform.

It seems like the correct approach would be to move the contents of `from vllm_rbln import envs` into the `register_ops()` function and declare them there. Could you please confirm this?

rebel-kblee:
I think` patches/deepseek_mtp.py` could be change? Because of 0.24 made this forward return a 2-tuple, and `model_returns_tuple()` is hard-coded True for DeepSeekMTPModel. Can you check this?
I think the contents of the other diffs all seem correct. 👍

rebel-cjlee:
> vLLM >= 0.24 resolves platform plugins during import vllm - so a module-scope import makes import vllm_rbln re-enter the plugin loader before register is bound, and vLLM silently falls back to CpuPlatform.
>
> It seems like the correct approach would be to move the contents of from vllm_rbln import envs into the register_ops() function and declare them there. Could you please confirm this?

@rebel-kblee  You're right, the timing change is real and I confirmed it. Doesn't look like it actually happens now though: our import chain only touches modules that are already loaded by then, so we still get `RblnPlatform`. Let me know if I'm missing something.

Still a thin margin, one careless module-scope import and `vllm serve` quitely ends up on `CpuPlatform`. I'm reworking the test suite right now, so I'll add a subprocess test asserting `import vllm` gives `RblnPlatform`.

rebel-eunji:
@rebel-cjlee @rebel-kblee 
Thanks for your analysis. I just made a separate PR for that.
https://github.com/RBLN-SW/vllm-rbln/pull/876

And I saw your commit that sets VLLM_RBLN_USE_VLLM_MODEL before import vllm.
So I suggest add a guard to catch if the env var is set after import vllm.
Sometimes I made a mistake in examples/experimental scripts.
https://github.com/RBLN-SW/vllm-rbln/pull/877

rebel-cjlee:
> I think patches/deepseek_mtp.py could be change? Because of 0.24 made this forward return a 2-tuple, and model_returns_tuple() is hard-coded True for DeepSeekMTPModel. Can you check this?
I think the contents of the other diffs all seem correct. 👍

@rebel-eunik Could you see this issue?

rebel-cjlee:
@rebel-eunji @rebel-kblee 
I think we can just review and fix the DeepSeek MTP patch (checked by @rebel-eunik), and then proceed with the merge. Please let me know your thoughts. If you approve, I will merge the PR after the issue about DeepSeek MTP is resolved.

rebel-kblee (approved):
I couldn’t find any other issues without what I commented when I checked, so I think it should be fine!
LGTM!

rebel-cjlee:
@rebel-eunji I rebased onto your PR related to op registration, and resolved the conflicts. Could you take a look about it?

rebel-eunji:
Thanks. It looks like the custom op was simply moved to vllm_rbln/model_executor/layers/fused_moe/unquantized_fused_moe_method.py.
Looks good to me.

rebel-eunik:
> I think patches/deepseek_mtp.py could be change? Because of 0.24 made this forward return a 2-tuple, and model_returns_tuple() is hard-coded True for DeepSeekMTPModel. Can you check this?

Confirmed and fixed—DeepSeek is now working correctly with MTP. This change was caused by [vllm PR#45895](https://github.com/vllm-project/vllm/pull/45895).

rebel-cjlee (approved):

rebel-cjlee:
@rebel-eunji Could you check pytest failures in optimum path?
It looks like that PR #876 caused them. When it was rolled back, all tests passed.

rebel-cjlee:
@rebel-jonghewk Could you review these changes? If everything is okay, we can go ahead and merge.

rebel-jonghewk (approved):
lgtm

PR #830, opened 2026-07-27
fix: mimimax + eagle3 fixes

rebel-chanheo:


rebel-cjlee:
FYI, some of the changes here overlap with #821.

rebel-chanheo:
> FYI, some of the changes here overlap with #821.

Thanks. I will stack only the changes I have made on top of #821.

PR #831, opened 2026-07-27
fix(eagle3): enable EAGLE3 for MiniMax-M2.5 on the device-tensor path

rebel-chanheo:
### Summary

EAGLE3 does not work on MiniMax-M2.5 today: warmup segfaults, and once past that the drafter is fed the wrong hidden states and the sampler can write out-of-range ids. This PR removes those blockers so the path runs, and compiles the one op that would otherwise make it slower than no speculation at all.

Stacked on #821 (host/device tensor mixing in the multi-token draft loop).

### What was blocking

**Boot — SIGSEGV during KV warmup.** Upstream builds the draft→target scatter index inside `compute_logits` as `arange(draft_vocab_size) + d2t`. Both operands are input-independent, so the subgraph folds to an *anonymous* constant. Weight-free loading fills tensors by `data_ptr` with a name fallback over `named_parameters() + named_buffers()`; a folded constant matches neither and is skipped silently, then executes on placeholder data — arbitrary indices into a 200064-wide scatter.

`patches/llama_eagle3.py` registers `target_ids` as a named buffer (initialised to `arange`, so in range even if never filled), fills it in `load_weights`, and keeps the subgraph live with an input-derived zero.

**Wrong aux layers.** `eagle_aux_hidden_state_layer_ids` was read only at the top level of the config. MiniMax-M2.5-Eagle3 nests it under `eagle_config`, so the lookup fell back to a default triple and the target emitted hidden states from layers the drafter was never trained on.

**Wrong hidden states fed back.** #821 moved the gather to sampled positions inside `model_wrapper`; this override still gathered `positions` but not `hidden_states`, so every request read the leading request's rows.

**Stale host shadow.** The flash-attention builder reads `_seq_lens_cpu`, not `seq_lens`. Left unmirrored, every step after the first attends at the wrong sequence length — no crash, just lower acceptance.

**Out-of-range draft ids.** A drafter samples over `draft_vocab_size` and maps back through `d2t`; an unmapped id lands past the target vocab. Only `d_ids < 0` was rejected, so such an id rode a clamped token's probability through acceptance and was written out verbatim. Fixes the segfault in rebellions-sw/fsw-inference#430.

### Why one perf change is here

**`eagle.py` — compile the aux-state projection.** Eager, this single `Linear(num_aux * target_hidden → hidden)` costs 19.6 ms/step and scales with *weight* size rather than token count — the weight is materialised on every call. Compiled it is 0.30 ms, 26x. Without it EAGLE3 is slower than running without speculation, so it belongs with the enablement rather than in a later PR.

Decode has one shape once padded to the batch bucket, and the graph is built during `dummy_run`, so no request pays the compile. Prefill token counts are unbounded and stay eager.

### Also

- `rbln_model_runner.py` — return empty draft lists when the drafter is skipped. Wrapping `None` in a `DraftTokenIds` passes the core's `is not None` guard and kills the engine.
- `eagle.py` / `rbln_model_runner.py` — `record_function_or_nullcontext` scopes across the draft phase. No-ops unless `VLLM_CUSTOM_SCOPES_FOR_PROFILING=1`.

### Testing

MiniMax-M2.5 + `thoughtworks/MiniMax-M2.5-Eagle3`, DP4 / EP / weight-free / device-tensor, `max-num-seqs 4`, `num_speculative_tokens 3`.

- Warmup segfaults on every run without the buffer patch; boots with it.
- `d2t` mapping verified against the checkpoint: 32000 entries, `arange + d2t` in [9, 200051], all distinct, all < 200064.
- SWE-bench agent, 26 instances: acceptance 30% (0.90 of 3 drafts per step), no repetition failures.

### Related

- rebellions-sw/fsw-inference#431 — analysis and run records
- rebellions-sw/fsw-inference#430 — the segfault fixed by the bounds check


rebel-cjlee:
I haven't reviewed it yet, but we're planning a version bump to `0.24.0` soon. So, it would be great if you could check those changes and update this PR accordingly after the bump. Would that work for you?

rebel-chanheo:
> I haven't reviewed it yet, but we're planning a version bump to `0.24.0` soon. So, it would be great if you could check those changes and update this PR accordingly after the bump. Would that work for you?

Ok,  I'll rebase and re-run the enablement checks after the bump.

rebel-minkyu:
@rebel-chanheo Regarding the SIGSEGV in the upstream `compute_logits`, I guess the llama_eagle3 patches are resolving it. Since this way requires patches for each eagle3 models, would it be possible to handle it without the patches?
I've recently seen the same eagle3 issue, and initially handled it in [this way](https://github.com/RBLN-SW/vllm-rbln/commit/1c1f38f432116de51d36d934c1f0ed1be1810859). It just skips the problematic full->scatter logic and applies the d2t mapping directly after argmax, since (draft_logits -> full + d2t scatter -> argmax) and (draft_logits -> argmax -> d2t mapping) give the same id.
I'm not entirely sure if the attached way works properly with your test configs as well. Could you review it again by considering the trade-offs of the two approaches?

rebel-chanheo:
@rebel-minkyu 
Thanks, this makes sense. Since RBLN uses the draft logits only for greedy argmax, mapping the selected ID afterward should avoid the model-specific full-vocab scatter patch. I'm validating it now with the existing MiniMax setup, and I'll switch to this approach if the checks pass.

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -182,14 +241,12 @@ def propose(
         # (i.e., not the first proposal).
         if self.num_speculative_tokens > 1 and num_rejected_tokens is not None:
             common_attn_metadata.seq_lens -= num_rejected_tokens
+            self._mirror_seq_lens_to_host(common_attn_metadata, -num_rejected_tokens)
```
rebel-cjlee: This is dead after #860. That PR keeps all attention metadata on the host on does the H2D copy once inside the builder when the backend metadata is constructed. So, the builder reads `common_attn_metadata.seq_lens` directly and moves it to device itself, with no host/device shadow to sync. `_seq_lens_cpu` isn't passed anymore. `propose()` already mutates `seq_lens` in place, which is the exact tensor the builder consumes.
`cad._seq_lens_cpu` is unset and `_mirror_seq_lens_to_host` just early-returns.

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -431,7 +532,6 @@ def _build_dummy_attn_metadata(
             causal=True,
         )
 
-    @torch.inference_mode()
```
rebel-cjlee, line 434: Dropping `@torch.inference_mode()` here probably doesn't matter either way, but there's no comment on it so just checking. Was it intentional?

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -74,10 +120,17 @@ def propose(
...
+        # VLLM_CUSTOM_SCOPES_FOR_PROFILING=1. Read them under
+        # RBLN_RUNTIME_FORCE_SYNC=1: with async dispatch, device time lands in
+        # whichever scope happens to block.
         if self.method == "eagle3":
             assert isinstance(
                 self.model, (Eagle3LlamaForCausalLM, Eagle3DeepseekV2ForCausalLM)
             )
+            # Eager on purpose, and only fast with TORCH_RBLN_DEPLOY=ON: the
+            # non-deploy NaN/Inf scan walks this 56.6 MB weight on every call.
+            # See the commit message for the numbers.
```
rebel-cjlee: These feel like profiling/debug notes rather than comments about the code. I think we can just drop them.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1579,6 +1582,15 @@ def take_draft_token_ids(self) -> DraftTokenIds | None:
...
             return None
+        if self._draft_token_ids is None:
+            # No drafter ran this step -- `execute_model` skips it when
+            # `max_seq_len + num_spec > effective_drafter_max_model_len`.
+            # Wrapping the None gets past the core's `is not None` guard and
+            # kills the engine with "'NoneType' object is not iterable"; empty
+            # lists clear this step's drafts explicitly, whereas returning None
+            # would leave the previous step's drafts to be verified at the
+            # wrong position.
+            return DraftTokenIds(req_ids, [[] for _ in req_ids])
```
rebel-cjlee: This works fine as a guard. But upstream actually avoids the `None` altogether. In `execute_model`, when the input doesn't fit the drafter, it fills `_draft_token_ids` with zeros instead of leaving it `None` ([code](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/worker/gpu_model_runner.py#L4573-L4584)).
```python
            if not input_fits_in_drafter:
                # Zero out draft tokens so the scheduler doesn't schedule
                # stale drafts from the previous step.
                # For Nemotron-H: it is necessary to zero out the draft tokens,
                # otherwise the stale tokens will corrupt Mamba recurrent
                # state and logprobs for sequences near max_model_len.
                self._draft_token_ids = torch.zeros(
                    1, device=self.device, dtype=torch.int32
                ).expand(len(self.input_batch.req_ids), self.num_spec_tokens)
                self._draft_probs = None
                self._draft_prob_req_ids = None
                self._copy_draft_token_ids_to_cpu(scheduler_output, zeros_only=True)
```

That's why upstream's `take_draft_token_ids` has no `None` check at all.

What do you think about adding the same `if not input_fits_in_drafter` block in our code to match upstream? Would make syncing easier on the next version bump.

vllm_rbln/v1/spec_decode/eagle.py
rebel-chanheo: Thanks. removed!

vllm_rbln/v1/spec_decode/eagle.py
rebel-chanheo: Agreed on the numbers and the commit-message pointer — dropped.

I kept one line: `TORCH_RBLN_DEPLOY` defaults to `OFF` and nothing in the repo sets it, so "this path is only cheap with deploy mode on" isn't visible from the code. Happy to drop that too if you'd rather have it in the PR description instead.

vllm_rbln/v1/spec_decode/eagle.py
rebel-chanheo: Not intentional — thanks for catching it. Restored on dummy_run.

Upstream's `llm_base_proposer.dummy_run` carries `@torch.inference_mode()` and ours overrides it, so this just brings the override back in line.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-chanheo: Makes sense. Only the `_draft_token_ids` line applies here; `_draft_probs`, `_draft_prob_req_ids` and `_copy_draft_token_ids_to_cpu` don't exist on this runner. `take_draft_token_ids` already .tolist()s a tensor, so the None check drops out entirely like upstream.

rebel-chanheo:
@rebel-cjlee @rebel-minkyu 
I believe all the review comments have now been addressed. Could you PTAL and approve the PR if everything looks good? Thank you!

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -74,10 +77,16 @@ def propose(
         mm_embed_inputs: tuple[list[torch.Tensor], torch.Tensor] | None = None,
         num_rejected_tokens: torch.Tensor | None = None,
     ) -> torch.Tensor:
+        # `drafter/*` scopes below are no-ops unless
+        # VLLM_CUSTOM_SCOPES_FOR_PROFILING=1. Read them under
+        # RBLN_RUNTIME_FORCE_SYNC=1: with async dispatch, device time lands in
+        # whichever scope happens to block.
```
rebel-cjlee: These `drafter/*` scopes doesn't exist anymore. There's no `record_function_or_nullcontext` left in this file, so the comment points at nothing. Let's just drop it.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1652,6 +1661,10 @@ def propose_draft_token_ids(
             assert isinstance(self.drafter, RBLNEagleProposer)
             assert isinstance(sampled_token_ids, torch.Tensor)
 
+            # The "draft" phase covers more than propose(): these two helpers
+            # run first. prepare_next_token_ids_padded holds nine of the eleven
+            # integer ops that fall back to the host (`why: dtype-not-fp16`),
+            # each one a device round-trip.
```
rebel-cjlee: The "nine of the eleven integer ops" bit is a profiling snapshot that'll go stale as soon as the code shifts a little. Maybe move that into the PR notes/issue and keep just something like "these two helpers run before propose() and do host round-trip" or honestly just drop it. It's pure performance observation, so removing it altogether seems fine too.

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -379,7 +411,27 @@ def model_wrapper(
...
+                # row. Its index is input-independent, so the subgraph folds
+                # into an anonymous constant that weight-free apply cannot
+                # resolve by name; it then executes on placeholder indices and
+                # that out-of-bounds write is the SIGSEGV in KV warmup. Stay in
+                # draft-vocab space and map after the argmax instead
+                # (`_to_target_token_ids`) -- no per-model patch needed.
+                #
+                # Narrowed like the eagle3 branch in `propose()`: only these two
+                # carry `draft_id_to_target_id`, and the assert is what tells
+                # mypy `logits_processor` is callable rather than a Tensor.
```
rebel-cjlee: nit: can we drop the second paragraph? The "narrowed like the eagle3 branch" part is obvious from the code, and the assert is self-explanatory enough. mypy will catch it anyway if someone removes it later.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2985,6 +2998,11 @@ def warmup_model(self) -> None:
 
             # 5. specdec (medusa)
             if self.speculative_config and self.speculative_config.method == "medusa":
+                # Narrow like every other drafter branch in this file: only
+                # RBLNMedusaProposer.dummy_run takes no arguments, and without
+                # the assert the union resolves to the eagle proposer, whose
+                # dummy_run requires three.
+                assert isinstance(self.drafter, RBLNMedusaProposer)
```
rebel-cjlee: nit: I'd just drop this comment. We're already inside the `method=="medusa"` branch right above, so asserting it's the medusa proposer is self-evident.

rebel-cjlee (approved):
The code itself looks good to me. My only feedback is on a few comments. Nothing blocking, just cleanup. Thanks!

rebel-cjlee:
@rebel-jonghewk @rebel-eunji Could you take a look at the rejection sampler changes?
(cc, @rebel-chanheo)

vllm_rbln/v1/spec_decode/eagle.py
rebel-chanheo: Done

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-chanheo: Dropped it entirely

vllm_rbln/v1/spec_decode/eagle.py
rebel-chanheo: Done

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-chanheo: Done

rebel-chanheo:
@rebel-jonghewk @rebel-eunji  Could you please review this PR when you have a chance? I’d appreciate your feedback. Thanks!

rebel-jonghewk (approved):

PR #836, opened 2026-07-28
feature: release RBLN offload temp storage on worker shutdown

rebel-jonghewk:
## Problem

`torch.rbln.offload()` writes host-side tensor backing into a per-process temp dir under
`RBLN_OFFLOAD_DIR` (default `~/.cache/rbln_cache/offload`). The rebel runtime removes that dir when
it tears down, and the teardown hangs off a Python `atexit` hook — the last thing to run in the
worker process.

vLLM does not give a worker that much time. `MultiprocExecutor._ensure_worker_termination` waits a
few seconds, sends SIGTERM, waits again, then `p.kill()`. Teardown sits behind the KV-transfer and
profiler shutdown in `RBLNWorker.shutdown()`, so a worker that stalls anywhere in between is
SIGKILLed before the dir is ever removed. One DP4 run left 4.7 GB behind, and it is only reclaimed
lazily — by the stale-dir sweep the runtime does when some later process creates its *next* offload
dir.

## Change

`RBLNWorker.shutdown()` now calls `torch.rbln.release_offload_temp_storage()` as its first step,
before anything that can stall. Guarded by `has_torch_rbln`; failures are logged, not raised, so
cleanup can never break shutdown.

This moves the reclaim from the end of teardown to the start of shutdown — the same work, early
enough to win the race against the grace timer.

## Scope

Only this process's own temp dir is touched; the call never scans the offload base dir, so a
sibling worker's or another user's dir is out of reach. Dirs left by a run that was killed outright
are still handled by the existing sweep.

## Related

* torch-rbln: RBLN-SW/torch-rbln#175 — adds `torch.rbln.release_offload_temp_storage()`
* rebel-compiler: rebellions-sw/rebel_compiler#12421 — adds `rbln_release_offload_temp_storage`

Merge order is bottom-up: rebel-compiler, then torch-rbln (needs a rebel-compiler wheel containing
the new symbol), then this PR.


rebel-jonghewk:
@rebel-chanheo Could you let me know once the new RC tag for torch-rbln is released?

rebel-chanheo:
> @rebel-chanheo Could you let me know once the new RC tag for torch-rbln is released?

@rebel-jonghewk torch-rbln==0.4.0rc0 has been released!

https://pypi.org/project/torch-rbln/0.4.0rc0/

rebel-cjlee (approved):

PR #838, opened 2026-07-28
feature(pp): pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer

rebel-ykchoi:
### 🚀 Summary of Changes

#### Summary

Pipeline parallelism (`pipeline_parallel_size > 1`) on the RBLN path worked for plain decode but broke down the moment decode batching, speculative decoding, disaggregated serving, or NIXL KV-cache transfer entered the picture: a step's decode batch could overflow the compiled per-stage bucket, spec-decode misclassified the prefill/decode phase inconsistently across ranks, the KV connector saved only part of the model's layers, and a NIXL reader would connect to producer stages it does not read from and index per-stage descriptors out of range. This PR reorganizes the scheduler/runner so PP is a first-class, correct, and efficient mode, and extends prefill→decode KV transfer over NIXL to work when either side runs with pipeline parallelism.

The work is one story told in three movements: (1) give the scheduler a single PP-aware decode-admission budget, make speculative decoding rank-consistent on top of it, consolidate prefill/decode phase authority onto the scheduler, and turn batch balancing on by default for PP; (2) make the NIXL handshake advertise per-stage layer bands so a reader pulls from — and builds descriptors for — only the producer stages it owns; (3) correct speculative-decode memory sizing and data-parallel batch padding for the extra query positions a spec step carries.

##### A. PP-aware scheduler & runner

##### B. NIXL KV-cache transfer under PP

**5. Per-stage layer-band handshake.** Each pipeline stage advertises its own band of layers in the handshake metadata; a reader connects to and pulls from only the producer stages whose layers fall within the band it owns. This covers both the **symmetric layout** (prefill and decode share a pipeline size, stage matched to stage) and the **fan-in layout** (a reader without pipeline parallelism aggregates several producer stages).

**6. Overlap-aware remote descriptor build.** Per-stage remote descriptors are built assuming a producer exposes no more regions than the reader owns — which holds only for the stages the reader actually reads from. Stages outside the reader's band (which can expose more regions when the layer count does not divide evenly across stages) are **skipped before that build**, so it never indexes out of range.

**7. Non-last-rank KV output via the two-phase path.** The KV-connector output for a non-last PP rank is surfaced through the runner's two-phase `sample_tokens()` path; `RBLNWorker.execute_model` returns `None` on non-last ranks instead of re-emitting the same finished send/recv notifications from an empty output, so every rank saves its own layers' KV exactly once.

**8. Unsupported combinations rejected at handshake.** TP together with PP, sliding-window attention together with PP, and unequal prefill/decode block sizes are rejected during the handshake. Transfers without pipeline parallelism keep the existing single-stage path unchanged.

##### C. Speculative-decode corrections

**9. Memory sizing & DP batch padding.** Both stem from a spec step carrying `num_speculative_tokens + 1` query positions:
- Available-memory estimation counts the extra decode graph warm-up compiles per `(batch bucket, query length)` for spec decode, so the command-stream-buffer reservation matches warm-up and the KV-cache-block estimate does not over-reserve device memory.
- When DP ranks disagree on whether the current step is a speculative (multi-token) decode step, the batch is padded up to the largest decode bucket (mirroring the existing prefill-step policy), keeping every rank on an already-compiled graph. No-op for single-bucket configs.

#### Testing

- **Unit** — PP decode-admission budget and demand-spread contrast (`test_scheduler.py`), anchor-reconciled deferral predicate and multi-accept payload (`test_rbln_model_runner.py`), backfill/slide query-window math (`test_query_backfill.py`), PP disaggregation (`test_pd_disaggregation.py`), the NIXL per-stage layer-band handshake and overlap-skip (`test_rbln_nixl_handshake_pp.py`), handshake metadata PP fields (`test_rbln_nixl_metadata.py`), worker-side handshake (`test_rbln_worker_handshake.py`), and the non-last-rank `execute_model` return contract (`test_rbln_worker.py::TestExecuteModel`).
- `pre-commit run --hook-stage manual` clean (ruff, mypy 3.10–3.13, typos, license); full runner/scheduler/pd/connector unit suites green.

#### Risk / compatibility

- The spec-decode changes are no-ops for single-bucket / non-spec configurations.
- The `VLLM_RBLN_PP_BALANCE_DECODE_BATCH` default flip is confined to PP; the escape hatch preserves the old behavior.

#### Affected modules

Platform, Scheduler, Runner, NIXL KV connector, Worker.

---

* Resolves #
* Related to #
https://github.com/RBLN-SW/vllm-rbln/pull/777
---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/torch_compile/unit/test_forward_context.py
+++ b/tests/torch_compile/unit/test_forward_context.py
@@ -46,3 +46,62 @@ def test_forward_context(vllm_config, attn_metadata_mock: MagicMock):
             f"Expected 'dp_metadata' class name is RBLNDPMetadata, \
                     got {get_forward_context().dp_metadata.__class__.__name__}"
         )
+
+
+def test_num_tokens_and_reqs_across_dp_bit_pack(monkeypatch):
+    """Round-trip the bit-packed (num_tokens, num_reqs, is_prefill, is_idle)
+    all-reduce. Intercepts the inner all-gather so no real DP group is needed,
+    and checks each field unpacks correctly -- including the new is_idle mask
+    and the (prefill -> num_reqs None) behavior."""
+    from vllm_rbln.forward_context import RBLNDPMetadata
+
+    token_bits = 16
+
+    def peer_encoded(num_tokens, num_reqs):
+        return num_tokens | (num_reqs << token_bits)
+
+    def make_fake_inner(peer_value):
+        def fake_inner(encoded, dp_size, dp_rank):
+            arr = [0, 0]
+            arr[dp_rank] = encoded
+            arr[1 - dp_rank] = peer_value
+            return torch.tensor(arr, dtype=torch.int32)
+
+        return fake_inner
+
+    # Decode: this rank (rank0) = idle (num_tokens=1, num_reqs=1, is_idle);
+    # peer (rank1) = busy (num_tokens=8, num_reqs=2).
+    monkeypatch.setattr(
+        RBLNDPMetadata,
+        "num_tokens_across_dp",
+        staticmethod(make_fake_inner(peer_encoded(8, 2))),
+    )
+    tokens, reqs, idle = RBLNDPMetadata.num_tokens_and_reqs_across_dp(
+        num_tokens=1,
+        num_reqs=1,
+        dp_size=2,
+        dp_rank=0,
+        is_prefill=False,
+        is_idle=True,
+    )
+    assert tokens.tolist() == [1, 8]
+    assert reqs is not None and reqs.tolist() == [1, 2]
+    assert idle.tolist() == [1, 0]
+
+    # Prefill on this rank -> num_reqs_across_dp is None; is_idle stays 0.
+    monkeypatch.setattr(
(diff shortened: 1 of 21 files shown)


rebel-ykchoi:
Superseded by a 3-way split for reviewability (same content, plus one new fix):

- #856 — feat(pp): PP-aware scheduler & runner (section A) → `dev`
- #857 — feat(nixl): PP KV-cache transfer (section B), stacked on #856
- #858 — fix(dp): spec-decode memory sizing + DP batch padding (section C) **and** a new idle-data-parallel-step alignment fix not in this PR, based on #856 and parallel to #857

Closing in favor of those.

PR #845, opened 2026-07-29
feature(custom-op): defer op registration to rebel-compiler with a fallback

rebel-eunji:
### 🚀 Summary of Changes

This is a temporary patch that lets the custom ops move from vllm-rbln to
rebel-compiler without releasing both at once. Every definition in this package
now goes through `vllm_rbln/custom_ops.py`, which defers to rebel-compiler when
the installed compiler already registers the op and falls back to the copy here
when it does not. The copies are removed once the minimum supported compiler
registers all of them.

The custom ops fall into three groups:

1. only in rebel-compiler
2. only in vllm-rbln
3. in both

Only the third group can conflict, and two of its ops have a signature
difference:

| op | difference |
|---|---|
| `custom_moe_glu` | `n_group` / `topk_group` — present in rebel-compiler, absent in vllm-rbln |
| `custom_moe_glu_mxfp4` | `n_group` / `topk_group` — present in rebel-compiler, absent in vllm-rbln |

The difference is harmless: both parameters are optional, no call site in
vllm-rbln passes them, and the compiler's converter never reads them
(`moe_glu_routing` collects the five required inputs and then only looks at
`expert_map` and the bias slot), so they cannot reach the kernel either way.

When the two signatures differ, `custom_ops.py` reports it at import time:
```
[vllm-rbln] WARNING 07-28 02:54:05 [custom_ops.py:142] rbln_custom_ops::custom_moe_glu_mxfp4 has a different signature in vllm-rbln than in the installed rebel-compiler. The compiler's definition is used; the copy in this package is stale.
[vllm-rbln] WARNING 07-28 02:54:05 [custom_ops.py:142]   rebel-compiler: (Tensor hidden_states, Tensor gate_proj_blocks, Tensor gate_proj_scales, Tensor gate_proj_bias, Tensor up_proj_blocks, Tensor up_proj_scales, Tensor up_proj_bias, Tensor down_proj_blocks, Tensor down_proj_scales, Tensor down_proj_bias, Tensor masked_routing_weight, Tensor alpha, Tensor limit, Tensor? expert_map=None, SymInt? n_group=None, SymInt? topk_group=None) -> Tensor
[vllm-rbln] WARNING 07-28 02:54:05 [custom_ops.py:142]   vllm-rbln:      (Tensor hidden_states, Tensor gate_proj_blocks, Tensor gate_proj_scales, Tensor gate_proj_bias, Tensor up_proj_blocks, Tensor up_proj_scales, Tensor up_proj_bias, Tensor down_proj_blocks, Tensor down_proj_scales, Tensor down_proj_bias, Tensor masked_routing_weight, Tensor alpha, Tensor limit, Tensor? expert_map=None) -> Tensor
```

### Argument names follow rebel-compiler

Two fallback copies had a routing argument whose name differed from
rebel-compiler's, so they were renamed to match. rebel-compiler owns these ops,
which makes its name the contract: optimum-rbln passes this argument by keyword
(`masked_routing_weight=`), so a copy with a different name breaks that call
whenever the fallback is the definition in use. Nothing changes for vllm-rbln
itself, which passes the argument positionally.

| op | before | now | matches |
|---|---|---|---|
| `custom_moe_glu_mxfp4` | `masked_routing_weights` | `masked_routing_weight` | rebel-compiler `dev` and rebellions-sw/rebel_compiler#12049 |
| `custom_moe_glu_group_dequantize` | `masked_routing_weight` | `masked_routing_weights` | the released rebel-compiler (a `dev` build) |

The two go in opposite directions because rebel-compiler `dev` is inconsistent
today: singular for `mxfp4`, plural for `group_dequantize`. #12049 makes both
singular. Matching the compiler that is actually installed keeps the
signature-drift warning quiet, so `group_dequantize` stays plural for now and is
renamed to the singular form once the pin moves to a build that includes #12049.
A comment above the definition in `fp8.py` records that.

### Import order

rebel-compiler registers its ops before vllm-rbln registers any, because
`vllm_rbln/custom_ops.py` runs `import rebel` at module import time and every op
module imports `custom_ops` at the top, before its own decorators run.

```
import rebel
 └ rebel/__init__.py:28        from . import ops
    └ ops/__init__.py:16       from . import torch_custom_ops
       └ torch_custom_ops/__init__.py   imports the submodules → decorators run
```

This matters because `torch.library.custom_op` does not reject a duplicate: the
later registration silently replaces both the schema and the implementation.
Doing the import here rather than relying on `platform.py` keeps the outcome the
same no matter which module loads first.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
5. Verify output: `...`
6. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-jongho (approved):
LGTM

vllm_rbln/custom_ops.py
```diff
@@ -0,0 +1,141 @@
...
+
+#: Ops this package had to define because the compiler does not provide them.
+FALLBACK: set[str] = set()
+
+#: Schema of this package's copy of each op, whether or not it was used. This is
+#: what the drift warning compares against the compiler's registration.
+LOCAL_SCHEMAS: dict[str, str] = {}
+
+
+def fallback_ops() -> frozenset[str]:
```
rebel-jonghewk: when is this used? Is it currenlty used in the code?

rebel-eunji: It was my mistake. I had to remove it clearly..

vllm_rbln/custom_ops.py
```diff
@@ -0,0 +1,128 @@
...
+
+import rebel  # noqa: F401
+import torch
+
+from vllm_rbln.logger import init_logger
+
+logger = init_logger(__name__)
+
+#: Ops taken from rebel-compiler.
+FROM_COMPILER: set[str] = set()
```
rebel-jonghewk: nit: How about `ALREADY_DEFINED` instead of `FROM_COMPILER`? The value doesn't have to come from rebel-compiler — it could be defined anywhere else, right?

rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/845/changes/e174a62829db4a0db95d34317044aea8b95ffa6d
Thanks

vllm_rbln/custom_ops.py
```diff
@@ -0,0 +1,129 @@
...
+#: definition -- rebel-compiler's -- is the one in use.
+ALREADY_DEFINED: set[str] = set()
+
+#: Schema this package declares for each op, whether or not its copy was used.
+#: This is what the drift warning compares against the existing registration.
+DECLARED_SCHEMAS: dict[str, str] = {}
+
+
+def custom_op(name: str, **kwargs: Any) -> Callable[[Callable], Any]:
+    """Define `name` unless rebel-compiler already did. Drop-in for
```
rebel-jonghewk: comments has `rebel-compiler` but we can maybe remove it?

vllm_rbln/custom_ops.py
```diff
@@ -0,0 +1,129 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections.abc import Callable
+from typing import Any
+
+import rebel  # noqa: F401
```
rebel-jonghewk, line 19: I think we should add a comment here. This is the code that ensures RBLN's custom ops are registered first, right?

rebel-eunji: RIght. I added one line comment to specify the importance of `import rebel`.
[3fb7639f62641a9876f1a79f322a4bb701d037e7](https://github.com/RBLN-SW/vllm-rbln/pull/845/changes/3fb7639f62641a9876f1a79f322a4bb701d037e7)

vllm_rbln/custom_ops.py
rebel-eunji: I removed the `rebel-compiler`.

rebel-jonghewk (approved):

PR #846, opened 2026-07-29
feature(optimum): add Qwen3-Reranker score API support

rebel-eunji:
## Summary

Add Qwen3-Reranker support through vLLM's native `LLM.score()` API on the Optimum-RBLN path.

The implementation exposes the original Qwen3-Reranker checkpoint as a pooling model while reusing the existing Qwen3 generation graph. It computes relevance from the final-position `yes` and `no` vocabulary logits:

```text
score = p_yes / (p_yes + p_no)
      = sigmoid(logit_yes - logit_no)
```

Reusing the generation graph allows the `generate()` and `score()` paths to share the same compiled RBLN artifact without loading or converting classifier weights separately.

### 1. Identify the original Qwen3-Reranker configuration

The model is loaded using the same overrides expected by upstream vLLM:

```python
hf_overrides={
    "architectures": ["Qwen3ForSequenceClassification"],
    "classifier_from_token": ["no", "yes"],
    "is_original_qwen3_reranker": True,
}
```

The implementation recognizes this combination as the original, unconverted Qwen3-Reranker checkpoint.

`classifier_from_token` is also used to resolve the `no` and `yes` vocabulary token IDs. Both labels must be present and must each map to a single token.

### 2. Run the reranker through the generation graph

Although callers use `runner="pooling"` and `LLM.score()`, the score must be computed from vocabulary logits produced by `lm_head`.

Therefore:

```text
Qwen3ForSequenceClassification
        ↓
Qwen3ForCausalLM
        ↓
RBLNQwen3ForCausalLM
```

The architecture is remapped to `Qwen3ForCausalLM`, rather than `Qwen3Model`, so that `lm_head` remains in the compiled graph.

This also keeps Qwen3-Reranker separate from Qwen3-Embedding:

| Model | Pooling source | Required graph |
| --- | --- | --- |
| Qwen3-Embedding | Hidden states | `Qwen3Model` |
| Qwen3-Reranker | `yes`/`no` vocabulary logits | `Qwen3ForCausalLM` |

Without this distinction, the reranker could be mistaken for an embedding model and remapped to `Qwen3Model`, which would remove the `lm_head` needed to calculate its score.

### 3. Convert label logits into a relevance score

The generation graph returns vocabulary logits for the final sequence position:

```text
[batch, 1, vocab_size]
```

The reranker selects the configured label logits and reduces them to one value per request:

```text
score_logit = logit_yes - logit_no
```

The classifier pooler then applies sigmoid:

```text
sigmoid(score_logit)
```

This is mathematically equivalent to applying a two-way softmax over the `no` and `yes` logits.

### 4. Adapt pooling inputs to the generation graph

Pooling requests arrive right-padded, while the Qwen3 generation graph requires left-padded inputs.

The reranker model therefore:

- recovers each request's actual length from its positions;
- moves its tokens to the right side of the shared input width;
- builds the corresponding left-padded attention mask;
- rejects inputs exceeding the compiled maximum sequence length.

Prefix caching is also disabled because the public API uses a pooling runner, for which prefix caching is not supported.

## Why the vLLM Registry Guard Was Removed

This PR removes the Optimum-RBLN `validate_arch_supported()` guard that directly checked the architecture against:

```python
vllm.model_executor.models.ModelRegistry.get_supported_archs()
```

### The guard duplicated validation already performed by vLLM

vLLM resolves and validates the model architecture while constructing `ModelConfig`, before the Optimum-RBLN model runner is created.

An architecture that vLLM cannot resolve is therefore rejected before it reaches the RBLN runner.

Maintaining a second copy of that validation in Optimum-RBLN did not add another effective safety boundary.

### The direct registry check did not model vLLM's architecture resolution

Not every architecture accepted by vLLM appears directly in `ModelRegistry.get_supported_archs()`.

For example, the Qwen3-Reranker is presented as:

```text
Qwen3ForSequenceClassification
```

Upstream vLLM accepts it through its architecture normalization and `convert=classify` rules, deriving it from the registered Qwen3 causal language model.

The removed guard inspected only the raw registry and could therefore reject architectures that vLLM itself had already resolved successfully.

### RBLN-specific validation remains

The RBLN runner still calls:

```python
get_rbln_model_info(model_config)
```

before applying any internal architecture remapping.

This preserves the validation that matters on the RBLN side: the resolved architecture must have an Optimum-RBLN implementation.

The responsibilities are now separated as follows:

| Validation | Owner |
| --- | --- |
| Can vLLM resolve this model architecture? | vLLM |
| Does Optimum-RBLN have an implementation for it? | `get_rbln_model_info()` |
| Is the reranker configuration internally valid? | Qwen3-Reranker model initialization |

Removed or deprecated architectures also continue to receive vLLM's dedicated error messages rather than being replaced by a less specific RBLN-side registry error.

## Why `score()` and `generate()` Can Produce Different Values

The two APIs implement the same relevance formula, but upstream vLLM evaluates the label projection at different precisions when the model uses BF16.

### Upstream `generate()` path

The generative path applies the full vocabulary `lm_head` using the model dtype:

```text
BF16 hidden state
    ↓
BF16 full-vocabulary lm_head
    ↓
BF16 logits for 151,669 tokens
    ↓
FP32 log_softmax
    ↓
yes/no probability
```

The `yes` and `no` logits have already been rounded to BF16 before `compute_logprobs()` promotes them to FP32.

Converting these logits to FP32 afterward cannot recover the information lost during the BF16 projection.

### Upstream `score()` path

For pooling models, upstream vLLM promotes the classifier head to `head_dtype`, which is FP32.

The effective classifier weight is constructed from the two relevant `lm_head` rows:

```text
score.weight = lm_head[yes] - lm_head[no]
```

The score path then evaluates one FP32 classifier projection:

```text
BF16 hidden state
    ↓
FP32 classifier projection
    ↓
FP32 score logit
    ↓
sigmoid
```

This path still contains the BF16 error from the hidden states, but it avoids independently rounding the `yes` and `no` vocabulary logits.

### Precision comparison

| Path | Label projection | BF16 rounding point |
| --- | --- | --- |
| Upstream `generate()` | Full-vocabulary BF16 matmul | Before the `yes`/`no` logits are subtracted |
| Upstream `score()` | Single FP32 classifier matvec | No additional rounding after the hidden states |
| Optimum-RBLN `score()` | Generation graph's vocabulary logits | Same projection behavior as the generation graph |

At a logit magnitude of approximately 18, one BF16 step is roughly:

```text
18 × 2⁻⁸ ≈ 0.07
```

Rounding the `yes` and `no` logits independently before subtracting them can therefore introduce a visible error in:

```text
logit_yes - logit_no
```

This explains why upstream `score()` can be numerically closer to an FP32 reference than upstream `generate()` when serving the model in BF16.

When the full model runs in FP32, this asymmetry disappears and the two APIs produce matching results.

## Optimum-RBLN Precision Trade-off

Optimum-RBLN intentionally reuses the compiled generation graph for `score()`.

As a result, the implementation reads the `yes` and `no` logits after the graph's `lm_head` projection and inherits the same low-precision characteristics as the generative path.

The reduced score logit is converted to `head_dtype` before sigmoid:

```python
score_logits = logits[:, true_id] - logits[:, false_id]
return score_logits.unsqueeze(-1).to(model_config.head_dtype)
```

This prevents the sigmoid result itself from being represented in BF16—for example, a score near `0.9995` being rounded to `1.0`.

However, this conversion cannot recover precision already lost while producing the vocabulary logits.

### Selected trade-off

This PR prioritizes:

- sharing one compiled artifact between `generate()` and `score()`;
- reusing the existing Qwen3 generation graph;
- avoiding a separately compiled sequence-classification graph;
- avoiding runtime dependency on classifier weights from the original checkpoint.

The trade-off is that Optimum-RBLN's `score()` path does not receive upstream vLLM's FP32 classifier-projection accuracy advantage in BF16.

Matching that behavior would require a different design: extracting the two `lm_head` rows, constructing a dedicated FP32 classifier weight, and applying it directly to the final hidden state.

That alternative is intentionally outside the scope of this PR.

## Usage

```python
from transformers import AutoTokenizer
from vllm import LLM

model = "Qwen/Qwen3-Reranker-0.6B"

llm = LLM(
    model=model,
    runner="pooling",
    block_size=4096,
    max_model_len=32768,
    hf_overrides={
        "architectures": ["Qwen3ForSequenceClassification"],
        "classifier_from_token": ["no", "yes"],
        "is_original_qwen3_reranker": True,
    },
)

chat_template = AutoTokenizer.from_pretrained(model).chat_template

outputs = llm.score(
    queries,
    documents,
    chat_template=chat_template,
)

scores = [output.outputs.score for output in outputs]
```

A runnable example is included:

```bash
python examples/optimum/pooling_models/qwen3_reranker_score_offline.py
```

## Tests

CPU-only tests cover:

- Qwen3-Reranker detection before and after architecture remapping;
- separation between Qwen3-Embedding and Qwen3-Reranker;
- mapping to the Qwen3 generation implementation;
- generation-architecture and chunked-prefill classification;
- `classifier_from_token` validation;
- label token ID resolution;
- left-padding conversion;
- over-length input rejection;
- prefix-caching behavior;
- classifier pooler construction;
- equivalence between:
  - `sigmoid(logit_yes - logit_no)`;
  - the two-way `yes`/`no` softmax probability.

## Scope

This PR supports original Qwen3-Reranker checkpoints identified by:

```text
is_original_qwen3_reranker=True
classifier_from_token=["no", "yes"]
```

Offline-converted sequence-classification checkpoints containing a standalone `score.weight` are not supported by this implementation.

The generation example's chat-template and log-probability fixes were split into separate changes and are not part of this PR.

## Related Pull Requests

- #854 — fixes the generation-based Qwen3-Reranker example
- #879 — adds `logprob_token_ids` handling to the RBLN sampler (replaces #853, which was branched off the wrong base)

## Type of Change

- [ ] Release (`release`)
- [x] Feature (`feature`)
- [x] Model support (`model`)
- [ ] Core engine changes (`core`)
- [ ] Bug fix (`fix`)
- [ ] Performance improvement (`perf`)
- [x] Refactor or code cleanup (`refactor`)
- [ ] Documentation (`docs`)
- [ ] Other (`other`)

## Checklist

- [x] PR title follows the repository's Conventional Commits format.
- [x] The implementation includes a runnable example.
- [x] CPU-only unit tests cover the model routing and score calculation.
- [x] The precision difference between the upstream APIs is documented.
- [x] The Optimum-RBLN implementation trade-off is documented.
- [x] RBLN device validation has been completed.


rebel-jonghewk (approved):
Can you check that an error is raised when the arch is Qwen3ForSequenceClassification but the reranker flags are not set?

rebel-eunji:
@rebel-jonghewk 
I added robust guard and pytest to check missing flag of hf_overrides.

PR #853, opened 2026-07-30
fix(sampler): honour SamplingParams.logprob_token_ids

rebel-eunji:
## Summary

Fix `RBLNSampler` to honor `SamplingParams.logprob_token_ids`.

Previously, requests that specified only `logprob_token_ids` returned an empty `logprobs` list. This PR aligns the RBLN sampler behavior with upstream vLLM.

## Problem

`logprob_token_ids` requests log probabilities for specific tokens regardless of their rank. This is useful for workloads such as reranking, where callers need scores for a fixed set of label tokens such as `"yes"` and `"no"`.

However, `RBLNSampler` computed log probabilities only when `max_num_logprobs` was set. When a request specified only `logprob_token_ids`:

1. `max_num_logprobs` remained `None`.
2. The sampler skipped log-probability computation.
3. `logprobs_tensors` was set to `None`.
4. The response contained an empty `logprobs` list.

This could result in errors such as:

```text
IndexError: list index out of range
```

## Root Cause

The request plumbing was already present:

- `RBLNInputBatch` populated `logprob_token_ids`.
- The value was passed through `SamplingMetadata`.

The missing behavior was in `RBLNSampler`, which did not read or process `logprob_token_ids`.

## Changes

This PR mirrors the relevant behavior of upstream `Sampler.forward`:

- Compute raw log probabilities when either:
  - `max_num_logprobs` is set, or
  - `logprob_token_ids` is provided.
- Gather log probabilities for the explicitly requested token IDs.
- Prefer `logprob_token_ids` when both options are provided.
- Keep the dictionary-based gather operation outside TorchDynamo using `@torch.compiler.disable` to avoid recompilation based on per-request dictionary contents.

## Expected Behavior

Given:

```python
sampling_params = SamplingParams(
    temperature=0,
    max_tokens=1,
    logprob_token_ids=[true_token, false_token],
    allowed_token_ids=[true_token, false_token],
)
```

Both requested token IDs should be included in the output log probabilities, even if one of them falls outside any practical top-k window.

### Before

```text
CompletionOutput(
    index=0,
    text='yes',
    token_ids=[9693],
    cumulative_logprob=0.0,
    logprobs=[],
    finish_reason=length,
    ...
)
```

### After

`output.outputs[0].logprobs` is populated with the requested token probabilities.

## Validation

- [x] `pre-commit run --all-files`
- [x] Confirmed the implementation matches upstream vLLM sampler semantics
- [x] Verified the equivalent `SamplingParams` behavior on vLLM's CPU platform
- [ ] Validate the fixed path on an RBLN device

On the upstream CPU implementation, the returned scores matched a Transformers reference to seven decimal places.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation


vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -296,8 +315,18 @@ def forward(
         # return int32 (while PyTorch argmax and topk return int64).
         sampled = sampled.long()
 
+        # Logprobs for an explicit set of token ids, used by the scoring APIs.
+        # Unlike the top-k window below, this finds the tokens whatever their
+        # rank, so a label token the model finds very unlikely is still returned.
```
rebel-eunji:
```suggestion
```

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -309,6 +335,10 @@ def forward(
                 raw_logprobs, num_logprobs, token_ids=sampled
             )
 
+        # Asking for both is not ambiguous: the explicit ids win, as upstream.
```
rebel-eunji:
```suggestion
        # If we have both num_logprobs and logprob_token_ids, prefer
        # logprob_token_ids as it's more specific
```

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -270,7 +286,10 @@ def forward(
         # This is different from the V0 sampler, which uses the logits that
         # is used for sampling (after penalties and temperature scaling).
         num_logprobs = sampling_metadata.max_num_logprobs
-        if num_logprobs is not None:
+        raw_logprobs: torch.Tensor | None = None
+        # `logprob_token_ids` asks for specific tokens regardless of rank, so it
+        # needs the raw logprobs even when no top-k window was requested.
```
rebel-eunji:
```suggestion
```

PR #854, opened 2026-07-30
fix(examples): match the Qwen3-Reranker chat template contract

### LINKED ISSUE #841: examples/optimum/run_qwen3_reranker.py builds prompts that contain neither the query nor the document
## What happens

Running the example unmodified gives the same score for both pairs, even though the pairs are
different and both documents are relevant:

```
$ python examples/optimum/run_qwen3_reranker.py \
    --model_id ./qwen3-reranker-0.6b-b1 --max_seq_len 8192 --num_input_prompt 2

scores: [0.0001381069574226601, 0.0001381069574226601]
```

fp32 PyTorch with the prompt from the model card gives 0.9995 and 0.9972 for the same pairs.

## Why

`chat_template.jinja` in `Qwen/Qwen3-Reranker-*` takes the query and the document from `query`
and `document` roles:

```jinja
{%- set query_text = messages | selectattr("role","eq","query") | map(attribute="content") | first -%}
{%- set document_text = messages | selectattr("role","eq","document") | map(attribute="content") | first -%}
```

`format_instruction()` passes `system`/`user` instead, so both come out empty and every pair
renders to the same prompt (`len({tuple(p.prompt_token_ids) for p in prompts}) == 1`). The
template also emits the assistant/`<think>` block itself, so appending `SUFFIX` duplicates it.

## Fix

```python
def format_instruction(instruction, query, doc):
    return [
        {"role": "system",   "content": instruction},
        {"role": "query",    "content": query},
        {"role": "document", "content": doc},
    ]


def process_inputs(pairs, instruction, max_length, suffix_tokens, tokenizer):
    messages = [format_instruction(instruction, query, doc) for query, doc in pairs]
    token_ids = tokenizer.apply_chat_template(
        messages, tokenize=True, add_generation_prompt=False, return_dict=False
    )
    return [TokensPrompt(prompt_token_ids=ids[:max_length]) for ids in token_ids]
```

`SUFFIX` should not be appended. This reproduces the model-card prompt exactly.

rebel-eunji:
## Summary

Fix the Optimum Qwen3-Reranker generation example to follow the chat template shipped with `Qwen/Qwen3-Reranker-*`.

The example now:

- passes `system`, `query`, and `document` messages using the roles expected by the template;
- uses the prompt rendered by the template without manually appending or truncating it;
- requests the `yes` and `no` log probabilities explicitly;
- includes an irrelevant query-document pair to verify score separation.

This example-only fix was split from #846 so it can be reviewed independently from Qwen3-Reranker `score()` API support.

Resolves #841.

### Incorrect message roles

The Qwen3-Reranker chat template reads its inputs from the following roles:

- `system`
- `query`
- `document`

The existing example instead passed a `system` message and a combined `user` message.

As a result, the template could not find the query or document and rendered both fields as empty:

```text
<Query>:
<Document>:
```

Different query-document pairs could therefore produce identical prompts and meaningless scores.

### Duplicated prompt suffix

The chat template already emits the trailing assistant/think block required to place the model at the answer position.

The example appended the same suffix manually, resulting in a duplicated assistant/think block.

It also manually truncated token IDs before appending the suffix. For sufficiently long inputs, this could remove part of the prompt generated by the template.

### Incomplete label log probabilities

The example used `logprobs=20` to retrieve the `yes` and `no` token probabilities.

However, the log probabilities are selected from the full vocabulary before `allowed_token_ids` is applied. For a confidently irrelevant document, the `yes` token can fall outside the top-20 window.

In that case, the example used its fallback value instead of the actual `yes` probability, producing an incorrectly inflated relevance score.

### Match the chat template contract

Build messages with the roles expected by Qwen3-Reranker:

```python
[
    {"role": "system", "content": instruction},
    {"role": "query", "content": query},
    {"role": "document", "content": document},
]
```

The rendered chat template output is passed directly to vLLM as text.

This removes the need for:

- the manually maintained `SUFFIX`;
- manual tokenization and truncation;
- `TokensPrompt`;
- tokenizer padding configuration;
- `return_dict=False`;
- `enable_thinking=False`.

### Request label probabilities explicitly

Replace:

```python
logprobs=20
```

with:

```python
logprob_token_ids=[true_token, false_token]
```

This requests the `yes` and `no` probabilities directly, regardless of their ranks in the full vocabulary.

### Add an irrelevant example pair

A deliberately irrelevant query-document pair is now included by default. This makes the low-score case visible whenever the example is run.

### Clarify the example filename

Rename:

```text
qwen3_reranker_offline.py
```

to:

```text
qwen3_reranker_generate_offline.py
```

The new name clarifies that this example calculates reranker scores through generation. It also avoids confusion with upstream vLLM's similarly named `score()` example.

The Optimum CI workflow has been updated to use the new filename.

## Validation

Run:

```bash
python examples/optimum/pooling_models/qwen3_reranker_generate_offline.py
```

Expected scores from the FP32 reference:

```text
scores: [0.99949825, 0.99936193, 8.978e-06]
```

The first two query-document pairs are relevant and should receive high scores.

The third pair is deliberately irrelevant and should receive a score near:

```text
8.978e-06
```

Before requesting the label token probabilities explicitly, this case could incorrectly return approximately `1.45e-02`.

Result: passed.

## Dependency

This PR depends on #879.

`SamplingParams.logprob_token_ids` is accepted by the current API, but `RBLNSampler` does not yet return the requested token probabilities. PR #879 adds that sampler support.

Merging this PR before #879 would cause the updated example and its CI step to fail when reading the missing log probabilities.

## Scope

This PR only fixes the existing generation-based Qwen3-Reranker example.

Qwen3-Reranker support through the native `LLM.score()` API is intentionally excluded and will be handled separately.

## Related Issues and Pull Requests

- Resolves #841
- Depends on #879
- Split from #846

## Type of Change

- [ ] Release (`release`)
- [ ] Feature (`feature`)
- [ ] Model support (`model`)
- [ ] Core engine changes (`core`)
- [x] Bug fix (`fix`)
- [ ] Performance improvement (`perf`)
- [x] Refactor or code cleanup (`refactor`)
- [ ] Documentation (`docs`)
- [ ] Other (`other`)

## Checklist

- [x] PR title follows the repository's Conventional Commits format.
- [x] The PR is linked to an existing issue.
- [x] The test method and expected results are documented.
- [x] The CI workflow uses the renamed example.
- [x] `pre-commit run --all-files` passes.


examples/optimum/pooling_models/qwen3_reranker_generate_offline.py
rebel-jonghewk:
```suggestion
true_score = math.exp(logprobs[true_token].logprob)
false_score = math.exp(logprobs[false_token].logprob)
```

is the else still necessary?

.github/workflows/rbln_optimum_ci.yaml
```diff
@@ -132,7 +132,7 @@ jobs:
 
       - name : Run Qwen3-reranker (V1)
         run: >
-          uv run --no-sync python3 examples/optimum/pooling_models/qwen3_reranker_offline.py
+          uv run --no-sync python3 examples/optimum/pooling_models/qwen3_reranker_generate_offline.py
           --max_seq_len 32768
           --num_input_prompt $((2 * ${{ env.NUM_INPUT_PROMPT }}))
```
rebel-jonghewk: do we have to change the default NUM_INPUT_PROMPT to test this case?

rebel-eunji: Thanks. From now on, num_input_prompt refers to the number of pairs.

examples/optimum/pooling_models/qwen3_reranker_generate_offline.py
rebel-eunji: thanks. I fixed it.

rebel-jonghewk (approved):
lgtm

PR #856, opened 2026-07-30
feature(pp): pipeline parallelism for the refactored RBLN scheduler/runner

rebel-ykchoi:
> **Step 1 of 3** — split of #838 (*pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer*). This PR carries **section A** of #838: the PP-aware scheduler & runner. The NIXL transfer PR is stacked on this branch; the data-parallel decode-shape PR is parallel, also based on this branch.

#### Summary

Pipeline parallelism (`pipeline_parallel_size > 1`) on the RBLN path worked for plain decode but broke down the moment decode batching or speculative decoding entered the picture: a step's decode batch could overflow the compiled per-stage bucket, and spec-decode misclassified the prefill/decode phase inconsistently across ranks. This PR reorganizes the scheduler/runner so PP is a first-class, correct, and efficient mode — one PP-aware decode-admission budget, rank-consistent speculative decoding on top of it, prefill/decode phase authority consolidated onto the scheduler, and batch balancing on by default for PP.

#### What changed

#### Testing

- **Unit** — PP decode-admission budget and demand-spread contrast (`test_scheduler.py`), anchor-reconciled deferral predicate and multi-accept payload (`test_rbln_model_runner.py`), backfill/slide query-window math (`test_query_backfill.py`).
- `pre-commit run --hook-stage manual` clean (ruff, mypy 3.10–3.13, typos, license); runner/scheduler unit suites green.

#### Risk / compatibility

- The `VLLM_RBLN_PP_BALANCE_DECODE_BATCH` default flip is confined to PP; the escape hatch (`=0`) preserves the old behavior.

#### Affected modules

Platform, Scheduler, Runner.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

### DIFF
--- a/tests/torch_compile/unit/v1/core/test_decode_batch_size.py
+++ b/tests/torch_compile/unit/v1/core/test_decode_batch_size.py
@@ -1,0 +1,42 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for ``decode_batch_size`` -- the single source of truth for the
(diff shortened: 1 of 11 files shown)


rebel-cjlee:
@rebel-ykchoi We are planning a version bump up to `0.24.0` early next week, and I understand there will be quite a few changes related to kv connector and disaggreated serving.
It would be great if you could check those changes and update this PR accordingly after the version bump.

rebel-ykchoi:
Superseded by #900 — the same change re-ported onto the vLLM 0.24 base. Closing in favor of the new PR.

PR #857, opened 2026-07-30
feature(nixl): pipeline parallelism for prefill/decode KV-cache transfer

rebel-ykchoi:
> **Step 2 of 3** — split of #838 (*pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer*). This PR carries **section B** of #838: NIXL KV-cache transfer under PP. **Stacked on #856** (PP-aware scheduler & runner) — review/merge that first; the diff here is only the NIXL changes.

#### Summary

Extend prefill→decode KV-cache transfer over NIXL to work when either side runs with pipeline parallelism. Before this, a reader would connect to producer stages it does not read from and index per-stage descriptors out of range. Each pipeline stage now advertises its own band of layers in the handshake, and a reader connects to — and builds descriptors for — only the producer stages whose layers fall within the band it owns.

#### What changed

**1. Per-stage layer-band handshake.** Each pipeline stage advertises its own band of layers in the handshake metadata; a reader connects to and pulls from only the producer stages whose layers fall within the band it owns. This covers both the **symmetric layout** (prefill and decode share a pipeline size, stage matched to stage) and the **fan-in layout** (a reader without pipeline parallelism aggregates several producer stages).

**2. Overlap-aware remote descriptor build.** Per-stage remote descriptors are built assuming a producer exposes no more regions than the reader owns — which holds only for the stages the reader actually reads from. Stages outside the reader's band (which can expose more regions when the layer count does not divide evenly across stages) are **skipped before that build**, so it never indexes out of range.

**3. Non-last-rank KV output via the two-phase path.** The KV-connector output for a non-last PP rank is surfaced through the runner's two-phase `sample_tokens()` path; `RBLNWorker.execute_model` returns `None` on non-last ranks instead of re-emitting the same finished send/recv notifications from an empty output, so every rank saves its own layers' KV exactly once.

**4. Unsupported combinations rejected at handshake.** TP together with PP, sliding-window attention together with PP, and unequal prefill/decode block sizes are rejected during the handshake. Transfers without pipeline parallelism keep the existing single-stage path unchanged.

#### Testing

- **Unit** — PP disaggregation (`test_pd_disaggregation.py`), the NIXL per-stage layer-band handshake and overlap-skip (`test_rbln_nixl_handshake_pp.py`), handshake metadata PP fields (`test_rbln_nixl_metadata.py`), worker-side handshake (`test_rbln_worker_handshake.py`), and the non-last-rank `execute_model` return contract (`test_rbln_worker.py::TestExecuteModel`).
- `pre-commit run --hook-stage manual` clean; connector/worker unit suites green.

#### Risk / compatibility

- Unsupported PP combinations fail fast at the handshake rather than mis-transferring.

#### Affected modules

NIXL KV connector, Worker.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

### DIFF
--- a/tests/torch_compile/unit/v1/core/test_pd_disaggregation.py
+++ b/tests/torch_compile/unit/v1/core/test_pd_disaggregation.py
@@ -669,6 +669,9 @@ def _build_connector_worker(
 
     vllm_config = MagicMock()
     vllm_config.cache_config = CacheConfig(block_size=block_size)
+    # _check_pp_constraints compares pipeline_parallel_size <= 1; give it a real
+    # int (MagicMock would raise TypeError). 1 == the non-PP default these tests assume.
+    vllm_config.parallel_config.pipeline_parallel_size = 1
     kv_cache_config = MagicMock()
     kv_cache_config.num_blocks = num_blocks
     if kv_cache_specs is not None:
@@ -685,6 +688,9 @@ def _build_connector_worker(
         self.kv_cache_config = kv_cache_config_
         self.kv_buffer_device = kv_buffer_device
         self._block_size = {}
+        # The real NixlConnectorWorker.__init__ sets this to None; register_kv_caches
+        # reads it after super().register_kv_caches(). Stub it so the harness matches.
+        self.xfer_handshake_metadata = None
 
     modules_patch = (
         # Mask `nixl_rbln` from sys.modules so `import nixl_rbln` raises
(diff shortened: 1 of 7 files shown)


rebel-ykchoi:
Superseded by #901 — the same change re-ported onto the vLLM 0.24 base. Closing in favor of the new PR.

PR #858, opened 2026-07-30
fix(dp): decode memory sizing, batch padding, and idle-step alignment under data parallelism

rebel-ykchoi:
> **Step 3 of 3** — split of #838 (*pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer*). This PR carries **section C** of #838 (spec-decode memory sizing & DP batch padding) **plus a new fix not in #838**: aligning the idle data-parallel step to the busy ranks' decode shape. **Based on #856** (PP-aware scheduler & runner) and independent of the NIXL PR, so it can be reviewed in parallel — the diff here is only the data-parallel decode-shape changes.

#### Summary

Two data-parallel decode-shape corrections, both about the shape a rank runs when the ranks in the group do not all do the same thing in a step. The first (already described in #838) fixes speculative-decode memory sizing and pads the batch so ranks stay on an already-compiled graph. The second (**new, added here**) fixes the *idle* rank: when one rank has no work it still runs a dummy step to stay in lockstep, and it must run the same shape the busy ranks settled on rather than an oversized (and sometimes never-compiled) one.

#### What changed

**1. Speculative-decode memory sizing & DP batch padding.** Both stem from a speculative step carrying `num_speculative_tokens + 1` query positions instead of 1:
- Available-memory estimation reserves a fixed command-stream buffer per compiled decode graph. Warm-up compiles one decode graph per `(batch bucket, query length)`, and speculative decoding adds a second query length for every bucket. Those extra graphs are now counted so the reservation matches warm-up and the KV-cache-block estimate does not over-reserve device memory.
- When data-parallel ranks disagree on whether the current step is a speculative (multi-token) decode step, the batch is padded up to the largest decode bucket (mirroring the existing prefill-step policy), keeping every rank on an already-compiled graph. No-op for single-bucket configs.

**2. Idle data-parallel step aligned to the busy ranks' decode shape (new — not in #838).** When one data-parallel rank has no work it still runs a dummy step so the group stays in step. That dummy previously guessed its own shape, which could pull the whole group onto a larger, mismatched graph. It now copies the shape the busy ranks settled on and runs the same graph they run; when every rank is idle it uses the smallest decode graph.
- The shape decision hands back a request count and a token count. Normally the token count is request-count × query-length, so the idle rank divides to recover its query length. But in two fall-back cases the token count is a padding target rather than that product — when a peer is still reading a prompt, and when the busy ranks disagree on query length. There the division overstated the query length, and in the prompt case asked for a shape that had never been prepared, forcing it to be built the first time it was used.
- The shape decision now returns the intended per-request query length directly. On those two cases it is 1, so the idle rank runs the smallest prepared graph and carries the padding target separately — the same graph a busy rank doing a single-token decode already uses. The other cases keep deriving it as before.
- The draft-model path is not yet idle-aware — follow-up.

#### Testing

- **Unit** — the cross-rank token/request/idle pack round-trip (`test_forward_context.py`); the shape-decision exclusion, all-idle, and fall-back paths incl. the new idle-step cases (`test_rbln_model_runner.py`); dummy-step entry (`test_rbln_worker.py`).
- `pre-commit run --hook-stage manual` clean; runner/forward-context unit suites green.

#### Risk / compatibility

- Both changes only affect data-parallel, multi-bucket setups; single-bucket / single-DP paths are unchanged.
- The memory-sizing correction only ever reserves *fewer* blocks (removes an over-reservation).

#### Affected modules

Runner, forward context, Worker, spec-decode drafter.

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

### DIFF
--- a/tests/torch_compile/unit/test_forward_context.py
+++ b/tests/torch_compile/unit/test_forward_context.py
@@ -46,3 +46,62 @@ def test_forward_context(vllm_config, attn_metadata_mock: MagicMock):
             f"Expected 'dp_metadata' class name is RBLNDPMetadata, \
                     got {get_forward_context().dp_metadata.__class__.__name__}"
         )
+
+
+def test_num_tokens_and_reqs_across_dp_bit_pack(monkeypatch):
+    """Round-trip the bit-packed (num_tokens, num_reqs, is_prefill, is_idle)
+    all-reduce. Intercepts the inner all-gather so no real DP group is needed,
+    and checks each field unpacks correctly -- including the new is_idle mask
+    and the (prefill -> num_reqs None) behavior."""
+    from vllm_rbln.forward_context import RBLNDPMetadata
+
+    token_bits = 16
+
+    def peer_encoded(num_tokens, num_reqs):
+        return num_tokens | (num_reqs << token_bits)
+
+    def make_fake_inner(peer_value):
+        def fake_inner(encoded, dp_size, dp_rank):
+            arr = [0, 0]
+            arr[dp_rank] = encoded
+            arr[1 - dp_rank] = peer_value
+            return torch.tensor(arr, dtype=torch.int32)
+
+        return fake_inner
+
+    # Decode: this rank (rank0) = idle (num_tokens=1, num_reqs=1, is_idle);
+    # peer (rank1) = busy (num_tokens=8, num_reqs=2).
+    monkeypatch.setattr(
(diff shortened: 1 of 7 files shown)


rebel-ykchoi:
Superseded by #902 — the same change re-ported onto the vLLM 0.24 base. Closing in favor of the new PR.

PR #778, opened 2026-07-15
feature: add persistent swa_attn_masks buffer for vaddr reuse

rebel-daeyang:
## Background
`swa_attn_masks` was rebuilt as a new device tensor on every decode step, allocating a fresh device buffer (vaddr) each time — an unnecessary per-step overhead.

## Change
Route `swa_attn_masks` through a persistent reused buffer (`_swa_attn_masks_buf`, via `_to_device_inplace`) — the same pattern already used for `cache_seq_lens` / `cache_offsets`.

## Effect
Removes the per-step fresh device allocation for `swa_attn_masks`. Applies only to decode with batch size > 1 (`swa_attn_masks` is built only when `b_size > 1`).

## Reviewer
* **Primary**: (지정 필요)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #860, opened 2026-07-30
perf(attention): reuse device buffers instead of allocating per step

rebel-cjlee:
#### Remove per-step device allocation from attention metadata construction

`RBLNFlashAttentionMetadataBuilder.build()` allocated a fresh device tensor for every metadata field via `.to(self.device)`. Since `build()` runs on every step, that is one device allocation per tensor per step, and the RBLN allocator does not reuse them; the cost lands directly in prepare time. Models with SWA pay it for more tensors across more KV cache groups, so they are hit hardest.

This takes the same approach as #548, #778 (`_to_device_inplace`): copy into persistent device buffers instead of reallocating. The difference is how the buffers are looked up. #548 keeps one buffer per attribute, so it can only hold a single shape at a time; a shape mismatch reallocates and overwrites that slot. Prefill and decode use different layouts, so every prefill/decode transition reallocates the tensors that appear in both phases, which is exactly what the buffer was meant to avoid. Here the buffers live in a dict keyed on `(slot, shape, dtype)`, so both layouts stay resident and allocations drop to zero after warm-up.

#### Unify attention metadata on host tensors

`CommonAttentionMetadata` carried 4 tensors for two values (`query_start_loc`/`query_start_loc_cpu`, `seq_lens`/`_seq_lens_cpu`), and call sites disagreed about which to read. The metadata builder read the `_cpu` twins while `RBLNEagleProposer` mixed both, and `query_start_loc` was staged to the device through a `CpuGpuBuffer` on every step even though nothing ever read the device copy.

All metadata arithmetic now stays on the host, with a single tensor per value, and the H2D copy happens once at the point where the backend metadata is constructed. This removes a per-step H2D copy from the runner and a per-`propose` D2H copy from the EAGLE drafter, and it makes the device-copy boundary a single place that is easy to keep correct. `_seq_lens_cpu` and `_num_computed_tokens_cpu` are no longer passed at all (upstream has deprecated both for removal).

#### Result

On gpt-oss-120b, mean prepare time drops by roughly **90% for decode** and **25% for prefill**. Device time is unchanged, so the gain is entirely host-side. The improvement is also present on models without SWA, but smaller, since fewer tensors are staged per step.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Run the offline benchmark with metrics and the runtime timer enabled, and compare the reported mean prepare time against the base branch. Worth checking both a sliding-window model and a non-sliding-window one, and both a fresh compile and a cached-model run.

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/torch_compile/unit/v1/spec_decode/test_eagle.py
+++ b/tests/torch_compile/unit/v1/spec_decode/test_eagle.py
@@ -144,8 +144,6 @@ def make_common_attn_metadata(
         query_start_loc=query_start_loc,
         query_start_loc_cpu=query_start_loc.cpu(),
         seq_lens=seq_lens,
-        _seq_lens_cpu=seq_lens.cpu(),
-        _num_computed_tokens_cpu=torch.zeros_like(seq_lens),
         num_reqs=seq_lens.shape[0],
         num_actual_tokens=total_num_tokens,
         max_query_len=int((query_start_loc[1:] - query_start_loc[:-1]).max().item()),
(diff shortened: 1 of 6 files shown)


rebel-daeyang (approved):
lgtm

PR #870, opened 2026-07-31
feature: compute KV cache available memory per chiplet

rebel-jongholee:
### 🚀 Summary of Changes

`rebel_compiler` [#10678](https://github.com/rebellions-sw/rebel_compiler/pull/10678) added the dynamic `num_blocks` API. This PR uses it.

Today the KV cache is sized from a single whole-card memory estimate. That estimate has no notion of chiplets, so on a 4-chiplet card it can exceed the per-chiplet budget and the engine allocates a cache that cannot fit — the failure in [vllm-rbln-internal#42](https://github.com/rebellions-sw/vllm-rbln-internal/issues/42).

Instead of estimating, this marks the KV `num_blocks` dimension dynamic at compile time, asks each compiled artifact's `kv_cache_memory_profile()` how many blocks actually fit **per chiplet**, reallocates the KV tensors at that size, and tells the scheduler the new number.

Opt-in and inert unless enabled: with `VLLM_RBLN_USE_DYNAMIC_KV_CACHE` off (the default) `estimate_available_memory` returns byte-identical values to `dev` across the full `num_key_value_heads` range, and the `EngineCore` patch is not installed.

**Affected modules**: Worker (`v1/worker/rbln_worker.py`, new `v1/worker/kv_profile.py`, `v1/worker/utils.py`), Runner (`v1/worker/rbln_model_runner.py`), Platform (`envs.py`, `platform.py`), new `patches/dynamic_kv.py`, new `docs/dynamic_kv_cache.md`.

The per-chiplet budget needs the card's real DRAM capacity, so this PR adds the first sysfs readers in the tree. `RBLN_DEVICES` indexes what the container exposes, not the sysfs card numbering, so both readers resolve against `/dev/rbln*`. Both are only reached with the flag on.

---

* Related to https://github.com/rebellions-sw/vllm-rbln-internal/issues/42
* Uses https://github.com/rebellions-sw/rebel_compiler/pull/10678

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

The configuration below is the one that needs **no compiler change**, so it can be verified today:

```bash
export VLLM_RBLN_USE_VLLM_MODEL=1 VLLM_RBLN_USE_DEVICE_TENSOR=1
export VLLM_RBLN_DISABLE_OFFLOAD=1 VLLM_RBLN_SAMPLER=1
export RBLN_DEVICES=0
export VLLM_CACHE_ROOT=<a fresh dir per configuration>

# before: dies in warm-up
export VLLM_RBLN_USE_DYNAMIC_KV_CACHE=0
# after:  serves
export VLLM_RBLN_USE_DYNAMIC_KV_CACHE=1

vllm serve Qwen/Qwen2.5-3B-Instruct \
  --tensor-parallel-size 1 --gpu-memory-utilization 0.9 \
  --block-size 8192 --max-num-batched-tokens 8192 \
  --max-model-len 32768 --max-num-seqs 16 --trust-remote-code
```

Use a fresh `VLLM_CACHE_ROOT` per configuration: the compile cache hash does not include dynamism, so a static and a dynamic build share one signature.

**Expected** — `before`: `available_memory_estimate = 59.39 GiB`, 211 blocks, fails with `SYS_EBUSY: Lack of device memory`. `after`: 157 blocks, serves, answers correctly. The binding chiplet goes 34.65 GiB (1.100× of the gmu budget) → 31.39 GiB (0.998×).

On a compiler without [#10678](https://github.com/rebellions-sw/rebel_compiler/pull/10678) the run now fails at start-up with an explicit message naming the missing symbol, instead of an ImportError after the whole compile and warm-up.

**Supported layouts** — the dynamic path asserts, rather than assumes, what the compiler's dynamic-input validator requires: every attention layer must dispatch to `paged_flash_causal_attention_naive_{prefill,decode}` (no sliding window), no cross-layer KV sharing, and KV base deduplication off. MLA and speculative decoding are refused outright in the worker's constructor — MLA because `num_blocks` is dim 0 of its KV shape rather than dim 1, spec decode because the merge cannot attribute regions to the drafter versus the target. A model outside that set fails loudly at start-up with the flag on, and is untouched with it off.

tests/torch_compile/unit adds coverage for the new guards, the sysfs card-index resolution and capacity clamp, the replication-factor gate, and the env defaults — including tests that pin the default path to dev's numbers. That directory is not part of the PR pytest job, so it has to be run locally:

```bash
pytest tests/torch_compile/unit
```

Clear VLLM_RBLN_USE_DYNAMIC_KV_CACHE from the shell first — conftest.py calls load_general_plugins() during pytest_configure, so a flag left set from the steps above is read before any fixture can clear it and three tests that assert the defaults fail.

---

### 📋 Checklist

---

### 💬 Notes

**Compiler version.** This PR changes no lock file — `uv.lock` is identical to `dev`, which pins `rebel-compiler 0.11.2.dev158+gbaabdcac.prod`. The Qwen figures above were re-measured on exactly that build: same 157 blocks, same 59.39 GiB estimate with the flag off, same `SYS_EBUSY` failure without it. Earlier measurements on `0.11.1.dev531+ge4aaf3a1` and `0.11.1.dev582+g8a786155.prod` gave the same block count, so neither bump moves the answer.

**Merge order.** The MiniMax 4-card configurations in #42 (TP4+EP, DP4+EP) additionally need `rebel_compiler` `4dff83c9` — without it compilation dies with `RecursionError` in `_validate_paged_dynamic_inputs`. That commit alone is sufficient (verified: identical block count and residency to a build carrying an extra local commit); filed separately. **MiniMax PP4 and Qwen work with the current compiler**, so this PR is not blocked, but "MiniMax works" can only be claimed after that commit lands.

**Known limitations** (details in the issue comment):

1. **TP≥2 does not return the pre-resize KV cache.** The outgoing cache stays resident on TP≥2; TP=1 returns it, and the free is not observable from inside the process. The per-chiplet budget therefore charges those bytes whenever TP>1, which costs blocks instead of overflowing. MiniMax TP4+EP measured 284 blocks at 1.015× of its budget on the binding chiplet; replaying that run's profile with the charge applied gives 276 blocks at 0.998×. **Not yet re-measured on device.**
2. **Block count is not deterministic.** On Qwen gmu 0.9, 1 run in 3 retained an extra 1490 MiB arena per chiplet and landed at 1.043× of the requested budget.
3. **Two configurations warn and serve the pre-compile estimate**, because both are an explicit request from the caller: compile and warm-up skipped (`--enforce-eager`, `VLLM_RBLN_COMPILE_MODEL=0`, `VLLM_RBLN_ENABLE_WARM_UP=0`), and `--num-gpu-blocks-override`. `mark_dynamic` is still applied and logged in both, so that log line is never evidence that the block count came from the device.


rebel-minhopark:
Hello, new compiler nightly package is uploaded (https://nexus.mgmt.rbln.in/#browse/browse:pypi-compiler-nightly-prod:rebel-compiler%2F0.11.1.dev582%2Bg8a786155.prod)
You can change compiler version using uv lock update!

rebel-minhopark (commented):
One usability concern about `VLLM_RBLN_COMPILE_KV_CACHE_NUM_BLOCKS`: should this be a knob the user has to set at all?

As it stands, enabling the feature takes two env vars, and getting the second one wrong fails in ways that are hard to self-diagnose:

- Setting only `VLLM_RBLN_USE_DYNAMIC_KV_CACHE` silently produces the static
  result — and the `mark_dynamic` log line still appears, which reads as if the
  feature were active. That is listed as known limitation 3, so it is already
  understood as a sharp edge.
- There is no documented value the user is supposed to pick. The code says
  "8 is enough -- warm-up points every dummy request at block 0" and
  "a smaller value costs less", but never that 8 is *required*. So the user is
  being asked to choose a number whose only correct answer, per the PR's own
  reasoning, is "as small as works".
- Choosing it too large is not a config error but a runtime one: on TP>=2 the
  compile-time cache is not returned and gets charged against the per-chiplet
  budget, so a large N quietly costs blocks, or trips
  "the retained compile-time KV cache leaves no budget on ...".

Since the shrink is already ANDed with `VLLM_RBLN_USE_DYNAMIC_KV_CACHE` and with
the absence of `--num-gpu-blocks-override`, and is ignored with a warning when it
is not below vLLM's block count, all the conditions for a safe default are already
computed inside the worker. Would it work to default the value to 8 whenever
`VLLM_RBLN_USE_DYNAMIC_KV_CACHE` is on — i.e. make the env var an override for debugging rather than a required second step? That would reduce the public surface to one flag, remove the misleading half-enabled state entirely, and still leave the knob for anyone who needs to tune it.

If there is a configuration where a small default is wrong and the user genuinely
has to pick the number, I would rather see that case documented here — that is the
part I could not find in the PR or the code comments.

rebel-eunji:
Could you remove excessive claude generated comments? It is hard to read..

rebel-jongholee:
Thanks for the review @rebel-minhopark, @rebel-eunji .

I've kept only `VLLM_RBLN_USE_DYNAMIC_KV_CACHE` and turned the other two into constants (`COMPILE_KV_CACHE_NUM_BLOCKS`, the unprofiled reserve). A configuration where the pre-compile estimate is at or below the hint has nothing to shrink, and instead of quietly carrying on it now refuses to start. That is the "case where a small default is wrong" you asked about, and it shows up with a large `--block-size`. Rather than asking for a number, the message points at `--block-size`, `--gpu-memory-utilization` and the device count, and it's written up in `docs/dynamic_kv_cache.md`. 

The hint is the smallest value that works: one step lower, dynamo specializes the dimension away and leaves a static artifact. Lower is better anyway, narrowing that refusal window and cutting what TP>=2 pays for the compile-time cache. Only where the user has explicitly chosen something, such as `--enforce-eager` or `--num-gpu-blocks-override`, do we warn and fall back to the estimate, and that's deliberate, so it's documented too. So the state where the flag is on but you silently get the static result is gone. The `mark_dynamic` line that misled you was firing once per layer; it's one line per call now.

That said, this PR doesn't turn the feature on by default. Its scope is making the dynamic KV `num_blocks` from rebel_compiler#10678 usable from vllm-rbln. Making it behave seamlessly the way `--gpu-memory-utilization` does on GPU needs more work on the compiler side first, and I'd rather verify that thoroughly and flip the default in its own PR. In the same spirit, the two changes to the default-path estimate that were still outside the flag are now inside it, so with the flag off this branch behaves exactly like dev.

Sorry for the hard read, and thanks for saying so. Since this path isn't on by default yet, I'd left comments in place to keep context for the experiments that follow, but I've cleaned them up. I'll keep PRs to dev tidy from the start going forward.

Thanks for the nightly too. I verified on `0.11.1.dev582+g8a786155.prod` and bumped `uv.lock`, but dev's lock moved to `0.11.2.dev158+gbaabdcac.prod` in the meantime and the two entries collided. The compiler this feature needs is checked by symbol at runtime, so there's no reason to pin it in the lock. I reverted the branch-side change, left `uv.lock` matching dev, and re-verified on `0.11.2.dev158`.

There is still a compiler-side issue: MiniMax TP4+EP / DP4+EP don't compile at all, hitting a `RecursionError` in `_validate_paged_dynamic_inputs`. My plan is to merge this PR, which enables the feature behind an option, and follow up with the compiler fix and the default switch as described above.

docs/dynamic_kv_cache.md
```diff
@@ -0,0 +1,102 @@
...
+
+The following are rejected at start-up when the flag is on, and are unaffected
+when it is off. Run with `VLLM_RBLN_USE_DYNAMIC_KV_CACHE=0` to use them.
+
+| Configuration | Why |
+| --- | --- |
+| MLA models | `num_blocks` is dimension 0 of MLA's KV shape rather than dimension 1, and MLA dispatches `paged_flash_causal_mla_naive_*`, which the compiler's dynamic-input validator does not admit. `VLLM_MLA_DISABLE=1` also works. |
+| Speculative decoding | The drafter's profile is merged next to the target's, and the merge cannot tell an artifact re-reporting a shared KV tensor from one owning its own. |
+| KV transfer connectors | The connector registers the KV cache's physical views during warm-up, and the reallocation invalidates them. |
+| Sliding-window attention, cross-layer KV sharing, KV base deduplication | The compiler requires every dynamic KV input to reach exactly one `paged_flash_causal_attention_naive_{prefill,decode}` call. |
```
rebel-jonghewk, line 60: I'll look into whether we can relax this restriction.

vllm_rbln/patches/dynamic_kv.py
```diff
@@ -0,0 +1,212 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Hand the dynamic-KV block count to the scheduler after warm-up."""
+
+# NOTE(RBLN): `EngineCore._initialize_kv_caches` is the only hook that works for
+# both single-process and TP>1 -- it warms the workers up before returning, and
+# the caller builds the `Scheduler` only afterwards. Both facts are re-checked
+# below, so a vLLM upgrade that reorders them fails loudly instead of mis-sizing.
+
+from __future__ import annotations
```
rebel-eunji: nit: Is it required?

vllm_rbln/patches/dynamic_kv.py
```diff
@@ -0,0 +1,212 @@
...
+# Held at import time: the registry replaces the attribute outright.
+engine_core_original_initialize_kv_caches = EngineCore._initialize_kv_caches
+
+# vLLM release this was written against; the structural checks below are the real
+# guard, so a mismatch only warns.
+_VERIFIED_VLLM_VERSION = "0.24.0"
+
+
+def _dynamic_kv_enabled() -> bool:
+    return bool(envs.VLLM_RBLN_USE_DYNAMIC_KV_CACHE)
```
rebel-eunji: It's exessive.
Just use `condition=lambda: envs.VLLM_RBLN_USE_DYNAMIC_KV_CACHE` directly like other patches.

vllm_rbln/patches/dynamic_kv.py
```diff
@@ -0,0 +1,212 @@
...
+    from vllm.config import VllmConfig
+
+logger = init_logger(__name__)
+
+# Held at import time: the registry replaces the attribute outright.
+engine_core_original_initialize_kv_caches = EngineCore._initialize_kv_caches
+
+# vLLM release this was written against; the structural checks below are the real
+# guard, so a mismatch only warns.
+_VERIFIED_VLLM_VERSION = "0.24.0"
```
rebel-eunji: Is it required? Hard coding the already pinned vllm version looks not good...

docs/dynamic_kv_cache.md
rebel-jongholee: Thanks. It's three separate limits actually
- cross-layer sharing: needs >1 use on a dynamic input
- sliding-window, MLA: needs kernels other than paged_flash_causal_attention_naive_*
- base dedup: might be a different problem. In that mode the per-layer KV tensor may not be a graph input at all, so it may not even reach the validator. I haven't confirmed this.

vllm_rbln/patches/dynamic_kv.py
rebel-jongholee: Agreed, fixed.

vllm_rbln/patches/dynamic_kv.py
```diff
@@ -0,0 +1,188 @@
...
+def patched_initialize_kv_caches(
+    self: EngineCore, vllm_config: "VllmConfig"
+) -> KVCacheConfig:
+    kv_cache_config = engine_core_original_initialize_kv_caches(self, vllm_config)
+
+    if getattr(self, "scheduler", None) is not None:
+        raise RuntimeError(
+            "the scheduler already exists when _initialize_kv_caches returns, so "
+            "its block pool predates the resize; revisit the dynamic-KV patch."
+        )
```
rebel-eunji: Is it required? In upstream vllm, initialize_kv_caches is called only once.

vllm_rbln/patches/dynamic_kv.py
rebel-jongholee: No. It was just a record of the version I verified on, which belongs in the PR description, not in the code. Removed. Thanks.

vllm_rbln/patches/dynamic_kv.py
rebel-jongholee: No. Dropped the TYPE_CHECKING guard too and import VllmConfig directly, like the rest of the tree does.

vllm_rbln/patches/dynamic_kv.py
rebel-jongholee: Right, Removed, along with a few other checks in the same shape.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+                "VLLM_RBLN_USE_DYNAMIC_KV_CACHE does not support speculative "
+                "decoding; the merged profiles cannot be attributed per artifact."
+            )
+
+        if not USE_DEVICE_TENSOR:
+            raise RuntimeError(
+                "VLLM_RBLN_USE_DYNAMIC_KV_CACHE requires "
+                "VLLM_RBLN_USE_DEVICE_TENSOR=1; without it the artifact carries "
+                "no dynamic KV dimension."
+            )
```
rebel-cjlee: These are all known at config time. Moving this into `platform.py` makes it fail before the model loads.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+                "the EngineCore._initialize_kv_caches patch is not installed, so "
+                "nothing would resize the KV cache or re-announce the count."
+            )
+
+    def _assert_dynamic_kv_compiler_support(self) -> None:
+        """Guard: the installed rebel-compiler must carry the #10678 API.
+
+        Both symbols are reached only after the compile and warm-up, so probing
+        them up front saves an older compiler a full compile before it dies.
+        """
```
rebel-cjlee: This is a pure import / `hasattr` check with no runtime state. So, it can be moved in `platform.py`.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+
+    def _assert_dynamic_kv_transfer_absent(self) -> None:
+        """Guard: a KV connector and dynamic KV cannot be combined."""
+        # NOTE(RBLN): `finalize_kv_cache_registrations` hands the connector the
+        # KV cache's physical views during warm-up; the resize frees them.
+        if has_kv_transfer_group():
+            raise RuntimeError(
+                "VLLM_RBLN_USE_DYNAMIC_KV_CACHE cannot be combined with a KV "
+                "transfer connector; the resize invalidates its registrations."
+            )
```
rebel-cjlee: If you check `vllm_config.kv_transfer_config is not None` instead of `has_kv_transfer_group()`, this one can also move up into `platform` with other guards.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+
+    def compute_dynamic_kv_num_blocks(self) -> int | None:
+        """Ask the compiled artifacts how many KV blocks fit this device.
+
+        Runs after warm-up and reallocates nothing; the engine takes the minimum
+        across ranks and hands it back through `apply_dynamic_kv_num_blocks`.
+        None means the path is not in play.
+        """
+        if not envs.VLLM_RBLN_USE_DYNAMIC_KV_CACHE:
+            return None
```
rebel-cjlee: This condition looks unreachable. The only caller is `patched_initialize_kv_cahces` via `collective_rpc`, and that patch is installed only when `conditon=lambda: envs.VLLM_RBLN_USE_DYNAMIC_KV_CALCHE` is true.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+        self, kv_cache_config: KVCacheConfig
+    ) -> KVCacheConfig:
+        """Return a small-KV-cache copy of the config, or it unchanged.
+
+        The cache allocated here is what `warmup_model()` traces, so its
+        `num_blocks` becomes the hint of the `mark_dynamic`'d dim.
+        """
+        compile_num_blocks = COMPILE_KV_CACHE_NUM_BLOCKS
+        if not envs.VLLM_RBLN_USE_DYNAMIC_KV_CACHE:
+            return kv_cache_config
```
rebel-cjlee, line 529: Same with `compute_dynamic_kv_num_blocks`

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+
+    def _maybe_shrink_kv_cache_for_compile(
+        self, kv_cache_config: KVCacheConfig
+    ) -> KVCacheConfig:
+        """Return a small-KV-cache copy of the config, or it unchanged.
+
+        The cache allocated here is what `warmup_model()` traces, so its
+        `num_blocks` becomes the hint of the `mark_dynamic`'d dim.
+        """
+        compile_num_blocks = COMPILE_KV_CACHE_NUM_BLOCKS
```
rebel-cjlee: Moving it to L556 would be better.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+        offenders = [
+            type(runtime).__name__
+            for runtime in runtimes
+            if type(runtime).__name__ != "DynamoRuntime"
+        ]
+        if offenders:
+            raise RuntimeError(
+                "VLLM_RBLN_USE_DYNAMIC_KV_CACHE needs every KV-holding runtime to "
+                f"be a rebel DynamoRuntime; found {sorted(set(offenders))}."
+            )
```
rebel-cjlee: I'm wondering if this check is actually necessary. Are there any cases where a different runtime is used here?

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+        """Guard: nothing may shrink the KV cache unless something will resize it.
+
+        Both rpcs are only ever driven by the `EngineCore._initialize_kv_caches`
+        patch.
+        """
+        if not envs.VLLM_RBLN_USE_VLLM_MODEL:
+            raise RuntimeError(
+                "VLLM_RBLN_USE_DYNAMIC_KV_CACHE requires "
+                "VLLM_RBLN_USE_VLLM_MODEL=1; see docs/dynamic_kv_cache.md."
+            )
```
rebel-cjlee: I think it looks like a dead code.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -407,7 +497,691 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
+
+        from vllm.v1.engine.core import EngineCore
+
+        from vllm_rbln.patches.dynamic_kv import patched_initialize_kv_caches
+
+        if EngineCore._initialize_kv_caches is not patched_initialize_kv_caches:
+            raise RuntimeError(
+                "the EngineCore._initialize_kv_caches patch is not installed, so "
+                "nothing would resize the KV cache or re-announce the count."
+            )
```
rebel-cjlee: This raise also looks like a dead code.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -79,6 +97,58 @@
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import GrammarOutput, SchedulerOutput
 
+# Blocks the KV cache is shrunk to for the compile: a trace hint for the
+# mark_dynamic'd dim, not a capacity. Smaller is strictly better -- it narrows the
+# window where the estimate is not above it and the run is refused, and TP>=2 is
+# charged the retained compile-time cache per block. But 2-block artifacts abort
+# on device during the warm-up prefill (SYS_TASK_ABORTED) once
+# max_num_batched_tokens grows; 4 is the smallest value measured safe.
+COMPILE_KV_CACHE_NUM_BLOCKS = 4
```
rebel-huijong: I'm not sure what these comments are getting at. Can't we just infer this from max_model_len and block_size? If the block count is smaller than that anyway, it's a misconfigured setup.

vllm_rbln/v1/worker/rbln_worker.py
rebel-huijong, line 1: As @rebel-cjlee said, it seems like most of this can be checked using just the config, so please clean it up. Also, for the runtime check part, I think checking whether VLLM_RBLN_USE_VLLM_MODEL is true should be sufficient.

rebel-huijong (changes requested):
Overall, I'd like to see this cleaned up first by separating where the core logic should live (worker / runner) from where simple checks should live (platform / config).

Beyond that, regarding the implementation itself, current implementation differs from vLLM upstream's init flow of determine_available_memory (profile_run) → init KV → warmup as we are doing everything in warmup. I'd like you to consider splitting profile_run out into determine_available_memory similar to upstream.

Also, this is a question coming from my own lack of understanding of the full behavior it looks like nothing additional happens after the final KV block count is determined. Is it correct that the first request still runs fine without any significant performance issues with just this?

docs/dynamic_kv_cache.md
rebel-jonghewk: @rebel-jongholee Do you have a list of model configurations that have been tested?

rebel-jongholee: @rebel-jonghewk Tested so far all with VLLM_RBLN_USE_DYNAMIC_KV_CACHE=1. The two failures are this PR's own guards firing as designed. Common settings: max-model-len 8192, block-size 1024, max-num-batched-tokens 512.

PASS
- `MiniMaxAI/MiniMax-M2.5/2.7`:  DP4+EP at gmu 0.8/0.9/0.95
(TP4+EP passes too, but TP has memory imbalance issue for now)
- `Qwen/Qwen2.5-3B-Instruct`: TP1 at gmu 0.6/0.75/0.9
- `mistralai/Mistral-Nemo`: TP1 (not an SWA model)
FAIL
- `deepseek-ai/DeepSeek-V3.2`: TP4, MLA guard at config creation
- `openai/gpt-oss-120b` TP4, layout guard after model load

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Done.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: The flag guard there was the dead one, removed.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Right, removed. platform guard runs first at platform.py

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: It was duplicated guard. Removed.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Moved to platform.py. but it can be removed if the compiler version is stable.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Moved to platform.py, slightly stricter than has_kv_transfer_group().

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Moved to platform.py

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Removed. Same for the guard in apply_dynamic_kv_num_blocks.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: Removed. It was duplicated guard too.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: I think this number is ugly too. It's the size of a throwaway cache used for the compile, so smaller is better (to cover tight memory setups that only fit in a few blocks).
However, `1` doesn't work. Dynamo specializes a size-1 dim, so mark_dynamic doesn't survive. `2` compiles fine, but we found the device dying with SYS_TASK_ABORTED during warmup for some {block_size, max_num_batched_tokens} combinations. It doesn't show up in the recommended setup (mnbt <= block_size) and looks like a compiler issue rather than something this PR does, so I'll open an issue separately. Still, having seen it, we couldn't keep 2, so we bumped it to 4, the smallest passing value I measured.

The cost is that setups whose estimate is 4 blocks or fewer get refused (it was 2 or fewer before), but I think such tiny KV setups usually don't exist.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jongholee: The checks that only read config moved into platform validation so they fail before the model loads (26ae3df1), and the worker keeps only the checks that read runtime state. I also looked into splitting profile_run out as you suggested, and I agree that's where this should end up. In upstream, profile_run runs before the KV cache  allocation, so it can measure free memory first and allocate the cache once. 

But on the RBLN, that order doesn't work. We have to allocate the KV cache before compiling, but we only know how many blocks fit after compiling. So we compile with a small KV first and then scale up num_blocks.

```
upstream : load → determine (run & measure free mem) → init KV (final N) → warmup
before PR : load → determine (formula estimate = N) → init KV (N) → compile + warmup
after PR  : load → determine (estimate for placeholder) → init KV (4 blocks) → compile + warmup → ask artifact → resize to N → one decode step(no re-compile)
goal      : load → determine (compile small + ask artifact = N) → init KV (final N) → warmup
```

I agree with removing COMPILE_KV_CACHE_NUM_BLOCKS and aligning with the upstream flow. If the final block count were known during determine_available_memory(), the engine could size the cache correctly from the start. That would eliminate the EngineCore patch, the resize step, and the block count re-announcement, allowing upstream's built-in checks and num_gpu_blocks_override to work directly with the real number.

After this PR merges there is more work planned anyway (issues on the compiler side like MLA/SWA support and the small compile cache not being returned on TP>=2, plus making this feature default), and I'll address what I can in follow-up PRs. Just to clarify, the current approach still compiles exactly once, so there is no double compile time and no broken warmup.

rebel-jongholee:
> Overall, I'd like to see this cleaned up first by separating where the core logic should live (worker / runner) from where simple checks should live (platform / config).
> 
> Beyond that, regarding the implementation itself, current implementation differs from vLLM upstream's init flow of determine_available_memory (profile_run) → init KV → warmup as we are doing everything in warmup. I'd like you to consider splitting profile_run out into determine_available_memory similar to upstream.
> 
> Also, this is a question coming from my own lack of understanding of the full behavior it looks like nothing additional happens after the final KV block count is determined. Is it correct that the first request still runs fine without any significant performance issues with just this?

The checks that only read config moved into platform validation so they fail before the model loads (https://github.com/RBLN-SW/vllm-rbln/commit/26ae3df180646808f95376d57e7b24d4155c2352), and the worker keeps only the checks that read runtime state. I also looked into splitting profile_run out as you suggested, and I agree that's where this should end up. In upstream, profile_run runs before the KV cache allocation, so it can measure free memory first and allocate the cache once.

But on the RBLN, that order doesn't work. We have to allocate the KV cache before compiling, but we only know how many blocks fit after compiling. So we compile with a small KV first and then scale up num_blocks.
```
upstream : load → determine (run & measure free mem) → init KV (final N) → warmup
before PR : load → determine (formula estimate = N) → init KV (N) → compile + warmup
after PR  : load → determine (estimate for placeholder) → init KV (4 blocks) → compile + warmup → ask artifact → resize to N → one decode step(no re-compile)
goal      : load → determine (compile small + ask artifact = N) → init KV (final N) → warmup
```
I agree with removing COMPILE_KV_CACHE_NUM_BLOCKS and aligning with the upstream flow. If the final block count were known during determine_available_memory(), the engine could size the cache correctly from the start. That would eliminate the EngineCore patch, the resize step, and the block count re-announcement, allowing upstream's built-in checks and num_gpu_blocks_override to work directly with the real number.

After this PR merges there is more work planned anyway (issues on the compiler side like MLA/SWA support and the small compile cache not being returned on TP>=2, plus making this feature default), and I'll address what I can in follow-up PRs. Just to clarify, the current approach still compiles exactly once, so there is no double compile time and no broken warmup.

rebel-cjlee (approved):

rebel-huijong (approved):
Thanks for moving all the setup checks over to platform.py so cleanly. For what's left around the overall flow, let's think it through together and work toward a cleaner solution in the next PR.

PR #877, opened 2026-08-02
fix(platform): fail fast when device-tensor envs change after import

rebel-eunji:
### 🚀 Summary of Changes

device_name/device_type/dist_backend are frozen at class definition from VLLM_RBLN_USE_VLLM_MODEL and VLLM_RBLN_USE_DEVICE_TENSOR, so setting either env after `import vllm` silently leaves the platform on the wrong device. check_and_update_config now re-derives the attributes from the current env and raises if they disagree with the frozen ones.

Both the frozen value and the guard go through _use_device_tensor(), so the two env vars are tracked from a single source.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/.github/workflows/rbln_trigger_fsw_merge_ci.yaml
+++ b/.github/workflows/rbln_trigger_fsw_merge_ci.yaml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         vllm_version:
-          - "0.22.0+cpu"
+          - "0.24.0+cpu"
     steps:
       - name: Dispatch to fsw-integration (vllm ${{ matrix.vllm_version }})
         uses: Wandalen/wretry.action@v3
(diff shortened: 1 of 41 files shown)


rebel-cjlee:
@rebel-eunji Direction makes sense, but a lot of our tests intentionally flip these envs via monkeypatch, so this breaks them (`tests/torch_compile`) fails quite a bit as-is. Could we think a bit more about the implementation direction first - freeze at import and fail on any change, vs. making the value live so a change actually takes effect?

PR #878, opened 2026-08-03
model: Minimax m2.7 KV fp8 quantization

rebel-thkim:
### 🚀 Summary of Changes

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-cjlee:
Don't we need to bump the version of compiler?

rebel-thkim:
> Don't we need to bump the version of compiler?

Yes, we need it. But stable compiler is not prepared yet. @rebel-cjlee

rebel-cjlee:
@rebel-thkim Then, could you convert this to a draft, and mark it as ready when the compiler is ready?

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -365,6 +384,45 @@ def __init__(
...
+        scale = scale.reshape(-1).to(device=self.device, dtype=torch.float32)
+        numel = scale.numel()
+        if numel not in (1, self.num_kv_heads):
+            raise ValueError(
+                "RBLNFlashAttention expects a per-tensor (len 1) or per-head "
+                f"(len num_kv_heads={self.num_kv_heads}) fp8 KV scale, "
+                f"got length {numel}."
+            )
+        return scale
+
```
rebel-kblee: @rebel-mhkang do we need this normalization? I think it would be consider at custom op side

vllm_rbln/v1/attention/ops/flash_causal_attention_naive.py
```diff
@@ -66,6 +72,9 @@ def flash_causal_attention_naive_decode(
     seq_idx: torch.Tensor,
     block_tables: torch.Tensor,
     sinks: torch.Tensor | None = None,
+    k_quantize_scale: torch.Tensor | None = None,
+    v_quantize_scale: torch.Tensor | None = None,
+    cache_dtype: torch.dtype | None = None,
```
rebel-kblee, line 77: @rebel-thkim If it is None, does the compiler automatically convert it to bf16 or others?

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-thkim: The reshape/cast itself is indeed handled on the custom op side (the converter coerces to the kernel's rank-0 scalar). What this helper adds is (a) absorbing vLLM's per-scheme scale shapes — the default is a rank-0 scalar (), compressed-tensors per-tensor is [1], and the attn_head strategy is [num_kv_heads] — into one fixed op-level contract so the compiler doesn't need to track vLLM's quantization schemes, and (b) failing fast with a readable error when a model carries a scale layout the compiled kernel doesn't support yet (the current fp8 kernel is per-tensor only), instead of a shape-verify error deep inside relay.

vllm_rbln/v1/attention/ops/flash_causal_attention_naive.py
rebel-thkim: None does not trigger any automatic conversion — it means the kv_cache tensor's own dtype is authoritative: a bf16/fp16 cache takes the existing non-fp8 kernel path (the scales are dropped), handled natively as before. The fp8 kernel is selected only when cache_dtype is set, or as a fallback when the cache operand is 8-bit — in that fallback the reinterpret target defaults to e4m3, which is exactly why we pass cache_dtype explicitly on the fp8 path (an e5m2 cache would otherwise be silently read as e4m3). An explicit dtype other than e4m3fn/e5m2 fails compile with NotImplementedError.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-thkim: It seems like we could just pass it directly to `layer._k_scale` without the two, right? But right now, when the scale of [head] is set in the compiler, it doesn't display a separate not_impl_error, does it? @rebel-jaehob

tests/native/v1/attention/test_flash_attention.py
```diff
@@ -500,9 +500,32 @@ def test_unsupported_head_size_raises(self, cfg):
...
+        with pytest.raises(NotImplementedError, match="does not support"):
+            make_impl(cfg, kv_cache_dtype="nvfp4")
+
+    @pytest.mark.parametrize(
+        "kv_cache_dtype,expected",
+        [
+            ("auto", None),
+            ("fp8", torch.float8_e4m3fn),  # upstream alias of e4m3
+            ("fp8_e4m3", torch.float8_e4m3fn),
+            ("fp8_e5m2", torch.float8_e5m2),
```
rebel-kblee: this format is supported in current compiler?

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -62,8 +64,22 @@
...
 
+def _fp8_cache_dtype(kv_cache_dtype: str) -> torch.dtype | None:
+    """Real element dtype the uint8 fp8 KV-cache container holds, or None on
+    the non-fp8 "auto" path (the cache tensor's own dtype is real). Upstream's
+    kv_cache_dtype_str_to_dtype gives the uint8 byte-container dtype instead,
+    so it cannot be used here. "fp8" is an alias of e4m3 upstream."""
+    return {
+        "fp8": torch.float8_e4m3fn,
+        "fp8_e4m3": torch.float8_e4m3fn,
+        "fp8_e5m2": torch.float8_e5m2,
```
rebel-kblee: do we support "fp8_e5m2" dtype with current compiler?

rebel-cjlee:
@rebel-thkim 
`forward` dispatches to five different ops (sliding window, causal normal, flash causal, normal, flash) and this PR only wires the scales and cache dtype into the flash causal one. What happens on the other four when `kv_cache_dtype` is `fp8`? The cache tensor is `uint8` for every layer, so I think SWA or normal attention would just get a `uint8` cache and no scales.

`supported_kv_cache_dtypes` on the backend class doesn't seem to gate this either, since our platform returns the backend directly and upstream never calls `validate_configuration` for us. So the impl `__init__` check is the only gate and it lets any layer type through. Should we guard the combinations we haven't wired yet, and also the `VLLM_RBLN_USE_CUSTOM_KERNEL=1` path?

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -62,8 +64,22 @@
...
+    }.get(kv_cache_dtype)
+
+
 @register_backend(AttentionBackendEnum.FLASH_ATTN)
 class RBLNFlashAttentionBackend(AttentionBackend):
+    supported_kv_cache_dtypes: ClassVar[list[CacheDType]] = ["auto", "fp8", "fp8_e4m3"]
```
rebel-kblee: if supported, do we need to add "fp8_e5m2" in here

rebel-thkim:
@rebel-cjlee  You're right — thanks for catching this.

rebel-thkim:
@rebel-kblee  I checked with the compiler team, and they said it's supported, so I've left it open in that direction.

rebel-kblee (approved):

PR #879, opened 2026-08-04
fix(sampler): honour SamplingParams.logprob_token_ids

rebel-eunji:
## Summary

Fix `RBLNSampler` to honour `SamplingParams.logprob_token_ids`.

Requests that specify only `logprob_token_ids` currently return an empty `logprobs` list. This PR aligns the RBLN sampler with upstream vLLM.

Recreated from #853, which was branched off `release: v0.11.0 2nd` instead of `dev`. Same change, rebased onto current `dev`.

## Problem

`logprob_token_ids` requests log probabilities for specific tokens regardless of their rank. Reranking workloads need this: the caller wants scores for a fixed set of label tokens such as `"yes"` and `"no"`.

`RBLNSampler` computed log probabilities only when `max_num_logprobs` was set. When a request specified only `logprob_token_ids`:

1. `max_num_logprobs` remained `None`.
2. The sampler skipped log-probability computation.
3. `logprobs_tensors` was set to `None`.
4. The response contained an empty `logprobs` list, which surfaces as `IndexError: list index out of range`.

The request plumbing was already in place — `optimum_input_batch` builds the per-request `logprob_token_ids` mapping and passes it through `SamplingMetadata`. The missing piece was in `RBLNSampler`, which never read it.

## Changes

Mirror the relevant behaviour of upstream `Sampler.forward`:

- Compute raw log probabilities when either `max_num_logprobs` is set or `logprob_token_ids` is provided.
- Gather log probabilities for the explicitly requested token IDs.
- Prefer `logprob_token_ids` when both options are given.
- Wrap the dictionary-based gather in `@torch.compiler.disable`, since it walks a per-request python dict and would otherwise make TorchDynamo recompile on its contents.

## Expected Behaviour

```python
sampling_params = SamplingParams(
    temperature=0,
    max_tokens=1,
    logprob_token_ids=[true_token, false_token],
)
```

returns log probabilities for both tokens instead of an empty list.


rebel-jonghewk (approved):
Just a quick question. 
Is it okay if we don't add inital params when calling `super().__init__()` ? 
```
super().__init__(logprobs_mode=logprobs_mode, use_fp64_gumbel=use_fp64_gumbel)
```

rebel-eunji:
@rebel-jonghewk 
Thanks. I'll fix it together in this PR because it is minor fix.

PR #861, opened 2026-07-31
other(test): consolidate optimum tests under tests/optimum

rebel-cjlee:
### 🚀 Summary of Changes

Preparation step for the upcoming `VLLM_RBLN_USE_VLLM_MODEL=1` test renewal: consolidates the scattered optimum-path tests under `tests/optimum/` so the two backends have separate test trees. **Moves only - no test logic changes.**

| Before | After |
|---|---|
| `tests/v1/` | `tests/optimum/v1/` |
| `tests/lora/` | `tests/optimum/lora/` |
| `tests/entrypoints/` | `tests/optimum/entrypoints/` |
| `tests/optimum_compile/` | `tests/optimum/optimum_compile/` |
| `tests/optimum_accuracy/` | `tests/optimum/optimum_correctness/` |

Everything moved is optimum-only today: the `v1` / `lora` / `optimum_compile` trees import `vllm_rbln.*.optimum*` directly, and the `entrypoints` tests either pin `MODES = ["0"]` (`# FIXME MODE=1 is skipped for now.`) or leave `VLLM_RBLN_USE_VLLM_MODEL` at its `False` default while loading optimum-rbln pre-compiled artifacts. `tests/torch_compile/` is untouched.

Notes:
- All 43 test-file moves are pure `git  mv` renames with zero content diff. Each directory's `conftest.py` and `utils.py` moved with its tests, so autouse fixture scope and `from .utils import` relative imports are unchanged.
- Added `tests/optimum/__init__.py` - without it the package root shifts to `tests/optimum`, renaming modules from `tests.v1.core.utils` to `v1.core.utils` and exposing a top-level `v1` package on `sys.path`. With it, module paths keep their pre-move `tests.optimum.v1.*` form.
- Updated CI paths in `rbln_vllm-rbln-pytest.yaml` and `rbln_optimum_ci.yaml`. Matrix job names and skip tokens are unchanged, so required status check names are unaffected.

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
[...]

PR #800, opened 2026-07-20
fix(optimum): allow non-fp32 dtypes for models that support them

rebel-eunji:
### 🚀 Summary of Changes

This PR allows non-fp32 dtype.
To guard the change, it adds assertions that verify tensors are cast to the intended dtype.

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #880, opened 2026-08-04
fix(optimum): allow non-fp32 dtypes for models that support them

rebel-eunji:
## Summary

Re-applies #800, which was reverted in #862 to unblock CI. Same change, rebased onto current `dev`.

The optimum path forced `model_config.dtype` to fp32 for every model. Only optimum-rbln classes that declare `_supports_non_fp32 = False` actually need that, so this gates the override and lets the rest keep the dtype the caller asked for.

### Gate the fp32 override

```python
if forces_fp32_dtype(vllm_config.model_config):
    model_config.dtype = torch.float32
```

`forces_fp32_dtype()` resolves the optimum-rbln class for the architecture and reads its `_supports_non_fp32` flag.

### Keep vLLM and the compiled artefact in agreement

The dtype now varies per model, so it has to be carried through the compile and load paths:

- `from_pretrained` receives `model_config.dtype`, so the artefact is compiled with the requested dtype.
- `dtype` joins the compile-cache path hash, so a bf16 and an fp32 build of the same model no longer collide.
- For a pre-compiled model, `dtype` is read back out of `rbln_config.json` and written into `model_config.dtype`. The artefact is the source of truth: vLLM must feed inputs in exactly the dtype it was compiled with.
- `dtype` inside `additional_config`'s `rbln_config` is rejected with a message pointing at vLLM's own `dtype` argument. optimum-rbln takes the compile dtype as a load argument, not as an `rbln_config` field.

### Drop the casts that assumed the override

Several call sites re-cast tensors because everything was fp32 anyway:

- `RBLNOptimumMultimodalMixin` no longer casts the flattened mm embeddings to `inputs_embeds.dtype`.
- `RBLNOptimumQwenVLForConditionalGeneration` no longer casts cached mm embeddings to `self.dtype`.
- `HybridAttentionImageStrategy` no longer casts the prefill attention mask to int64. It is not a graph input — optimum-rbln only uses it to drop padding.
- PaliGemma's attention mask is cast to `self.dtype`, the model's compute dtype, rather than `rbln_model_config.torch_dtype`. This one *is* used in the computation, so it has to match.

## Rebase Notes

Two things moved on `dev` since the original PR:

- `platform.py` now imports the optimum helpers inside `check_and_update_config`, so `forces_fp32_dtype` is imported there rather than at module scope.
- The optimum test suite moved under `tests/optimum/` (#861), so `test_from_vllm.py` lands at `tests/optimum/optimum_compile/converter/test_from_vllm.py`.

Nothing else conflicted; the remaining hunks re-applied unchanged.

## Why It Was Reverted

Per #862:

> The PR raises mismatch with optimum-rbln golden because the `pixel_values` is generated as `model_config.dtype`, however the optimum-rbln generates `pixel_values` as fp32 and downcasts it.

That is a consequence of this change, not of the rebase: once a multimodal model keeps its bf16 dtype, vLLM's preprocessing emits `pixel_values` in bf16, while optimum-rbln's reference path builds them in fp32 and downcasts afterwards. The two round differently.

This PR does not address that. Please confirm the golden comparison is settled before merging.

## Tests

`tests/optimum/optimum_compile/` passes locally (92 passed), covering the re-added cases:

- `dtype` parsed out of `rbln_config`;
- `dtype` in `additional_config` rejected;
- `dtype` included in the compile-cache path hash.

One unrelated failure in that directory, `test_registry.py::test_registered_class_exists_in_optimum[Qwen3_5ForConditionalGeneration]`, is pre-existing on `dev` — the local optimum-rbln 0.11.1a2 does not export `RBLNQwen3_5ForConditionalGeneration`.

## Type of Change

- [x] Bug fix (`fix`)

## Related PRs
- https://github.com/RBLN-SW/optimum-rbln/pull/674 -> This should be merged with this PR.


vllm_rbln/model_executor/models/optimum/model_base.py
```diff
@@ -241,7 +241,10 @@ def init_model(self) -> None:
                 json.dumps(spec.rbln_config, indent=2, default=str),
             )
             model = spec.model_cls.from_pretrained(
-                self.model_config.model, rbln_config=spec.rbln_config, config=hf_config
+                self.model_config.model,
+                rbln_config=spec.rbln_config,
+                config=hf_config,
+                dtype=self.model_config.dtype,
```
rebel-thkim, line 248: Would there be any issues if I passed `self.model_config.dtype` to models (VLMs) that contain submodules?

vllm_rbln/utils/optimum/converter/from_optimum.py
```diff
@@ -116,6 +118,18 @@ def sync_from_optimum(
...
+    # inputs in exactly the dtype the artefact was compiled with.
+    if params.dtype is not None:
+        compiled_dtype = getattr(torch, params.dtype)
+        if vllm_config.model_config.dtype != compiled_dtype:
+            logger.info(
+                "Updating model_config.dtype from %s to %s based on rbln_config.json",
+                vllm_config.model_config.dtype,
+                compiled_dtype,
+            )
+            vllm_config.model_config.dtype = compiled_dtype
```
rebel-thkim, line 131: I'm not entirely sure what this code block means, but there aren't any issues when the dtype in the submodule differs from that in the parent module, right?

vllm_rbln/model_executor/models/optimum/model_base.py
rebel-eunji: Currently, vllm does not consider the dtypes of submodules are different. I'll go over it.

vllm_rbln/utils/optimum/converter/from_optimum.py
rebel-eunji: I found out that manually changing dtype is impossible in optimum-rbln for now because of update_rbln_config.

Every submodule inherits the global dtype.

https://github.com/RBLN-SW/optimum-rbln/blob/fb6878e7afad54c6d54d6a61ac0ba98e0ab53816/src/optimum/rbln/modeling_base.py#L484

vllm_rbln/v1/worker/optimum_model_runner.py
```diff
@@ -306,6 +306,12 @@ def is_ec_producer_only(self) -> bool:
     def load_model(self) -> None:
         with set_current_vllm_config(self.vllm_config, check_compile=False):
             self.model = get_optimum_model(vllm_config=self.vllm_config)
+        # The converter keeps model_config.dtype in sync with the compiled
+        # artefact, so the runner reads self.dtype everywhere; catch drift here.
+        assert self.model.dtype == self.dtype, (
+            f"model_config.dtype ({self.dtype}) does not match the compiled "
+            f"artefact's dtype ({self.model.dtype})"
+        )
```
rebel-jonghewk: I'm wondering if an assertion is needed here. My understanding is that when the dtype doesn't match, we currently perform a conversion, which incurs some overhead. If so, a warning might be enough instead of hard-failing.

rebel-eunji: The assertion works for catching bug. As you said, the vllm dtype will be changed silently to dtype of compiled model before the assertion.

rebel-eunji: I updated the assertion log to clarify that it is an internal bug check ensuring that the invariant the converter must maintain has not been violated.

rebel-jonghewk (approved):

PR #889, opened 2026-08-05
fix(sampler): remove torch.compile.disable and remove unsued op

rebel-eunji:
## Summary

`torch.compile` was regressing CPU sampler latency for two reasons:

- `apply_temperature` used out-of-place division as a workaround for a TorchInductor issue with in-place division. This allocated a logits-sized temporary tensor on every call.
- Multiple `@torch.compiler.disable` escape hatches fragmented the sampler into disconnected graphs, preventing compilation from providing meaningful performance benefits.

This PR removes `torch.compile` from the CPU sampler path and cleans up the workarounds that were required to support it.

## Changes

- Stop wrapping `self.sampler` with `torch.compile` in `optimum_model_runner.py`.
- Remove the now-unnecessary `@torch.compiler.disable` annotations.
- Remove compile-specific workaround operations from:
  - `sample/ops/logprobs.py`
  - `sample/ops/penalties.py`
- Restore `apply_temperature` to use in-place division (`div_`), matching upstream vLLM and avoiding the temporary tensor allocation.

Because `apply_temperature` now mutates logits in place, this PR also includes two changes to ensure correctness:

- `RBLNInputBatch._make_sampling_metadata_rbln` resets the temperature values for padding rows to `1.0` on every metadata rebuild. Padding rows share a pooled logits buffer that is not refreshed on every step. Without this reset, a stale temperature left by a removed request could be applied repeatedly, eventually causing `inf` or `NaN` values.
- `benchmarks/benchmark_sampler.py` restores logits from `reference_logits` before each warmup and benchmark iteration. Otherwise, in-place division would repeatedly rescale the same tensor, affecting both sampled outputs and latency measurements.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -222,24 +213,29 @@ def sample(
...
-        return logits
+        # `forward` hands the logits over in the model dtype rather than casting
+        # them to float32, so the constants the builtin processors write into
+        # them have to follow. `MinTokensLogitsProcessor` writes its -inf with
+        # `index_put_`, which rejects a source dtype that differs from the
+        # destination — unlike `__setitem__`, which converts — so a float32
+        # constant raises on a half-precision logits buffer.
+        for processor in sampling_metadata.logitsprocs.all:
+            if (
+                isinstance(processor, MinTokensLogitsProcessor)
```
rebel-jonghewk: what's the reason for overriding `apply_logits_processors` instead of fixing where `neg_inf_tensor` is built?

rebel-jonghewk (approved):

rebel-jonghewk:
Also, could you check whether there are enough test cases to cover the various sampler-related issues?

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: This was required to support the min_tokens feature and is not related to removing the CPU torch.compile.
I'll add this fix in a separate PR.

rebel-cjlee (approved):

rebel-eunji:
I've just replaced min_tokens with ignore_eos=True in openai pytest because the min_tokens was used for keep generating texts blocking eos_token.

> min_tokens – Minimum number of tokens to generate per output sequence before EOS or stop_token_ids can be generated

I'll merge this PR and open new PRs for robust pytests and min_tokens.

PR #896, opened 2026-08-06
fix(worker): read the DP device mapping vLLM 0.24 puts on the config

rebel-chanheo:
### 🚀 Summary of Changes

`_init_device_env`'s explicit branch reads `RBLN_DEVICES` directly. That was correct through vLLM 0.23, where the launcher narrowed the env var itself before spawning each DP engine:

```python
# v0.23.0  vllm/v1/engine/utils.py
def set_device_control_env_var(vllm_config, local_dp_rank):
    value = get_device_indices(evar, local_dp_rank, world_size, local_world_size)
    with patch.dict(os.environ, values=((evar, value),)):
        yield
```

applied under `is_dp and (needs_device_env_isolation or use_ray)`, where `needs_device_env_isolation = not (is_cuda_alike() or is_xpu())` — RBLN qualifies. So the worker always saw a value already scoped to its shard.

0.24 replaced that with `set_assigned_physical_gpu_ids_for_dp_rank()` at the same call site and under the same condition. It leaves the env var alone and puts the per-rank mapping on the config instead:

```python
# v0.24.0  vllm/v1/engine/utils.py
vllm_config.parallel_config.assigned_physical_gpu_ids = physical_gpu_ids
```

Nothing narrows `RBLN_DEVICES` any more, so every rank sees the whole deployment's list:

```
AssertionError: device_ids: ['4', '5', '6', '7'] should have device count: 1
```

(DP=4, TP=1 on `RBLN_DEVICES=4,5,6,7`; `world_size` is 1, so each rank wants one device and gets four.)

This reads the mapping when it is present and falls back to the env var when it is not.

### Why the mapping rather than slicing here

Upstream derives it with `local_dp_rank` and `local_world_size`, so multi-node, hybrid LB and Ray land on the right shard without this file knowing the topology. Slicing here from the global `data_parallel_rank` would get those wrong.

It also covers non-MoE DP, which slicing would miss: the mapping is written in the launcher, before the engine process resets `data_parallel_size` and `data_parallel_rank` to 1/0 for dense models ([core.py](https://github.com/vllm-project/vllm/blob/v0.24.0/vllm/v1/engine/core.py#L1188-L1200)).

`world_size` is TPxPP in both 0.23 and 0.24 — the env var is what changed meaning, not the size. An earlier revision of this PR claimed otherwise; that was wrong.

`getattr` rather than attribute access, so the file still imports against a vLLM that predates the field.

### Testing

`TestInitDeviceEnv` had no case where an explicit device list met DP > 1 — the reason this could break without CI noticing. Four added:

| test | without the fix |
|---|---|
| `assigned_physical_ids_win_over_the_env_var` | fails, production message |
| `assigned_physical_ids_expand_by_num_devices` | fails, production message |
| `no_assignment_falls_back_to_the_env_var` | passes (pins the fallback) |
| `assigned_physical_ids_wrong_count_still_asserts` | passes (pins the guard) |

83 passed / 4 skipped in the file.

End-to-end: this is the behaviour every DP4 EAGLE3 serving run on MiniMax-M2.5 has had locally — HumanEval 164/164, MT-Bench 80/80, ShareGPT 100/100 and a full SWE-bench pass. Without it, the server does not boot.

Regression from #829 (vLLM 0.24 bump).

### DIFF
--- a/tests/native/v1/worker/test_rbln_worker.py
+++ b/tests/native/v1/worker/test_rbln_worker.py
@@ -38,6 +38,7 @@ def _make_vllm_config(
     data_parallel_size=1,
     data_parallel_rank=0,
     world_size_across_dp=1,
+    assigned_physical_gpu_ids=None,
     quantization=None,
     enforce_eager=False,
     profiler=None,
@@ -51,6 +52,7 @@ def _make_vllm_config(
             data_parallel_size=data_parallel_size,
             data_parallel_rank=data_parallel_rank,
             world_size_across_dp=world_size_across_dp,
+            assigned_physical_gpu_ids=assigned_physical_gpu_ids,
             disable_custom_all_reduce=False,
         ),
         model_config=SimpleNamespace(
@@ -108,6 +110,7 @@ def make_worker(monkeypatch):
         data_parallel_size=1,
         data_parallel_rank=0,
         world_size_across_dp=None,
+        assigned_physical_gpu_ids=None,
         num_devices=1,
         num_ray_nodes=1,
         has_torch_rbln=False,
@@ -120,6 +123,7 @@ def make_worker(monkeypatch):
             data_parallel_size=data_parallel_size,
             data_parallel_rank=data_parallel_rank,
             world_size_across_dp=wsd,
+            assigned_physical_gpu_ids=assigned_physical_gpu_ids,
         )
         monkeypatch.setattr(WorkerBase, "__init__", _fake_super_init)
         monkeypatch.setattr(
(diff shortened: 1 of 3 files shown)


rebel-cjlee:
@rebel-chanheo Could you also add tests in `tests/native`? I think adding them to `TestInitDeviceEnv` in `tests/native/v1/test_rbln_worker.py` would be great.
FYI, `tests/torch_compile` is scheduled for removal soon.

rebel-chanheo:
> @rebel-chanheo Could you also add tests in `tests/native`? I think adding them to `TestInitDeviceEnv` in `tests/native/v1/test_rbln_worker.py` would be great. FYI, `tests/torch_compile` is scheduled for removal soon.

Ok — same 4 cases added to `TestInitDeviceEnv` in `tests/native/v1/worker/test_rbln_worker.py`.

rebel-cjlee (approved):

PR #899, opened 2026-08-07
fix(metrics): record E2E per execute_model pass instead of per sampler call

rebel-daeyang:
#### Problem

**DP=1 chunked prefill** — E2E was only recorded where the sampler span ended, so every intermediate chunk was dropped: `_sample()` returns early, before `profile_sampler()`, when `is_intermediate_chunked_prefill` is true. On a 4-prompt × 4-chunk run `PREFILL + SAMPLE` counted 16 while `PREFILL E2E` kept 4 — and each of those 4 covered only that request's **last chunk**.

**DP>1 chunked prefill** — On top of the same loss, the E2E anchor sat at the model span start, so the `num_tokens_across_dp` all_reduce in `_determine_batch_padding()` fell outside the measured window. Rank skew from that host-blocking barrier was invisible in E2E, and it is also where `PREFILL E2E` came out *below* `mean_device_time`.

#### Solution

Move the E2E boundary from the model span to the whole **pass** (`execute_model` entry → `sample_tokens` return).

```
before   [ model ]· · ·[ sampler ]       E2E = model start → sampler end (nothing recorded without a sampler)

after    start_e2e ┃ prepare + DP barrier + [ model ] + postprocess + [ sampler ] + bookkeeping ┃ end_e2e
                   └─────────────────────────────────────E2E────────────────────────────────────┘
```

* Both levels are now recorded **per pass**, so their call counts match and subtracting the two means gives `E2E − (MODEL + SAMPLE)` = the host overhead around the compiled graph.
* The boundary is taken by two decorators on the runner (`@e2e_starts` / `@e2e_ends`). `execute_model` returning `None` is exactly equivalent to `sample_tokens` following (see `EngineCore.step()`), so on a non-None return the pass is closed in `execute_model` instead — that is the path taken by a non-last PP rank and by pooling models.
* prepare, the DP barrier, postprocess and bookkeeping now fall inside E2E. Engine-side work *between* passes (`schedule()`, `update_from_output()`, the DP unfinished-reqs barrier) is deliberately excluded, so E2E stays below the latency a serving benchmark reports by that margin.

---

* No linked issue (self-contained metrics instrumentation fix)

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit: `pytest tests/torch_compile/unit/v1/worker/test_metrics_v2.py` (20 passed)
2. E2E: run gpt-oss-20b (4 layers) with `VLLM_RBLN_METRICS=1`. For chunked prefill use `--max-batched 256 --input-len 1024` (4 chunks), `--num-prompts 4 --max-tokens 32`.
3. What to check: for each phase, `Total call counts` must be equal under `+ SAMPLE` and `E2E`, and the `E2E` mean must be greater than or equal to the `+ SAMPLE` mean.

---

| case | phase | count (`+ SAMPLE` vs `E2E`) | model+sample | E2E | host overhead |
|---|---|---|---|---|---|
| DP=1 chunked | PREFILL | 16 vs 16 | 14.33 ms | 15.45 ms | 1.11 ms (7.2%) |
| | DECODE | 31 vs 31 | 7.94 ms | 10.14 ms | 2.20 ms (21.7%) |
| DP=2 chunked, DP0 | PREFILL | 16 vs 16 | 16.90 ms | 18.50 ms | 1.59 ms (8.6%) |
| | DECODE | 31 vs 31 | 8.12 ms | 11.00 ms | 2.87 ms (26.1%) |
| DP=2 chunked, DP1 | PREFILL | 16 vs 16 | 16.81 ms | 23.01 ms | 6.19 ms (26.9%) |
| | DECODE | 31 vs 31 | 7.93 ms | 11.00 ms | 3.07 ms (28.0%) |

Before this change, `PREFILL E2E` on chunked prefill reported 4 instead of 16. In the DP=2 rows, `model+sample` is effectively identical across the two ranks (16.90 / 16.81) while `PREFILL E2E` diverges (18.50 / 23.01) — that gap is the DP barrier skew, which this change makes visible for the first time.

---

### 📋 Checklist

---

### DIFF
--- a/tests/torch_compile/unit/v1/worker/test_metrics_v2.py
+++ b/tests/torch_compile/unit/v1/worker/test_metrics_v2.py
@@ -1,0 +1,364 @@
+# Copyright 2025 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for vllm_rbln.v1.worker.metrics_v2.
+
+The invariants that matter here are structural rather than numeric:
+
+* every pass counted in ``MODEL + SAMPLE`` must also be counted in ``E2E``, so
+  that subtracting their means gives the host overhead.
+* ``E2E`` must never fall below ``MODEL + SAMPLE``, since it encloses it.
+* neither level may absorb the engine time between two passes.
+"""
+
+import pytest
+
+import vllm_rbln.v1.worker.metrics_v2 as mv2
+from vllm_rbln.v1.worker.metrics_v2 import (
+    _e2e_ends,
+    _e2e_starts,
+    _NoopPerformanceContext,
+    _PerformanceContext,
+)
+
+MS = 0.001  # tests work in seconds; use milliseconds for readability
+
+
+class FakeClock:
+    """Stand-in for the ``time`` module: only ``perf_counter`` is used."""
+
+    def __init__(self) -> None:
+        self.now = 0.0
+
+    def perf_counter(self) -> float:
+        return self.now
+
+    def advance(self, seconds: float) -> None:
+        self.now += seconds
+
+
+@pytest.fixture
+def clock(monkeypatch):
+    fake = FakeClock()
+    monkeypatch.setattr(mv2, "time", fake)
+    # Keep the spans on nullcontext even if a real rebel runtime is installed.
+    monkeypatch.setattr(mv2, "_REBEL_HAS_CAPTURE", False)
+    return fake
+
+
+@pytest.fixture
+def ctx(monkeypatch):
+    # _rank_tag() imports vllm.distributed; the tag is irrelevant here.
+    monkeypatch.setattr(mv2, "_rank_tag", lambda: "")
+    return _PerformanceContext("test")
+
+
+def report_lines(ctx, monkeypatch):
+    """print_stats() 가 실제로 찍는 리포트 줄들."""
+    monkeypatch.setattr(mv2.envs, "VLLM_RBLN_METRICS_DIR", "", raising=False)
+    captured: list[str] = []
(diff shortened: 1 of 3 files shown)


rebel-cjlee:
@rebel-daeyang Before we proceed with the review, could you please update the PR description to English? As part of our team's policy, we maintain all PRs in English to align with standard open-source conventions.

rebel-daeyang:
@rebel-cjlee Thanks for the heads-up. I've updated the description to English; the code is unchanged.

rebel-cjlee:
Right now `capture_reports()` (the runtime timer) is scoped to the model/sample compute spans only. Would it make sense to extend it to the E2E span as well? I may be missing the intent here.

rebel-daeyang:
`capture_reports()` only collects reports emitted by rebel runtime executions, and the host path E2E is meant to expose emits none of them — input prep, DP all_reduce, postprocess, bookkeeping. Extending the capture scope wouldn't reach any of that, so I think the boundary has to sit in the runner.

rebel-cjlee:
@rebel-daeyang I don't understand it below:
> the host path E2E is meant to expose emits none of them

Doesn't the E2E path include the execution of the runtime executor (model, sample)?

rebel-daeyang:
Extending `capture_reports()` to the E2E span and extracting the model/sampler runtime metrics from within it is a nicer shape than keeping two separate spans.

What I'd want to confirm first is whether those runtime metrics actually come out correctly that way. The report values may be unreliable on their own right now.
Since the E2E numbers on `dev` are wrong today, could we merge this fix first and do your approach as a follow-up PR?

rebel-cjlee:
Please let me know once the verification is complete. If there are no issues with extending the span, would it be okay if I handle the refactoring follow-up?

rebel-cjlee (approved):

rebel-daeyang:
Sure, I'll let you know when the verification is complete. And yes, please go ahead with the refactoring follow-up.

rebel-daeyang:
@rebel-cjlee I tried the idea out before implementing it, and ran into a few issues along the way. The straightforward version does not work as expected, since `capture_reports()` keeps a single thread-local slot rather than a stack, so an inner capture shadows the outer one and the E2E list comes out empty. There is also a separate problem in how the reports are aggregated today, independent of this change.

I wrote up the measurements and split the follow-up into three items, A, B and C, in #909. B is on the `rebel_compiler` side and I will take that one.

Could you take a look at A and C and let me know whether you would prefer me to work on them, or whether it would be better for you to pick them up?

PR #900, opened 2026-08-07
feature(pp): pipeline parallelism for the refactored RBLN scheduler/runner

rebel-ykchoi:
> **Step 1 of 3** — split of #838 (*pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer*). This PR carries **section A** of #838: the PP-aware scheduler & runner. The NIXL transfer PR is stacked on this branch; the data-parallel decode-shape PR is parallel, also based on this branch.

#### Summary

Pipeline parallelism (`pipeline_parallel_size > 1`) on the RBLN path worked for plain decode but broke down the moment decode batching or speculative decoding entered the picture: a step's decode batch could overflow the compiled per-stage bucket, and spec-decode misclassified the prefill/decode phase inconsistently across ranks. This PR reorganizes the scheduler/runner so PP is a first-class, correct, and efficient mode — one PP-aware decode-admission budget, rank-consistent speculative decoding on top of it, prefill/decode phase authority consolidated onto the scheduler, and batch balancing on by default for PP.

#### What changed

**5. Drafter access gated on the last PP rank.** The speculative drafter is constructed only on the last PP rank, so the dummy-run, attention-backend initialization, and medusa warm-up paths gate on the drafter object rather than on `speculative_config` (which every rank carries). A non-last PP rank under speculative decoding skips the draft leg cleanly instead of dereferencing an absent attribute.

#### Testing

- **Unit** — the PP decode-cap machinery, admission budget, and the no-mixed-batching / held-prefill-not-starved scheduler behavior (`test_rbln_scheduler.py`); the decode-cap policies and budget (`test_utils.py`); the scheduler-stamped step phase on the spec-decode input path (`test_rbln_model_runner_inputs.py`); runner and drafter helpers (`test_rbln_model_runner.py`, `test_eagle.py`).
- `pre-commit run --hook-stage manual` clean (ruff, mypy 3.10–3.13, typos, license); runner/scheduler unit suites green.

#### Risk / compatibility

- The `VLLM_RBLN_PP_BALANCE_DECODE_BATCH` default flip is confined to PP; the escape hatch (`=0`) preserves the old behavior.

#### Affected modules

Platform, Scheduler, Runner.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)


rebel-cjlee:
nit
The comments in this change are far more verbose than they need to be, and in several spots the same explanation is repeated across multiple places. The logic is subtle enough to deserve good comment, but let's keep each explanation in one place and trim the reset down to the essentials.

vllm_rbln/envs.py
```diff
@@ -70,6 +70,12 @@
     VLLM_RBLN_USE_MOE_TOKENS_MASK: bool = True
     VLLM_RBLN_DISPATCH_ALL2ALL: bool = False
     VLLM_RBLN_COMBINE_ALL2ALL: bool = False
+    # --- PIPELINE PARALLELISM ---
+    # Spread each step's decode batch across PP microbatches
+    # (ceil(demand / pp_size)) instead of packing into one; avoids PP-depth
+    # collapse. See v1/core/utils.py DynamicDecodeCapPolicy. Off => static
+    # max_num_seqs // pp_size cap.
+    VLLM_RBLN_PP_BALANCE_DECODE_BATCH: bool = True
```
rebel-cjlee, line 78: Is there a real case where this is set to `False`? If we don't expect anyone to actually turn it off, the flag is just dead configuration surface. So, it might be cleaner to drop it and always balance. If it's meant as a temporary escape hatch, could we note when we plan to remove it?

vllm_rbln/platform.py
```diff
@@ -299,6 +299,48 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                        max_num_seqs - per_stage * pp_size,
+                    )
+                logger.info(
+                    "PP=%d: max_num_seqs=%d (max concurrent sequences) -> "
+                    "per-PP-stage decode batch = %d (compiled).",
+                    pp_size,
+                    max_num_seqs,
+                    per_stage,
+                )
+
```
rebel-cjlee:
```suggestion
            # Under PP the compiled per-stage decode batch is max_num_seqs // pp_size
            # (see decode_batch_size). Fail fast on an impossible config.
            pp_size = parallel_config.pipeline_parallel_size
            if pp_size > 1:
                max_num_seqs = scheduler_config.max_num_seqs
                if max_num_seqs < pp_size:
                    raise ValueError(
                        f"pipeline_parallel_size={pp_size} requires max_num_seqs >= {pp_size} "
                        f"(got {max_num_seqs}); per-stage decode batch would floor to 0."
                    )
                if max_num_seqs % pp_size != 0:
                    logger.warning(
                        "max_num_seqs=%d is not a multiple of pipeline_parallel_size=%d; "
                        "%d decode slot(s) will be unused.",
                        max_num_seqs, pp_size, max_num_seqs % pp_size,
                    )

```

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-cjlee, line 1: A lot of the comments here just restate what the next line already make clear, which makes the file heavier than it needs to be. Suggest trimming those and keeping comments for the non-obvious parts.

e.g, the 3-line block above `self._is_prefill_step = False` (~L226) - the name already says it's the prefill stpe, and the prefill/decode step distinction isn't new here.

rebel-cjlee, line 1: The same value goes by two names - `is_prefill_step` on the scheduler side vs `is_prefill_phase()` / `_set_step_phase` in the runner. And the setter (`_set_step_phase`) is a one-off that the dummy path bypasses with a raw field write. Suggest collapsing it to one property, named `is_prefill_step` everywhere, with both writes assigning though the setter:
```python
@property
def is_prefill_step(self) -> bool:
    return self._is_prefill_step

@is_prefill_step.setter
def is_prefill_step(self, value: bool) -> None:
    self._is_prefill_step = value

# _update_states (at the top, before any is_prefill_step read / early return)
def _update_states(self, scheduler_output):
    self.is_prefill_step = scheduler_output.is_prefill_step
    ...

# _dummy_run / execute_dummy_batch
self.is_prefill_step = is_prefill
```
This drops `_set_step_phase`, gives both write paths (scheduler stamp + dummy) the same entry point, and keeps one name from the scheduler output all the way through the runner. `_update_states` runs unconditionally before any read/early-return in `execute_model`, so stamping there is safe (or place out of `_update_states`).

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -93,7 +93,10 @@ def propose(
         )
 
         assert self.runner is not None
-        is_prefill = self.runner.is_prefill
+        # Use the scheduler-stamped step phase (rank-consistent) rather than the
+        # runner's per-rank is_prefill property, which can diverge across PP
+        # ranks under spec-decode. Drives draft batch padding + draft attn build.
+        is_prefill = self.runner.is_prefill_phase()
```
rebel-cjlee, line 96: Also redundant here and `is_prefill_phase` is self-explanatory.

vllm_rbln/v1/core/utils.py
```diff
@@ -0,0 +1,346 @@
...
+
+class DynamicDecodeCapPolicy:
+    """Cap that spreads the decode *demand* across the PP pipeline.
+
+    Avoids PP-depth collapse: after a pipeline drain (e.g. a
+    prefill stall) every decode becomes reschedulable at once, and the static
+    ``max_num_seqs // pp_size`` cap lets them pack into a single microbatch ->
+    the other ``pp_size - 1`` stages idle (depth-1, ~2x decode latency). Sizing
+    each step's batch to ``ceil(demand / pp_size)`` instead spreads the decode
+    demand across ``pp_size`` microbatches so the pipeline stays full.
```
rebel-cjlee: I have a question about this: Does splitting the batch help if a bigger batch isn't slower?

If I'm reading `DynamicDecodeCapPolicy` right, it splits a batch that the static cap would run in one step. With `max_num_seqs=8`, `pp=2` (cap=4) and 4 running decodes, `[A, B, C, D]` becomes `[A, B]` in one step and `[C, D]` in the next, so both PP stages have work.

This only helps if a batch of 4 takes more time per stage than a batch of 2. We halve the batch to keep the second stage busy, so the two even out unless the bigger batch really is slower.

For decode, most of the time goes to reading the weights, and that is the same work for a batch of 2 or 4. So both sizes take about the same time, and that is why batching helps throughput at all. If that is true here, splitting and not splitting give the same speed, and this policy does nothing in the case we hit most often.

So, do we have numbers for how long one decode step takes with a batch of 2, 4, and 8? That is the whole question here: if a batch of 4 takes about as long as a batch of 2, splitting gains nothing; if it takes twice as long, splitting roughly doubles throughput.
If a batch of 4 does turn out to be clearly slower than a batch of 2 on our hardware, then the balancing is worth having and you can ignore all of the above.

rebel-huijong (commented):
I like the approach of dynamically adjusting the micro batch sizes based on the currently running requests, rather than just capping the PP micro batch size at `max_num_seqs // pp_size`. 

However, as noted in @rebel-cjlee's reviews, the code changes feels a bit heavy for the feature, it would be great if it could be simplified, and it looks like there's definitely room to do so.

vllm_rbln/v1/core/utils.py
rebel-ykchoi: Good question. The framing here is the standard throughput↔interactivity trade-off: batching decode raises throughput but each decode step's latency grows with the batch — you're moving along the usual Pareto curve, and this holds on GPUs as well. Our own batch-size sweep on this device shows the same shape: decode latency increases with batch size (a batch of 4 is measurably slower than 2, and 8 slower still).

So the premise that "a batch of 2 and a batch of 4 take about the same time, so splitting does nothing" doesn't hold — a bigger decode batch really is slower.

Given that, splitting a fixed set of active decodes across PP microbatches helps for two reasons. Using your example (max_num_seqs=8, pp=2, 4 running decodes):
1. Smaller microbatches are genuinely cheaper. Packing all 4 into one microbatch costs L(4); splitting into [A,B] and [C,D] (ceil(4/2)=2) costs L(2) < L(4) per microbatch, so per-step latency drops rather than staying flat.
2. The pipeline stays full. With a single microbatch in flight on PP=2, one stage is always idle (a depth-1 pipeline). Two microbatches overlap across the two stages, so both stay busy.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: On the property suggestion — I held off, but let me lay out the reasoning and get your read rather than decide unilaterally.

`_is_prefill_step` is per-step mutable state: it's re-stamped every step (scheduler-authoritative in `execute_model`, the dummy's own phase on the warm-up / DP-idle path). My concern with exposing it as a property (especially with a setter) is that `self.is_prefill_step` then reads as a stable attribute — which can invite a stale read on a step that didn't stamp it, and hides the "stamped once per step" intent. A query method (`is_prefill_phase()`) seemed to signal that volatility more honestly. Does that reasoning hold for you, or do you still lean toward the property + unified name — i.e. is the naming/one-write-path cleanup worth trading that signal?

Your other point I think is clearly right regardless: the dummy path was setting the field raw, bypassing the setter. I've made `_set_step_phase` take the bool and be the sole writer, so both paths go through it now.

If you'd rather take it all the way to the property/rename, I'm happy to — just wanted to check the "stable-attribute vs volatile-state" trade-off with you first.

rebel-cjlee: On property vs method: I get that the concern is temporal validity, not mutability. But method-vs-property doesn't actually buy us that. It's a weak signal(`()` doesn't say "only valid this step"), it's already per-step mutable state behind `_is_prefill_step` so a property adds no new volatility, and every reader stamps it first anyway (top of `execute_model`, `_dummy_run`), so stale reads are largely hypothetical.

If stale reads are the real worry I"d enforce the contract rather than signal it through a name. e.g. reset to `None` at step start so an unstamped read fails loudly.

Still lean property + one name (`is_prefill_step`) end-to-end, optionally with that reset. But if you'd rather keep a method, that's fine too. In that case let's at least match the name.

vllm_rbln/v1/core/utils.py
rebel-cjlee: Thanks, the measured curve settles the perf question. No argument that balancing is worth having.

One follow-up on the implementation weight: the decision is binary (static cap vs `ceil(demand/pp)`), so the policy layer (protocol, static/dynamic class, controller factory) feels heavy for what it encodes. Unless a third strategy is planned, I think it collapses to one method in `rbln_scheduler.py`: the cap is a one-liner (`hard_cap if not balance else min(hard_cap, ceil(demand/pp))`) and the per-step count is a local counter plus a small `can_admit` helper. `decode_batch_size` stays as the shared helper, and the scheduler tests already covers the behavior.

One more question: Is decode-under-PP actually a deployment we're targeting?

rebel-ykchoi: Agreed the policy/controller layering was too much for a binary decision — I'd actually already collapsed it in the latest push, close to what you're describing:

- Dropped the protocol + static/dynamic policy classes + controller factory. What's left is one small DecodeBatchBudget(hard_cap, soft_cap), built per step by a for_step(max_num_seqs, pp, demand) classmethod. decode_batch_size stays the shared helper.
- Also dropped the `VLLM_RBLN_PP_BALANCE_DECODE_BATCH` flag: the soft cap is now unconditionally ceil(demand/pp), a no-op at pp == 1, so there's no balance branch left. The scheduler side is just for_step(...) + admit/discard/reset/can_admit, and the existing scheduler tests cover it.

Does that land where you wanted, or would you still prefer to inline it further (local counter in schedule(), drop the class)? Happy either way — wanted your read before finalizing.

On decode-under-PP: not a deployment we're targeting right now, but it's plausibly on the horizon, so I'd lean toward keeping this already-simplified form rather than stripping the balancing out entirely. Open to your view.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Good call — switched it to a property with one name end-to-end. is_prefill_step is now a getter/setter backed by _is_prefill_step, and readers just do self.is_prefill_step; the old is_prefill_phase() method and separate _set_step_phase writer are gone. I also moved the stamp to the very top of both entry points — right after the re-entrancy guard in execute_model, and the first statement after the asserts in _dummy_run — so it's unambiguously the step's first action. Tests updated to match.

vllm_rbln/v1/core/utils.py
```diff
@@ -0,0 +1,346 @@
...
+        """Un-admit ``n`` decode requests dropped from this step's batch.
+
+        Unlike ``reset()`` (whole batch dropped by a prefill eviction), this
+        removes only the requests that left the batch while others remain
+        admitted -- currently the PRIORITY-policy preemption in the running
+        loop, which evicts an already-scheduled decode to free KV blocks.
+        Keeping the count in step with the batch keeps the subsequent
+        ``can_admit()`` gate from stopping early on a stale (over)count.
+        """
+        self._count -= n
```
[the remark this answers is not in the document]
rebel-ykchoi: Good catch — guarded it explicitly. discard() now asserts the invariant before decrementing:
```
# A discard without a matching admit would drive the count negative,
# silently disabling both caps (can_admit only tests >=) -> overschedule.
assert self._count >= n, (
    f"discard({n}) with count={self._count}: discard without a matching admit"
)
self._count -= n
```
with a unit test (test_decode_budget_discard_underflow_asserts) covering both the bare discard() and the admit/discard/discard sequence.

For the record on the current code: the only discard() call site is the PRIORITY-policy preemption in the running loop, and it's inside the if preempted_req in scheduled_running_reqs: branch — so the victim was always admitted earlier in the same step and the count can't go negative today. The assert is there to keep it that way if a future path starts discarding.

Note the class has since been collapsed to a single DecodeBatchBudget(hard_cap, soft_cap) (the policy/controller layering is gone), but the guard and its test carried over unchanged.

vllm_rbln/envs.py
rebel-ykchoi: You were right — there was no real case for False, so I dropped the flag entirely rather than keeping it as an escape hatch. `VLLM_RBLN_PP_BALANCE_DECODE_BATCH` is gone from envs.py and there are no references left in the tree.

Balancing is now unconditional: the per-step budget takes a soft cap of ceil(demand / pp). No balance branch is needed because at pp == 1 that reduces to demand, which the max_num_seqs hard cap already bounds — so the non-PP path behaves exactly as before, with nothing to configure.

This also let the machinery collapse (see the utils.py thread): the static-vs-dynamic choice was the only thing the policy/controller layer existed to make, so with the flag gone it folded into one small DecodeBatchBudget(hard_cap, soft_cap).

vllm_rbln/platform.py
rebel-ykchoi: Applied, thanks — took the suggestion nearly verbatim: the 10-line NOTE is down to the two lines you wrote, and the ValueError / warning strings are the short forms from your snippet.

One deviation: I kept the logger.info line that reports `pipeline_parallel_size`, `max_num_seqs`, and the resulting `per-stage decode batch`. Under PP the batch that actually gets compiled is the derived `max_num_seqs // pp_size rather` than `max_num_seqs` itself, so having it in the startup log makes the bucket sizes downstream self-explanatory. Happy to drop it if you'd rather keep the startup output lean.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Agreed, and trimmed. The block you pointed at is now a one-liner:
```
# Step phase; see is_prefill_step for authority and lifecycle.
self._is_prefill_step: bool = False
```
I went through the rest of the file on the same rule — drop anything that restates the next line, keep the non-obvious parts (the RBLN-specific constraints stay marked with NOTE(RBLN)). That's ~96 comment lines reworked in this file, and I applied the same pass across the rest of the change rather than just here.

vllm_rbln/v1/spec_decode/eagle.py
rebel-ykchoi: Dropped — the three lines are gone; it's just the read now:
```
assert self.runner is not None
is_prefill = self.runner.is_prefill_step
```
The "why not re-derive per rank" rationale wasn't worth repeating at each call site, so it lives in one place (the is_prefill_step getter docstring on the runner) and the readers stay bare. With the property rename in the latest push the call site reads as plain attribute access, which makes it even more self-evident.

vllm_rbln/platform.py
rebel-cjlee: Makes sense, the derived per-stage batch isn't obvious from the config, so surfacing it once is fair. Fine to keep.
One thing: `check_and_update_config` runs per rank, so, it can print once per PP rank. I think `logger.info_once` is better here.

vllm_rbln/v1/core/utils.py
rebel-cjlee: No need to inline further. admit/discard/reset/can_admit as methods reads better than a bare counter inside schedule(). Good to finalize it.

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -38,16 +42,28 @@
...
+    prefix caching, and a step-level prefill/decode phase anchor."""
 
     kv_cache_copy_ops: list[KVCacheCopyOp] = field(default_factory=list)
+    # NOTE(RBLN): Whether this step runs the prefill graph (True) or the decode
+    # graph (False). RBLN has no mixed batching, so a step is uniformly one or
+    # the other. The scheduler is the single authority: it stamps this from the
+    # pre-advance Request state so every PP rank reads an identical value. The
+    # runner consumes it via its is_prefill_step property, which documents why a
+    # per-rank re-derivation would diverge.
+    is_prefill_step: bool = False
```
rebel-cjlee: Curious whehter the `is_prefill_step` field on `RBLNSchedulerOutput` could go away, and the runner just derive the phase from scheduler output instead:
```python
# add it to runner side
def _step_is_prefill(scheduler_output) -> bool:
    # prefill graph = more than one non-draft token per request
    return any(
        num_base_tokens(
            scheduler_output.num_scheduled_tokens,
            scheduler_output.scheduled_spec_decode_tokens,
            req_id,
        ) > 1
        for req_id in scheduler_output.num_scheduled_tokens
    )
```
A few things I like about it:
- `scheduler_output` is identical on every rank, so this stays rank-consistent without a stamped field. The per-rank divergence we're avoiding doesn't come back.
- It keys off query shape (non-draft tokens per request), which is what RBLN actually picks the compiled graph on, so the classification matches the graph directly.
- No new field on `RBLNSchedulerOutput`, which is thing that I would like to avoid.

I haven't fully checked the edges though. The one I'm unsure about is a single-token intermediate prefill chunk: shape-wise it looks like decode, so `is_intermediate_chunked_prefill` (= `is_prefill_step` AND `discard_mask`) would flip, and I'm not sure whether that case is even reachable. Could you take a look and see if this holds up?
Happy to be wrong on it.

rebel-cjlee: Also, we can replace `is_prefill` with it. Then, we can use new `is_prefill` in runner to determine the prefill phase.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-cjlee: This `num_new_tokens <= 0` was `== 0` before. I changed it to `<= 0` earlier because of speculative decoding: `num_computed_tokens` could get ahead of `num_tokens_with_spec`, so `num_new_tokens` went negative (a step with only drafts and no anchor token). `== 0` did not catch that negative, so I widened it to `<= 0` to skip it. Note 5 is the leftover of that.

Now `should_defer_spec_step` handles that same case earlier and defers it, so a negative `num_new_tokens` no longer reaches this line. Two things follow:
- Note 5 is **now stale**. It reads like this block still handles handles the spec overshoot, but `should_defer` does.
- If a negative ever does reach here (a case `should_defer` missed), `<= 0` skips it quietly every step, which turns into a silent hang, `== 0` lets if fall through to `allocate_slots` and fall loudly instead.

So I think we should put it back to `== 0` and drop Note 5.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2996,9 +3075,11 @@ def warmup_model(self) -> None:
             # 4. rejection sampler warmup
             self._warmup_sampler_decode_batches()
```
rebel-cjlee: `_warmup_sampler_decode_batches()` runs on every rank here, but `self.rejection_sampler` is only built on the last PP rank (in `__init__`). Its early return checks `speculative_config` / `num_spec_tokens` / pooling / the env flag, not the rank, and `num_spec_tokens` is set on all ranks. So on a non-last PP rank with spec decode, it skips the early return and then touches self.rejection_sampler, which does not exist there, so it fails with AttributeError at start-up.

`_dummy_sampler_run()` just above uses `self.sampler`, which is built on all ranks, so it does not crash, but it is still wasted work on non-last ranks.

The construction is already gated on the last rank, so the warm-up should match. Wrapping this block in if `get_pp_group().is_last_rank`: fixes the crash, drops the wasted work, and keeps construction and warm-up on the same rule.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -918,8 +955,14 @@ def _get_block_table(kv_cache_gid: int):
...
             if self.speculative_config and spec_decode_common_attn_metadata is None:
-                if isinstance(self.drafter, RBLNEagleProposer):
-                    if self.drafter.kv_cache_gid == kv_cache_gid:
+                # NOTE(RBLN): the drafter lives only on the last PP rank, so gate
+                # on the drafter object via getattr, not speculative_config (which
+                # every rank has). Here non-last ranks (drafter absent -> None)
+                # take the else branch, whose spec_decode_common_attn_metadata is
+                # unused -- only the last rank drafts.
+                drafter = getattr(self, "drafter", None)
+                if isinstance(drafter, RBLNEagleProposer):
```
rebel-cjlee: The `getattr(self, "drafter", None)` fallback repeats at several call sites because `self.drafter` is only assigned on the last PP rank in `__init__`. If we declare it up front (`self.draft: ... | None = None`) before the last-rank `if`, the attribute always exists and every site becomes a plan `self.drafter is not None` / `isinstance(...)`.

Same theme as the sampler warm-up (both are last-rank-only spec objects), but the fix differs: drafter just needs the up-front declaration since `getattr` already works, while the sampler warm-up must gate the call on the last rank (declaring `rejection_sampler = None` alone would not help, the warm-up still calls into it).

rebel-cjlee:
While looking into a hang I worked through the same spec+PP issues on a separate branch ([d1a0ee0](https://github.com/RBLN-SW/vllm-rbln/commit/d1a0ee04ee3ff98677371d2ea35926b959a2f993#diff-0728608c4c8fc124c724ef9835f6b1d43ca664e13ca9ced9b4f328c0dd819669)). It lines up closely with this PR, so I mostly used it to confirm the root cause. Sharing it as a reference, not a replacement.

Two things these that might be worth a look, both about last-rank-only spec objects (I left inline notes on these too):
- The sampler warm-up is gated on the last PP rank, so a non-last rank never calls into `rejection_sampler`, which isn't built there.
- `self.drafter` is declared as `None` up front in `__init__`, so the call sites drop the `getattr` and just check `self.drafter`.

Please feel free to take whatever seems useful, entirely at your discretion. I'd be glad to walk through any part of it if that would be helpful.

vllm_rbln/platform.py
rebel-ykchoi: Right — the line is the same on every rank (it just restates
`max_num_seqs // pp_size`), so one copy is all it's worth. Switched it to
`logger.info_once`.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-ykchoi: It holds up — adopted, and the field is gone from `RBLNSchedulerOutput`.

On the single-token edge: it cannot arrive as a prefill step. With one token left
`is_prefill()` already reports decode, which is where a 1-token prompt, a chunked
prefill's last token, a full remote-KV match and a full prefix-cache match all
land. A chunk can never exceed what is left, so the only disagreement possible is
a prefill handed exactly one token, and no clamp in `schedule()` cuts a chunk that
short (the one I can't rule out by construction is encoder-input truncation on a
multimodal request). So `is_intermediate_chunked_prefill` does not flip.

One deviation from your sketch: rather than `_step_is_prefill` on the runner side,
both views live in `v1/core/utils.py` — `is_prefill(request)` moved there next to
the new `step_is_prefill(scheduler_output)` and the `num_base_tokens` it is built
from. They have to agree, so the argument for why they do belongs in one place,
and the section comment there states the precondition and what a break would cost.
That also covers your follow-up: the runner property is `is_prefill` again,
stamped from `step_is_prefill(scheduler_output)`.

The agreement rests on those clamps rather than on anything the code enforces, so
the scheduler tests now check both views on every step of a run, including chunk
sizes that leave a single-token tail and the step that admits a request whose KV
arrived whole. I considered asserting it in `schedule()` instead, but that is
per-step work in the hot path for something the tests can pin.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-ykchoi: Agreed on both counts — restored `== 0` and dropped Note 5.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Right — took this from your `d1a0ee0`: both step-4 blocks are now inside `if get_pp_group().is_last_rank:`, so warm-up follows the same rule as construction.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Adopted from your `d1a0ee0` — declared up front as `... | None = None`, and the six read sites (including the `hasattr` in `load_model`) are now plain `self.drafter is not None` / `isinstance(...)` checks.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-cjlee: Looks good. The single-token tail cases are the ones that matter here, and the multimodal encoder-input truncation is out of scope, so I'm fine leaving that as the documented edge. Thanks for centralizing both views in one place.

rebel-cjlee (approved):

rebel-cjlee:
@rebel-ykchoi Can you try rebasing your branch on the latest `dev` and checking whether the tests are passed?

rebel-ykchoi:
@rebel-cjlee 
Heads-up from the latest `dev` merge: `013b14d6` added an assert to this block —

    if self.speculative_config and self.speculative_config.method == "medusa":
        assert isinstance(self.drafter, RBLNMedusaProposer)

I kept the `isinstance(self.drafter, ...)` form, so the merge shows that assert
removed. It covers both halves of the condition, and with the drafter declared up
front it is the safe one: on a non-last rank the drafter is None, so the assert
would fire there.

PR #901, opened 2026-08-07
feature(nixl): pipeline parallelism for prefill/decode KV-cache transfer

rebel-ykchoi:
> **Step 2 of 3** — split of #838 (*pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer*). This PR carries **section B** of #838: NIXL KV-cache transfer under PP. The PP-aware scheduler and runner it builds on are already on `dev`, so the diff here is only the NIXL changes.

#### Summary

Extend prefill→decode KV-cache transfer over NIXL to work when either side runs with pipeline parallelism. Before this, a reader would connect to producer stages it does not read from and index per-stage descriptors out of range. Each pipeline stage now advertises its own band of layers in the handshake, and a reader connects to — and builds descriptors for — only the producer stages whose layers fall within the band it owns.

#### What changed

**1. Per-stage layer-band handshake.** Each pipeline stage advertises its own band of layers in the handshake metadata; a reader connects to and pulls from only the producer stages whose layers fall within the band it owns. This covers both the **symmetric layout** (prefill and decode share a pipeline size, stage matched to stage) and the **fan-in layout** (a reader without pipeline parallelism aggregates several producer stages).

**2. Overlap-aware remote descriptor build.** Per-stage remote descriptors are built assuming a producer exposes no more regions than the reader owns — which holds only for the stages the reader actually reads from. Stages outside the reader's band (which can expose more regions when the layer count does not divide evenly across stages) are **skipped before that build**, so it never indexes out of range.

**3. Non-last-rank KV output via the two-phase path.** The KV-connector output for a non-last PP rank is surfaced through the runner's two-phase `sample_tokens()` path; `RBLNWorker.execute_model` returns `None` on non-last ranks instead of re-emitting the same finished send/recv notifications from an empty output, so every rank saves its own layers' KV exactly once.

**4. Unsupported combinations rejected at handshake.** TP together with PP, sliding-window attention together with PP, and unequal prefill/decode block sizes are rejected during the handshake. Transfers without pipeline parallelism keep the existing single-stage path unchanged.

**5. Handshake and notification robustness on the PP read path.** A dropped prefix-hit notification is logged and recorded (matching the transfer path in the same function) rather than surfacing raw; the handshake decode wraps its payload and agent-metadata errors and points a compatibility-hash mismatch at its opt-out config; the per-stage read-state registration computes its region ids once; and a read about to apply prefix caching asserts the local/remote/group block-id lengths agree, turning a group-count mismatch into a clear failure instead of an opaque index error.

**6. Host-staged save path re-synced with the base connector.** Two scheduler methods were near-copies of the base kept for one divergence each, so base behaviour added after those copies had been frozen out, and the two were entangled: the save bookkeeping lived in the copied metadata builder and the other method existed only to clear it. Now only the hook the base offers for the differing part is overridden — staging a prefill's blocks in a single save once every chunk has landed — and the rest is inherited. Two consequences a reviewer should look at: blocks are **accumulated as the chunks allocate them**, since a prompt spanning more than one block no longer arrives in one scheduler step and the previous one-shot assumption aborted the prefill engine; and a prefill that ends **stopped** rather than length-capped is now handed to the decode side instead of having its blocks freed with its KV never transferred. This path is used only when KV is staged through host memory.

#### Testing

- **Unit** — the chunked-prefill save accumulation and the finished-request outcomes for a stopped vs an aborted producer (`test_rbln_nixl_scheduler.py`); the NIXL per-stage layer-band handshake and overlap-skip read path (`test_rbln_nixl_handshake_pp.py`), handshake metadata PP fields (`test_rbln_nixl_metadata.py`), the PP-aware connector scheduler and worker (`test_rbln_nixl_scheduler.py`, `test_rbln_nixl_worker.py`).
- `pre-commit run --hook-stage manual` clean; connector/worker unit suites green.

#### Risk / compatibility

- Unsupported PP combinations fail fast at the handshake rather than mis-transferring.
- Behaviour change: a producer prefill finishing on a stop condition now transfers its KV instead of being dropped from the transfer, so the decode side no longer has to re-prefill it.

#### Affected modules

NIXL KV connector, Worker.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

### DIFF
--- a/vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
+++ b/vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
@@ -1116,6 +1116,11 @@ class RblnNixlConnectorWorker(NixlPullConnectorWorker):
             " (full prefix hit, notif only)" if prefix_hit else "",
         )
 
+        # Publish once and fail as one request: a handle visible while a later
+        # stage is still being prepped settles the request early, and the stages
+        # landing after that are a second completion with its metadata gone.
+        # A failed stage means recompute, so in-flight handles are released.
+        handles: list[int] = []
         for global_rank in self._overlapping_ranks[engine_id]:
             if prefix_hit:
                 agent_name = self._remote_agents[engine_id][global_rank]
@@ -1162,7 +1167,7 @@ class RblnNixlConnectorWorker(NixlPullConnectorWorker):
                     notif_msg=notif_id,
                 )
                 self.nixl_wrapper.transfer(handle)
-                self._recving_transfers[req_id].append(handle)
+                handles.append(handle)
             except Exception as e:
                 self._log_failure(
                     failure_type="transfer_setup_failed",
@@ -1172,4 +1177,10 @@ class RblnNixlConnectorWorker(NixlPullConnectorWorker):
                     dst_engine_id=engine_id,
                     remote_pp_rank=global_rank,
                 )
+                for submitted in handles:
+                    self.nixl_wrapper.release_xfer_handle(submitted)
                 self._handle_failed_transfer(req_id, handle)
+                return
+
+        if handles:
+            self._recving_transfers[req_id].extend(handles)
--- a/vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
+++ b/vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
@@ -12,12 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+from collections import defaultdict
 from typing import TYPE_CHECKING, Any
 
 import msgspec
 import numpy as np
 import rebel
 import torch
+import zmq
 from rebel.kv_cache import aligned_tensor
 from vllm.config import VllmConfig
 from vllm.distributed.kv_transfer.kv_connector.utils import (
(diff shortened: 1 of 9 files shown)


vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
```diff
@@ -663,3 +696,480 @@ def _compute_desc_ids(
...
+            except Exception as e:
+                self._log_failure(
+                    failure_type="transfer_setup_failed",
+                    req_id=req_id,
+                    msg="Marking blocks as invalid",
+                    error=e,
+                    dst_engine_id=engine_id,
+                    remote_pp_rank=global_rank,
+                )
+                self._handle_failed_transfer(req_id, handle)
```
rebel-cjlee, line 1182: On a stage failure here, `_handle_failed_transfer` puts the req in `_failed_recv_reqs`, but the loop keeps going and any handles from earlier successful stages stay in `_recving_trasfers[req_id]`. So the same req ends up in both places.

Then `get_finished` reports it twice: once from `_failed_recv_reqs` (which also pop `_recving_metadata[req_id]`), and again later when those leftover handles finish. The second time, `assert meta is not None` fires because the metadata was already popped, so a recoverable transfer failure crashes the engine.

The base path is safe because it has one transfer per req;, so a failed handle never sits in `_recving_transfers`. The multi stage loop breaks that assumption.

Could you check it?

rebel-cjlee:
@rebel-ykchoi Looks good to me overall. Two notes on the tests. There is quite a lot of test code and some of it overlaps, for example the raise cases in `TestPpHandshakeFanout` and `TestPpConstraints` cover similar ground, so a small tidy-up there would help.

And it would be good to cover the stage-failure path in `_read_blocks_for_req`. That one goes with the inline comment above: right now a partial-stage failure can double-report the req and crash `get_finished`, so the test should land together with that fix rather than as coverage on its own.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/worker.py
rebel-ykchoi: Confirmed, fixed in `9099a3ee`. The stage loop now collects its handles and publishes them only once the whole set is submitted; on a failure it releases the ones already in flight and stops, so the request is reported failed exactly once -- which is what keeps the second completion (against metadata that is already gone) from happening at all. The prefix-hit branch is unchanged: it creates no handles.

`test_a_failed_stage_leaves_no_handles_behind` fails without the fix, and it runs the real failure path so it also pins the single failure report.

rebel-ykchoi:
Tidied up in `019b7bbf`: the fan-out over `pp_size` 1/2/3, the overlapping-stage registration over an even and an uneven layer split, and the shard region ids over two region layouts are now one parametrized test each -- three near-identical bodies varying one input became one.

The raise cases are not duplicates, though. Removing any one guard leaves exactly one failing test: `cross_layers` / `mamba` / `swa` for the three `_check_pp_constraints` branches, and the handshake's own SWA and remote-TP guards for the two in `TestPpHandshakeFanout`. They read alike because a guard test is three lines; I added a comment naming the guard each one covers.

rebel-cjlee (approved):

PR #902, opened 2026-08-07
fix(dp): decode memory sizing, batch padding, and idle-step alignment under data parallelism

rebel-ykchoi:
> **Step 3 of 3** — split of #838 (*pipeline parallelism across the RBLN scheduler, runner, and NIXL KV-cache transfer*). This PR carries **section C** of #838 (spec-decode memory sizing & DP batch padding) **plus fixes not in #838**: aligning the idle data-parallel step to the busy ranks' decode shape, dropping the step a fully drained group used to compute and discard, and deciding the draft's batch through the same interface as the model's. **Based on #900** (PP-aware scheduler & runner) and independent of the NIXL PR, so it can be reviewed in parallel — the diff here is only the data-parallel decode-shape changes.

#### Summary

Two data-parallel decode corrections: the KV-cache memory estimate, and the padded batch a rank runs when the ranks in a group are not all doing the same thing in a step. RBLN compiles a fixed set of decode graphs, so every step has to land on one of them; under data parallelism the fused-MoE all-reduce additionally requires every rank to agree on the token dimension. The first fix (already described in #838) sizes the estimate by the graphs warm-up actually compiles. The second (**new, added here**) decides the padded batch in one place and keeps an *idle* rank on the shape the busy ranks settled on.

#### What changed

**1. Memory sizing by the number of compiled decode runtimes.** Available-memory estimation reserves a fixed command-stream buffer per compiled decode graph. Warm-up compiles one decode graph per `(batch bucket, query length)`, and speculative decoding adds a second query length for every bucket, as does the specialized MoE-decode fallback. Those extra graphs are now counted, so the reservation matches warm-up and the KV-cache-block estimate does not over-reserve device memory.

**2. The step's padded batch is decided in `v1/worker/dp_utils.py`.** One rule, one result, one place:

- **`determine_batch_execution_and_padding`** — the rule, and it is pure: given this rank's counts and, under data parallelism, what every rank reported, it returns a `BatchDescriptor` (`num_reqs_padded`, `query_len`, `num_tokens_padded` — the padded dimensions a compiled graph is selected by) and the `BatchRoute` that produced it.
- **`synchronize_dp_ranks`** — the transport: each rank's `(num_tokens, num_reqs, is_prefill, is_idle)` packed into a single `int32` all-reduce and split back out as a `DPStatus`.
- **`coordinate_batch_across_dp`** — composes the two. The runner keeps a thin `_determine_batch_execution_and_padding` that skips the collective when there is no other rank to agree with, and the callers consume the descriptor's fields instead of re-deriving them.
- **`determine_draft_batch_execution_and_padding`** — the draft's rule, through the same interface: same `ShapeConfig`, same `BatchDescriptor`. It runs inside a step whose ranks have already reduced, so it reads the `DPStatus` that step published instead of reducing again — one all-reduce per step.

The module name and vocabulary follow upstream: `vllm/v1/worker/dp_utils.py` (`_run_ar`, `coordinate_batch_across_dp`), `BatchDescriptor` in `vllm/forward_context.py`, which plays the same role for cudagraph dispatch, and `_determine_batch_execution_and_padding`, which upstream carries on both the model runner and its proposers — the same "model and draft through one interface" shape as here.

Seven routes decide a step, and separating them is what the shapes need — two of them can land on the same numbers for different reasons, so the numbers alone cannot say which rule ran:

| route | when | result |
|---|---|---|
| `LOCAL` | single rank | local bucket; no token padding, no collective |
| `UNSPECIALIZED` | specialized MoE decode off | local bucket, token dimension `max_num_tokens` |
| `ANY_PREFILL` | a peer is still reading a prompt | prefilling rank keeps its own shape; decoding peers take the max bucket; token dimension `max_num_tokens` |
| `ALL_IDLE` | the group has drained | the step returns without staging or running anything |
| `QLEN_ASYM` | busy ranks disagree on query length | max bucket; token dimension `num_reqs_padded * max query length` |
| `AGREED` | the ordinary decode step | bucket of the busiest rank |
| `PINNED` | warm-up compiling a dictated dimension | local bucket, token dimension as pinned |

The metrics patch reads the pass phase from this decision, so it moves with the rename; the phase rule itself is unchanged here — a decode whose padded token dimension equals the prefill one ran the prefill-sized graph.

That rule now has two finer signals available, and they are complementary. The padded dimension collapses different causes into one phase: a decode padded because a peer is prefilling (`ANY_PREFILL`) and one padded because the config never specializes MoE decode (`UNSPECIALIZED`) — the first is rank skew, the second is static — while a step where the busy ranks disagree on query length (`QLEN_ASYM`) is indistinguishable from a plain decode even though it runs the top bucket's graph. The **route** separates those three. The **`DPStatus`** behind it then says *why* a route was taken and what it cost this rank: which peers were idle, which were prefilling, and how far this rank's own counts sat below the dimension it ended up running — i.e. how much of a padded step was this rank's work rather than padding. Read together they turn one phase counter into per-cause accounting, which is what a DP-skew question actually needs. Left out of this change: the phase enum and the patch belong to the metrics work; here the patch only follows the decision under its new name.

The route is reported and logged rather than branched on: callers run on the numbers. The one exception is `ALL_IDLE`, whose answer is not a shape (see 3). `PINNED` is the one route that dictates rather than decides — warm-up is symmetric, so it cannot reach the dimensions DP asymmetry asks for and pins them instead; the descriptor still carries the single value every caller reads.

**3. Idle data-parallel step aligned to the busy ranks' shape (new — not in #838).** When a rank has no work it still runs a dummy step so the group stays in lockstep. It is excluded from the decision and then adopts what the busy ranks decided, so it runs the graph they run instead of guessing a larger — sometimes never-compiled — one.

When *every* rank reports idle the step now runs nothing at all. Every rank reaches that answer from the same status, so all of them stop together: no peer is waiting inside a forward, and the work a step with no tokens owes the KV connector rides on the scheduler's own no-forward path, not on the dummy step. Previously each rank computed a full decode it then discarded.

**4. The draft decides through the same interface.** Its bounds stay its own — the token dimension follows the busiest rank rather than a length the group agreed on, and the drafting loop, which every rank runs one token per request deep whatever it staged, is a restatement of the status at that length rather than a second rule — so **the shapes it lands on are unchanged**. What goes away is the bookkeeping around them: the draft no longer holds a padding rule of its own, its callers no longer choose between a warm-up pin and a decided dimension after the fact, and the transport tensor is built where it is decided rather than defensively copied. An idle rank is the one case the status cannot speak for — it reported the minimal entry to stay out of the decision, then runs the warmed `1 + num_spec` decode length rather than the length the step decided for the model leg, since only the agreeing route hands an idle rank the busy ranks' length — and the rule reads that from the idle bit the status already carries instead of taking it from the caller.

**5. An idle rank skips a drafting pass nobody waits for.** A DP-idle rank runs the draft so that one whose forward joins the cross-DP all-gather stays in step with the busy ranks. Fused MoE is the only reader of the step's padded token dimension, so a dense draft joins nothing there: on an idle rank its first pass and its drafting loop both produce output the step throws away, and the whole chain is skipped instead. The draft is where that is known — the published status carries the idle bit, and whether its own forward has fused MoE it detects once at load time — so the runner does not have to be told what kind of draft it holds in order to call it.

#### Testing

- **Unit** — `test_dp_utils.py`: each route's exact output pinned for two bucket layouts (the same route yields different dimensions depending on the config's decode buckets), the contract that holds on every route (the token dimension is never smaller than the staged tokens, the padded request count never smaller than the actual one), the bit-packed all-reduce round-tripped with the inner all-gather intercepted so no real group is needed, and the draft's rule beside the model's — the same dimensions it pinned before the move, plus an idle rank running a length it never reported while still landing on the busy ranks' padding, and the drafting loop decided from the restated status rather than from a prefilling peer's fallback. `test_rbln_model_runner.py`: the descriptor a step resolves to on the single-rank and the coordinated path, that a dummy step stages the query length the decision returned, that a fully drained group stages and runs nothing, and that an idle step hands the draft the warmed decode length even when the step decided another one. `test_eagle.py`: fused MoE detected (and not) in the draft at load time, and the drafting chain skipped on an idle rank for a dense draft while a fused-MoE draft still runs it. `test_rbln_worker.py`: the draft-runtime memory count.
- `pre-commit run --all-files` clean; `tests/native/v1/worker`, `tests/native/v1/spec_decode` and `tests/native/patches` green (605 passed, 6 skipped).

#### Risk / compatibility

- A single-rank deployment takes the `LOCAL` route: no cross-rank collective, no token padding, and the same local bucket as before.
- The memory-sizing correction only ever reserves *fewer* blocks (it removes an over-reservation).
- The routes that change a shape are reachable only under data parallelism; which one a step takes is in the runner's debug log, so an unexpected dimension can be traced back to the rule that chose it.

#### Affected modules

Runner, forward context, worker, spec-decode drafter, the metrics patch, and the new `v1/worker/dp_utils.py`.

---

* Related to #838

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)


vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -495,7 +496,9 @@ def dummy_run(
         num_reqs_padded, dp_padded, num_tokens_across_dp = (
             self._determine_draft_batch_padding(num_reqs, num_tokens, is_prefill)
         )
-        num_padded_tokens = override_padded or dp_padded
+        num_padded_tokens = (
+            override_padded if override_padded is not None else dp_padded
+        )
```
rebel-cjlee, line 511: I feel like we might not need this change. What do you think?

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -557,7 +560,9 @@ def dummy_run(
         num_reqs_padded, dp_padded, num_tokens_across_dp = (
             self._determine_draft_batch_padding(num_reqs, num_reqs, False)
         )
-        num_padded_tokens = override_padded or dp_padded
+        num_padded_tokens = (
+            override_padded if override_padded is not None else dp_padded
+        )
```
rebel-cjlee, line 580: Same with line 499~501.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Every rank contributes into its own slot and zero elsewhere, so the sum is a
+    gather. Runs on the CPU group: the values are tiny and the caller is on the
+    host path.
+    """
+    slots = [0] * dp_size
+    slots[dp_rank] = value
+    tensor = torch.tensor(slots, device="cpu", dtype=torch.int32)
+
+    from vllm.distributed.parallel_state import get_dp_group
```
rebel-cjlee: How about moving this import in the module level?

rebel-cjlee (commented):
There are quite a few conflicts due to PR #945. It seems like many code changes are needed. Could I review this again once the conflicts are resolved?
Also, the metrics part might need some cleanup. If necessary, I will update that part myself.

vllm_rbln/v1/spec_decode/eagle.py
rebel-ykchoi: Both of these lines are gone now. The warm-up pin is handed to the decision as pinned_num_tokens_padded, and the rule returns the token dimension already resolved, so there is nothing left to choose between at the call site:

```
batch_desc, num_tokens_across_dp = determine_draft_batch_execution_and_padding(
    cfg=self.runner.shape_config,
    status=self.runner.dp_status,
    dp_rank=self.dp_rank,
    num_reqs=num_reqs,
    num_tokens=num_tokens,
    is_prefill=is_prefill,
    pinned_num_tokens_padded=override_padded,
)
num_reqs_padded = batch_desc.num_reqs_padded
num_padded_tokens = batch_desc.num_tokens_padded
```

The padded dimensions the draft lands on are unchanged.

vllm_rbln/v1/spec_decode/eagle.py
rebel-ykchoi: Gone as well, for the same reason — the drafting loop reaches the rule the same way, with first_pass=False.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Moved.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -284,16 +284,32 @@ def load_model(self):
...
         assert "rbln" in device_name
 
-        specialized_moe_decode = int(self.model_runner.specialized_moe_decode)
+        has_specialized_moe_decode = self.model_runner.specialized_moe_decode
         decode_batch_buckets_count = (
             self.model_runner.bucketing_manager.decode_batch_buckets_count
         )
 
-        num_runtimes = 1 + decode_batch_buckets_count + specialized_moe_decode
+        spec_enabled = getattr(self, "speculative_config", None) is not None
```
rebel-cjlee: WorkBase always has `specualtive_config` field as `SpeculativeConfig | None`. I don't think `getattr` is needed.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -428,7 +444,10 @@ def determine_available_memory(self) -> int:
                 n_model_bytes=n_model_bytes,
             )
 
+            # Draft runtimes: one per bucket, plus the specialized-MoE fallback.
             num_draft_runtimes = 1 + decode_batch_buckets_count
+            if has_specialized_moe_decode:
+                num_draft_runtimes += 1
```
rebel-cjlee, line 453: It looks like the number of runtimes is still not accurate. For better accuracy, we'd probably need to separate them by spec method. However, since we haven't been calculating it in detail anyway, I think it's fine to leave it as is for now.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from enum import Enum
+
+import torch
+import torch.distributed as dist
+from vllm.distributed.parallel_state import get_dp_group
+
+
+class BatchRoute(str, Enum):
```
rebel-cjlee: The `str` mixin is unused. The only string use is `route.value` at `rbln_model_runner.py:2997`, and `'%s' % route` renders `BatchRoute.X` with or without it. Plain `Enum` is a no-op change here.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+    """Which rule decided this step's shape.
+
+    Not derivable from the numbers: two rules can land on the same
+    ``(num_reqs_padded, query_len, num_tokens_padded)`` -- an unspecialized DP
+    decode and a decode alongside a prefilling peer both give ``(8, 1,
+    max_num_tokens)`` -- so a log line or a test that only checked the values
+    could not tell a right-numbers-wrong-rule step from a correct one. Nothing
+    branches on it: callers run on the numbers and log the route.
```
rebel-cjlee: "Nothing branches on its" is no longer true. `rbln_model_runner.py:2136` branches. It also points at a design seem: members are labels ("the shape was decided this way") while `ALL_IDEL` is control flow ("there is no shape").

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+  testable without a runner.
+- ``coordinate_batch_across_dp``: the two together, for callers under DP.
+- ``determine_draft_batch_execution_and_padding``: the draft's rule. Its shapes
+  follow different bounds, but it runs inside a step that already agreed, so it
+  reads the published status instead of reducing again.
+
+``(num_reqs, query_len)`` are the free variables; tokens are their product. The
+result therefore always states ``query_len`` outright: callers stage from it and
+never divide ``num_tokens_padded`` to recover it.
+"""
```
rebel-cjlee, line 32: 18-20 - the ranks do not have to run the same graph. `QLEN_ASYM` hands each rank its own `query_len`, so with peers at 4 and 1 tokens one rank runs `(4, 4)` and the other `(4, 1)` and nothing hangs. Only `num_tokens_padded` is common, because that is what the fused-MoE all-gather sizes off. What actually has to hold is two things; `num_tokens_padded` is the same on every rank, and each rank's `num_reqs_padded * query_len` fits inside it. The second half is the one a caller can violate, so it is worth writing down.

22-31 - I'd drop the function list. Line 31 already has the wrong name. A hand-maintained list like this will keep drifting, and each function already carries the same explanation in its own docstring.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+        # rank because every rank sees the same status.
+        cheapest = cfg.decode_batch_buckets[0]
+        return (
+            BatchDescriptor(
+                num_reqs_padded=cheapest,
+                query_len=1,
+                num_tokens_padded=cheapest,
+            ),
+            BatchRoute.ALL_IDLE,
+        )
```
rebel-cjlee: This descriptor is never read: the only caller returns at `rbln_model_runner.py:2136` before touching `batch_desc`. `not busy` is exactly `all(status.is_idle)`, which the runner can read off the status this function already returns. That would avoid generating a descriptor nobody uses.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+    max_num_tokens: int
+    """Prefill-sized token dimension, used as the padding target."""
+    specialized_moe_decode: bool
+
+
+@dataclass(frozen=True)
+class BatchDescriptor:
+    """Describes the padded batch this rank runs, i.e. which compiled graph the
+    step dispatches to. Same role as upstream's ``BatchDescriptor``, with the
+    dimensions RBLN compiles for."""
```
rebel-cjlee, line 108: Could `route` just live on the descriptor as a field, instead of riding along as the second item of the tuple? It describes the batch this rank runs, so it feels like it belongs there.

If it did, this function would return one thing, and the two callers that write `_route` (`rbln_model_runner.py:1445, `patches/metrics.py:313`) wouldn't have to unpack a value they throw away. Tests would just read `desc.route`. Nothing is broken as is, so feel free to skip it. It's two dead unpacks and three shorter signatures.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2160,18 +2208,31 @@ def _dummy_run(
...
+                        num_padded_tokens=num_tokens_padded_override,
+                    )
+            else:
+                # DP-idle step: run the draft so one whose forward joins a DP
+                # all-gather stays in step with the busy ranks. Its length is the
+                # warmed decode one, not the length this step decided: only the
+                # agreeing route hands an idle rank the busy ranks' length, and on
+                # the others it would get its own 1, which warm-up skips for the
+                # draft.
+                self.drafter.dummy_run(num_reqs, 1 + self.num_spec_tokens, False)
```
rebel-cjlee: On a backfill forced step the busy ranks draft at query length 1, so `dp_utils.py:321-323` derives a token dimension of `num_reqs_padded * 1`. This line still runs the idle rank's draft at `1 + num_spec_tokens`, so it stages `num_reqs_padded * (1 + num_spec)` into a dimension sized 1. With a MoE draft that reaches `MoERunner.foward`, where `t > max_pad` skips the pad and the reshape to `R * max_pad` no longer lines up. A dense draft returns early at `eagle.py:491-500` and is unaffected.

base was consistent here only by accident: its idle rank reported itself busy at `1 + num_spec`, so the group's dimension never dropped below what this rank ran. Excluding idle from the decision removed that, and the hardcoded length wasn't adjusted to match.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2160,18 +2208,31 @@ def _dummy_run(
...
+                # TODO(RBLN): a decode step whose drafts the drafter-length skip
+                # zeroed runs the draft's first pass at query length 1, which
+                # this skips.
+                if is_prefill or query_len == 1 + self.num_spec_tokens:
+                    self.drafter.dummy_run(
+                        num_reqs,
+                        query_len,
+                        is_prefill,
+                        num_padded_tokens=num_tokens_padded_override,
+                    )
```
rebel-cjlee: This gate drops the draft's first pass at query length 1 from warm-up, but `rbln_scheduler.py:916-924` produces exactly that batch: when any scheduled decode request can't backfill inside its current KV block, it forces every request in the batch to one token and pops them out of `scheduled_spec_decode_tokens`. The runner then takes the `spec_decode_metadata is None` branch, and `propose()` runs its first pass at query length 1, on a shape warm-up no longer compiles. base called `drafter.dummy_run` on every `_dummy_run`, so it was covered.

The TODO above attributes this to the drafter length skip, but that path looks different: the zeroed drafts still go out as `num_spec` token ids and `rbln_scheduler.py:423` treats a non empty list as schedulable, so the next step stays at `1 + num_spec`. The backfill guard is the one that reaches query length 1, so the note points at the wrong producer.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -910,19 +925,21 @@ def _build_attention_metadata(
...
         max_query_len: int,
-        num_tokens_padded: int | None = None,
-        num_reqs_padded: int | None = None,
+        num_reqs_padded: int,
         logits_indices: torch.Tensor | None = None,
     ) -> tuple[PerLayerAttnMetadata, CommonAttentionMetadata | None]:
         """
         :return: tuple[attn_metadata, spec_decode_common_attn_metadata]
+
+        Unlike the base runner this takes no padded token count: the metadata
```
rebel-cjlee: "the base runner" reads two ways and one of them is false. If it means upstream's, fine. Naming upstream explicitly would remove the ambiguity.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2054,17 +2070,47 @@ def _get_nans_in_logits(
...
+
+        warmup=False (serving-time DP-idle step, execute_dummy_batch): this rank
+        has no real work. It must still join the cross-DP all-reduce so busy
+        peers don't hang, but must NOT drive the shape decision -- it contributes
+        a minimal num_reqs=1 entry (excluded from the shape decision via
+        is_idle), then adopts the busy-decided shape and runs the same compiled
+        graph the busy ranks run.
         """
         num_tokens = num_tokens_per_req * num_reqs
         assert num_tokens <= self.max_num_tokens
```
rebel-cjlee: `num_tokens` is reassigned at 2143 once the decided `query_len` is known, so on the idle adopt path this assert covers the requested count rather than what actually gets staged. It's safe today since that passes `num_reqs=1`, just nothing the check no longer guards the slice at line 2159.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2074,84 +2120,86 @@ def _dummy_run(
...
+                num_tokens,
+                is_idle,
+                pinned_num_tokens_padded=num_tokens_padded_override,
+            )
+        )
+        if route is BatchRoute.ALL_IDLE:
+            # Every rank reported idle, so every rank stops here and no collective
+            # inside the forward is left half-done. What the connector has to do
+            # when a step runs no tokens rides on the scheduler's own path.
+            return
```
rebel-cjlee: `execute_dummy_batch` never enters the KV connector lifecycle, so the sentence about what the connector has to do points at a path this code doesn't take.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+            4,
+        )
+        assert across.tolist() == [4, 1]
+
+    def test_an_idle_rank_may_run_a_length_it_did_not_report(self):
+        # An idle rank reported the minimal entry so it would not drive the
+        # decision, then runs the warmed decode length. Its own token count is the
+        # one that cannot be checked against the status; a busy rank's still is.
+        idle = _status(num_tokens=[1, 8], num_reqs=[1, 4], is_idle=[1, 0])
+        desc, _ = self._determine(idle, num_reqs=1, num_tokens=3)
```
rebel-cjlee, line 780: This pins a descriptor that can't actually run. `num_reqs_padded * query_len` is 4*3 = 12, but `num_tokens_padded` is 8, and 12 rows is what the stager allocates. The comment says the all gather buffer matches the peers, which is true, but the local tensor also has to fit inside it.

So this locks in the shape the idle draft path produces at `rbln_model_runner.py:2235`. If that gets fixed, this golden has to change with it.

tests/native/v1/spec_decode/test_eagle.py
```diff
@@ -587,17 +526,43 @@ def compute_logits(self, sample_hidden_states):
...
+        # is what lets an idle rank skip drafting altogether.
+        self._stub_super_load_model(monkeypatch)
+        proposer = make_eagle_proposer(num_speculative_tokens=1)
+        monkeypatch.setattr(
+            proposer.vllm_config.speculative_config, "enforce_eager", True
+        )
+        proposer.load_model(target_model=object())
+        assert proposer.draft_has_moe is False
+
+    def test_a_moe_draft_is_detected(self, monkeypatch):
```
rebel-cjlee, line 567: This never runs the real detection. `modules` is stubbed to a plain iterable at line 534, so `nn.Module.modules()` is never walked, and `Mock(spec=MoERunner)` sets `__class__` so `isinstance` passes without a real runner anywhere.

Worth guarding, because a false negative here means an idle rank skips a MoE draft and the busy ranks wait forever. Two lines would cover the structural part: `issubclass(MoERunner, nn.Module)` (true today through `PluggableLayer`) and `issubclass(RBLNMoERunner, MoERunner)`. If upstream ever stops making the runner a submodule, both tests here stay green.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+
+        assert desc.query_len >= 1
+        assert desc.num_reqs_padded >= 1
+        # The staged batch fits the compiled batch dimension.
+        assert desc.num_reqs_padded >= kwargs["num_reqs"]
+        # What this rank stages fits in the graph's token dimension. Which exact
+        # dimension each route picks is pinned by the golden cases above, so this
+        # stays the one property that has to hold everywhere.
+        if desc.num_tokens_padded is not None:
+            assert desc.num_tokens_padded >= kwargs["num_reqs"] * desc.query_len
```
rebel-cjlee: `num_reqs` is the count before padding, but what the stager allocates is `num_reqs_padded * query_len`, and that's also what the fused MoE gather pads to. So the check passes on descriptors that don't actually fit.

I ran the stronger form over every case here on both bucket layouts and it holds, so it looks like a safe swap.

One thing though, it won't catch the case that matters. This sweep only runs `determine_batch_execution_and_padding`, and the descriptor that breaks the contract is a draft one. Extending the sweep to `determine_draft_batch_execution_and_padding`, or asserting it in `BatchDescriptor.__post_init__`, would cover both sides.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+        assert desc.num_reqs_padded == 1
+        assert desc.query_len == 69
+
+    def test_single_dp_prefill_reports_num_reqs(self):
+        desc, route = _determine(_cfg(), num_reqs=1, num_tokens=69, is_prefill=True)
+        assert desc.num_reqs_padded == 1
+
+
+# (route, kwargs) reaching every route, for the route-independent contract.
+INVARIANT_CASES = [
```
rebel-cjlee, line 497: The comment says every route, but `PINNED` isn't in the list. Worth adding, since it's the only route where the token dimension comes from the caller instead of being derived, so it's the only one that can break the contract by construction.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+                num_reqs=[1, 1, 1, 1],
+                is_prefill=[0, 1, 0, 0],
+                is_idle=[1, 0, 0, 0],
+            ),
+        )
+        assert (desc.num_reqs_padded, desc.query_len, desc.num_tokens_padded) == (
+            8,
+            1,
+            512,
+        )
```
rebel-cjlee: I think `is_idle` does nothing on this route. It's first read at `dp_utils.py:202`, and `ANY_PREFILL` returns at line 187. I ran it both ways and got `(8, 1, 512)` either way, which is also exactly what line 221's asserts. So this reads as a copy of line 221 with one input the code never looks at.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+        assert (desc.num_reqs_padded, desc.num_tokens_padded) == (1, 4)
+
+
+class TestPrefillIsNormalized:
+    # D1: num_reqs_padded == num_reqs on a prefill step, whatever the route. Today
+    # ANY_PREFILL reports the top bucket instead and the caller overrides it back;
+    # the decision owns it now, so the caller-side override can go.
+    @pytest.mark.parametrize("specialized", [True, False])
+    @pytest.mark.parametrize("buckets", [SINGLE, LADDER])
+    def test_prefill_reports_num_reqs(self, specialized, buckets):
```
rebel-cjlee, line 476: The `LADDER` half of this can't fail. With `num_reqs=1`, `find_bucket(1)` is 1 too, so passthrough and bucketing give the same answer. `SINGLE` does separate them (1 vs 8). `num_reqs=3` would separate them on `LADDER` as well (3 vs 4), so right now two of the four cases aren't doing anything.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+
+    @staticmethod
+    def _peer(num_tokens, num_reqs, *, prefill=False, idle=False):
+        token_bits, req_bits = 16, 13
+        value = num_tokens | (num_reqs << token_bits)
+        if prefill:
+            value |= 1 << (token_bits + req_bits)
+        if idle:
+            value |= 1 << (token_bits + req_bits + 1)
+        return value
```
rebel-cjlee: Nit: the bit layout is spelled out three times now, here, in `_encode` at line 625, and in `peer_encoded` at line 933. One independent encoder is the right call for a round trip test, three is a lot. `_encode` already takes `is_idle`, so these two could just call it.

tests/native/v1/worker/test_dp_utils.py
```diff
@@ -0,0 +1,975 @@
...
+
+    def test_places_self_at_own_rank_slot(self, fake_dp_collective):
+        # A non-zero dp_rank proves the source writes at index dp_rank, not slot
+        # 0; a slot-0 test could not tell placement from a constant.
+        fake_dp_collective({0: 11, 1: 22, 3: 33})
+        out = dp_utils._run_ar(7, dp_size=4, dp_rank=2)
+        assert out.cpu().tolist() == [11, 22, 7, 33]
+
+
+def test_synchronize_dp_ranks_bit_pack(monkeypatch):
```
rebel-cjlee: `TestDPStatus` has six round trip tests and not one of them asserts `status.is_idle`. This is the only place that does, and it sits outside the class, so reading `TestDPStatus` gives the impression the new bit isn't covered. (line 880 does push the bit through `coordinate_batch_across_dp`, so transport is exercised, it just isn't asserted as a field.

Could move in as two methods, one for the idle round trip and one for prefill and idle not bleeding into each other. The second half here mostly repeats line 800.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+
+    if not cfg.specialized_moe_decode:
+        return (
+            BatchDescriptor(
+                num_reqs_padded=local_num_reqs_padded,
+                query_len=query_len,
+                num_tokens_padded=cfg.max_num_tokens,
+            ),
+            BatchRoute.UNSPECIALIZED,
+        )
```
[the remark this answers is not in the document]
rebel-ykchoi: Right — the drained-group check sat behind the specialization split, so an
unspecialized DP group that had drained still staged and ran the target.

Fixed in 2e7fa6cd: the check now runs before that split, right after the
single-DP return, and the whole rule answers `(None, BatchRoute.ALL_IDLE)`
instead of naming a shape. There is no shape for a step no rank runs, so the
caller skips execution entirely rather than dispatching a graph whose output is
discarded.

    busy = [i for i, idle in enumerate(status.is_idle) if not idle]
    if not busy:
        return None, BatchRoute.ALL_IDLE

The runner returns early on `None`, and `execute_model` asserts it is non-None
so a real step can never reach the model without a decided shape.

Coverage: `TestDrainedGroup::test_answers_with_no_shape` is parametrized on
`specialized=[True, False]`, so the unspecialized case you named fails without
this change. The runner-side skip is covered in
`test_rbln_model_runner.py`, also parametrized on `specialized`.

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,443 @@
...
+
+    num_tokens_padded = cfg.max_num_tokens
+    if cfg.specialized_moe_decode and not is_prefill:
+        if any_prefill:
+            num_reqs_padded = cfg.decode_batch_buckets[-1]
+        else:
+            num_reqs_padded = cfg.find_bucket(max(status.num_reqs))
+            num_tokens_padded = num_reqs_padded * max(
+                tokens // reqs for tokens, reqs in zip(tokens_per_rank, status.num_reqs)
+            )
```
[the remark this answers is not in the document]
rebel-ykchoi: Correct, and the fix is on the runner side rather than in what the status
publishes.

The idle rank no longer picks its own draft length. It reports the minimal entry
so it cannot drive the decision, then stages the length that decision handed
back, and the draft leg now runs on that same length (bbc14472):

    self.drafter.dummy_run(num_reqs, query_len, False)

So on your example — busy ranks at one request, query length 1 — the idle rank
adopts query length 1 for both the target and the draft, and the pad the group
settled on describes every rank's draft. The hard-coded `1 + num_spec_tokens`
is gone; nothing derives an idle draft shape from a length the group never
agreed on.

Two things back it up:

- `BatchDescriptor.__post_init__` asserts `num_reqs_padded * query_len <=
  num_tokens_padded` on every descriptor, the draft's included. A pass whose
  rows exceed the dimension the group settled on now fails at construction
  instead of silently dropping rows in the fused-MoE dispatch.
- A dense draft has no all-gather to stay in step with, so it does not run at
  all on an idle step (cb463cc5) and answers its own pass from its own counts.
  Only a draft that actually joins the collective is kept in step.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2054,17 +2070,47 @@ def _get_nans_in_logits(
         except IndexError:
             return {}
 
+    def _stage_dummy_seq_lens(
+        self, num_reqs: int, num_tokens_per_req: int, is_prefill: bool
+    ) -> tuple[np.ndarray, int]:
```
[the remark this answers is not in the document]
rebel-ykchoi: Agreed. The second caller disappeared while this was being written — the
decision moved ahead of staging — and the helper stayed behind with a docstring
still claiming two.

Inlined in 65383d2c. The two arrays it wrote (`seq_lens_np`,
`num_tokens_no_spec`) are now visible where the dummy step is built, next to
the length the decision handed back.

It also improved the tests: the cases that stubbed the helper out now run the
staging they were standing in for.

tests/native/v1/worker/test_rbln_model_runner.py
```diff
@@ -322,70 +351,130 @@ def _runner(*, is_prefill, bucket=8):
...
+        )._determine_batch_execution_and_padding(3, 30)
+        assert batch_desc.num_tokens_padded is None
+        assert across is None
+
+
+class TestShapeConfigWiring:
+    # ShapeConfig is built once in __init__ and every route reads it. The tests
+    # around here hand-build one, which leaves the construction unpinned: a field
+    # wired to the wrong source would keep every one of them green.
+
```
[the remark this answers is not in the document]
rebel-ykchoi: The concern is right; the answer is to drive the field through the rule rather
than to drop the case. `ShapeConfig` is built once in `__init__` and every route
reads it, while the cases around it hand-build one — so a field wired to the
wrong source keeps all of them green, and nothing else in the suite would catch
it.

Rewritten in b8c59946: no assignment is compared to its source any more. Both
cases build a real runner and feed `runner.shape_config` to
`determine_batch_execution_and_padding`, then assert on the answer.

- `test_a_real_runners_shapes_reach_the_decision` picks `num_reqs=3`, which sits
  between two buckets, so a `find_bucket` wired to a pass-through returns 3 and
  the case fails. The token dimension and the specialization flag show up in the
  same answer.
- `test_the_bucket_list_reaches_the_routes_that_read_it` builds a runner with a
  peer and a bucket ladder, then reaches the two routes that read opposite ends
  of `decode_batch_buckets` — the top on a prefilling peer, the same list where
  the busy ranks disagree on a query length. A one-entry ladder would make the
  two ends indistinguishable, so the case asserts the ladder has more than one.

Both fail if a field is wired to the wrong source, and neither restates an
assignment.

vllm_rbln/v1/worker/rbln_worker.py
rebel-ykchoi: You're right. Fixed in 715af642:

    spec_enabled = self.speculative_config is not None

The `getattr` implied the attribute might be absent, which was never the case
and made the line read as a guard against something that cannot happen.

vllm_rbln/v1/worker/rbln_worker.py
rebel-ykchoi: Agreed — left as is. The count is unchanged in this PR; only the comment moved.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Right. `route.value` is the only string use and it reads the same either way,
so the mixin bought nothing while implying the members were meant to be handled
as strings.

    class BatchRoute(Enum):

Changed in 2e7fa6cd.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Both parts were right, and the second one is what fixed the first. The route was
supposed to be a label, and `ALL_IDLE` had been carrying a control-flow meaning
that the other members do not have — which is exactly why a caller had to branch
on it.

So the branch is gone from the route (2e7fa6cd). A drained group has no shape, so
the rule stops assembling one:

    busy = [i for i, idle in enumerate(status.is_idle) if not idle]
    if not busy:
        return None, BatchRoute.ALL_IDLE

The caller now branches on `batch_desc is None`, i.e. on the absence of the thing
it needs, and the route is back to being read only for the log line. `ALL_IDLE`
stays as the label that says which rule produced that absence — the log would
otherwise report a skipped step with no reason.

The docstrings say so now:

    Nothing branches on it: callers run on the numbers and log the route, and the
    one route that has no numbers says so by coming back without a descriptor.

    ALL_IDLE: Every rank drained: there is no shape to run, so the descriptor is
    None.

`execute_model` asserts the descriptor is non-None, so the absence cannot reach
the model on a real step.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Both applied in 715af642. The docstring stated an invariant that is stronger
than the code's, and the list had already drifted.

The two invariants now stand where the claim was:

    The ranks need not land on the *same* pair: a rank whose peers drafted while
    it did not runs its own query length and nothing hangs. Two things have to
    hold instead:

    - ``num_tokens_padded`` is the same on every rank, because that is what the
      fused-MoE all-gather sizes off.
    - each rank's ``num_reqs_padded * query_len`` fits inside it. A caller can
      break this one by running a pass at a length the ranks never settled on, so
      the rules state it where they return.

Your point that the second half is the one a caller can violate is also what
placed the check: `BatchDescriptor.__post_init__` asserts
`num_reqs_padded * query_len <= num_tokens_padded` on every descriptor, so a
pass at an unsettled length fails at construction instead of quietly losing rows
in the fused-MoE dispatch.

The function list is deleted. Each function carries the same explanation in its
own docstring, and a hand-maintained copy only goes stale — as it had.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Right that the descriptor was dead. I took the first half of the suggestion and
not the second.

The rule now returns without one (2e7fa6cd):

    busy = [i for i, idle in enumerate(status.is_idle) if not idle]
    if not busy:
        return None, BatchRoute.ALL_IDLE

What I did not do is move the test to the caller as `all(status.is_idle)`. The
rule owns the order the conditions are tried in — a prefilling peer is answered
before idleness is even looked at — and a caller re-deriving one of those
conditions from the raw status would be a second place that has to agree with
that order. Keeping the decision in one function is the reason it is a pure
function of the counts at all.

So the caller reads the absence rather than the condition:

    if batch_desc is None:
        return

`execute_model` asserts non-None, so a real step cannot proceed without a shape,
and the label stays on the route for the log line.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Thanks — I looked at it and kept the tuple, for a reason that only became clear
while resolving the drained-group thread.

A drained group has no shape, so the rule returns `(None, ALL_IDLE)`. The route
still has to be reported: the log line for a skipped step would otherwise say
nothing about why it was skipped. If the route lived on the descriptor, that case
would have to build a descriptor purely to carry the label — the object you
pointed out nobody reads, back again for a different reason.

The other half is what the descriptor is. It names a compiled graph: three
integers, compared by value, and frozen so that comparison means "the same
graph". The route is provenance of the decision, not a dimension of it. On the
descriptor, two batches that dispatch to the identical graph would compare
unequal because different rules produced them, and a test asserting on a shape
would have to state the rule as well.

The cost of keeping them apart is the two unpacks you counted, and one of them
does read the value — the runner logs it. I'd rather pay that than make the
shape carry its own history.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Confirmed, including the mechanism: `t > max_pad` skips the pad and the reshape
lands on rows that were never gathered. The hard-coded length was the defect.

Fixed in bbc14472. The idle rank now drafts at the length the step decided:

    self.drafter.dummy_run(num_reqs, query_len, False)

The reason it was hard-coded was warm-up, not the step: the draft leg only ran
when `is_prefill or query_len == 1 + num_spec_tokens`, so query length 1 was
never compiled for the draft and the idle rank could not have used the decided
length even where the route handed it one. That gate is gone — warm-up already
iterates `query_lens = [1, 1 + num_spec]` over every bucket, and the draft leg now
runs on each, so both decode lengths are compiled and the backfill step you
describe lands on a warmed graph.

The invariant is also checked rather than reasoned about now:
`BatchDescriptor.__post_init__` asserts `num_reqs_padded * query_len <=
num_tokens_padded` on every descriptor the draft builds, so a pass staged into a
smaller dimension fails at construction instead of reaching `MoERunner.forward`.

And your last sentence pointed at a second thing: a dense draft has no collective
to stay in step with, so it no longer runs on an idle step at all (cb463cc5).

Coverage in `TestDummyRunDraftParticipation`:
`test_idle_draft_runs_the_decided_length` and
`test_warmup_compiles_the_draft_at_every_query_length`.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Both correct, and I verified the producer you name: when a scheduled decode
request cannot backfill inside its current KV block the guard forces every
request in the batch to one token and pops them out of
`scheduled_spec_decode_tokens`, so the runner takes the
`spec_decode_metadata is None` branch and `propose()` runs its first pass at
query length 1.

The gate is removed (bbc14472). Warm-up already iterates
`query_lens = [1, 1 + num_spec]` over every decode bucket; the draft leg now runs
on each of them, so that first pass lands on a compiled shape.

The TODO went with it. Your reading of why it was wrong holds: the zeroed drafts
still go out as `num_spec` token ids and `rbln_scheduler.py:423` treats a
non-empty list as schedulable, so that path stays at `1 + num_spec` and never
reaches query length 1. Attributing the gap to it would have sent the next reader
to the wrong place.

`test_warmup_compiles_the_draft_at_every_query_length` fails if the gate comes
back.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Right, "base" could be read as this repository's own, which would make the
sentence false. Named explicitly in 715af642:

    Unlike upstream's runner this takes no padded token count: the metadata
    handed to the drafter is built from the unpadded counts here, where upstream
    builds it padded and unpads it again on the way out. The padded request count
    is still needed -- the decode branch of the builder pads the batch to it.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Right — the assert ran before the decision and then the value it checked was
replaced, so it was guarding a number the slices never used.

Moved below the decision in 715af642, so it checks the length that is actually
staged:

    query_len = batch_desc.query_len

    num_scheduled_tokens = np.array([query_len] * num_reqs, dtype=np.int32)
    num_tokens = int(num_scheduled_tokens.sum())
    # The decided length, not the requested one, is what the buffers below are
    # sliced to.
    assert num_tokens <= self.max_num_tokens

The buffer slices below it are now covered by the same check they depend on.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: Correct — that sentence described a path this function does not take. Removed in
715af642; the comment now says only what the early return is for:

    if batch_desc is None:
        # A drained group: every rank read the same status, so they all stop
        # here and no collective inside the forward is left half-done.
        return

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: The golden was wrong in exactly that way, and you were right that it was locking
in the idle-draft shape rather than describing a valid one.

The status is fixed so the case is consistent: the peer decodes four requests
three tokens deep (`num_tokens=[1, 12]`), which is the length the idle rank runs
too, so the answer is `(4, 3, 12)` and the rows fit the dimension (7863ebb3).

The larger point — that a golden can state a shape nothing could run — is now
enforced on the object rather than left to review. `BatchDescriptor.__post_init__`
asserts `num_reqs_padded * query_len <= num_tokens_padded` (5e04de72), so a case
that writes down `4 x 3` into 8 tokens fails at construction and cannot be
committed.

And the path itself changed: the idle rank drafts at the length the step decided,
not `1 + num_spec` (bbc14472), so there is no longer a shape here that only that
hard-coded length could produce.

I also added `test_a_busy_rank_at_one_token_stays_inside_the_bound` for the case
where the bound is tight — every rank forced to one token, so what a rank stages
is exactly the dimension they settle on.

tests/native/v1/spec_decode/test_eagle.py
rebel-ykchoi: Right on both counts: neither case walks a real module tree, so the structure the
detection rests on was untested — and the failure mode you name is the bad one,
since a fused-MoE draft reading as dense means an idle rank skips it and the busy
ranks wait on a collective nobody joins.

Added as you suggested (13ee19f7):

    def test_the_detection_rests_on_the_runner_being_a_submodule(self):
        assert issubclass(eagle_module.MoERunner, torch.nn.Module)
        assert issubclass(RBLNMoERunner, eagle_module.MoERunner)

The comment above it says what the two stubbed cases cannot see, so the next
reader does not take them for coverage of the walk itself.

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: Taken, in the stronger of the two forms you offered: the contract now lives on
the object (5e04de72).

    def __post_init__(self) -> None:
        assert self.num_tokens_padded is None or (
            self.num_reqs_padded * self.query_len <= self.num_tokens_padded
        ), (
            f"a batch of {self.num_reqs_padded} x {self.query_len} does not fit "
            f"the {self.num_tokens_padded} tokens the group settled on"
        )

That covers both rules and anything built by hand, including the draft
descriptor you point at — which is the one that was actually breaking it, so the
sweep alone would have stayed green while the bug shipped. It also caught a
golden in this file that stated a shape nothing could run.

Thanks for running the stronger form over both bucket layouts first; that told
me the assert could go in without weakening a route.

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: Added (7863ebb3):

    (
        "pinned",
        dict(
            num_reqs=2,
            num_tokens=2,
            status=_status(num_tokens=[2, 2], num_reqs=[2, 2]),
            pinned_num_tokens_padded=MAX_NUM_TOKENS,
        ),
    ),

Your reason is the right one — it is the only route where the dimension is handed
in rather than derived, so it is the only one a caller can break by construction.

One route stays out of the sweep by definition: `ALL_IDLE` returns no descriptor,
so there is no shape to check. It is covered in `TestDrainedGroup` instead.

The token half of the contract also moved to `BatchDescriptor.__post_init__`, so
every case here — pinned included — passes it just by being constructed.

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: Right — `ANY_PREFILL` returns before `is_idle` is read, so the case differed from
the one above it only in a field the rule never looks at. It could not fail
unless its neighbour failed too.

Removed, and replaced with one that can distinguish the rule (7863ebb3):

    def test_top_bucket_is_not_this_ranks_own_bucket(self):
        # The one above cannot tell the rule apart from "my own bucket": with a
        # single-bucket ladder both answers are 8. On a ladder this rank's own
        # count would fit bucket 1, and taking it would put the decoding ranks on
        # a graph the prefill-sized token dimension cannot fill.
        ...
        assert desc.num_reqs_padded == 8  # buckets[-1], not find_bucket(1) == 1

That is the property the route actually has, and on a single-bucket ladder it is
invisible — which is why the previous case could not have caught a `find_bucket`
in place of `decode_batch_buckets[-1]`.

An idle rank on the agreeing route does change the answer, and that one is
covered separately as `test_idle_rank_adopts_the_busy_query_len`.

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: Right — two of the four parametrizations could not fail. Changed to `num_reqs=3`
as you suggested (7863ebb3):

    def test_prefill_reports_num_reqs(self, specialized, buckets):
        # Three requests: on a ladder the bucket rule would round up to 4, so a
        # prefill that came back padded is visible here as well as on SINGLE.
        ...
        assert desc.num_reqs_padded == 3
        assert desc.query_len == 23

Now all four separate the two behaviours: 3 vs 4 on the ladder, 3 vs 8 on the
single-bucket layout.

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: Agreed — one independent encoder is the point, three is just three places to
update. Folded in 7863ebb3: `_encode` is now the only place the layout is
spelled out, and it carries the note about mirroring the source.

    def _encode(
        num_tokens: int, num_reqs: int, is_prefill: bool, is_idle: bool = False
    ) -> int:
        # Mirrors the source bit layout: tokens in bits 0..15, reqs in 16..28,
        # prefill flag at bit 29, idle flag at bit 30.

`_peer` stays as the keyword-argument spelling its four call sites read better
with, but it holds no layout of its own any more:

    def _peer(num_tokens, num_reqs, *, prefill=False, idle=False):
        return _encode(num_tokens, num_reqs, prefill, idle)

tests/native/v1/worker/test_dp_utils.py
rebel-ykchoi: Right — the new field was the one nobody asserted, and the class that owns the
round trips read as if it were not covered at all.

Moved in as the two methods you described (7863ebb3):

    def test_the_idle_flag_round_trips(self, fake_dp_collective):
        # An idle rank reports the minimum, so the flag is what tells it apart from
        # a rank genuinely running one request one token deep.
        ...
        assert status.is_idle == (True, False)
        assert status.num_tokens == (1, 8)
        assert status.num_reqs == (1, 2)
        assert status.num_tokens_across_dp.cpu().tolist() == [1, 8]

    def test_the_phase_and_idle_flags_do_not_bleed(self, fake_dp_collective):
        # Adjacent bits: a prefilling rank must not come back idle, and the peer's
        # flags must not land on this rank.
        ...
        assert (status.is_prefill, status.is_idle) == ((True, False), (False, True))

The second one is worth its own case rather than being folded into the earlier
prefill round trip: the two flags sit in adjacent bits, so the failure it guards
against is one leaking into the other, and the prefill case cannot see that with
the idle bit clear. The part that did repeat the earlier case is dropped.

tests/native/test_dp_metadata.py
```diff
@@ -1,189 +0,0 @@
...
-        "extra",
-        [
-            {"num_padded_tokens": 16},  # missing num_tokens_across_dp
-            {"num_tokens_across_dp": torch.tensor([8, 8, 8, 8], dtype=torch.int32)},
-        ],
-    )
-    def test_dp_requires_both_args(self, extra):
-        parallel_config = SimpleNamespace(data_parallel_size=4)
-        with pytest.raises(AssertionError):
-            RBLNDPMetadata.make(parallel_config, num_tokens=8, **extra)
```
rebel-cjlee, line 189: Deleting all of `test_dp_metadata.py` also dropped `TestMake`. The collective tests moving into `test_dp_utils.py` makes sense, but `RBLNDPMetadata.make()` is untouched by this PR and still lives, so its logic ends up with no real coverage. `test_forward_context.py` only monkeypatches `make()` out, Could we keep `TestMake`? (maybe move it into `test_forward_context.py`?, it's up to your choice)

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2889,54 +2935,67 @@ def _attach_kv_cache_bindings(
...
+                self.parallel_config.data_parallel_rank,
+                route.value,
+                int(self.is_prefill),
+                int(is_idle),
+                num_reqs,
+                num_tokens // num_reqs,
+                batch_desc.num_reqs_padded,
+                batch_desc.query_len,
+                batch_desc.num_tokens_padded,
+            )
```
rebel-cjlee, line 2996: If `debug_once` dedupes on the format string, this only ever fires once for the whole run, so a later route like `QLEN_ASYM` never shows up and the route tracing doesn't really work. Could you double check how `debug_once` dedupes here?

vllm_rbln/v1/worker/dp_utils.py
```diff
@@ -0,0 +1,463 @@
...
+        )
+
+    # The busy ranks agree, so decide from them and let an idle rank adopt the
+    # result -- its own (1 request, 1 token) must not shrink the graph.
+    num_reqs_padded = cfg.find_bucket(max(status.num_reqs[i] for i in busy))
+    return (
+        BatchDescriptor(
+            num_reqs_padded=num_reqs_padded,
+            query_len=max_query_len,
+            num_tokens_padded=num_reqs_padded * max_query_len,
```
rebel-cjlee, line 249: base bumped this (`max_tokens_per_req` in the old code) is `num_spec + 1` when `use_aux_hidden_state_outputs` and every rank is at qlen=1, with a comment about a compiler failure on that exact shape (it was in `_determine_batch_padding`, the `use_aux_hidden_state_outputs` branch, before this moved to `dp_utils`). That bump is gone here and I don't see it anywhere else. The draft side qlen=1 got covered, but this is the target model, and warm-up compiles the qlen=1 decode graph unconditionally.

Could you run a warm-up once with eagle3 aux + MoE + specialized_moe_decode to confirm it still compiles? If it does, dropping the workaround was safe and we can note that. If not, we probably need the bump back.

rebel-cjlee (approved):
Looks great. The revision addressed the feedback cleanly, and the idle draft and all-idle fixes look solid, the new descriptor invariant especially.

I left a few small comments, none blocking. The main one is the eagle3 aux bump that's gone now, worth quick warm-up with eagle3 aux + MoE + specialized_moe_decode to confirm it still compiles. The other two are minor, the dropped `TestMake` coverage and a `debug_once` question. Would appreciate a look at those, especially the warm-up, but this is good to go from my side.

vllm_rbln/v1/worker/dp_utils.py
rebel-cjlee: #998 removes the lines that I pointed..., just take a quick look please!

tests/native/test_dp_metadata.py
rebel-ykchoi: You're right, and nothing else covered it — the set_forward_context tests stub make() out, so its assertions and the non-DP single-slot tensor ran nowhere.

 The four cases are back in tests/native/test_forward_context.py, above the fixture that stubs the method, with a line at the stub saying where the real one runs. Checked they bite: three mutations in make() each turn exactly one of them red.

vllm_rbln/v1/worker/dp_utils.py
rebel-ykchoi: Took a look — #998 removes exactly the lines in question (the use_aux_hidden_state_outputs bump of max_tokens_per_req to num_spec_tokens + 1), so the two agree now. Rebased onto it; that hunk was the only conflict, resolved by keeping the function gone.

 I also ran the warm-up you asked for — eagle3 aux hidden states, a MoE target, specialized MoE decode on — and confirmed it compiles.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-ykchoi: The dedupe key includes the arguments, not just the format string: logger.debug_once forwards to _print_debug_once in vllm/logger.py, an @lru_cached function taking (logger, msg, *args). So a later route with different numbers is a new key and does print — confirmed against the installed logger, and both DP ranks' lines show up in a real run.

 One nuance: the cache is @lru_cache's default 128 entries, so with this many arguments it is once per distinct combination rather than once per process.

PR #903, opened 2026-08-07
fix(scheduler): restore the deferred block-free fence

rebel-ykchoi:
> Bump follow-up fix, independent of the #838 pipeline-parallelism split. Restores a scheduler fence that the 0.24 bump dropped when the copied `schedule()` was re-synced.

#### Summary

The RBLN scheduler copies upstream's `schedule()`. Re-syncing that copy during the 0.24 bump dropped the deferred block-free fence, so the blocks of a request aborted mid-step could return to the pool before that step's output was processed. With several decode microbatches in flight (pipeline parallelism) and a KV connector present, a connector load can then refill blocks that an in-flight step is still writing. This restores the fence.

#### What changed

**1. Deferred block-free fence restored.** `sched_step_seq` is advanced on every non-empty step and each request records the step it was last scheduled in. When a request is aborted, its blocks are held in `deferred_frees` instead of being freed immediately, and released only once the corresponding step's output has been processed. The fence engages only when more than one batch is in flight (`pipeline_parallel_size > 1`) and the instance is a KV consumer; otherwise freeing stays immediate, exactly as before.

#### Testing

- **Unit** — the fence holds an aborted request's blocks until the in-flight step drains, and freeing stays immediate without multiple in-flight batches (`test_rbln_scheduler.py::TestDeferredBlockFree`).
- `pre-commit run --hook-stage manual` clean (ruff, ruff-format, typos, mypy 3.10-3.13, license); scheduler unit suite green.

#### Risk / compatibility

- No behavior change for `pipeline_parallel_size == 1` or non-KV-consumer instances — freeing stays immediate.
- Mirrors the upstream fence position (just before `_update_after_schedule`).

#### Affected modules

Scheduler.

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

### DIFF
--- a/tests/native/v1/core/test_rbln_scheduler.py
+++ b/tests/native/v1/core/test_rbln_scheduler.py
@@ -1263,9 +1263,12 @@ class TestDeferredBlockFree:
             use_kv_connector=MockKVConfig(),
         )
         manager = sched.kv_cache_manager
-        # 32 tokens over a 16-token block with 8-token sub-blocks: the second
-        # block is partial once the step advances, which is what gets indexed.
-        request = create_requests(1, num_tokens=32)[0]
+        # 24 tokens over a 16-token block with 8-token sub-blocks: one full block
+        # plus an 8-token partial. Ordinary caching hashes the full block only, so
+        # the partial one's hash can come from nothing but this path -- a multiple
+        # of the block size would leave no partial block at all and the assert
+        # below would hold either way.
+        request = create_requests(1, num_tokens=24)[0]
         sched.add_request(request)
         sched.schedule()
         partial_block = manager.coordinator.get_blocks(request.request_id)[0][-1]
--- a/tests/native/v1/core/test_rbln_scheduler.py
+++ b/tests/native/v1/core/test_rbln_scheduler.py
@@ -26,6 +26,7 @@ from vllm.v1.request import RequestStatus
 
 from tests.native.v1.core.utils import (
     EOS_TOKEN_ID,
+    MockKVConfig,
     _drain,
     advance_to_decode,
     create_rbln_scheduler,
@@ -1173,3 +1174,55 @@ class TestDecodeCapMachinery:
             "discard() must fire exactly once for the preempted already-scheduled "
             f"decode, got {discard_calls}"
         )
+
+
+class TestDeferredBlockFree:
+    # The ``defer_block_free`` fence (``sched_step_seq``): blocks of a request
+    # aborted mid-step must not return to the pool until that step's output is
+    # processed, because with several batches in flight (PP) a connector load can
+    # refill blocks the in-flight step is still writing. On only when >1 batch is
+    # in flight AND the instance is a KV consumer. The fence lives in the copied
+    # schedule() and had no native unit coverage.
+
+    def test_deferred_free_fenced_by_inflight_step(self):
+        sched = create_rbln_scheduler(
+            pipeline_parallel_size=2, use_kv_connector=MockKVConfig()
+        )
+        assert sched.defer_block_free
+
+        request = create_requests(1)[0]
+        sched.add_request(request)
+        output = sched.schedule()
+        assert output.total_num_scheduled_tokens > 0
+        assert sched.sched_step_seq == 1
+        assert request.last_sched_seq == 1
+
+        block_pool = sched.kv_cache_manager.block_pool
+        free_before = block_pool.get_num_free_blocks()
+        sched.finish_requests(request.request_id, RequestStatus.FINISHED_ABORTED)
+        # Fenced: blocks stay out of the pool until the in-flight step drains.
+        assert sched.deferred_frees, "blocks must be fenced, not freed"
+        assert block_pool.get_num_free_blocks() == free_before
+
+        sched.update_from_output(output, make_model_runner_output(output, 0))
+        assert sched.processed_step_seq == 1
+        assert not sched.deferred_frees
+        assert block_pool.get_num_free_blocks() > free_before
+
+    def test_no_deferred_free_without_multiple_inflight_batches(self):
+        # Single batch in flight (no PP): freeing stays immediate even with a
+        # KV connector present.
+        sched = create_rbln_scheduler(use_kv_connector=MockKVConfig())
+        assert not sched.defer_block_free
+
+        request = create_requests(1)[0]
+        sched.add_request(request)
+        output = sched.schedule()
+        assert sched.sched_step_seq == 0
+
+        block_pool = sched.kv_cache_manager.block_pool
+        free_before = block_pool.get_num_free_blocks()
+        sched.finish_requests(request.request_id, RequestStatus.FINISHED_ABORTED)
+        assert not sched.deferred_frees
+        assert block_pool.get_num_free_blocks() > free_before
+        sched.update_from_output(output, make_model_runner_output(output, 0))
--- a/tests/native/v1/core/test_rbln_scheduler.py
+++ b/tests/native/v1/core/test_rbln_scheduler.py
@@ -1210,8 +1210,9 @@ class TestDeferredBlockFree:
         assert block_pool.get_num_free_blocks() > free_before
 
     def test_no_deferred_free_without_multiple_inflight_batches(self):
-        # Single batch in flight (no PP): freeing stays immediate even with a
-        # KV connector present.
+        # A guard rather than a test of the fence itself: with a single batch in
+        # flight (no PP) the whole mechanism has to stay inert even though a KV
+        # connector is present, so freeing is immediate and neither counter moves.
         sched = create_rbln_scheduler(use_kv_connector=MockKVConfig())
         assert not sched.defer_block_free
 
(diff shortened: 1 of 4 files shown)


rebel-ykchoi:
@rebel-cjlee , @rebel-minkyu 
Could I get a review on this one? It should be a quick read — one restored line
plus its tests.

`RBLNScheduler.schedule()` was missing the `sched_step_seq` advance, so the
deferred block-free fence never deferred: a finished request's KV blocks could
return to the pool while an in-flight step was still writing them. The line is
back, identical to the base.

`defer_block_free` is only set under PP with a KV consumer, so it is a no-op
elsewhere; the tests cover both directions.

Thanks!

tests/native/v1/core/test_rbln_scheduler.py
```diff
@@ -941,3 +942,55 @@ def test_prefill_chunk_schedules_no_spec_tokens(self):
...
+        sched.add_request(request)
+        output = sched.schedule()
+        assert sched.sched_step_seq == 0
+
+        block_pool = sched.kv_cache_manager.block_pool
+        free_before = block_pool.get_num_free_blocks()
+        sched.finish_requests(request.request_id, RequestStatus.FINISHED_ABORTED)
+        assert not sched.deferred_frees
+        assert block_pool.get_num_free_blocks() > free_before
+        sched.update_from_output(output, make_model_runner_output(output, 0))
```
rebel-cjlee, line 1232: Two things about this test.

First, it passes with or without the fix. It does catch someone dropping `self.defer_block_free` from the new condition, so it works as a regression guard, but it isn't a test of this change. Is the guard what you meant it to be? If so, it's worth covering the `total_num_scheduled_tokens > 0` half too, since nothing here catches that one.

Second, the last line has no assertion after it, so all it checks is that the call doesn't raise. If that's on purpose please say so in a comment. Otherwise I'd drop it.

tests/native/v1/core/test_rbln_scheduler.py
```diff
@@ -941,3 +942,55 @@ def test_prefill_chunk_schedules_no_spec_tokens(self):
...
+
+    def test_deferred_free_fenced_by_inflight_step(self):
+        sched = create_rbln_scheduler(
+            pipeline_parallel_size=2, use_kv_connector=MockKVConfig()
+        )
+        assert sched.defer_block_free
+
+        request = create_requests(1)[0]
+        sched.add_request(request)
+        output = sched.schedule()
```
rebel-cjlee, line 1195: The `total_num_scheduled_tokens > 0` part of the new condition has no test. You can delete it and both new tests still pass, since neither one ever schedules an empty step.

It's the part I'd most want covered. The two counters are meant to move together, and `update_from_output` only bumps `processed_step_seq` when the step has tokens. So if an empty step bumped `sched_step_seq` here, `sched_step_seq` would run ahead and never get caught up. The fence would never clear and those blocks would never come back to the pool.

Three lines cover it:
```python
sched = create_rbln_scheduler(
    pipeline_parallel_size=2, use_kv_connector=MockKVConfig()
)
sched.schedule()  # nothing queued
assert sched.sched_step_seq == 0
```

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -958,6 +958,11 @@ def schedule(self, throttle_prefills: bool = False) -> RBLNSchedulerOutput:
             )
             scheduler_output.ec_connector_metadata = ec_meta
 
+        # Advance the fence only for non-empty steps (those that actually
+        # write KV and have their output processed later in update_from_output).
+        if self.defer_block_free and total_num_scheduled_tokens > 0:
```
rebel-cjlee, line 1049: Something I noticed while reading. The deferred branch calls `pop_blocks_for_free()`, but `RBLNKVCacheManager` only overrides `free()`, and that's where the sub block cleanup lives. So on the deferred path it looks like none of it runs.

If that's right, `_req_sub_hashes` keeps an entry for a request that's already gone, and the partial block drops out of the LRU. A leak rather than a crash. This branch was dead before this PR, so it would be new exposure.

This should show it, if you want to check:
```python
sched = create_rbln_scheduler(
    enable_prefix_caching=True,
    block_size=16,
    sub_block_size=8,
    pipeline_parallel_size=2,
    use_kv_connector=MockKVConfig(),
)
request = create_requests(1, num_tokens=32)[0]
sched.add_request(request)
output = sched.schedule()
sched.finish_requests(request.request_id, RequestStatus.FINISHED_ABORTED)
sched.update_from_output(output, make_model_runner_output(output, 0))
assert request.request_id not in sched.kv_cache_manager._req_sub_hashes
```

tests/native/v1/core/test_rbln_scheduler.py
rebel-ykchoi: Yes — a guard, and it says so now: with a single batch in flight the whole
mechanism has to stay inert even with a connector present. It is the only test
covering the `defer_block_free` half — dropping that clause fails this one alone.
The closing `update_from_output` asserts the inertness too (release counter still
0, nothing deferred) instead of only not raising, and the other half is covered by
a new empty-step test.

tests/native/v1/core/test_rbln_scheduler.py
rebel-ykchoi: Added, close to your three lines, plus an assert that `defer_block_free` is on so
a misconfigured scheduler cannot make it pass for the wrong reason. Checked it the
way you found the gap: dropping `total_num_scheduled_tokens > 0` fails that test
alone.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-ykchoi: Right, and your snippet reproduces it: `_req_sub_hashes` still held the aborted
request after the drain. Two details from digging in: `_pending_indexing` was
getting consumed by `do_pending_indexing()` (whose docstring assumes `free()` took
the entry), so what leaked was the hashes; and the partial block was indexed there
with `mark_cached=False`, which is why it fell out of the LRU.

`pop_blocks_for_free()` now runs the same finalization as `free()` — extracted to a
helper, called before the blocks leave the manager since indexing reads them
through the coordinator. Covered on both doors: a manager-level test next to the
`free()` ones, and a scheduler-level one for the fenced path.

tests/native/v1/core/test_rbln_scheduler.py
```diff
@@ -1173,3 +1174,105 @@ def spy(self, n=1):
...
+            enable_prefix_caching=True,
+            block_size=16,
+            sub_block_size=8,
+            pipeline_parallel_size=2,
+            use_kv_connector=MockKVConfig(),
+        )
+        manager = sched.kv_cache_manager
+        # 32 tokens over a 16-token block with 8-token sub-blocks: the second
+        # block is partial once the step advances, which is what gets indexed.
+        request = create_requests(1, num_tokens=32)[0]
```
rebel-cjlee: I think `num_tokens=32` makes the last assert vacuous. With `block_size=16` that's exactly two full blocks, so `remainder` is 0 and `_index_partial_block` skips the group, meaning the synthetic hash never gets assigned. The block still ends up with a `block_hash` becuase `cache_blocks` hashes both full blocks, so the assert passes either way.

`num_tokens=24` would match what the comment says. `remainder` becomes 8 so a partial block exists, and `num_full_blocks` is 1 so ordinary caching only touches the first block. That leaves the synthetic hash as the only source for the second one.

Easy check if you want, drop the `_finalize_sub_block_state` call and keep only that assert.

The other two asserts do fail without the fix, so the test still works, and the manager lever test covers the partial block properly with 12 tokens over 8. So it's just this line and the comment above it.

vllm_rbln/v1/core/rbln_kv_cache_manager.py
```diff
@@ -605,8 +608,23 @@ def free(self, request: Request) -> None:
...
+        sub-block state for the request."""
+        self._finalize_sub_block_state(request)
         super().free(request)
 
+    def pop_blocks_for_free(self, request: Request) -> list[KVCacheBlock]:
+        """Same finalization as ``free()`` on the deferred path.
+
+        The scheduler takes this one when an in-flight step may still write the
+        blocks, so it holds them and returns them to the pool later. The request
+        is gone either way, so its sub-block state has to be settled here too.
```
rebel-cjlee: `The request is gone either way` isn't quite right. `_preempt_request` also calls `_free_request_blocks`, so a preempted request lands here and then goes back on the waiting queue. Behavior looks fine since the state rebuilds on resume, so it's just the sentence, plus the same wording in the new manager test's comment.

rebel-cjlee (approved):

tests/native/v1/core/test_rbln_scheduler.py
rebel-ykchoi: Right, 32 fills the two blocks exactly, so the partial-block indexing skips the group and ordinary caching satisfies the assert on its own. Moved to 24: with the finalization dropped that assert now fails where it passed before. The comment says what the number is chosen for.

vllm_rbln/v1/core/rbln_kv_cache_manager.py
rebel-ykchoi: Correct — preemption releases the blocks too, and that request goes back on the waiting queue. Reworded the docstring and the test comment around the blocks leaving rather than the request ending.

PR #904, opened 2026-08-07
perf(compile): enable static output for decode model graphs

rebel-daeyang:
### 🚀 Summary of Changes

**Problem — a decoder model that ends in a CCL op pays a large cost without static output.**

Under TP the vocab-parallel logits are gathered across ranks, so the model's final output tensor is also the destination buffer of an `all_gather`. That one tensor is both a **user output and a CCL operand**.

Without static output the runtime allocates a fresh device buffer for that output on every run, so its device address can change. **Whenever the output lands on a new device address, the CCL op has to be re-mapped to it.** That mapping is a synchronizing collective across all ranks and is expensive — tens to hundreds of milliseconds, growing with the number of CCL operands and the rank count.

**Solution.** Pass `use_static_output=True` for the two model-side graphs. The runtime then reuses one device buffer per output index, the output address stops changing, and the re-mapping no longer happens after the initial one.

### Why the sampler is deliberately excluded

Static output is only safe while a graph's output is fully consumed before that same graph runs again.

* **Model forward / `compute_logits`** — logits are read by sampling inside the same step. `execute_model()` and `sample_tokens()` stay a synchronous pair (they hand off through `ExecuteModelState`), so step N+1's forward cannot overwrite step N's logits before step N's sampler has read them. Safe to enable.
* **Sampler graphs** — their token ids leave the step. Under asynchronous scheduling the `.tolist()` that would force the D2H can be deferred past the next forward, which would then overwrite the shared buffer and silently corrupt results. Left off.

So the decision is per graph, not a global switch: no env var and no config toggle, and the unsafe combination cannot be selected by accident.

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

### DIFF
--- a/vllm_rbln/compilation/compiler.py
+++ b/vllm_rbln/compilation/compiler.py
@@ -83,6 +83,7 @@ def compile(
     global_device_id: int | None = None,
     use_cache: bool = True,
     cache_dir: str = "",
+    use_static_output: bool = False,
 ) -> CompiledTarget:
     _ensure_torch_dynamo_configured()
 
@@ -106,6 +107,8 @@ def compile(
     set_option("mode", mode)
     set_option("use_global_ctx", use_global_ctx)
     set_option("global_device_id", global_device_id)
+    if use_static_output:
+        options["use_static_output"] = True
     if use_cache and not envs.VLLM_DISABLE_COMPILE_CACHE:
         set_option("cache_dir", cache_dir or os.path.join(envs.VLLM_CACHE_ROOT, "rbln"))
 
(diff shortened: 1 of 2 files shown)


vllm_rbln/compilation/compiler.py
```diff
@@ -106,6 +107,8 @@ def set_option(key: str, value: Any) -> None:
     set_option("mode", mode)
     set_option("use_global_ctx", use_global_ctx)
     set_option("global_device_id", global_device_id)
+    if use_static_output:
+        options["use_static_output"] = True
```
rebel-cjlee: Can't we just pass it even when it's `False`?
If so, `set_option("use_static_output", use_static_output)` would be simpler.

rebel-daeyang: Great, I applied the feedback.

rebel-jonghewk (approved):
lgtm

rebel-cjlee (approved):

PR #907, opened 2026-08-09
other(test): fix native DP/spec-decode specs and narrow env scrubbing

rebel-cjlee:
### 🚀 Summary of Changes

Three test-suite fixes. No production code changes.

#### Add `arctic-inference` to the `test` extra (`pyproject.toml`, `uv.lock`)

vLLM's `suffix` speculative decoding requires it at runtime; without it the suffix case in `test_spec_decode_e2e.py` cannot run.

#### Fix test specs (`test/native/`)

- `test_dp_e2e.py`: `gpt-oss-20b` at `data_parallel_size=4` needs `enable_expert_parallel=True` to build a valid engine config.
- `test_dp_specs.py`: check the platform rules against `spec.model` instead of a hardcoded model, so each DP spec is validated with the model it declares.
- `test_spec_decode_e2e.py`: set `num_gpu_block_override=32` for ngram and suffix.

#### Scrub only `VLLM_RBLN_*` from the environment (`test/native/utils.py`)

`scrub_env` deleted every `RBLN_*` name, so compiler environments passed on the command line were silently dropped. The suite owns vllm-rbln's knobs, not the host description or the compiler flags.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] ❓ Other (`other`): test-only - specs, test dependency, harness env handling.

---

### 🧪 How to Test

1. `uv sync --extra test`
2. `pytest tests/native/test_utils.py tests/native/test_envs.py -q` → 138 passed
3. `RBLN_VERBOSE=6 VLLM_RBLN_SORT_BATCH=1 pytest tests/native/test_utils.py --collect-only`
   → header reads `native: scrubbed VLLM_RBLN_SORT_BATCH`; `RBLN_VERBOSE` survives
4. `pytest tests/native/v1/spec_decode/test_spec_decode_e2e.py --model-compile -m model_compile`
5. `pytest tests/native/distributed/ --model-compile -m model_compile` (needs 4 NPUs)

---

---

### 📋 Checklist

---

### 💬 Notes

`NATIVE_ENV` still pins `RBLN_ROOT_IP` / `RBLN_LOCAL_IP`.

---

### DIFF
--- a/tests/native/distributed/test_dp_e2e.py
+++ b/tests/native/distributed/test_dp_e2e.py
@@ -31,7 +31,7 @@ from tests.native.utils import (
 DP_MODELS: list[CompileModelSpec] = [
     CompileModelSpec(
         "openai/gpt-oss-20b",
-        {"data_parallel_size": 4, "max_num_seqs": 2},
+        {"data_parallel_size": 4, "max_num_seqs": 2, "enable_expert_parallel": True},
         {"VLLM_RBLN_SUB_BLOCK_CACHE": "0", "VLLM_RBLN_DECODE_BATCH_BUCKET_LIMIT": "2"},
     ),
 ]
--- a/uv.lock
+++ b/uv.lock
@@ -150,6 +150,12 @@ wheels = [
 ]
 
 [[package]]
+name = "arctic-inference"
+version = "0.2.0"
+source = { registry = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/simple/" }
+sdist = { url = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/packages/arctic-inference/0.2.0/arctic_inference-0.2.0.tar.gz", hash = "sha256:242755ae3662527dd842b4c3bd0c921e5a969e270daff0c9436b412505af4c4d" }
+
+[[package]]
 name = "astor"
 version = "0.8.1"
 source = { registry = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/simple/" }
@@ -2908,6 +2914,7 @@ runtime = [
     { name = "rebel-compiler" },
 ]
 test = [
+    { name = "arctic-inference" },
     { name = "pytest" },
     { name = "pytest-asyncio" },
     { name = "pytest-cov" },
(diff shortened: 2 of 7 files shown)


tests/native/distributed/test_dp_e2e.py
```diff
@@ -31,7 +31,7 @@
 DP_MODELS: list[CompileModelSpec] = [
     CompileModelSpec(
         "openai/gpt-oss-20b",
-        {"data_parallel_size": 4, "max_num_seqs": 2},
+        {"data_parallel_size": 4, "max_num_seqs": 2, "enable_expert_parallel": True},
```
rebel-huijong, line 34: were we just launching 4 vllms? lol

uv.lock
```diff
@@ -149,6 +149,12 @@ wheels = [
     { url = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/packages/apache-tvm-ffi/0.1.12/apache_tvm_ffi-0.1.12-cp312-abi3-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl", hash = "sha256:dc0acd7eeb0e451d5e3f686af3ba0b495fdbf97b5b54cf9a0f770cdafe0e691a" },
 ]
 
+[[package]]
+name = "arctic-inference"
+version = "0.2.0"
+source = { registry = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/simple/" }
+sdist = { url = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/packages/arctic-inference/0.2.0/arctic_inference-0.2.0.tar.gz", hash = "sha256:242755ae3662527dd842b4c3bd0c921e5a969e270daff0c9436b412505af4c4d" }
+
```
rebel-huijong, line 157: @rebel-minhopark 
QQ, sorry if this is a basic question, but why do we download all these packages from mgmt instead of pypi.org?

rebel-huijong (approved):

uv.lock
rebel-minhopark: This index is proxy of pypi.org. Much faster than pypi.org in uDC.

PR #912, opened 2026-08-10
fix(optimum): reject block_size == max_model_len for Qwen3.5

rebel-seinpark:
### 🐛 Problem

Qwen3.5 (`Qwen3_5ForConditionalGeneration`) forces `attn_impl="flash_attn"` in
`get_param_qwen3_5` (its full_attention layers require it). But it delegates the
KV-cache partition length to `get_param_qwen2_vl`, which sets
`kvcache_partition_len = block_size` **only when `block_size != max_model_len`**.

So a `block_size == max_model_len` request reaches optimum-rbln as
`attn_impl="flash_attn"` **without** `kvcache_partition_len`. optimum-rbln then
(`modeling_attention_utils.py`):

1. silently substitutes `DEFAULT_FLASH_ATTN_PARTITION_LENGTH` (16384), discarding
   the requested `block_size`, and
2. fails `validate_attention_method` — flash attention needs
   `max_seq_len % partition == 0` **and** `max_seq_len // partition >= 2` — with an
   error that references **16384**, a value the caller never set.

Net effect: the requested `block_size` is silently ignored (or the compile fails
with a confusing message). `block_size == max_model_len` is simply not a valid
flash-attention configuration.

### 🔧 Fix

`get_param_qwen3_5` now rejects, up front and in terms of `block_size`, any
`block_size` that does not evenly partition `max_model_len` into at least 2 parts:

```python
if max_model_len // block_size < 2:
    raise ValueError(... "Set block_size < max_model_len ...")
```

Once the guard passes, `block_size != max_model_len` holds, so the base builder
sets `kvcache_partition_len = block_size` and the requested value is what actually
gets compiled. optimum-rbln already enforces the 1024..32768 partition range, so
that is not duplicated here.

Adds NPU-free unit tests for `get_param_qwen3_5` (valid partition, the
2-partition minimum, and the three rejection cases) under
`tests/optimum/optimum_compile/converter/`.

### ✅ Type of Change

* [x] 🐛 Bug fix

### 📝 Note

The optimum smoke test `TestQwen35VL` (currently disabled on `dev`) should be
re-enabled with a valid `block_size < max_model_len` once Qwen3.5 compiles
end-to-end — tracked as a separate follow-up.


vllm_rbln/model_executor/models/optimum/compilation/multimodal/qwen.py
```diff
@@ -55,6 +55,13 @@ def get_param_qwen3_5(
 ) -> dict:
     # Qwen3.5's linear_attention layers use gated_delta_net; the full_attention
     # layers require flash attention, so force it here.
+    if max_model_len // block_size < 2:
+        raise ValueError(
+            f"Qwen3.5 uses flash attention, so block_size ({block_size}) must "
+            f"partition max_model_len ({max_model_len}): it must leave at least "
+            f"2 partitions (block_size <= max_model_len // 2). "
+            f"Set block_size < max_model_len (e.g. {max_model_len // 2})."
+        )
```
rebel-eunji, line 63: It works, but it looks awkward. And the error message does not align with if branch. 

```suggestion
    num_partitions, remainder = divmod(max_model_len, block_size)
    if num_partitions < 2 or remainder:
        raise ValueError(
            f"Qwen3.5 forces flash attention, which requires block_size ({block_size}) "
            f"to divide max_model_len ({max_model_len}) into at least 2 equal partitions. "
            f"Use a divisor of max_model_len that is at most {max_model_len // 2}."
        )
```

rebel-eunji (approved):

vllm_rbln/model_executor/models/optimum/compilation/multimodal/qwen.py
rebel-seinpark: I don't think checking remainder is needed. It will be check in optimum-rbln.

rebel-eunji: Then how about `block_size == max_model_len` logic?
If block_size == max_model_len, raise error.

rebel-seinpark: `num_partitions < 2` is already block this case. I will apply your error message. Thx!

PR #914, opened 2026-08-10
other: add VLLM_RBLN_NUM_HIDDEN_LAYERS to build only the first N layers

rebel-cjlee:
### 🚀 Summary of Changes

`VLLM_RBLN_NUM_HIDDEN_LAYERS=N` builds only the first N decoder layers and leaves the rest as `PPMissingLayer`. 0 keeps the whole model.

This replaces `hf_overrides={"num_hidden_layers": N}`, which needed a `load_weights` override per model to skip the extra checkpoint layers. Upstream's `is_pp_missing_parameter` already does that, so no loader patch is needed. The HF config is also left untouched, so layer indices still match the checkpoint and DeepSeek-V3's MTP block keeps pointing past the stack instead of at a real layer.

Under PP the N layers are spread across stages; `N < pp_size` is rejected because an empty stage caps every other stage at one KV block.

`tests/native` runs at 3 layers by default (`--num-hidden-layers`, 0 for the whole model), `hf_runner` included. Two comparisons had to loosen: at 3 layers the top two tokens can be indistinguishable in bf16, and greedy is an argmax. `check_outputs_almost_equal` accepts a divergence only when the two picks are the step's leading pair and each run rates them within 0.125 nats.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] ❓ Other (`other`): dev/test knob, not a serving path

---

### 🧪 How to Test

1. `pytest tests/native/patches/test_distributed_utils.py tests/native/test_output_checks.py` — 20 tests, no checkpoint, no device.
2. `pytest tests/native --model-compile` — runs at 4 layers; `--num-hidden-layers 0` restores the whole model.
3. `VLLM_RBLN_NUM_HIDDEN_LAYERS=3 vllm serve <model>` — 3 layers built, remaining weights skipped.

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/native/compile/models.py
+++ b/tests/native/compile/models.py
@@ -21,5 +21,7 @@ from tests.native.model_specs import CompileModelSpec
 MODELS: list[CompileModelSpec] = [
     CompileModelSpec("Qwen/Qwen3-30B-A3B", {"tensor_parallel_size": 8}),
     CompileModelSpec("Qwen/Qwen1.5-MoE-A2.7B", {"tensor_parallel_size": 8}),
-    CompileModelSpec("openai/gpt-oss-20b", {"tensor_parallel_size": 8}),
+    CompileModelSpec(
+        "openai/gpt-oss-20b", {"tensor_parallel_size": 8}, num_hidden_layers=4
+    ),
 ]
--- a/tests/native/conftest.py
+++ b/tests/native/conftest.py
@@ -415,6 +415,14 @@ def hf_runner():
     return HfRunner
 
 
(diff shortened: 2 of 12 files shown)


rebel-cjlee:
@rebel-huijong please take a look

rebel-huijong (approved):
LGTM

FYI, load format will be possible when this [torch-rbln PR](https://github.com/RBLN-SW/torch-rbln/pull/174) gets merged.

PR #915, opened 2026-08-11
refactor(sampler): encode greedy requests as top_k=1 instead of one-hot logits

rebel-eunji:
### 🚀 Summary

This PR changes how mixed greedy/random batches are handled by the RBLN sampling operators.

Greedy requests are now encoded as:

```text
top_k = 1
```

instead of approximating or constructing an argmax distribution by modifying the logits. `top_p` is irrelevant for a greedy row: vLLM already pins it to the neutral `1.0`, and an all-greedy batch elides it to `None`.

---

### 🔍 Background

Upstream vLLM handles greedy and random requests separately and merges their results with `torch.where`.

vllm-rbln processes a mixed batch through a single sampling operator:

- `rbln::top_k_top_p` for regular sampling
- `rbln::rejection_sample` for speculative decoding

Although an all-greedy batch can use `rbln::argmax`, these sampling operators do not provide a separate per-row greedy mode.

Additionally, vLLM normalizes a greedy request's sampling metadata to:

```text
top_k = vocab_size
top_p = 1.0
temperature = 0.0
```

The `top_k` and `top_p` values are therefore indistinguishable from an unfiltered random request. Only `temperature == 0.0` identifies a greedy row.

---

### ⚠️ Previous Behavior

The two sampling paths handled greedy rows differently.

#### Speculative decoding

The rejection sampler converted each greedy row's logits into an exact one-hot distribution by:

1. Finding the argmax
2. Allocating a tensor filled with `-inf`
3. Scattering `0.0` into the argmax position

This required vocabulary-sized operations for every greedy sampling step.

#### Regular sampling

The regular sampler replaced the zero temperature with `1e-3`, attempting to make the softmax distribution approach a one-hot distribution.

This was only an approximation. For example, when the top two logits differ by `1e-4`, scaling by `1e-3` leaves the argmax probability at approximately 52%.

---

### ✅ New Behavior

Greedy rows are identified using `temperature == 0.0` and passed to the sampling operators with `top_k=1`.

Restricting the candidate set to one token guarantees that only the argmax can be sampled, without modifying the logits.

`build_op_top_k_top_p` encodes the per-request `top_k` / `top_p` inputs as follows:

```text
batch      | filters (random rows)    | top_k             | top_p
-----------+--------------------------+-------------------+----------------
all greedy | -                        | tensor of 1s      | None
all random | none (pure multinomial)  | None              | None
           | top-k only               | metadata tensor   | None
           | top-p only               | None              | metadata tensor
           | top-k and top-p          | metadata tensor   | metadata tensor
mixed      | none (pure multinomial)  | 1 / vocab_size    | None
           | top-k only               | 1 / metadata      | None
           | top-p only               | 1 / vocab_size    | 1.0 / metadata
           | top-k and top-p          | 1 / metadata      | 1.0 / metadata
```

In the mixed rows, `a / b` reads: `a` at greedy rows, `b` at random rows. `top_k` is never `None` there, because a greedy row is encoded as `top_k == 1`.

`None` marks an unused filter: the compiler spells it as a scalar neutral, which lets the primitive drop the dead criterion at initgen time — a per-row tensor of the same neutral value cannot be dropped.

The all-greedy row is reached only from the rejection sampler — spec decode has no separate greedy op, while the regular sampler answers an all-greedy batch with the `rbln::argmax` fast path.

---

#### Shared sampling-parameter builder

Added `build_op_top_k_top_p()` in:

```text
vllm_rbln/v1/sample/ops/top_k_top_p.py
```

The helper:

- Passes an all-random batch's metadata pair through as-is, `None` included
- Materializes `top_k` for a mixed batch (`vocab_size` where no request sets it) and rewrites its greedy rows to `top_k=1`
- Encodes an all-greedy batch as a tensor of 1s with `top_p=None`
- Uses the logits batch size so padded Optimum rows receive valid values

Both the regular sampler and rejection sampler now use this helper.

#### Rejection sampler

- Removed the vocabulary-sized one-hot logits rewrite
- Leaves greedy logits unchanged during temperature processing
- Passes per-request `top_k` and `top_p` directly to `rbln::rejection_sample`

#### Regular sampler

- Uses `top_k=1` for greedy rows in mixed batches
- Replaces the previous `1e-3` temperature workaround with a neutral value of `1.0`
- Removes the dead `greedy_sampled` branch
- Keeps the existing `rbln::argmax` fast path for all-greedy batches

---

#### Correctness

- Greedy sampling in mixed batches is now exact rather than probabilistic.
- Greedy rows retain their real processed logits and log probabilities instead of receiving synthetic `-inf`/`0` values.

#### Performance

The rejection sampler no longer performs the following vocabulary-sized operations for greedy rows:

- `max`
- `full_like(-inf)`
- `scatter_`
- Row gather and scatter-back

The RBLN operator cost itself is unchanged; the improvement is in host-side preprocessing.

#### Compiled graph variants

An unused filter is deliberately passed as `None`, as agreed with the compiler side: the compiler spells it as a scalar neutral, which lets the primitive drop the dead criterion at initgen time. A per-row tensor of the same neutral value cannot be dropped.

The trade-off is one compiled graph per `None`/tensor combination — `(k, p)`, `(k, None)`, `(None, p)`, `(None, None)`. Warmup should exercise the combinations a deployment expects to serve.

---

### 🧪 Tests

Added coverage for:

- One test per row of the batch/filter table above, named after it (`test_all_greedy`, `test_all_random_pure_multinomial`, `test_mixed_top_k_only`, ...)
- Exact greedy behavior with nearly tied logits
- Rejection sampling with matching and mismatching draft tokens
- Requests with fewer draft tokens than the padded speculative length
- Padded Optimum metadata rows
- Metadata produced from real `SamplingParams` and `InputBatch`

Run:

```bash
pytest tests/torch_compile/unit/v1/sample/ -q
pytest tests/optimum/v1/worker/test_optimum_input_batch.py -q
```

Expected result: all tests pass.

---

N/A

---

### ✅ Type of Change

- [ ] 🚀 Release (`release`)
- [ ] ✨ Feature (`feature`)
- [ ] 🧬 Core engine changes (`core`)
- [x] 🛠 Bug fix (`fix`)
- [x] ⚙️ Performance improvement (`perf`)
- [ ] 📄 Documentation (`docs`)
- [ ] ❓ Other (`other`)

---

### 📋 Checklist

- [x] PR title follows the Conventional Commits format
- [x] Tests and expected results are documented
- [x] Relevant unit tests have been added
- [x] Documentation is not required for this internal sampling change

---

### 💬 Reviewer Notes

Please pay particular attention to:

1. The `temperature == 0.0` contract used to identify greedy rows
2. The `top_k == 1` greedy encoding (a greedy row's `top_p` is left neutral, never rewritten)
3. Padded Optimum batch rows receiving valid `top_k` values (`top_p` may be elided to `None`)
4. The `None` elision contract: an unused filter reaches the op as `None`, at the cost of one compiled graph per `None`/tensor combination


rebel-minkyu (approved):
Thank you for cleaning this up! FYI, tests/torch_compile will soon be removed, and replaced by tests/native. Those sampler tests will then be moved to tests/native or tests/optimum. cc. @rebel-cjlee

rebel-eunji:
@rebel-eunik I resovled the conflicts. Could you review this PR?

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -184,27 +180,27 @@ def sample(
...
         assert sampling_metadata.temperature is not None
 
         temperature = sampling_metadata.temperature
         if not sampling_metadata.all_random:
-            temperature = torch.where(
-                temperature < _SAMPLING_EPS, _SAMPLING_EPS, temperature
-            )
+            # A greedy row is answered by its top_k=1 filter (see
+            # build_op_top_k_top_p), so its temperature only needs to avoid the
+            # division by zero -- 1.0 leaves the row's logits untouched.
```
rebel-eunji:
```suggestion
```

rebel-eunik: nit: The clipping part looks a bit unnatural. Leaving a comment about it would help prevent confusion later.

rebel-eunik (approved):
Looks good to me, aside from the requested comment.

vllm_rbln/v1/sample/ops/top_k_top_p.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
```
rebel-jonghewk, line 1: can you add a `__init__.py` for ops pkg? the module maybe dropped from the wheel.

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: I agree with that. Let's resolve that in following PR.

vllm_rbln/v1/sample/ops/top_k_top_p.py
rebel-eunji: Great Idea! I applied it.

rebel-jonghewk (approved):
lgtm

PR #917, opened 2026-08-12
fix(mega-cache): key the bundle on VLLM_RBLN_NUM_HIDDEN_LAYERS

rebel-cjlee:
### 🚀 Summary of Changes

`VLLM_RBLN_NUM_HIDDEN_LAYERS` (#914) landed after the mega cache (#805) and was never added to either env partition, so it never reached the bundle key.

It truncates the decoder stack at model build time and is invisible to `vllm_config.compute_hash()`, so a truncated build and a full build key the same bundle. Classifying it as compile-relevant also unbreaks `test_partition_covers_all_declared_vars`, which has been failing on `dev` since #914.

Tests move to `tests/native`, trimmed from 88 to 64 cases:
| File | What it covers |
|---|---|
| `v1/worker/test_mega_cache.py` | signature composition + env keying, real-`VllmConfig` port sweep, path safety, save/load round trip, `warmup_model()` wiring |
| `test_envs.py` | every declared variable is classified compile / non-compile, and only once |
| `compilation/test_compiler.py` | `mega_cache_only` is set with `cache_dir` and dropped with it |

`tests/torch_compile/unit/v1/worker/test_mega_cache.py` is left in place.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)
* [x] ❓ Other (`other`): test migration to `tests/native`

---

### 🧪 How to Test

1. Run `pytest tests/native/v1/worker/test_mega_cache.py tests/native/test_envs.py tests/native/compilation/test_compiler.py`
2. Verify output: all pass (65 + 122 + 26). Full lane: `pytest tests/native` → 1080 passed, 20 skipped, 1 xfailed.
3. Edge case tested: reverting the `envs.py` line fails the partition guard and `test_graph_env_invalidates[VLLM_RBLN_NUM_HIDDEN_LAYERS]`; reordering load/save in `warmup_model()` or flush/serialize in `save()` fails the ordering assertions.

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-jonghewk (approved):
1. Agreed, and I don't have a better shape in mind than what you're describing. Would appreciate it if you took a look — happy to pick it up once you've thought it through.

2. I think it has to be the second one — merging dev before CI doesn't close the hole on its own. CI only re-runs on push/synchronize, so once dev moves nothing re-triggers and the stale green check stays.

What makes me hesitate is the cost: every PR pays an extra update + full CI run right before merging, and simultaneous merges have to re-run. How often do you think this actually bites? If it's rare, the post-merge dev CI may be enough.

PR #919, opened 2026-08-12
refactor(sampler): keep the sampled token ids in int32

rebel-eunji:
### 🚀 Summary of Changes

The compiler is dropping the int64 widening from the sampling ops: the primitives sample in int32, and SamplerOutput.sampled_token_ids is int32 too, so int64 only ever existed between those two points.

The plain sampler needs no change. Upstream vLLM already accounts for a backend whose sampling ops return int32 — FlashInfer is one — so it casts with `.long()` before the logprobs gather, the one place an index dtype is required, and narrows back with `.to(torch.int32)` afterwards. Both casts stay correct whichever dtype the op returns.

The rejection sampler is different: it allocates its own output buffer to keep the op's shape static, so that buffer's dtype is ours to choose. It becomes int32, which also lines up with TorchRejectionSamplerImpl and upstream's own buffer.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-jonghewk (approved):
lgtm

Issue #909, opened 2026-08-10
metrics_v2: move the runtime report capture to the E2E span and fix report aggregation


rebel-daeyang:
## Context

PR #899 moved the E2E metric boundary to the whole `execute_model` -> `sample_tokens` pass. The review suggested a follow-up: extend `capture_reports()` to that same E2E span and pull the model/sampler runtime metrics out of it, instead of keeping two separate spans. It also flagged that the report values may not be trustworthy on their own yet.

I measured both before writing any code. The direction is right, but the straightforward version of it does not work, and the report values need two separate fixes. This issue records what was measured and splits the work.

## How it was measured

Qwen2.5-0.5B-Instruct, 4 hidden layers, DP=1, one device per rank, `--max-num-seqs 4`, `--input-len 512`, `--max-tokens 32`, device-tensor path, with `VLLM_RBLN_METRICS=1` and `RBLN_RUNTIME_TIMER=1`.

```
export VLLM_RBLN_USE_DEVICE_TENSOR=1 VLLM_RBLN_SAMPLER=1 VLLM_RBLN_ENABLE_WARM_UP=1
export RBLN_WEIGHT_FREE=1 RBLN_RUNTIME_TIMER=1
export VLLM_RBLN_METRICS=1 VLLM_RBLN_METRICS_DIR=$PWD/metrics

vllm-rbln-exec --task r --model qwen2.5-0.5b --rsd 1 --dp 1 \
  --block-size 1024 --max-model-len 4096 --max-num-blocks 32 \
  --max-batched 1024 --batch 4 --num-hidden-layers 4 --input-len 512 \
  --num-prompts 4 --max-tokens 32 --logprobs 0 \
  --trust-remote-code --use-cached-models --cache-results --cache-ignore --skip-validation
```

Three runs: the current code with a temporary dump of each span's raw report list, the same with the inner captures disabled so a single E2E-level capture collects everything, and one more with `RBLN_RUNTIME_FORCE_SYNC=1` as a reference for what the device numbers should look like.

## Finding 1: nesting `capture_reports()` shadows the outer list

[...]

PR #920, opened 2026-08-12
other(metrics): instrument the runner from a patch, on the wall clock

rebel-cjlee:
### 🚀 Summary of Changes

Move the runner metrics out of the runner and into a patch, and time when with `perf_counter` instead of the rebel runtime report capture.

Metrics used to live on the serving path: a context built in the runner's `__init__`, two context managers around the model and sampler calls, two decorators, and a `print_stats()` call in the worker's shutdown. It is a debugging aid, so it now lives in `vllm_rbln/patches/metrics.py` and is installed only when `VLLM_RBLN_METRICS` is set. The runner and the worker no longer contain any metrics code.

Dropping `capture_reports()` also drops the host, device, ccl, and prepare columns, Three wall-clock ranges replace them:
1. the pass - `execute_model` through `sample_tokens`
2. the model call
3. the sampler call

Range 2 and 3 are replaced as one sum, so `MODEL + SAMPLE` has the same call count as `E2E` and the difference of the two means is the host overhead around the graphs.

Decode steps that ran the prefill-sized graph are now reported as their own phase instead of being averaged into `DECODE`, using the rule from #940. The report sections are `MODEL + SAMPLE (<phase>)` and `E2E (<phase>)` for `PREFILL`, `DECODE`, and `PADDED DECODE`.

---

* Related to #909 
* Related to #940 (the padded-decode phase rule comes from there)

---

### ✅ Type of Change

* [x] ❓ Other (`other`): debugging instrumentation, off unless `VLLM_RBLN_METRICS=1`

---

### 🧪 How to Test

1. Unit tests:

   ```
   uv run --no-sync pytest tests/native/patches/test_metrics.py
   ```

2. Real run:

   ```
   VLLM_RBLN_USE_VLLM_MODEL=1 VLLM_RBLN_METRICS=1 VLLM_RBLN_METRICS_DIR=$PWD/metrics <your run>
   ```

3. Check the report printed at shutdown, and `metrics.json` in that directory:
  * `MODEL + SAMPLE (<phase>)` and `E2E (<phase>)` sections appear, with equal call counts for the same phase
  * the `E2E` mean is at or above the `MODEL + SAMPLE` mean
  * warm-up is not counted - the counts match the number of real steps
  * with `VLLM_RBLN_METRICS` unset, no patch is applied and nothing is printed 

---

---

### 📋 Checklist

---

### 💬 Notes

The padded-decode rule (`num_tokens_padded == self.max_num_tokens`) is copied into the patch. `num_tokens_padded` is a local of `execute_model`, so the patch wraps `_determine_batch_padding` to read it. Changing how the runner pads means revisiting that line. #940 puts the same rule inline in the runner; whichever lands second should drop the duplicate.

With speculative decoding, the draft model is not in `MODEL + SAMPLE`. The drafter owns its own `model_executable`, and only the runner's is wrapped. The draft time is still inside `E2E`, so the gap between the two sections is larger than the host overhead in that case.

---


vllm_rbln/v1/worker/metrics_v2.py
```diff
@@ -122,146 +107,95 @@ def _add(a: int | None, b: int | None) -> int | None:
...
+        ctx, self._capture_ctx = self._capture_ctx, None
+        self._reports = None
+        if ctx is not None:
+            ctx.__exit__(None, None, None)
 
     def print_stats(self) -> None:
-        self._flush_pending()  # record the final step if it had no sampler
         sections: dict[str, Metrics] = {}
-        for phase, m in sorted(self._metrics.items(), key=lambda x: not x[0]):
-            sections["PREFILL + SAMPLE" if phase else "DECODE + SAMPLE"] = m
```
rebel-daeyang, line 241: The `PREFILL/DECODE + SAMPLE` latency sections are gone. Could we keep them?
Seeing `model+sampler` latency at the vllm-rbln level is useful on its own, and that number was a plain `perf_counter` span, independent of the capture. I think merging the capture window didn't strictly require dropping it.

rebel-cjlee: Just to make sure I follow what you're proposing: keep the whole-pass span measuring latency + runtime timings (as this PR does), and on top of that measure model and sampler latency each on their own span again?

If that's the ask, my goal here is to keep this debug-only instrumentation minimal, and reviving the model/sampler spans is basically the thing #909 (item C) set out to remove. If we keep them, there's little reason to merge the capture into a single whole-pass in the first place, which is what this PR is for, so it'd largely undo the PR.

Is that what you had in mind?

rebel-daeyang (approved):
LGTM.
I checked that `MODEL + SAMPLE` and `E2E` are both reported correctly, and the `PADDED DECODE` section shows up as expected.

rebel-jonghewk (approved):

PR #923, opened 2026-08-13
fix(patches): backport vllm#46865 — give every sub-connector the request's real blocks

rebel-jinhwan:
## Summary

Backports `vllm-project/vllm#46865` into `vllm_rbln/patches/`. Without that fix, **an
offload connector (LMCache) inside a `MultiConnector` can never store anything** — no
error, no warning, the cache just stays empty forever.

Upstream merged it to `main` as `2285cfc` on 2026-07-09 and it **ships in 0.26.0**. It is
in neither the vLLM we currently pin (0.22.0) nor 0.24.0 — verified by reading the
installed source.

## The problem

`update_state_after_alloc` is the scheduler-side hook vLLM calls right after it admits a
request and allocates its KV blocks. Its second argument is **all of the request's
blocks**. vLLM documents the hook only as "the connector uses this info to determine if a
load is needed", so `MultiConnector` hands *empty* blocks to every child that did not win
the lookup:

```python
chosen = self._requests_to_connector.get(request.request_id, -1)
empty_blocks = blocks.new_empty()
for i, c in enumerate(self._connectors):
    c.update_state_after_alloc(request, blocks if i == chosen else empty_blocks, ...)
```

But an offload connector uses that same argument to record **which physical blocks hold
the request's KV**, and its store path reads exactly that: `allocated_tokens` bounds how
much may be stored, and the recorded block ids become the `LoadStoreOp.block_ids` of the
STORE op. So blank blocks do not mean "will not load" — they mean **"cannot store"**.

Worse, `_requests_to_connector` is only populated for a connector whose
`get_num_new_matched_tokens` returned a non-zero count. On a prefill instance with a cold
offload cache **nobody** returns non-zero: the NIXL child is the producer and has nothing
to pull, and the offload child's cache is empty. `chosen` stays `-1`, so *every* child
gets empty blocks, and the deployment cannot break out of the resulting cycle:

```
cold cache -> lookup returns 0 -> not chosen -> empty blocks
  -> no block ids recorded -> allocated_tokens == 0 -> no STORE op
  -> cache stays cold
```

## The change

`patched_update_state_after_alloc` gives every child the real blocks and lets
`num_external_tokens` alone mark the chosen one — identical to the upstream PR:

```python
chosen_connector = self._requests_to_connector.get(request.request_id, -1)
for i, c in enumerate(self._connectors):
    c.update_state_after_alloc(
        request, blocks, num_external_tokens if i == chosen_connector else 0
    )
```

`condition` applies the patch only while the installed vLLM's copy of that method still
contains `empty_blocks`. It reads the source instead of comparing versions because
vllm-rbln pins vLLM through its own lock: a build can report a version below 0.26.0 while
sitting on a `main` snapshot that already carries the fix, and a version test would then
re-patch code that is already correct. Reading the source is exact either way.

A **`TODO(vllm>=0.26.0)`** in the module docstring and in the `register_patch` `reason`
names the removal trigger and everything to remove (the module, its registration in
`vllm_rbln/patches/__init__.py`, and the test).

## Verification

R100 PD-disagg (MiniMax-M2.5, DP4, NIXL + LMCache mp under a single `MultiConnector`),
same configuration either side, 24 requests at concurrency 4. Maximum values observed by
instrumenting `GetStoreMetadata`:

| field | before | after |
|---|---:|---:|
| `allocated_tokens` | **0** | **53,248** |
| `min_available_tokens` | 0 | 49,285 |
| `num_chunks` | 0 | **1** (152 of 176 calls `>= 1`) |
| `num_vllm_hit_tokens` | 0 | 24,576 |
| server `cached_objects` | **0** | **152** |
| server `l1_used_gb` | 0.00 (0.0%) | 18.41 (4.6%) |

Hit-token counts leaving zero is the direct evidence that the cycle is broken.

## Tests

Three cases added in `tests/native/patches/test_multi_connector.py`:

- a non-chosen child receives the **real** blocks, with only `num_external_tokens` set to 0
- when **nobody is chosen** — the path this deployment actually hits — every child still
  receives the real blocks
- `condition` is a real check rather than a constant, so the patch can disable itself

`tests/native/patches` plus `tests/native/distributed/kv_connector`: 114 passed.

## References

- upstream PR: https://github.com/vllm-project/vllm/pull/46865 (`2285cfc`)
- analysis: https://github.com/rebellions-sw/lmcache-rbln/issues/121


vllm_rbln/patches/multi_connector.py
```diff
@@ -0,0 +1,76 @@
...
+    # Source check, not a version compare: vllm-rbln pins vLLM by lock, so a
+    # build can report < 0.26.0 while already carrying the fix.
+    try:
+        source = inspect.getsource(MultiConnector.update_state_after_alloc)
+    except (OSError, TypeError):
+        logger.warning(
+            "Cannot read MultiConnector source; applying vllm#46865 backport."
+        )
+        return True
+    return "empty_blocks" in source
```
rebel-cjlee: I think we can make this simpler by just checking the vLLM version instead of inspecting the source code. How about this?
1. If `vllm_version < "0.26.0": Apply the patch.
2. If `vllm_version >= "0.26.0"`: Raise an `assert` (or `AssertionError`).
This way, when we upgrade to `0.26.0` or later, it will explicitly fail and remind us to remove this temporary code.

e.g.
```python
assert vllm.__version__ < "0.26.0", "vLLM is updated to >= 0.26.0. Time to remove this temporary patch!"
```

rebel-jinhwan: Thanks, I agree this is much simpler. 
I updated it as you suggested

rebel-cjlee (approved):
Can you try rebasing your branch on latest `dev` or merging the latest `dev` into yours? Then CI will works.

PR #926, opened 2026-08-13
fix(mega-cache): key the bundle on the warm-up graph set

rebel-jonghewk:
### 🚀 Summary of Changes

`config_signature()` keyed the bundle on `VllmConfig.compute_hash()`, which drops `max_num_seqs` and both KV-block knobs (`num_gpu_blocks_override`, `gpu_memory_utilization`). Each changes a warm-up graph shape, so two runs differing only in those flags shared one bundle file.

The partial hit is what hurts. Prefill is shaped by `max_num_batched_tokens` (hashed) and hits; decode is shaped by `max_num_seqs` (not hashed) and misses. A cache hit registers no module in the compiler's `CompileState`, so the decode graph recompiles at `weight_mode=gen` and owns a second copy of the weights on the device — warm-up then dies on `SYS_EBUSY`, reported as `Not enough memory for N blocks of KV cache`. It does not self-heal: `save_cache_artifacts()` serializes only this session's misses, so the re-saved bundle drops what hit.

Fix: fold the three fields into the signature. A false mismatch costs one recompile, a false match costs a duplicate weight set.

---

* Resolves #
* Related to #805, #917 (same class of bug)

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

```
pytest tests/native/v1/worker/test_mega_cache.py -k warmup_graph_set
```

E2E (MiniMax-M2.5 dp4+EP, `VLLM_RBLN_NUM_HIDDEN_LAYERS=4`, `--num-gpu-blocks-override 196`, shared `VLLM_CACHE_ROOT`): serve at `--max-num-seqs 1`, then at `4`. Before, both log the same signature; after, they differ and each keeps its own bundle.

---

Before the fix, 4 layers (1/15th of the real weight footprint):

| | `--max-num-seqs 1` (fresh) | `--max-num-seqs 4` (same cache) |
|---|---|---|
| `config_signature` | `44edcbc6c8a2f970` | `44edcbc6c8a2f970` (same) |
| graphs hit | 0 | 1 per rank (prefill) |
| device bytes / rank | 10.368 GiB | **14.740 GiB** (+4.37 GiB ≈ one weight set) |
| bundle written | 2,186,275,519 B | **828,124,121 B** (−62%) |

At 62 layers the duplicate is ~51 GiB/rank (~12.8 GiB/chiplet) against a 35 GiB chiplet ceiling.

---

### 📋 Checklist

---

### 💬 Notes

**#870 can take the KV-block knobs back out.** Dynamic KV marks `num_blocks` dynamic and compiles against a 2-block dummy, so the block count leaves the compiled artifact and `num_gpu_blocks_override` / `gpu_memory_utilization` stop shaping any graph — hashing them then only invalidates bundles for nothing. It stays conditional: dynamic KV is opt-in and its gates (MLA, spec decode, KV connector, sliding window, cross-layer sharing) fall back to the static path where the block count is still baked in, so the two fields should be hashed only on that path. `max_num_seqs` stays either way — the decode bucket batch size is in the graph regardless.


tests/native/v1/worker/test_mega_cache.py
```diff
@@ -190,6 +190,46 @@ def test_graph_relevant_config_invalidates(self, overrides):
...
+
+        def buckets(n):
+            return get_bucketing_manager(
+                "exponential", max_batch_size=n
+            ).decode_batch_buckets
+
+        assert buckets(a) != buckets(b), "probe does not move the bucket set"
+        assert mega_cache.config_signature(
+            make_vllm_config(max_num_seqs=a)
+        ) != mega_cache.config_signature(make_vllm_config(max_num_seqs=b))
```
rebel-cjlee: Just one thing on this test.

With exponential, the first `assert buckets(a) != buckets(b)` can never really fail. The biggest bucket is always `max_batch_size`, so as long as `a != b` the sets are different on matter what. That makes the assert basically the same as `assert a != b`, which is always true. And the second assert (different `max_num_seqs` -> different signature) is already checked by `test_warmup_graph_set_config_invalidate`), since we hash the raw field.

The case where buckets actually overlap is the manual strategy. For example `manual=[1,2,4]` with `max_num_seqs` 8 vs 16 gives the same bucket set but a different `max_batch_size`. But even there we hash the raw `max_num_seqs`, so the signatures still come out different anyway. So switching this test to manual wouldn't really test the code. It'd turn into a design question: do we want the same graph to share the cache or not?

So my suggestion is to just drop this test. If we want to keep showing that `max_num_seqs` changes the graph, that belongs in the `bucketing` tests, not on the signature. And the manual overlap thing is better left as a separate design discussion.

rebel-cjlee: I think it needs check how the bucketing logic handles the "same bucketing set but different `max_num_seqs`" case (e.g., bucket=[1,2,4], max_num_seqs=8 ?), and whether it's even a real, usable scenario in practice.

rebel-cjlee (approved):

tests/native/v1/worker/test_mega_cache.py
rebel-jonghewk: Thanks for the review 👍

vllm_rbln/v1/worker/mega_cache.py
```diff
@@ -65,6 +65,29 @@ def _compile_env_factors() -> str:
     return hash_factors(factors)
 
 
+def _warmup_graph_set_factors(vllm_config) -> str:
```
rebel-eunji, line 78: Same class of gap, one axis still open: spec decode. Upstream `SpeculativeConfig.compute_hash()` keys only on the eagle3 aux-hidden-states factors and drops `num_speculative_tokens`, `method`, and the draft model — but each shapes a warm-up graph (`query_len = num_spec_tokens + 1` in the decode dummy runs, the eagle draft-loop unroll, which drafter graphs exist, drafter hidden size / max_model_len). Two runs differing only in `--num-speculative-tokens` would share a bundle that partially hits — the same duplicate-weight failure this PR fixes.

Follow-up: fold these in here.

```python
spec = getattr(vllm_config, "speculative_config", None)
draft = getattr(spec, "draft_model_config", None)
factors.update({
    "num_speculative_tokens": normalize_value(
        getattr(spec, "num_speculative_tokens", None)),
    "spec_method": normalize_value(getattr(spec, "method", None)),
    "draft_model": draft.compute_hash() if draft is not None else None,
    "draft_tp": normalize_value(
        getattr(spec, "draft_tensor_parallel_size", None)),
})
```

Notes: `draft_model_config.compute_hash()` is launch-stable, so it needs no `_stable_compute_hash` treatment; do not hash `draft_parallel_config` wholesale (same port leak). Runtime-only proposer knobs (`disable_by_batch_size`, `prompt_lookup_*`, suffix tunables) stay out — they never reach a compiled graph.

rebel-eunji: It's generated by my claude

rebel-eunji (approved):
LGTM 👍

PR #927, opened 2026-08-13
fix(sampler): match builtin logits-processor dtypes to model-dtype logits

rebel-eunji:
## Summary

Fix dtype mismatches in vLLM's builtin logits processors when the RBLN sampler applies them directly to model-dtype logits such as `bfloat16`.

Requests using `min_tokens` could previously fail with:

```text
RuntimeError: Index put requires the source and destination dtypes match,
got BFloat16 for the destination and Float for the source.
```

Similar mismatches could also affect `logit_bias` and `min_p`, whose internal tensors are created as `float32`.

This PR adds RBLN-specific, dtype-aware variants of the affected builtin logits processors and selects them only when the RBLN sampler is enabled.

It also fixes a `min_p` shape mismatch on bucket-padded decode steps by sizing the logits-processor batch update to the padded sampler bucket.

## Background

The default vLLM sampler converts logits to `float32` before applying logits processors. The RBLN sampler intentionally keeps logits in the model dtype to avoid copying and converting a `[num_requests, vocab_size]` tensor on every decoding step.

As a result, the RBLN sampler may pass model-dtype logits to processors whose internal tensors remain `float32`. The following in-place operations require compatible dtypes:

| Processor | Internal tensor | Operation |
|---|---|---|
| `MinTokensLogitsProcessor` | `neg_inf_tensor` | `index_put_` |
| `LogitBiasLogitsProcessor` | `bias_tensor` | indexed `+=` |
| `MinPLogitsProcessor` | `min_p` | `mul_` |

### Add dtype-aware builtin logits processors

Added `vllm_rbln/v1/sample/rbln_logits_processor.py` with RBLN-specific subclasses for the affected processors.

#### `RBLNMinTokensLogitsProcessor`

Synchronizes the scalar `-inf` tensor with the dtype of the incoming logits in both:

- `apply()`
- `apply_with_spec_decode()`

Per-call synchronization is required for speculative decoding because the same processor instance can receive two different dtypes in one decoding step:

- `apply()` receives model-dtype logits from the RBLN sampler.
- `apply_with_spec_decode()` receives `float32` target logits from rejection sampling.

Converted scalar tensors are cached by dtype, so each dtype is materialized only once.

#### `RBLNLogitBiasLogitsProcessor`

Synchronizes `bias_tensor` with the incoming logits dtype in `apply()` before performing indexed in-place addition.

The dtype is determined from the actual logits rather than `model_config.dtype`, because the compiled Optimum model's output dtype may differ from the dtype exposed by the vLLM model configuration.

#### `RBLNMinPLogitsProcessor`

Synchronizes the active `min_p` tensor with the incoming logits dtype in `apply()` before it is multiplied in place with the maximum probabilities.

As with logit bias, synchronization uses the actual logits dtype rather than relying on the model configuration.

### Add an RBLN logits-processor builder

Added `build_rbln_logitsprocs()`, which mirrors upstream vLLM's `build_logitsprocs()` while constructing the dtype-aware RBLN builtin processors.

The builder preserves upstream behavior for:

- Pooling models
- Speculative decoding
- Custom logits processors
- Installed logits-processor plugins

When speculative decoding is enabled, only the RBLN min-tokens processor is constructed, matching upstream behavior where `min_p` and `logit_bias` are unsupported with speculative decoding.

### Select the builder based on sampler implementation

Both model runners now select their logits-processor builder according to `VLLM_RBLN_SAMPLER`:

- `VLLM_RBLN_SAMPLER=1` uses `build_rbln_logitsprocs()`.
- `VLLM_RBLN_SAMPLER=0` uses upstream `build_logitsprocs()`.

This is applied in:

- `RBLNModelRunner`
- `RBLNOptimumModelRunner`

The RBLN-specific processors are therefore used only when processors receive model-dtype logits. The default vLLM sampler continues to use the upstream processors with its `float32` logits path.

### Size the logits-processor batch update to the padded sampler bucket

`refresh_metadata_rbln` now reports the padded `bucket_size` (instead of `num_reqs`) as `BatchUpdate.batch_size`.

`MinPLogitsProcessor` sizes its `min_p` tensor from `batch_update.batch_size`, and the RBLN sampler applies it to pooled logits with `bucket_size` rows. A `num_reqs`-sized `min_p` therefore failed on any decode step where `num_reqs < bucket_size` (e.g. `max_num_seqs=4` with 3 live requests):

```text
RuntimeError: The size of tensor a (4) must match the size of tensor b (3)
at non-singleton dimension 0
```

Padding slots hold default parameters (`min_p=0`), so the extra rows mask nothing.

### Align Optimum runner memory handling with vLLM

The Optimum model runner now uses vLLM's global `PIN_MEMORY` setting when:

- Constructing logits processors
- Allocating the sampled-token CPU buffer
- Batching multimodal inputs

This replaces the runner-local `pin_memory=False` value.

### Extend Optimum runner test utilities

Updated the Optimum runner test helpers to support:

- Configurable model dtype
- `min_tokens`
- `stop_token_ids`
- `logit_bias`
- `min_p`

The mock model and generated logits now use the configured model dtype, allowing both `float32` and `bfloat16` paths to be tested.

### Min-tokens sampling

Added:

```text
tests/optimum/v1/worker/test_sampler.py::
test_forward_min_tokens_masks_stop_tokens
```

The test runs with both `float32` and `bfloat16` logits and verifies that:

1. Stop tokens remain masked while fewer than `min_tokens` tokens have been generated.
2. A non-stop runner-up token is selected while masking is active.
3. The stop token becomes sampleable once the `min_tokens` boundary is reached.
4. The `bfloat16` path does not raise the mixed-dtype `index_put_` error.

### Logit bias

Added:

```text
tests/optimum/v1/worker/test_sampler.py::
test_forward_logit_bias_overrides_argmax
```

The test runs with both `float32` and `bfloat16` logits and verifies that:

1. A positive logit bias changes the greedy-sampling result.
2. The processor's `bias_tensor` is synchronized with the incoming logits dtype.
3. The indexed in-place addition does not fail for `bfloat16` logits.

### Min-p sampling

Added:

```text
tests/optimum/v1/worker/test_sampler.py::
test_forward_min_p_masks_low_probability_tokens
```

The test runs with both `float32` and `bfloat16` logits and verifies that:

1. Tokens below the configured min-p threshold are masked.
2. The remaining token is sampled deterministically.
3. The processor's `min_p` tensor is synchronized with the incoming logits dtype.
4. The in-place threshold calculation does not fail for `bfloat16` logits.

### Default vLLM sampler fallback

The three tests above are additionally parametrized over `VLLM_RBLN_SAMPLER` and verify that with `VLLM_RBLN_SAMPLER=0`:

1. The runner constructs the upstream builtin processors instead of the RBLN subclasses.
2. The internal tensors remain `float32`, matching the default sampler's `float32` logits path.
3. `min_tokens`, `logit_bias`, and `min_p` sampling behavior is unchanged.

### Padded-bucket sampling

Added `tests/optimum/v1/worker/test_sampler_bucket.py` and wired it into the optimum PR pipeline.

The suite covers the RBLN sampler's padded-bucket path (`num_reqs < bucket_size`): `min_p` (regression test for the broadcast failure above), `min_tokens`, `logit_bias`, logprobs, a finished request leaving a clean padding row, bucket-size transitions as the batch grows and shrinks, and mixed per-request sampling params. Every test runs against both bucket sources (the `get_bucket_sizes` ladder and compiled `decoder_batch_sizes`), and the `min_p` and mixed-params tests additionally run with `bfloat16` logits. `test_get_bucket_sizes` moved into this file from `test_sampler.py`.

### Speculative decoding

Added:

```text
tests/torch_compile/unit/v1/sample/test_torch_rejection_sampler.py::
test_min_tokens_masks_stop_tokens
```

The test constructs processors through `build_rbln_logitsprocs()` and passes the same `RBLNMinTokensLogitsProcessor` instance through:

1. `apply()` with model-dtype `bfloat16` logits.
2. `apply_with_spec_decode()` with the rejection sampler's `float32` target logits.

It also verifies that stop tokens are masked at the applicable draft positions before the `min_tokens` boundary and that rejection sampling selects the expected replacement token.

## Performance Impact

The change does not convert the full logits tensor to `float32`.

When an affected processor is inactive, no dtype conversion is performed.

When active:

- `RBLNMinTokensLogitsProcessor` performs a dictionary lookup and converts only a scalar `-inf` tensor once per dtype.
- `RBLNLogitBiasLogitsProcessor` converts `bias_tensor` only when its dtype differs from the incoming logits.
- `RBLNMinPLogitsProcessor` converts the active `min_p` tensor only when its dtype differs from the incoming logits.

This preserves the RBLN sampler's existing optimization of processing logits in the model dtype.

## Type of Change

- [ ] Release
- [ ] Feature
- [ ] Model support
- [ ] Core engine change
- [x] Bug fix
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist

- [x] The PR title follows the Conventional Commits format.
- [x] The root cause and expected behavior are documented.
- [x] Regression tests cover `min_tokens`, `logit_bias`, and `min_p`.
- [x] Speculative-decoding behavior is covered.
- [ ] This PR is linked to an existing issue.
- [ ] Relevant documentation has been updated, if applicable.


rebel-jonghewk (commented):
min_p still raises on any bucket-padded decode step for optimum. 

 refresh_metadata_rbln passes self.num_reqs to get_and_reset, so MinPLogitsProcessor.min_p is [num_reqs, 1]. The sampler then receives pooled_tensors[bucket_size]. With max_num_seqs=4 and 3 live requests the bucket is 4, and max_probabilities.mul_(self.min_p) broadcasts [4, 1] against [3, 1]:
```
RuntimeError: The size of tensor a (4) must match the size of tensor b (3) at non-singleton dimension 0
```

rebel-jonghewk (approved):
lgtm

PR #929, opened 2026-08-14
refactor(sampler): refactor sampler and add more pytests

rebel-eunji:
## Summary

This PR improves sampler-related test coverage and adjusts Dynamo recompile limits to better match the actual compilation behavior of the RBLN sampler.

There are two main changes:

1. **Recompile limit tuning**
   - Remove penalty-related warm-up configs from `WARM_UP_CONFIGS`
   - Set `torch._dynamo` recompile limits based on the actual number of compiled sampler variants per bucket
   - Apply the limit adjustment to both:
     - `RBLNOptimumModelRunner`
     - `RBLNModelRunner` will be refactored by @rebel-cjlee in following PR (TODO)

2. **Expanded worker-level sampler tests**
   - Add/extend tests to verify the exact token-level behavior of:
     - penalties
     - token masking
     - bad words handling
     - mixed greedy/random sampling
     - logprobs
     - deterministic top-k / top-p edge cases
   - Ensure behavior is validated across **multiple decode steps**, not just a single step

---

### 1. Penalty warm-up configs were unnecessary
Penalty paths run eagerly on CPU, so the penalty entries in `WARM_UP_CONFIGS` were not contributing meaningful compiled coverage. They only increased warm-up cost.

Those scenarios are still useful for benchmarking, so they are kept in:

- `benchmarks/benchmark_sampler.py`

### 2. Recompile limits should reflect actual compiled variants
Sampler compilation behavior is driven mainly by bucket shape and `(top_k, top_p)` input combinations.

Per bucket, the compiled sampler paths need:
- **4 variants** for `rbln_top_k_top_p_sample`
- **1 variant** for `rbln_greedy_sample`

This PR updates the limits accordingly and uses `max(...)` so existing larger limits are never reduced.

This is especially important for the native runner, which previously could hit `FailOnRecompileLimitHit` when bucket count became large.

---

#### `vllm_rbln/v1/worker/optimum_model_runner.py`
- Replaced the previous recompile-limit formula with a limit based on bucket count
- Also raises `accumulated_recompile_limit`

#### `vllm_rbln/v1/worker/rbln_model_runner.py`
- Added the same recompile-limit logic for the native runner

#### `vllm_rbln/v1/sample/rbln_sampler.py`
- Removed penalty benchmark configs from `WARM_UP_CONFIGS`

#### `benchmarks/benchmark_sampler.py`
- Added benchmark config definitions for penalty scenarios here instead

---

### New helper utilities
`tests/optimum/v1/sample/utils.py`
- Added `prefill_requests(...)`
- Added `run_decode_steps(...)`
- Added `sampled_token(...)`
- Extended `make_request(...)` to support:
  - `allowed_token_ids`
  - `bad_words_token_ids`
- Seeded `init_none_hash(...)` for request block hashing during output token appends

These helpers make the tests behave more like production decode flow by feeding sampled tokens back into each request.

---

### Penalty behavior
Added tests to verify:
- decode steps with penalties do **not** trigger recompilation
- presence/frequency penalties change greedy token selection as expected
- repetition penalty also applies to prompt tokens
- penalized and non-penalized requests remain isolated within the same padded batch

### Token masking
Added tests for:
- `allowed_token_ids`
- mask isolation in padded batches

### Bad words handling
Added tests for:
- single-token bad word masking
- multi-token continuation masking

### Sampling path behavior
Added tests for:
- mixed greedy/random batches
- deterministic top-k / top-p cases
- logprob ordering and value correctness against `torch.log_softmax`

### Existing test cleanup / improvement
- `test_forward_sampler_mode_and_structured_output`
  - now runs with both `use_rbln_sampler=True/False`
- `test_forward_sampling_parameters`
  - reduced parameter explosion by collapsing penalty combinations into:
    - `no_penalty`
    - `all_penalty`
- `test_no_nan_logits_with_padded_bucket`
  - now runs multiple decode steps instead of one
- shared sampler env setup moved into a fixture

---

## Files changed

- `benchmarks/benchmark_sampler.py`
- `tests/optimum/v1/sample/test_sampler.py`
- `tests/optimum/v1/sample/utils.py`
- `vllm_rbln/v1/sample/rbln_sampler.py`
- `vllm_rbln/v1/worker/optimum_model_runner.py`
- `vllm_rbln/v1/worker/rbln_model_runner.py`

---

## Related issues

- Related to #919
- Related to #927

---

### Main sampler test suite
```bash
pytest tests/optimum/v1/sample/test_sampler.py
```

Expected result:
- **113 passed**

### Regression checks for shared utility changes
```bash
pytest \
  tests/optimum/v1/sample/test_sampler_bucket.py \
  tests/optimum/v1/worker/test_optimum_model_runner.py \
  tests/optimum/v1/worker/test_prefix_cache_fallback.py
```

Expected result:
- **33 passed**

---

## Notes

- The only runtime behavior change on the native path is the recompile-limit adjustment.
- The new limit logic only **raises** limits when needed; it never lowers existing settings.
- `test_penalty_decode_steps_do_not_recompile` is intended to guard this behavior directly.


benchmarks/benchmark_sampler.py
```diff
@@ -28,6 +28,57 @@
 
 MAX_NUM_PROMPT_TOKENS = 64
 
+# Penalty configs are no longer part of WARM_UP_CONFIGS: penalties run eagerly
+# on the CPU, so they compile nothing and would only lengthen server warm-up.
+# As benchmark scenarios they still matter — they measure that eager cost.
```
rebel-eunji:
```suggestion
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -475,6 +475,17 @@ def __init__(
...
+        # (top_k, top_p) None/tensor combination (4) and rbln_greedy_sample
+        # one. Raise the per-code-object and global limits to fit; never lower.
+        torch._dynamo.config.recompile_limit = max(
+            torch._dynamo.config.recompile_limit,
+            4 * self.bucketing_manager.batch_buckets_count,
+        )
+        torch._dynamo.config.accumulated_recompile_limit = max(
+            torch._dynamo.config.accumulated_recompile_limit,
+            5 * self.bucketing_manager.batch_buckets_count,
+        )
```
rebel-cjlee: `compiler.py:41` already sets this. `_ensure_torch_dynamo_configured` writes `cache_size_limit`, which is just another name for `recompile_limit`. After this PR, two files write the same global under two different names. Can we keep it in one place?

If we move it into `compiler.py`, the problem is that `compiler.py` does not know the bucket count. One way to split it: the runner keeps computing the nu;mber, since only the runner knows the buckets, and `compilation/` gets a small function that raises the limit, so it stays the only file that touches `torch._dynamo.config`. If this feels bigger than what this PR should carry, feel free to leave it as it. I will make follow up PR for it.

rebel-cjlee: `rbln_top_k_top_p_sample`, `rbln_greedy_sample` and `rbln_rejection_sample` are all plain module level functions with no `self`. Each one has only one id match group, so both counters count the same entries. The real ceiling is `min(recompile_limit, accumulated_recompile_limit)`, and the smaller one always wins. Giving them 4N and 5N cannot help.
Please tell me if I'm missing something.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -475,6 +475,17 @@ def __init__(
             type(self.bucketing_manager).__name__,
             self.bucketing_manager.decode_batch_buckets,
         )
+        # Per batch bucket, rbln_top_k_top_p_sample compiles one entry per
+        # (top_k, top_p) None/tensor combination (4) and rbln_greedy_sample
+        # one. Raise the per-code-object and global limits to fit; never lower.
```
rebel-cjlee: `rbln_rejection_sample` is a third code object with its own budget, but it is not mentioned in the comment and not counted in the limit. Is that out of scope for this PR?

rebel-cjlee (approved):

rebel-jonghewk (approved):
lgtm

PR #933, opened 2026-08-15
fix(worker): fall back to full prefill when prefix-cached KV copy fails

rebel-eunji:
## Problem

Copying prefix-cached KV blocks can fail at runtime, for example due to device OOM.  
Previously, this failure propagated out of `execute_model` and could kill the entire engine with `EngineDeadError`, taking the server down during inference.

## Why runner-side fallback is safe

KV copy is only an optimization. When a prefix hit occurs, the scheduler still assigns the full prompt and the runner is the only place where the cached-prefix trim is applied.  
If KV copy fails, a full prefill writes the same KV values into the same destination blocks, so the engine can safely recover without any scheduler-side changes.

## Changes

- Introduce a typed `KVCacheCopyError` for KV copy failures.
- Attempt prefix-cached KV copy in `execute_model` before building forward inputs.
- If KV copy fails, rebuild the prefill inputs without trimming the cached prefix and run a full prefill instead.
- Keep other exceptions unchanged so real invariant violations still surface normally.
- Remove the old `reuse_prefix_cached_kv` path and consolidate the logic in the runner.

## Follow-up

A future improvement would be to report copy failures back to the scheduler so it can avoid repeatedly offering prefix hits under persistent memory pressure.


rebel-jonghewk (approved):
lgtm

PR #935, opened 2026-08-17
feature(APC): enable smaller sub block size for prefix caching

rebel-eunji:
### 🚀 Summary of Changes

Sub-block prefix caching previously tied `sub_block_size` to the prefill chunk size (`max_num_batched_tokens`). This constraint existed because the attention kernel could only write KV data within a single block per prefill step — if a chunk started mid-block, it had to fit entirely within that block's remaining capacity.

The **multi-block attention store** lifts this constraint: the kernel can split a prefill chunk's KV write across two consecutive blocks. This PR wires that capability into the scheduler by decoupling `sub_block_size` from the prefill chunk size.

**What changes:**

- Adds `sub_block_size` as an `RBLNConfig` field, so it takes `--rbln-sub-block-size`, an `additional_config` key, or `VLLM_RBLN_SUB_BLOCK_SIZE` (default `0` = auto = `max_num_batched_tokens`). It sits next to the existing `enable_sub_block_cache`, which decides whether the path runs at all.
- A value below the prefill chunk size is what makes a chunk span two blocks, so it needs the multi-block store. Availability is a property of the device rather than a feature flag: REBEL CR13 alone carries it, and the decoupled setting is rejected at startup on any other device, with the error naming the device that was found.
- `sub_block_size_in_use()` (`vllm_rbln/v1/core/utils.py`) holds every rule on the setting, so a caller gets a size or a `ValueError` and never a setting that was dropped on the way. It rejects a `sub_block_size` given together with `enable_sub_block_cache=False` or `enable_prefix_caching=False` instead of ignoring it, enforces `block_size >= max_num_batched_tokens >= sub_block_size`, and returns `None` — staying on vLLM's own manager — when a KV cache group's spec cannot be sliced per token.
- Updates `docs/sub_block_prefix_caching.md` to document the new contract and environment variable.
- Reworks `examples/optimum/prefix_caching/prefix_caching_long_offline.py` so it actually reaches the new path. The old prompts were far too short to ever cross a block boundary, so the example could not tell the two stores apart. The shared prefix and the per-prompt tail are now padded (see the comment above `PREFIX_PAD`), and `report_geometry()` prints the per-request `hit / off0 / chunk` geometry, warning when no request spilled into a second block.
- The same example now takes its latency from vLLM's own per-request stats (`RequestOutput.metrics`, kept alive by `disable_log_stats=False`) instead of a `time.time()` bracket around `generate()`. `queued_ts`/`scheduled_ts`/`first_token_ts`/`last_token_ts` are engine-core monotonic timestamps, so the with/without comparison is not polluted by script-side tokenization and output handling, and prefill time — the interval sub-block caching actually shortens — is reported separately from queueing and decode.

**Effect on prefix cache hit rate:**

With `sub_block_size` smaller than the prefill chunk size, the scheduler can match more sub-blocks at the prefix boundary, increasing the effective cache hit granularity. For example, with `block_size=1024`, `max_num_batched_tokens=512`, and `sub_block_size=128`, a request that gets a 3712-token full-block hit can resume its prefill at offset 640 within block 3. The 512-token chunk is then split across two blocks:

```
[vllm-rbln] INFO [flash_attention.py:90] seq_idx=[3712]
[vllm-rbln] INFO [flash_attention.py:90] block_table=[[1, 2, 3, 6, 0, 0, 0, 0]]
[vllm-rbln] INFO [flash_attention.py:90] static: block_size(partition_len)=1024  query_len(T)=512  M=2
[vllm-rbln] INFO [flash_attention.py:112] b=0 seq=3712 off0=640 start_blk=3 -> STEP1 store [B,M]:
[vllm-rbln] INFO [flash_attention.py:112]   seq_idx(update_block_idx)     = [6, 0]
[vllm-rbln] INFO [flash_attention.py:112]   seq_idx2(update_block_offset) = [640, 0]
[vllm-rbln] INFO [flash_attention.py:112]   update_block_size             = [384, 128]   (sum=512, T=512)
[vllm-rbln] INFO [flash_attention.py:112]   update_token_begin            = [0, 384]
```

384 tokens are written to block 6 (starting at offset 640) and the remaining 128 tokens spill into the next block. Without multi-block store support this cross-boundary write was impossible, so `sub_block_size` had to equal the chunk size to guarantee the write stayed within one block. When the cross-block write path is taken, the second entry in update_block_idx may also be 0 — this is a legitimately allocated physical block, but only the first update_block_size[1] tokens are valid; the rest of the block is padding.

---

* Related to rebel-compiler `feat-shm-machinelowering`

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

1. On a REBEL CR13, start vLLM with prefix caching and a sub-block size smaller than the prefill chunk:
   ```bash
   vllm serve <model> \
       --enable-prefix-caching \
       --max-num-batched-tokens=512 \
       --block-size=1024 \
       --rbln-sub-block-size=128
   ```
2. Send a request whose prompt shares a long prefix with a previously cached prompt.
3. Verify in the logs that `Sub-block prefix caching enabled: block_size=1024, sub_block_size=128` is printed at startup.
4. Verify that the attention kernel log shows a cross-block KV write (`update_block_size = [N, M]` with two non-zero entries summing to `query_len`).
5. **Unsupported device:** run the same command on a device other than REBEL CR13 and confirm startup fails with a `ValueError` that names the device, rather than quietly falling back.
6. **Conflicting config:** pass `--rbln-sub-block-size=128` together with `--rbln-enable-sub-block-cache=false` (or without `--enable-prefix-caching`) and confirm a `ValueError` is raised at startup.
7. **Invalid config:** set `--rbln-sub-block-size` to a value larger than `max_num_batched_tokens` and confirm a `ValueError` is raised at startup.
8. **Offline example:** run `examples/optimum/prefix_caching/prefix_caching_long_offline.py` and check that the `spill` column is `yes` for at least one request — otherwise no request crossed a block boundary and the multi-block store path never ran. The per-request latency block printed by `report_latency()` gives the prefill interval for both engines.

---

### 📋 Checklist

---


vllm_rbln/config.py
```diff
@@ -79,8 +79,12 @@ class RBLNConfig:
     """Use the custom RBLN kernels."""
 
     enable_sub_block_cache: bool = True
-    """Enable sub-block prefix caching. The sub-block size equals
-    max_num_batched_tokens (the prefill chunk size)."""
+    """Enable sub-block prefix caching, at `sub_block_size` granularity."""
+
+    sub_block_size: int = 0
```
rebel-cjlee:
```suggestion
    sub_block_size: int | None = Field(default=None, gt=0)
```
How about it here?
Upstream uses this shape, for example, `max_parallel_loading_workers: int | None = Field(default=None, ge=1)` in `vllm/config/parallel.py`. Then pydantic rejects a bad value at the input, so the `sub_block_size < 0` check in `sub_block_size_in_use()` is no longer needed. `None` also reads as "not set", while `0` currently means both "not set" and a size that is never valid.

This would be the first optional env var in the repo, so `tests/native/test_envs.py` needs two small edits, a `_PROBE_OVERRIDES` entry and a union aware annotation check.

vllm_rbln/v1/core/utils.py
```diff
@@ -243,23 +245,54 @@ def sub_block_size_in_use(
...
+            raise ValueError(
+                f"sub_block_size={wanted} asks for sub-block prefix caching and "
+                f"{off}=False turns it off."
+            )
+        return None
+
+    sub_block_size = wanted or max_num_batched_tokens
+    if not RBLNKVCacheManager.can_use_sub_block_caching(
+        kv_cache_config, sub_block_size
     ):
```
rebel-cjlee: This returns `None` even when the user asked for a size, so `--rbln-sub-block-size 2048` with `block_size=1024` turns sub-block caching off with no message. The error right below at line 284 is written for exactly that shape, but a size larger than `block_size` never reaches it. A size that is not a divisor, say 100 with `block_size=1024`, is silent in the same way. The docstring above promises a size or a `ValueError`, and the two rules around this one already raise when a given size cannot be honored, at line 272 and line 292.

How about raising here too when `wanted` is set, and keeping the `return None` for the derived default?
`test_equal_block_and_sub_block_size_disables` would flip to expect the error.

rebel-cjlee (commented):
Did you get a chance to run the `finishing-a-change` skill on this one? No problem if you did, it just does not always fire on its own, so I'm checking. It reads the diff against the repository rules before review, and a few of the comments above map straight to its checklist, for example "comments added to code you did not otherwise change".
If you have not, one pass it worth it. Claude catches a fair share of these before a review has to.

rebel-cjlee (approved):

PR #833, opened 2026-07-27
feature(platform): always use RBLN device tensors

rebel-eunji:
### 🚀 Summary of Changes

The CPU-tensor / device-tensor toggle is removed.
Device tensors are now always used.

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #936, opened 2026-08-18
docs(agent): set the conventions that keep agent-written changes reviewable

rebel-cjlee:
### 🚀 Summary of Changes

Coding agents here tend to write more than the task needs. The output looks reasonable, but it is verbose and over-engineered, and the same shapes come back in every PR:

- Abstractions and helpers built for a single call site
- Comments that restate the line below them, or explain the change to the reader
- A fallback or a silent recovery instead of fixing the cause
- Tests that pass whether the code is write or wrong, so they read as coverage without being coverage

Nothing in the repo says not to do this, so each agent guesses, and reviewers catch it by hand one time at a time. This writes the rules down once.

`AGENTS.md` holds the rules that apply to any change, and `CLAUDE.md` is a one line pointer to it:

- The model path split is the key part. `VLLM_RBLN_USE_VLLM_MODEL` divides the code in two, and only `__init__.py` and `platform.py` are allowed to branch on it.
- A short `Terms` section pins the words we already mix: model path, suite, lane.
- Scope, failure handling, comments, tests and reporting each get a short list. Every rule is written so a reviewer can point at it and say yes or no, and where a rule needs a limit it has a number rather than an adjective.

Three skills under `.claude/skills/` carry the steps that only matter at one moment, which keeps them out of the always loaded file:

- `writing-tests`: where a test goes in `tests/native/`, which mark it needs, and how to prove it fails before the fix
- `debugging`: the traps that make a reproduction lie, and the rules that stop a wrong cause from looking confirmed
- `finishing-a-change`: the check before calling a change done

What I expect from this:

- Smaller diffs, because the file says no helper for one call site and no files nobody asked for
- Fewer silent workarounds, because code either works or fails with a clear error
- Fewer empty tests, because a test has to fail before the fix and the agent has to show that it did
- Faster review;, because reviewers point at a rule instead of writing the same comment again
- Less repeated explaining, since new agents and new people read the same file

---

* Resolves #
* Related to #

---

### ✅ Type of Change

---

### 🧪 How to Test

1. Hooks: `uvx pre-commit run --files AGENTS.md CLAUDE.md .claude/skills/*/SKILL.md`
2. Spot check the facts against `tests/native/conftest.py` for the marks and session options, and `vllm_rbln/patches/registry.py` for the patch rules

---

---

### 📋 Checklist

---

### 💬 Notes

Neither model path changes. Docs only.

`tests/optimum/` rules are not settled. The marks and session options are native only. The test skill says so plainly, because an agent that has only seen the native rules would apply them where they gate nothing.

---

### DIFF
--- a/.claude/skills/debugging/SKILL.md
+++ b/.claude/skills/debugging/SKILL.md
@@ -1,0 +1,62 @@
+---
+name: debugging
+description: Use when investigating a bug, a failing test, or unexpected behavior in vllm-rbln — the environment and model-path traps that make a reproduction lie, and the evidence rules that keep a wrong cause from looking confirmed.
+---
+
+# Debugging vllm-rbln
+
+Two things make a debugging session go wrong here: reproducing something other than the reported bug, and confirming a cause that is not the cause. This skill is about those. It does not restate general debugging method.
+
+## 1. Before you trust the reproduction
+
+Get a reproducer before changing anything, and narrow it to the smallest input and configuration that still fails.
+
+**If you cannot reproduce it, stop and report that.** Do not fix a bug you have not seen. An untested fix for an unverified report costs a reviewer more than no patch.
+
+Check these before concluding anything, because each one makes a reproduction silently exercise something other than what you think:
+
+- **`tests/native/` scrubs `VLLM_RBLN_*`.** Exported values do not reach it; the session options are the only way in. A knob you "set" may never have applied.
+- **`--device-tensor` is session-wide.** `platform.py` resolves it at module scope, so a per-test attempt to change it does nothing and the test still passes.
+- **Marks decide the process.** `use_device` and `model_compile` items run in a spawned child. A failure that only appears in one of the two contexts is about the process, not the logic.
+
+## 2. Make the evidence discriminate
+
+Evidence that fits your hypothesis is easy to find and proves nothing — a wrong cause has confirming evidence too. Force the evidence to choose between explanations: an observation that every candidate predicts has narrowed nothing.
+
+- **State what would be true if you are wrong**, and check that specifically.
+- **Predict the output before you run.** Write down what you expect. Matching means your model of the system is right; a surprise means it is wrong, and that is the finding.
+- **The cause must explain every symptom**, not just the loudest one. A leftover unexplained detail usually means a second cause or the wrong one.
+
+Report which parts you verified and which you inferred. Never state a hypothesis as a fact.
+
+## 3. Fix the cause, not its surroundings
+
+Change the thing that is wrong — not a caller, not a wrapper, not a guard around the symptom.
+
+A workaround is allowed when the real fix is out of reach, as long as it is declared: name the cause, say why the fix is out of reach, and get agreement. A patch under `patches/` is that shape, with the `reason` recording what upstream cannot express and review deciding whether to accept it. What is not allowed is a guard added because you could not find the cause.
+
+## 4. If the fix does not work, discard it
+
+Revert it and form another hypothesis. Do not stack a second patch on a failed first one.
+
+A failed fix is information: the cause was probably wrong. Go back to step 2 rather than making the current theory work harder.
+
+Never widen an exception, add a default, or skip a branch to make the failure go away. Reaching for that means you are fixing without a cause.
+
+## 5. Verify against the reproducer
+
+Run the reproducer again, plus the tests around the code you touched, and read the output. A fix that has not been run is not a fix. Say so plainly if you could not run it.
+
+## 6. Check both model paths
+
+- Which path does the bug live on?
+- Does the same defect exist on the other path?
+- Does the fix change behavior on the other path?
+
+Only `__init__.py` and `platform.py` branch on the flag. If the fix seems to need a new `if envs.VLLM_RBLN_USE_VLLM_MODEL` anywhere else, the shape is wrong — stop and ask.
+
+## 7. Report the gaps
+
+State explicitly what you could not build, run, or reproduce; hardware or models you did not have; and anything you inferred rather than verified.
+
+Tests that broke after your change are your regressions. Debug them. Do not stash or revert to check whether they also fail on `main`.
--- a/.claude/skills/finishing-a-change/SKILL.md
+++ b/.claude/skills/finishing-a-change/SKILL.md
@@ -1,0 +1,67 @@
+---
+name: finishing-a-change
+description: Use before claiming a change is done, fixed, ready, or passing, and before writing a commit or PR — runs the checks, reviews the diff against the repository rules, and reports what was not done.
+---
+
+# Finishing a change
+
+By the time you get here the session is long and the instructions you read at the start have faded. Work through this list rather than from memory.
+
+## 1. Run the hooks
+
+```bash
+uvx pre-commit run --files <every file you changed>
+```
+
+`pre-commit` is not a project dependency; `uvx` is how it runs here. Read the output. Passing means you saw it pass.
+
+A new `.py` file must carry the Apache header, or `check-license-header` fails.
+
+## 2. Run the tests
+
+Run the tests covering what you changed, and include the command and its result in your reply. "Should pass" is not a result. If you could not run them, say that instead of implying you did.
+
+## 3. Read your own diff
+
+```bash
+git diff
+```
+
+Look for each of these:
+
+- **Files nobody asked for** — docs, examples, scripts, changelogs
+- **Scratch files** left over from iterating
+- **Single-use helpers** that should be inlined
+- **Abstractions** built for one call site
+- **Comments that narrate the code**, describe your changes, or address the reader
+- **Comments added to code you did not otherwise change**
+- **Swallowed failures** — `except Exception`, bare `except`, `except: pass`, an unrequested fallback or default
+- **`else` branches for cases that cannot happen** — should be `assert` or `raise`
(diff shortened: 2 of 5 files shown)


.claude/skills/debugging/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+
+## 1. Before you trust the reproduction
+
+Get a reproducer before changing anything, and narrow it to the smallest input and configuration that still fails.
+
+**If you cannot reproduce it, stop and report that.** Do not fix a bug you have not seen. An untested fix for an unverified report costs a reviewer more than no patch.
+
+Check these before concluding anything, because each one makes a reproduction silently exercise something other than what you think:
+
+- **`tests/optimum/` takes the model path from the default.** It has no suite-level conftest, so an exported `VLLM_RBLN_USE_VLLM_MODEL` flips it to the native path without failing. Check the environment first.
```
rebel-jonghewk: > **`tests/optimum/` takes the model path from the default.

This is not true in some cases. I think it should be modified.

.claude/skills/debugging/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+- **`tests/native/` scrubs `VLLM_RBLN_*`.** Exported values do not reach it; the session options are the only way in. A knob you "set" may never have applied.
+- **`--device-tensor` is session-wide.** `platform.py` resolves it at module scope, so a per-test attempt to change it does nothing and the test still passes.
```
rebel-jonghewk, line 19: Okay for now but once [#833](https://github.com/RBLN-SW/vllm-rbln/pull/833) is landed, it might need to be deleted.

.claude/skills/debugging/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+
+Run the reproducer again, plus the tests around the code you touched, and read the output. A fix that has not been run is not a fix. Say so plainly if you could not run it.
+
+## 6. Check both model paths
+
+- Which path does the bug live on?
+- Does the same defect exist on the other path?
+- Does the fix change behavior on the other path?
+
+The runtime branch belongs in `envs.py`, `__init__.py`, or `platform.py`. If the fix seems to need a new `if envs.VLLM_RBLN_USE_VLLM_MODEL` anywhere else, the shape is wrong — stop and ask.
```
rebel-jonghewk: What does `runtime` mean here? The term is kind of ambiguous.

.claude/skills/finishing-a-change/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+
+Repository-specific checks:
+
+```bash
+grep -rn VLLM_RBLN_USE_VLLM_MODEL vllm_rbln/
+```
+
+Only `envs.py`, `__init__.py`, and `platform.py` may appear. Anything else is a branch in the wrong place.
+
+- A new env var must appear under the right banner in **both** the `TYPE_CHECKING` block and the runtime dict of `envs.py`.
```
rebel-jonghewk: Also one of RBLN_COMPILE_ENV and RBLN_NON_COMPILE_ENV for now.

.claude/skills/finishing-a-change/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+- **Dead leftovers** — renamed `_var`, unused re-exports, `# removed` comments
+- **Anything not in English**
+
+Repository-specific checks:
+
+```bash
+grep -rn VLLM_RBLN_USE_VLLM_MODEL vllm_rbln/
+```
+
+Only `envs.py`, `__init__.py`, and `platform.py` may appear. Anything else is a branch in the wrong place.
```
rebel-jonghewk: I think there is a comment in rbln_model_runner  that mentions `VLLM_RBLN_USE_VLLM_MODEL`.

.claude/skills/finishing-a-change/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+# Finishing a change
+
+By the time you get here the session is long and the instructions you read at the start have faded. Work through this list rather than from memory.
+
+## 1. Run the hooks
+
+```bash
+uvx pre-commit run --files <every file you changed>
```
rebel-jonghewk, line 13: Is it okay to leave out `--hook-stage manual`? The CI includes it.

.claude/skills/debugging/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+- **Predict the output before you run.** Write down what you expect. Matching means your model of the system is right; a surprise means it is wrong, and that is the finding.
+- **The cause must explain every symptom**, not just the loudest one. A leftover unexplained detail usually means a second cause or the wrong one.
+
+Report which parts you verified and which you inferred. Never state a hypothesis as a fact.
+
+## 3. Fix the cause, not its surroundings
+
+Change the thing that is wrong — not a caller, not a wrapper, not a guard around the symptom.
+
+If the real fix is out of scope, do not substitute a local workaround. Say what the root cause is and what fixing it would take, then ask.
```
rebel-jonghewk: Can `register_patch` be seen as a workaround?

.claude/skills/debugging/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+- **`tests/optimum/` takes the model path from the default.** It has no suite-level conftest, so an exported `VLLM_RBLN_USE_VLLM_MODEL` flips it to the native path without failing. Check the environment first.
+- **`tests/native/` scrubs `VLLM_RBLN_*`.** Exported values do not reach it; the session options are the only way in. A knob you "set" may never have applied.
+- **`--device-tensor` is session-wide.** `platform.py` resolves it at module scope, so a per-test attempt to change it does nothing and the test still passes.
+- **Marks decide the process.** `use_device` and `model_compile` items run in a spawned child. A failure that only appears in one of the two contexts is about the process, not the logic.
+
+## 2. Make the evidence discriminate
+
+Evidence that fits your hypothesis is easy to find and proves nothing — a wrong cause has confirming evidence too. Force the evidence to choose between explanations.
+
+- **Name at least two candidate causes**, then look for the observation that separates them. If an observation is consistent with both, it has not narrowed anything.
```
rebel-jonghewk: > Name at least two candidate causes

This feels a bit narrow to me, it could differ depending on the debugging approach someone takes. I think it would be better to focus on more general content. What do you think?

rebel-jonghewk (commented):
@rebel-eunji @rebel-jongho Could you both take a look as well?
@rebel-eunji, could you focus on the Optimum-related parts, and @rebel-jongho, could you review the overall skills and agents format?

.claude/skills/debugging/SKILL.md
rebel-cjlee: Agreed. If #833 lands, the whole `--device-tensor` line should go with it.

.claude/skills/debugging/SKILL.md
rebel-cjlee: Agreed, it was doing three jobs at once. Dropped the word. The dict is not called environment_variables and the sentence just names the two files.

.claude/skills/debugging/SKILL.md
rebel-cjlee: Dropped the line.

.claude/skills/finishing-a-change/SKILL.md
rebel-cjlee: Right, `rbln_model_runner.py:3025` mentions it in a comment, so the grep returned 4 files. Dropped the grep and kept the rule as text: no new branch on the flag outside `__init__.py` and `platform.py`.

.claude/skills/finishing-a-change/SKILL.md
rebel-cjlee: Intended: `mypy-local` is "Don't run in CI", `mypy-3.10` to `3.13` are "Only run in CI". CI runs the matrix anyway, so the local command stays `--files`.

.claude/skills/finishing-a-change/SKILL.md
rebel-cjlee: Added. 3 places now: `TYPE_CHECKING`, `environment_variables`, and one of `RBLN_COMPILE_ENV` / `RBLN_NON_COMPILE_ENV`. `test_mega_cache.py` asserts the two sets cover everything.

.claude/skills/debugging/SKILL.md
rebel-cjlee: Agreed, a patch can be a workaround. The ban was too broad, so it now targets undeclared ones:
```
A workaround is allowed when the real fix is out of reach, as long as it is declared: 
name the cause, say why the fix is out of reach, and get agreement. A patch under `patches/` is
that shape, with the `reason` recording what upstream cannot express and review deciding
whether to accept it. What is not allowed is a guard added because you could not find the cause.
```

Step 4 still bans widening an exception or adding a default outright, since those hide the symptom rather than declare a tradeoff.

.claude/skills/debugging/SKILL.md
rebel-cjlee: Agreed, that bullet prescribed one method. Dropped it and folded the method-agnostic part into the opening sentence:
```
Force the evidence to choose between explanations: an observation that every candidate predicts has narrowed nothing.
```

Could you take another look?

rebel-jonghewk (approved):

PR #945, opened 2026-08-19
perf(eagle): reduce EAGLE drafting overhead

rebel-minkyu:
### 🚀 Summary of Changes

A series of EAGLE/EAGLE3 drafting optimizations. Each commit is independent and
described below. Previously, eagle3 overhead was so excessive that it could not be used practically, especially with DEVICE_TENSOR on.

1. [perf(eagle): fuse eagle3 initial projection into the target graph](https://github.com/RBLN-SW/vllm-rbln/commit/37543464f13e59cd358864a9e341b5104b1a1f1f)

- Moved EAGLE3's `combine_hidden_states` projection out of the eager drafter path
  and into the compiled target model graph. The target graph now returns
  `combined_hidden_states` instead of the raw `aux_hidden_states` list, so the
  drafter no longer concatenates/projects the aux states on the host.
- Previously, it was executed eagerly, and compiling it to a separate graph also has gain, but the fusion is much faster.
- Warm-up now compiles only the `num_spec + 1` decode query length for EAGLE
  (the 1-token decode graph is no longer reachable once the projection is fused).
- Dropped the now-unused `use_spec_decode` argument from
  `_build_attention_metadata`.

2. [perf(eagle): reuse DP status for padding determination](https://github.com/RBLN-SW/vllm-rbln/commit/4455dbe7ad4becf4202cc04ddcd6bf8f6662054c)

- `num_tokens_and_reqs_across_dp` contains a DP collective. The model runner now
  saves its result as `self.dp_status`, and the drafter reuses it instead of
  issuing its own all-reduce for every drafting pass.
- `num_tokens_and_reqs_across_dp` now returns `(num_tokens_across_dp,
  num_reqs_across_dp, any_prefill)` — `num_reqs_across_dp` is no longer `None`
  on the any-prefill path; the flag is returned explicitly instead.

3. [perf(eagle): apply input stager for eagle and optimize tensor copies](https://github.com/RBLN-SW/vllm-rbln/commit/45f52ce45a720e64024705a038db3d1ab59ad96c)

- The drafter now stages its inputs through the shared `InputStager` instead of
  slicing/`pad()`-ing per step. `InputStager` gained hidden-state staging and
  padding-aware token-index staging (buffer sized by `num_reqs_padded`).
- Merged `set_inputs_first_pass` into `_preprocess`, so input writing and
  padding happen in one place, and removed redundant `positions`/`hidden_states`
  copies in the drafting loop.
- Drafter compilation now uses `use_static_output=True` and shares the runner's
  `runtime_holder`.

4. [perf(eagle): perform eagle drafter greedy sampling in graph](https://github.com/RBLN-SW/vllm-rbln/commit/0342ad414f99cb5493eb8342d62110f3e5475d4d)

- The drafter graph now returns `torch.ops.rbln.argmax(logits)` instead of the
  full logits, so the greedy pick runs on device and only the draft ids come
  back to the host.
- Guarded with a `NotImplementedError` for `draft_sample_method != "greedy"`.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3027,7 +3033,11 @@ def warmup_model(self) -> None:
             # 2. decode
             query_lens = [1]
             if self.speculative_config:
-                query_lens.append(self.speculative_config.num_speculative_tokens + 1)
+                spec_width = self.speculative_config.num_speculative_tokens + 1
+                if self.speculative_config.use_eagle():
+                    query_lens = [spec_width]
+                else:
+                    query_lens.append(spec_width)
```
rebel-cjlee: Question about the warm-up change (`query_lens = [spec_width]` for EAGLE).

Reading the scheduler, there looks like a path where the whole decode batch drops to query len 1. in `rbln_scheduler.py`, when a decode ready request joins from the waiting queue with `num_computed_tokens % block_size < num_spec_tokens`, it lands in `unsafe_backfill_req_ids`. The block around line 916 then trims every scheduled decode to 1 token and pops its spec entry, so `scheduled_spec_decode_tokens` ends up empty. `_prepare_inputs` take the `use_spec_decode=False` branch after that and the step runs at query len 1. `TestSpecDecodeRetroactiveTrim ` in `test_rbln_scheduler.py` seems to cover exactly this case.

If I read that right, the target model would see a (batch, 1) shape that warm-up no longer compiles, and it would compile at runtime instead. What do you think?

vllm_rbln/v1/worker/input_stager.py
```diff
@@ -124,21 +131,50 @@ def _get_or_create_buffer(
...
+                device=self.device,
+            )
+            self._hidden_state_buffers[key] = buf
+
+        buf.fill_(layout.hidden_state_pad_value)
+        buf[: layout.num_reqs, : layout.query_len].copy_(
+            hidden_states,
+            non_blocking=True,
+        )
+        return buf
```
rebel-cjlee, line 157: `_stage_hidden_states` fills the whole buffer with the pad value on every call. On the prefill layout that buffer is (1, max_num_tokens, hidden_size), so it is a fairly large memset each step while only the first num_tokens rows get overwritten by the copy.

If that shows up as overhead, one option for the pad value 0 case is to zero just the padding instead:
```
    buf[layout.num_reqs :].zero_()
    buf[: layout.num_reqs, layout.query_len :].zero_()
    buf[: layout.num_reqs, : layout.query_len].copy_(hidden_states, non_blocking=True)
```

For prefill num_reqs equals num_reqs_padded, so the first line is a no op and often the second one is too. If the gain turns out to be small, leaving it as is seems fine too.

rebel-minkyu: The memset indeed takes at most 300us at prefill, and I've updated the code as suggested. Using `.zero_()` doesn't show much gain, so I left it using `fill` without an additional branch.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-minkyu: That's right. There's a compiler failure after eagle3 `combine_hidden_states` fusion on a case where `qlen=1` on every rank. Since it's quite a rare case, it's currently handled with padding. I'll remove it once the compiler issue is resolved.

rebel-cjlee (approved):

PR #947, opened 2026-08-19
refactor(weight-loader): drop the forked load_weights overrides

rebel-cjlee:
### 🚀 Summary of Changes

Deletes `vllm_rbln/model_executor/model_loader/weight_loader.py`

The file copies eight upstream `load_weights` functions and assigns them back onto the vLLM classes. The only thing the copies add is a layer skip:
```python
if layer_idx >= self.config.num_hidden_layers:
    continue
```

That skip existed because we used to shrink a model with `hf_overrides`. The config then said "3 layers", the checkpoint still had 32, and the upstream loader hit a `KeyError` on the layers we did not build.

`VLLM_RBLN_NUM_HIDDEN_LAYERS` (#914) works differently. It patches `get_pp_indices`, so `config.num_hidden_layers` keeps its real value and `make_layers` fills the dropped range with `PPMissingLayer`. Two things follow:

- The skip can never be true. It is dead code.
- Every upstream loader calls `is_pp_missing_parameter` before it touches a weight, so the dropped layers are already handled.

**This is not only a cleanup.** The copies were taken from older vLLM releases and never resynced, so they are now behind the pinned vLLM in ways nobody intended:

| Model | What the copy is missing |
| -- | -- |
| llama, llama4, qwen2 | Still uses the old `quant_config.get_cache-sacle()` path for kv-cache scales |
| qwen2 | Missing the `name not in params_dict` guard and the `default_weight_loader` branch |
| minimax_m2 | Missing the `.k_scale` / `.v_scale` remap |
| llama4 | Missing `num_redundant_experts` and `fused_moe_make_expert_params_mapping` |
| deepseek_v2 | Bound to `ForCausalLM`, so it bypasses `AutoWeightsLoader`; missing indexer fusion and `get_pp_missing_layer_names` |
| gpt_oss | Hardcodes `mxfp4_block = 32` instead of `OCP_MX_BLOCK_SIZE` (same value) |

#941 and #878 are both this same problem: they patch a copy so it catches up with a loader upstream already ships. As long as we keep the copies, that work comes back on every vLLM bump. Deleting them ends it.

---

* Resolves #
* Related to #914, #941, #878

---

### ✅ Type of Change

---

### 🧪 How to Test

1. Run uv run --no-sync pytest tests/native/patches/test_distributed_utils.py
2. Verify output: all pass. The one that matters is `test_dropped_layers_read_as_missing_to_the_weight_loaders`, written for this deletion. It proves a dropped layer reads as missing to any upstream loader.
3. Edge case tested: the `--model-compile` lane covers qwen3_moe, qwen2_moe, gpt_oss and minimax_m2. **DeepSeek-V3 W8A16, llama, and llama4 are in no lane.** DeepSeek-V3 W8A16 is the one to check by hand: #624 added `use_mha` and `fused_qkv_a_proj` handling to the copy, and while upstream 0.24 has that logic, it now runs through `AutoWeightsLoader` - a different entry point.

---

---

### 📋 Checklist

---

### 💬 Notes

If you shorten a model with `hf_overrides`, please switch to `VLLM_RBLN_NUM_HIDDEN_LAYERS`. After this PR, `hf_overrides={"num_hidden_layers": N}` is no longer supported: nothing skips the layers the config dropped. The env var is also safer.

What `VLLM_RBLN_NUM_HIDDEN_LAYERS` does not reach. It truncates through `get_pp_indices`, so it only affects layer stacks build by `make_layers`. Blocks built outside it (vision encoders, MTP, and EAGLE drafter blocks, and any model that builds its decoder stack by hand) keep their full depth. No model we support today is affected: all eight models above go through `make_layers`.

@rebel-kblee @rebel-thkim (cc, @rebel-jonghewk)
If any internal tool still passes `hf_overrides`, it needs to move to `VLLM_RBLN_NUM_HIDDEN_LAYERS`. I do not know who owns those tools, so I tagged the two of you as the authors of the related PRs. Please forward this to whoever should see it, or point me at them and I will.

---


rebel-kblee (approved):
👍 lgtm!

PR #951, opened 2026-08-19
feature(async): Support async scheduling

rebel-yskim:
### 🚀 Summary of Changes

Supports vLLM's `--async-scheduling` on the vLLM-native path, so the host can build step N+1 while the NPU is still running step N.

Upstream cannot be used as-is because `RBLNScheduler` sizes a running decode request as `num_tokens_with_spec + num_output_placeholders - num_computed_tokens`, which stays `<= 0` until `update_from_output(N)` appends N's real token. The engine's `batch_queue` therefore never fills, and each step's DP gloo all_reduce waits for the previous step's output. The pieces:

- **`RBLNAsyncScheduler`** composes `RBLNScheduler` with upstream `AsyncScheduler`, which bumps `num_output_placeholders` at schedule time. It defines no methods of its own; the MRO resolves `_update_after_schedule` and `_update_request_with_output` to `AsyncScheduler`.
- **Token feedback stays on device.** The scheduler stages `-1` placeholders; the runner overwrites them from `prev_sampled_token_ids`, remapped through `prev_req_id_to_index` because that tensor is in the *previous* step's batch order. A request that was not in the previous batch keeps its placeholder and gets its token from `token_ids_cpu`, which `_apply_pending_token_writeback` fills in.
- **The D2H is deferred** to the worker's async output thread via `AsyncRBLNModelRunnerOutput`, instead of blocking the step that produced the tokens.
- **The sampler hands back a copy.** Async holds the returned tensor past the step boundary, and holding the sampling graph's *own* output there stalls that graph's next launch. Both sampling paths now go through one 2-slot ring; the top-k/top-p path was previously uncompensated.
- **`platform.py` refuses async** when its premises are missing — `VLLM_RBLN_USE_DEVICE_TENSOR=1` and `VLLM_RBLN_SAMPLER=1` carry the in-flight tokens, speculative decoding needs more than one sampled column, and the optimum runner has no support — and logs which premise failed rather than running a silently synchronous engine.

**Affected modules:** Platform, Runner (vLLM-native), Scheduler, Sampler. `optimum_model_runner.py` is touched only by a `_bookkeeping_sync` signature change; the optimum path's behaviour is unchanged.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Unit lane:

Check, in this order:

1. **The arm took.** The async log names `RBLNAsyncScheduler` once per DP rank and contains no `Ignoring --async-scheduling` line. Without this the comparison is two sync arms.
2. **Text is identical.** Join the arms' outputs on the prompt and compare exactly. Greedy makes any difference a defect, not noise.
3. **Output health.** Adjacent-equal token ids near 0%. A broken build breaks both arms identically, so check 2 alone does not catch it.
4. **Native golden.** Pearson correlation over the cached golden, threshold 0.99, same value on both arms.

`tests/native`: 1143 passed, 13 skipped, 1 xfailed. Three files need `pytest-timeout`, which is not a project dependency; they do not collect without it.

---

### 💬 Notes

**One known non-determinism, and it is not async's.** A greedy `--run-iter` run fails its bit-exactness check under async. The cause is the RBLN MoE op: `custom_moe_glu` does not return the same values for a row when the collective's padded token count `T = dp_size × num_tokens_padded` changes, and `T` changes with peer timing, because a rank that decodes while a peer prefills is padded to the peer's width. The same request therefore gets ~1% different MoE output depending on when its peers happen to prefill, and greedy argmax flips at near-ties (median first divergence around token 222 of 256).

Async only reaches that skew sooner; it does not create it. **A sync arm driven into the same skew is byte-identical to the async arm**, which is the control that rules async out. Upstream vLLM has the same class of property on GPU — `coordinate_batch_across_dp` pads to the max across DP under cudagraphs, and `VLLM_BATCH_INVARIANT` is opt-in — so this is left as it is rather than papered over here.


rebel-eunji (commented):
Naming-consistency pass over the async additions. Three themes: the same object carrying different names in different classes, abbreviations where the file already has a word, and attribute names that a search for the concept cannot find. Suggestions are kept independently applyable; where one rename's uses fall inside another suggestion's range, the follow-ups are noted in prose.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -191,6 +193,83 @@ def _copy_pooler_output(
...
+        pending_token_writeback: Any = None,
+        req_ids: list[str] | None = None,
+        placeholder_pos: dict[str, int] | None = None,
+        logprobs_tensors: Any = None,
+    ):
+        self._model_runner_output = model_runner_output
+        self._invalid_req_indices = invalid_req_indices
+        self._device_index = device_index
+        # For the token_ids_cpu write-back, applied by the main thread.
+        self._pending = pending_token_writeback
```
rebel-eunji: nit: For consistency, I think keeping the original name looks better.

```suggestion
        self._pending_token_writeback = pending_token_writeback
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1451,6 +1631,61 @@ def execute_model(
                 intermediate_tensors,
             )
 
+            # Device-tensor async token feedback (decode fast path). staged input_ids
+            # holds the scheduler's placeholders; scatter the real previous token from
+            # prev_sampled_token_ids, remapped via prev_req_id_to_index.
+            prev_sampled = self.input_batch.prev_sampled_token_ids
+            prev_map = self.input_batch.prev_req_id_to_index
```
rebel-eunji: nit: For consistency, I think keeping the original name looks better.

```suggestion
            prev_req_id_to_index = self.input_batch.prev_req_id_to_index
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1258,6 +1352,47 @@ def _sample(
...
+                    continue
+                toks = tokens[i]
+                req_state = self.requests.get(req_id)
+                if req_state is not None:
+                    off = start - req_state.num_prompt_tokens
+                    if 0 <= off <= len(req_state.output_token_ids) - len(toks):
+                        req_state.output_token_ids[off : off + len(toks)] = toks
+                idx = ib.req_id_to_index.get(req_id)
+                if idx is not None:
+                    ib.token_ids_cpu[idx, start : start + len(toks)] = toks
```
rebel-eunji: nit: `ib` is a little bit confusing.. Using input_batch is better.

```suggestion
        while True:
            try:
                # Atomic against the output thread's append, so no lock.
                req_ids, tokens, placeholder_pos = (
                    self._pending_token_writeback.popleft()
                )
            except IndexError:
                return
            if not placeholder_pos:
                continue
            for i, req_id in enumerate(req_ids):
                if i >= len(tokens) or not tokens[i]:
                    continue
                start = placeholder_pos.get(req_id)
                if start is None:
                    continue
                sampled_ids = tokens[i]
                req_state = self.requests.get(req_id)
                if req_state is not None:
                    output_offset = start - req_state.num_prompt_tokens
                    if (
                        0
                        <= output_offset
                        <= len(req_state.output_token_ids) - len(sampled_ids)
                    ):
                        req_state.output_token_ids[
                            output_offset : output_offset + len(sampled_ids)
                        ] = sampled_ids
                req_index = self.input_batch.req_id_to_index.get(req_id)
                if req_index is not None:
                    self.input_batch.token_ids_cpu[
                        req_index, start : start + len(sampled_ids)
                    ] = sampled_ids
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1451,6 +1631,61 @@ def execute_model(
...
+                    host = tokfb_host[:n_fb]
+                    host.copy_(prev_sampled[:n_fb, 0])
+                    staged_model_inputs.input_ids[:n_fb, 0].copy_(host)
+                elif rows:
+                    # index_put_ demands equal dtypes, unlike the copy_ above.
+                    # The sampling op's width is not fixed -- `rbln::argmax`
+                    # still returns int64 -- so cast to the staged buffer's.
+                    staged_model_inputs.input_ids[dsts, 0] = prev_sampled[rows, 0].to(
+                        staged_model_inputs.input_ids.dtype
+                    )
```
rebel-eunji: The variable with meaning is better.

```suggestion
                prev_rows, cur_rows, identity = [], [], True
                for cur_row, req_id in enumerate(req_ids):
                    prev_row = prev_map.get(req_id)
                    if prev_row is None:
                        identity = False
                        continue
                    if prev_row != cur_row:
                        identity = False
                    prev_rows.append(prev_row)
                    cur_rows.append(cur_row)
                if identity:
                    num_feedback = len(prev_rows)
                    # staged input_ids is a host tensor, so this is a D2H
                    # either way; the intermediate buffer keeps it contiguous.
                    tokfb_host = self._tokfb_host
                    if tokfb_host is None or tokfb_host.shape[0] < num_feedback:
                        tokfb_host = torch.empty(
                            prev_sampled.shape[0], dtype=prev_sampled.dtype
                        )
                        self._tokfb_host = tokfb_host
                    host = tokfb_host[:num_feedback]
                    host.copy_(prev_sampled[:num_feedback, 0])
                    staged_model_inputs.input_ids[:num_feedback, 0].copy_(host)
                elif prev_rows:
                    # index_put_ demands equal dtypes, unlike the copy_ above.
                    # The sampling op's width is not fixed -- `rbln::argmax`
                    # still returns int64 -- so cast to the staged buffer's.
                    staged_model_inputs.input_ids[cur_rows, 0] = prev_sampled[
                        prev_rows, 0
                    ].to(staged_model_inputs.input_ids.dtype)
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -272,6 +351,17 @@ def __init__(
         else:
             self.sampler = Sampler(self.model_config.logprobs_mode)
 
+        # Host hop for the token feedback; see the identity branch in
+        # execute_model.
+        self._tokfb_host = None
```
rebel-eunji: `_tokfb_host` is hard to discover: neither "token" nor "feedback" matches it, and the only place the full meaning appears is in a comment. Suggest a more descriptive name so it can be found by what it actually is. (The uses in `execute_model`'s identity branch would follow, but are left out here to keep the two changes independently applicable.)

```suggestion
        self._token_feedback_host = None
```

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -114,6 +137,9 @@ def __init__(
         self._compiled_rbln_topk_topp_sampler = compile_sampler(
             rbln_top_k_top_p_sample, compile_context
         )
+        # Staging buffers for the sampled tokens; see _ring_copy.
+        self._tok_stage: list[torch.Tensor] = []
+        self._tok_slot = 0
```
rebel-eunji: `_tok_stage`/`_tok_slot` have two problems: "tok" doesn't surface when searching "token", and "stage" clashes with the runner's `InputStager`/`staged_model_inputs` vocabulary even though they're unrelated. Since the docstring already describes this as a ring, renaming to `_ring`/`_ring_idx` would match both the mechanism and search terms. (The same rename would apply to `RBLNSampler.__init__` and the accesses in `_ring_copy`.)

```suggestion
        self._sampled_token_ring: list[torch.Tensor] = []
        self._ring_slot = 0
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -191,6 +193,83 @@ def _copy_pooler_output(
...
+        invalid_req_indices: list[int],
+        device_index: int,
+        pending_token_writeback: Any = None,
+        req_ids: list[str] | None = None,
+        placeholder_pos: dict[str, int] | None = None,
+        logprobs_tensors: Any = None,
+    ):
+        self._model_runner_output = model_runner_output
+        self._invalid_req_indices = invalid_req_indices
+        self._device_index = device_index
```
rebel-eunji: I think `device_index` is not used.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-eunji: How about extracting this into a function?

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-eunji: nit: the comment is outdated.

vllm_rbln/platform.py
```diff
@@ -393,6 +433,14 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             scheduler_config.scheduler_cls = (
                 "vllm_rbln.v1.core.optimum_scheduler.RBLNOptimumScheduler"
             )
+            # Optimum model runner doesn't support async scheduling.
+            if scheduler_config.async_scheduling:
+                logger.warning(
+                    "Ignoring --async-scheduling: the optimum model runner does "
+                    "not support it. Running synchronously. Unset "
+                    "VLLM_RBLN_USE_VLLM_MODEL=0 to use the runner that does."
```
rebel-eunji: typo
```suggestion
                    "VLLM_RBLN_USE_VLLM_MODEL=1 to use the runner that does."
```

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -300,17 +325,14 @@ def forward(
         else:
             # Gather the logprobs and ranks of the topk and sampled token.
             logprobs_tensors = self.gather_logprobs(
-                raw_logprobs, num_logprobs, token_ids=sampled
+                raw_logprobs, num_logprobs, token_ids=sampled.long()
```
rebel-eunji, line 329: `logprob_token_ids` + sampling now crashes at `sampler.py` (around L315). 

After removing the global `.long()` cast, only the `gather_logprobs` path re-casts token IDs. `gather_specific_token_logprobs` (upstream `sampler.py` ~L214) still receives int32 token indices and passes them to `gather`, which requires int64 indices (`RuntimeError: Expected dtype int64`).  
This is a regression reachable in sync/async/optimum paths.

Please ensure token indices are cast to int64 before all gather-based logprob paths (including `gather_specific_token_logprobs`), not only the `gather_logprobs` branch.

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -133,7 +159,8 @@ def forward(
                 "per-request generators. Ignoring generators."
             )
 
-        return self._compiled_rbln_topk_topp_sampler(logits, temperature, k, p), None
+        out = self._compiled_rbln_topk_topp_sampler(logits, temperature, k, p)
+        return _ring_copy(self, out), None
```
rebel-eunji: optimum shares rbln_sampler. Could you change the code to call _ring_copy only for non-optimum?

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -166,9 +193,12 @@ def __init__(
         self._compiled_greedy_sample = compile_sampler(
             rbln_greedy_sample, compile_context
         )
+        # Staging buffers for the sampled tokens; see _ring_copy.
+        self._tok_stage: list[torch.Tensor] = []
+        self._tok_slot = 0
 
     def greedy_sample(self, logits: torch.Tensor) -> torch.Tensor:
-        return self._compiled_greedy_sample(logits)
+        return _ring_copy(self, self._compiled_greedy_sample(logits))
```
rebel-eunji: Same here

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -37,6 +37,29 @@
...
+    prev_sampled_token_ids and AsyncRBLNModelRunnerOutput. Holding the sampling
+    graph's own output there makes that graph's next launch block on the device.
+    Two slots, because the output thread reads one while the next step writes
+    the other.
+    """
+    ring = owner._tok_stage
+    if not ring or ring[0].shape != out.shape or ring[0].device != out.device:
+        ring = [
+            torch.empty(out.shape, dtype=out.dtype, device=out.device) for _ in range(2)
+        ]
```
rebel-eunji: This can be inefficient: it allocates a new tensor at every step.

rebel-yskim: Not every step — it's guarded by ring[0].shape != out.shape, and out.shape is the sampling bucket, so a steady decode run allocates twice on its first step and never again.

 It does reallocate when the sampled row count changes (prefill↔decode, bucket switch): two [bucket] int32 tensors, next to a full forward. No measurement says that shows up.

 If you want that at zero too, the grow-only pattern _token_feedback_host already uses — reallocate only when ring[0].shape[0] < num_tokens, return ring[slot][:num_tokens] — is five lines. Say the word and I'll add it.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Renamed. The same deque answered to `_pending_token_writeback` on the runner side and `_pending` here.

[9c0644fa](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/9c0644fada7451388207fe7029bf3cadce160574)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Done, together with the rest of that block's locals -- the whole thing moved into `_apply_async_token_feedback`.

[4553d758](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/4553d758fb63608bffc978af373cd113a1573616)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Done: `input_batch`, plus `placeholder_pos` / `sampled_ids` / `output_offset` / `req_index`.

[9c0644fa](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/9c0644fada7451388207fe7029bf3cadce160574)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Done -- `prev_rows` / `cur_rows` / `num_feedback`, in the extracted method.

[4553d758](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/4553d758fb63608bffc978af373cd113a1573616)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Renamed to `_token_feedback_host`, and the comment now names the method the buffer is used in rather than `execute_model`.

[4553d758](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/4553d758fb63608bffc978af373cd113a1573616)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Done: `_sampled_token_ring` / `_ring_slot`, here and in `RBLNSampler.__init__` and `_ring_copy`.

[9c0644fa](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/9c0644fada7451388207fe7029bf3cadce160574)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Right -- it came over from the CUDA wrapper, which needs it to pick a stream and an event. This one does a blocking copy on the output thread. Dropped, along with the `self.device.index` line at the call site.

[23313934](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/233139341ed0a7b1d6b820821ba001790b9f82b5)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Done -- `_apply_async_token_feedback`. The four preconditions are early returns now and the invariants moved into the docstring.

[4553d758](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/4553d758fb63608bffc978af373cd113a1573616)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Removed. That is the TODO this PR closes.

[23313934](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/233139341ed0a7b1d6b820821ba001790b9f82b5)

vllm_rbln/platform.py
rebel-yskim: Fixed, with one wording change: `Unset ... =1` reads against itself, so I used the same phrasing as the other switch hint in this file -- `Set VLLM_RBLN_USE_VLLM_MODEL=1 to use the runner that does.`

[7c557e6f](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/7c557e6fcd9a419fb70008cda6b8b60c04f107ab)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Confirmed and fixed. `gather_specific_token_logprobs` does `logprobs.gather(-1, sampled.unsqueeze(-1))`, which needs int64; the `token_ids_tensor[:, 0] = sampled` two lines above it casts on assignment, so the failure lands on the gather.

Cast at the call site rather than restoring the global `.long()`, which would bring back the int64 -> int32 round trip on every step. `gather_logprobs` keeps its `assert token_ids.dtype == torch.int64`.

[969cd8db](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/969cd8db05de911e915628b8a5ac9c0289a85249)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Gated -- but on `use_async_scheduling` rather than on the model path. The ring exists because async holds the sampled tensor past the step boundary, so that is the property it depends on, and AGENTS.md keeps `VLLM_RBLN_USE_VLLM_MODEL` branches out of here. optimum takes the default `False` and pays nothing.

[925f152e](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/925f152e00a754df982b44f3a2e72f592fb42850)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Same gate, via the `use_async_scheduling` flag.

[925f152e](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/925f152e00a754df982b44f3a2e72f592fb42850)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: nit: dtype checking doesn't seem necessary for now, but it'd be safer to check it in the code.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1342,6 +1623,10 @@ def _bookkeeping_sync(
             )
 
             self.input_batch.token_ids_cpu[req_idx, start_idx:end_idx] = sampled_ids
+            if self.use_async_scheduling:
+                # Remember where the placeholder landed. num_tokens_no_spec moves on
+                # every step, so the position cannot be re-derived later.
```
rebel-eunji, line 1541: nit: this relies on exactly one sampled token per step without stating it, which makes the neighboring generic-length code misleading.

```
assert num_sampled_ids == 1
```

rebel-eunji (approved):
It works well. LGTM 👍

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -37,6 +37,29 @@
 logger = init_logger(__name__)
 
 
+def _ring_copy(owner: nn.Module, out: torch.Tensor) -> torch.Tensor:
```
rebel-jonghewk: instead of putting the ring copy logic inside rbln_sampler, is it possible to add it in the runner? (right after _sample() returns?). For me, async scheduling seems to be a runner property.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -191,6 +193,81 @@ def _copy_pooler_output(
...
+                (self._req_ids, valid_sampled_token_ids, self._placeholder_pos)
+            )
+
+        output = self._model_runner_output
+        output.sampled_token_ids = valid_sampled_token_ids
+        if self._logprobs_tensors is not None:
+            # tolists() is where the logprobs D2H actually happens - on this
+            # thread, off the step's critical path.
+            output.logprobs = self._logprobs_tensors.tolists()
+        return output
```
rebel-jonghewk: How about separating this to a new file like.. v1/worker/async_output.py? model_runner.py seems to be growing

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Yes, and done -- the ring is now `RBLNModelRunner._stage_sampled_tokens`, applied to the sampler output at the end of `_sample()`.

You put your finger on why it was awkward where it was. In the sampler it needed a `use_async_scheduling` flag on both sampler classes purely to keep the optimum path from paying for a buffer it never reads, and a module-level helper reaching into `owner._sampled_token_ring` on someone else's object. In the runner it needs neither: the optimum path does not go through this runner at all, so the gate is structural.

It sits at the end of `_sample()` rather than at the call site because the intermediate-chunked-prefill early return above hands back a dummy tensor, not the sampling graph's output -- staging that would only make the ring thrash between its dtype and the sampler's.

[ad8a56e1](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/ad8a56e1acb0a0a8b4ecddc7d6241ef44f767850)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Added, and it came along with the move: the staleness check is now shape, dtype and device, so a buffer is never silently reused for a tensor of a different width -- `copy_` would have cast rather than complained.

The same commit also makes the buffers grow-only (allocated for a full batch, then sliced), which answers the allocation question above properly: a step whose batch shrank now reuses them instead of throwing both away.

[ad8a56e1](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/ad8a56e1acb0a0a8b4ecddc7d6241ef44f767850)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Added, inside the async branch. The comment above it says which way it would break -- `end_idx`, and with it `num_tokens_no_spec`, would be short if a step ever sampled more than the one -1 -- and names speculative decoding as the case that would, which the platform refuses under async.

[96a7c3a4](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/96a7c3a440c6fa7d529335288d6da946c505f454)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Done -- `vllm_rbln/v1/worker/async_output.py`. It shares nothing with the runner but its constructor arguments, so it moved as-is; the file-level docstring now says when vLLM calls `get_output()` and on which thread.

[8fa3f5b5](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/8fa3f5b5c7451e70161292c7ff0a2d0423d6e1b8)

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -300,17 +295,14 @@ def forward(
...
-                raw_logprobs, num_logprobs, token_ids=sampled
+                raw_logprobs, num_logprobs, token_ids=sampled.long()
             )
 
         # If we have both num_logprobs and logprob_token_ids, prefer
         # logprob_token_ids as it's more specific
         if logprob_token_ids_tensors is not None and num_logprobs is not None:
             logprobs_tensors = logprob_token_ids_tensors
 
-        # Use int32 to reduce the tensor size.
```
rebel-jonghewk, line 303: @rebel-eunji is it okay for optimum path?

vllm_rbln/v1/worker/async_output.py
```diff
@@ -0,0 +1,99 @@
...
+from vllm.v1.outputs import AsyncModelRunnerOutput, ModelRunnerOutput
+
+
+class AsyncRBLNModelRunnerOutput(AsyncModelRunnerOutput):
+    def __init__(
+        self,
+        model_runner_output: ModelRunnerOutput,
+        sampled_token_ids: torch.Tensor,
+        invalid_req_indices: list[int],
+        pending_token_writeback: Any = None,
```
rebel-jonghewk: nit: what is the dtype for pending_token_writeback?

vllm_rbln/v1/sample/rbln_sampler.py
rebel-eunji: It's okay. And there's a minor fix+refactoring suggestion for both type1 and type2. I'll make it as a following PR.

rebel-jonghewk (commented):
Can you check if everything is okay w/ async_scheduler + subblock prefix caching? 

```python
class TestSubBlockIndexingUnderAsyncScheduling:
    # Async scheduling runs schedule(N+1) before update_from_output(N). The
    # prefill's pending-indexing note (num_full_blocks_before) is replaced by
    # the donor's first decode schedule, whose num_computed_tokens already
    # counts the prompt -- so the drain scans past the prompt's full blocks and
    # their sub-blocks never reach the index. The sync arm is the control.
    @staticmethod
    def _sharer_schedule_after_donor_prefill(async_scheduling: bool):
        sched = create_rbln_scheduler(
            enable_prefix_caching=True,
            block_size=16,
            sub_block_size=8,
            max_num_batched_tokens=128,
            num_blocks=10000,
            async_scheduling=async_scheduling,
        )
        donor = make_request("0", list(range(16)), 16, max_tokens=20)
        sched.add_request(donor)
        out_prefill = sched.schedule()
        if async_scheduling:
            # The engine's batch queue schedules the donor's first decode
            # before the prefill's update_from_output arrives.
            out_decode = sched.schedule()
            assert out_decode.num_scheduled_tokens == {"0": 1}
        sched.update_from_output(
            out_prefill, make_model_runner_output(out_prefill, 0)
        )

        # Shares the donor's first 8-token sub-block, then diverges mid-block:
        # only the sub-block index can serve this prefix.
        sharer = make_request("1", list(range(8)) + [100 + i for i in range(16)], 16)
        sched.add_request(sharer)
        return sched.schedule()

    @pytest.mark.parametrize("async_scheduling", [False, True])
    def test_prompt_sub_blocks_survive_the_prefill_drain(self, async_scheduling):
        out = self._sharer_schedule_after_donor_prefill(async_scheduling)
        assert len(out.kv_cache_copy_ops) == 1
        assert out.kv_cache_copy_ops[0].num_tokens == 8
```

vllm_rbln/v1/worker/async_output.py
rebel-yskim: its type is `deque[tuple[list[str], list[list[int]], dict[str, int]]]`. I'll define it as `PendingTokenWriteback`, and use it instead of `Any`.

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: following fix (functionality issue) applied in [93be8e3](https://github.com/RBLN-SW/vllm-rbln/pull/951/commits/93be8e3c2621148841c06816a41018afe0692848), and re-visiting performance now. fyi, @rebel-jonghewk

vllm_rbln/v1/sample/rbln_sampler.py
```diff
@@ -277,17 +277,12 @@ def forward(
         sampled, processed_logprobs = self.sample(logits, sampling_metadata)
         if processed_logprobs is not None:
             raw_logprobs = processed_logprobs
-        # Convert sampled token ids to int64 (long) type to ensure compatibility
```
rebel-jonghewk, line 272: The dtype-cast removal quietly moved the step's device wait out of the metrics window, and I think patches/metrics.py has to move with it.

Before this PR, sampled.long() in RBLNSampler.forward was the step's first blocking read of the sampler output, so the wait for forward + sampler landed inside _sample — which is exactly what patches/metrics.py times as the SAMPLE half of MODEL + SAMPLE. With the casts gone, the first blocking read under sync scheduling is this tolist(), which sits outside every timed window. Three consequences:

- MODEL + SAMPLE degrades to dispatch cost (sub-ms), no longer device wall.
- E2E − (MODEL + SAMPLE), the "host overhead around the graphs" reading, silently absorbs the entire device wait — same trap class as the DP padding all_reduce sitting inside E2E but outside MODEL + SAMPLE.
- Sync-mode numbers before and after this commit are not comparable, and the perf CI parses these fields; the post-merge drop in MODEL + SAMPLE will read as a speedup that didn't happen.

How about wrapping _bookkeeping_sync in patches/metrics.py and add its wall to the graph slice ? 
cc @rebel-yskim @rebel-daeyang @rebel-cjlee

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: [93be8e3](https://github.com/RBLN-SW/vllm-rbln/pull/951/commits/93be8e3c2621148841c06816a41018afe0692848) caused blocking async (d2h block), so I've applied new changes: [2a2809c](https://github.com/RBLN-SW/vllm-rbln/pull/951/commits/2a2809c6d3ebb1532d6fff0c75a1914aae0ed76f).
Could you please check these? @rebel-jonghewk 
There are some patches in rbln_sampler. (keep buffer copy in rbln_sampler)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-yskim: Could you please check metric changes: [8c8c0be](https://github.com/RBLN-SW/vllm-rbln/pull/951/commits/8c8c0bed8acfc27bb138196204a6092b443622d2) ? @rebel-daeyang

vllm_rbln/v1/sample/rbln_sampler.py
rebel-jonghewk: @rebel-yskim I still think moving this to the model runner would be cleaner, but if there are functionality concerns, let's keep it as is for now. Could you add a TODO so we can revisit later?

vllm_rbln/platform.py
```diff
@@ -252,10 +252,43 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+                "path feeds one sampled token per step back into the next. "
+                "Running synchronously."
+            )
+            scheduler_config.async_scheduling = False
+
+        # TODO(yskim): Support PP on RBLN under async scheduling.
+        if (
+            scheduler_config.async_scheduling
+            and parallel_config.pipeline_parallel_size > 1
+        ):
```
rebel-cjlee: These three gates sit above the model-path branch, so they run on the optimum path too. The NOTE right below at 295 line spells out the convention for what belongs at this level.

Run optimum with `VLLM_RBLN_USE_DEVICE_TENSOR=0` and you get the line 258 warning instead: the first gate turns async off first, so line 438's actual reason ("the optimum model runner does not support it") never prints.

Could these three move into the native branch, just above the `scheduler_cls` selection at line 338? That puts the decision next to the only thing that consumes it and leaves optimum with just line 438. Nothing between the gates and line 338 reads `async_scheduling`, so behavior is unchanged but only the diagnostic.

vllm_rbln/platform.py
rebel-cjlee, line 1: Async refusal messages (line 258, 273, 286, 439)

All four open with "Ignoring --async-scheduling: ...", but vLLM resolves an unset `async_scheduling` to `True` in `vllm/config/vllm.py:964-1004`, before this hook runs. So any of them can fire on a run that never passed the flag, and the default optimum path hits line 439. "Disabling asynchronous scheduling: ..." stays true either way.

vllm_rbln/platform.py
```diff
@@ -252,10 +252,43 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        if scheduler_config.async_scheduling and not (
+            envs.VLLM_RBLN_USE_DEVICE_TENSOR and envs.VLLM_RBLN_SAMPLER
+        ):
             logger.warning(
-                "Asynchronous scheduling is not supported on RBLN. "
-                "Overriding scheduler_config.async_scheduling to False."
+                "Ignoring --async-scheduling: it requires "
+                "VLLM_RBLN_USE_DEVICE_TENSOR=1 (got %s) and VLLM_RBLN_SAMPLER=1 "
+                "(got %s), which carry the in-flight sampled tokens. Running "
+                "synchronously.",
```
rebel-cjlee: What breaks with `VLLM_RBLN_USE_DEVICE_TENSOR=1` and `VLLM_RBLN_SAMPLER=0`?
"which carry the in-flight sampled tokens" reads clearly for the first one, but I can't see what the second is carrying. When `VLLM_RBLN_SAMPLER=0`, the sampler still returns a device tensor, just a freshly allocated one rather than a compiled graph's resused output buffer, so the reason `_stage_into` exists doesn't apply, and the runner already skips staging for it. If I miss something, please let me know.

tests/native/test_platform.py
```diff
@@ -299,8 +303,63 @@ def test_only_a_lone_none_is_cleared(self):
...
+    ):
+        """Either env var off means async has no way to carry its in-flight tokens.
+
+        VLLM_RBLN_USE_DEVICE_TENSOR gates the feedback scatter that replaces the
+        scheduler's -1 placeholders; VLLM_RBLN_SAMPLER gates the ring the output
+        thread reads. Without them the runner decodes from a token that was never
+        sampled and returns wrong text with no error raised, so the platform
+        downgrades to sync rather than run the combination.
+        """
+        monkeypatch.setenv(switched_off, "0")
```
rebel-cjlee, line 348: `--device-tensor` sets `VLLM_RBLN_USE_DEVICE_TENSOR` for the whole session, so pinning only the variable being switched off leaves the other one to the lane. The gate refuses if either is 0, so under `--device-tensor 0` the `VLLM_RBLN_SAMPLER` case passes with the sampler left on too. That parametrization proves nothing there.

Pinning both would keep each case to one variable in either lane:
```suggestion
        for name in ("VLLM_RBLN_USE_DEVICE_TENSOR", "VLLM_RBLN_SAMPLER"):
            monkeypatch.setenv(switched_off, "0")
```

tests/native/test_platform.py
```diff
@@ -299,8 +303,63 @@ def test_only_a_lone_none_is_cleared(self):
 
 
 class TestSchedulerOverrides:
-    def test_async_scheduling_is_forced_off(self):
-        assert _build(async_scheduling=True).scheduler_config.async_scheduling is False
+    def test_async_scheduling_is_honored(self):
+        # The platform used to force this off unconditionally. It now follows
+        # vLLM's --async-scheduling, as long as the device-side token path is
+        # available (see below).
+        assert _build(async_scheduling=True).scheduler_config.async_scheduling is True
```
rebel-cjlee: This fails under `--device-tensor 0`. The conftest sets `VLLM_RBLN_USE_DEVICE_TENSOR=0` for the session, the first gate refuses, and `async_scheduling` comes back `False` while this asserts `True`. CI does not pass the option currently, so it stays green, but the conftest help calls running the suite once per value the intended workflow.

Setting both `VLLM_RBLN_USE_DEVICE_TENSOR` and `VLLM_RBLN_SAMPLER` to `"1"` with `monkeypatch.setenv` would make it lane independent, matching what the sibling test does for the negative cases.

tests/native/test_platform.py
```diff
@@ -265,16 +265,20 @@ def test_an_explicit_worker_is_respected(self):
...
             )
         )
-        assert (
-            config.scheduler_config.scheduler_cls
-            == "vllm_rbln.v1.core.rbln_scheduler.RBLNScheduler"
+        expected = (
+            "vllm_rbln.v1.core.rbln_scheduler.RBLNAsyncScheduler"
+            if config.scheduler_config.async_scheduling
+            else "vllm_rbln.v1.core.rbln_scheduler.RBLNScheduler"
         )
```
rebel-cjlee: The expected value is read from the same config being checked, so the assertion agrees with whatever the platform decided and cannot fail.

Separately, no test says what a plain `_build()` (no `--async-scheduling`) ends up with. That is the part this PR changes: on the native path it now comes out `True` and selects `RBLNAsyncScheduler`. One assertion on that would be worth having.

vllm_rbln/v1/worker/async_output.py
```diff
@@ -0,0 +1,96 @@
...
+        # The -1 placeholders the async path left in token_ids_cpu still have to be
+        # replaced by the real tokens, but not from this thread: the main thread
+        # reads token_ids_cpu in _preprocess. Queue the tokens instead and let the
+        # main thread apply them at the top of its next step.
+        self._pending_token_writeback.append(
+            (self._req_ids, valid_sampled_token_ids, self._placeholder_pos)
+        )
+
+        output = self._model_runner_output
+        output.sampled_token_ids = valid_sampled_token_ids
```
rebel-cjlee, line 96: `valid_sampled_token_ids` goes to two consumer here: the queue and the returned output. Is that a problem? The scheduler trims these lists in place when a request stop (`vllm/v1/core/sched/scheduler.py:1863`), so the queue holds lists someone else mutates. Copying on the queued side would decouple them cheaply if it is worth doing.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -262,10 +268,12 @@ def __init__(
         )
         self.runtime_holder: list = []
 
+        # bool(), not the field: vLLM resolves an unset value before this runs,
+        # but a directly-built config (a unit test) can still carry None.
+        self.use_async_scheduling = bool(self.scheduler_config.async_scheduling)
```
rebel-cjlee: Is this reachable? I put `aasert self.scheduler_config.async_scheduling is not None` here and the whole of `tests/native` passed, and the optimum builders pass an explicit bool. `None` is falsy so dropping `bool()` changes nothing at any use site, and mypy passes without it on both runners.

vllm_rbln/v1/worker/optimum_model_runner.py
```diff
@@ -255,7 +255,7 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         )
 
         # FIXME enable async scheduling for optimum model runner
-        self.use_async_scheduling = self.scheduler_config.async_scheduling
+        self.use_async_scheduling = bool(self.scheduler_config.async_scheduling)
```
rebel-cjlee: Same with `rbln_model_runner.py:271`.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-cjlee, line 266: It is not in this diff, but this PR is what makes it reachable: `num_output_placeholders` is always 0 without async, so the branch never ran before. The arithmetic there decides whether a request gets one more step, so getting it wrong either wastes a step or ends the request short of `max_tokens`, and it is RBLN's own logic rather than something `AsyncScheduler` covers. `create_rbln_scheduler(async_scheduling=True)` already exists, so counting the scheduled steps for a request with a small `max_tokens` against a sync arm would pin it. Could you add one?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -1241,24 +1259,146 @@ def _sample(
...
+                        <= len(req_state.output_token_ids) - len(sampled_ids)
+                    ):
+                        req_state.output_token_ids[
+                            output_offset : output_offset + len(sampled_ids)
+                        ] = sampled_ids
+                req_index = input_batch.req_id_to_index.get(req_id)
+                if req_index is not None:
+                    input_batch.token_ids_cpu[
+                        req_index, start : start + len(sampled_ids)
+                    ] = sampled_ids
```
rebel-cjlee: The bounds check guards the first store but not the second, so if it fails `output_token_ids` keeps its `-1` while `token_ids_cpu` gets the real tokens. Since `token_ids_cpu` is rebuilt from `output_token_ids` on re-entry, that `-1` comes back and reaches the model as an input token. `_update_states` trims `output_token_ids` on a KV load failure or async spec decode; spec decode is refused here, but the first looks reachable. Should both stores share the condition, or should the check be an assert?

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-cjlee, line 1: None of the async repairs have unit tests: `_repair_staged_input_ids`, `_repair_async_output_token_ids`, `_apply_pending_token_writeback`, and `AsyncRBLNModelRunnerOutput.get_output`. [4f5cabcf](https://github.com/RBLN-SW/vllm-rbln/pull/951/commits/4f5cabcf538dbe96e76c73a0c51b6919beef4581) fixed a row crossing bug in the first of those.

tests/native/v1/worker/test_rbln_model_runner_inputs.py
```diff
@@ -103,6 +103,9 @@ def test_parses_accepted_tokens_and_drops_placeholders(
             mr, "get_pp_group", lambda: SimpleNamespace(is_last_rank=True)
         )
         runner = make_model_runner()
+        # This class covers the synchronous bookkeeping path, and vLLM now
+        # resolves an unset --async-scheduling to enabled, so pin it.
+        runner.use_async_scheduling = False
```
rebel-cjlee: Could this be `make_model_runner(async_scheduling=False)`? `**config_overrides` forwards it to `EngineArgs`, so it goes through the real config path, and it won't silently go back to async if the attribute is ever renamed.

rebel-cjlee (commented):
Sorry for the slow reivew. Overall this looks good to me. I left small comments. The one I'd most like your read on is the writeback guard in `_apply_pending_token_writeback`, since the two stores can drift apart there. The rest are questions or suggestions, none of them blocking.

vllm_rbln/v1/worker/async_output.py
rebel-yskim: What has been keeping it harmless is only that the inner lists are one token wide, and that assumption was from spec-decode and PP gates. Too far away to rely on, so I've applied to copy the queued side. Please check.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Yes, it was reachable - but maybe shorter than the trim alone. The repair runs at the top of the step (ahead of `_update_states`), and async hands the queued tokens back only after the next forward is dispatched, so an entry can outlive one _update_states. The tokens turn out not to keep the -1, instead, the trim deletes it. 
Still, there would be same drift condition, so I've fixed it to share the condition - please check.

As you said in the comment, I thought that it would be the most critical point in the review, so I've applied this comment (+ async_output.py comment) at first - ~~I'm still applying other comments now. Would you please check them first? I'll apply other comments ASAP.~~ Done.

tests/native/test_platform.py
rebel-yskim: [621f179](https://github.com/RBLN-SW/vllm-rbln/pull/951/commits/621f1795d7e5279d82fc6f594020ce031c2cdca8)

vllm_rbln/v1/sample/rbln_sampler.py
rebel-daeyang: Confirmed that adding `_bookkeeping_sync`'s time makes MODEL + SAMPLE come out right in sync mode.
Under async mode, the step's own sampler wait moves out to `get_output`, so neither section reads as decode+sample latency any more and I'd rather revisit them in a follow-up PR.

vllm_rbln/platform.py
rebel-yskim: Moved. Checked the premise before trusting it: between the gates and the scheduler_cls selection nothing reads `async_scheduling` at all -- that selection is its only reader -- so this is diagnostics only, and both test lanes come out unchanged.

Optimum now prints just its own refusal.

[18718861](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/18718861cf5931cd6ed9481c229217688ae316da)

vllm_rbln/platform.py
rebel-yskim: All four say "Disabling asynchronous scheduling" now. You are right that the old wording could name a flag nobody passed, and the default optimum path is exactly where it happened.

[18718861](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/18718861cf5931cd6ed9481c229217688ae316da)

vllm_rbln/platform.py
rebel-yskim: You are not missing anything -- the message was lumping the two together. Split: VLLM_RBLN_USE_DEVICE_TENSOR carries the in-flight tokens, VLLM_RBLN_SAMPLER keeps the sampler on the device so they never reach the host mid-step. The runner already skips staging when the sampler is not RBLNSampler, as you say.

[18718861](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/18718861cf5931cd6ed9481c229217688ae316da)

tests/native/test_platform.py
rebel-yskim: Taken. Both are set on first, then one is switched off, so each case varies exactly one variable in either lane.

Worth recording how vacuous it was: dropping the VLLM_RBLN_SAMPLER term from the gate leaves the old case passing on the cpu lane and makes the new one fail.

[cfb18a25](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/cfb18a25ad4759511769f77f6f9a971baad066db)

tests/native/test_platform.py
rebel-yskim: Both parts done. The expectation is no longer read off the config under test -- a carrier is pinned off and the sync scheduler named outright.

Your second point became its own case rather than an extra assertion here: with both carriers on and no flag passed anywhere, async_scheduling resolves True and the native path selects RBLNAsyncScheduler. It fails if that selection is pointed back at the sync scheduler, so the two cases now pin one side each.

[c03661cf](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/c03661cfe408f8ae33402eddeb551b55323bb609)

tests/native/v1/worker/test_rbln_model_runner_inputs.py
rebel-yskim: Done -- it goes through EngineArgs now. Confirmed the kwarg actually lands: make_model_runner(async_scheduling=False) comes out with use_async_scheduling False.

[0bf60eda](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/0bf60eda25fad93025812cb092a18dc03d802e0b)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: Dropped, and the comment with it -- it claimed a unit test could still carry None, and none does. Reproduced both halves of your check: asserting `is not None` at both sites leaves tests/native green, and mypy passes without the cast. The production side is covered too, since vLLM's post-init resolves an unset value to an explicit bool in every branch before a runner exists.

One caveat on the mypy half: the hook runs with --follow-imports skip, so it never sees the upstream `bool | None`. The runtime evidence is what this rests on.

[1faf8ef9](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/1faf8ef9f9ff34846f9a57c612455f300c75d68c)

vllm_rbln/v1/worker/optimum_model_runner.py
rebel-yskim: Same change, same commit -- one line and its comment in each runner. Behaviour moves on neither path.

[1faf8ef9](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/1faf8ef9f9ff34846f9a57c612455f300c75d68c)

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-yskim: All four are covered now. _apply_pending_token_writeback and get_output went in earlier; this adds _repair_staged_input_ids and _repair_async_output_token_ids, on the same stub -- they need only prev_req_id_to_index and the staged tensor.

The row crossing you pointed at is the case worth naming. When the first request has no previous row, the ones that do sit at rows 1 and 2, and their previous rows are also 1 and 2 -- so the row lists match and only the any_skipped guard keeps the write off row 0. Comparing the lists alone is not enough, which is what the original bug was. Reverting that guard fails this case and nothing else.

[6cd00f0c](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/6cd00f0c0db50d8b2364fa97e741be1284b8baee)

vllm_rbln/v1/core/rbln_scheduler.py
rebel-yskim: Added, with sync as the control -- step count and each request's own max_tokens, single and mixed batches. Making the check more conservative adds a step and making it more eager ends a request a token short; as written it does neither.

The serial drain could not reach the branch, so the test keeps one output in flight and asserts the placeholders are above zero. The draft-rejection slack stays uncovered while async refuses spec decode.

[e2fbd0e1](https://github.com/RBLN-SW/vllm-rbln/pull/951/changes/e2fbd0e1f42c9ea81c986819d6e97b554aaa25e1)

rebel-cjlee (approved):

PR #955, opened 2026-08-20
fix(dp): rendezvous the DP ranks at the end of warm-up

rebel-yskim:
## Symptom

`PREFILL E2E` reports a p99 several times its median, while `PREFILL + SAMPLE` stays flat on all four ranks. There is no request queueing in that run, so the gap had no obvious explanation.

## Cause

Every forward pass blocks in a CPU all-reduce on the DP group, inside `_determine_batch_padding`. That all-reduce sits **inside** the E2E window and **outside** the MODEL + SAMPLE window, so a wait there looks exactly like slow prefill.

The dummy runs inside `warmup_model()` go through that same all-reduce, and they do leave the ranks aligned — to 0.2 ms. But the sampler warm-up that follows them (`_dummy_sampler_run`, `_warmup_sampler_decode_batches`) is not collective and runs for seconds, so ranks leave warm-up hundreds of ms apart again. Stamping four checkpoints, with a running count of DP all-reduces executed:

The first forward of a run is always a prefill, so that skew lands on exactly one prefill pass per rank. Traced, all four ranks leave that first all-reduce at the *same instant*; only their start times differ:

With 8 prefill passes per run, that single pass is both the p90 and the p99 — at n=8 numpy's p90 interpolates 30% of the way into the maximum.

## Fix

Barrier once at the end of `warmup_model()`, so the skew stays in initialisation. Nothing is made faster: the engine is not honestly ready until every rank is, and the wait simply stops hiding inside the first request.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -21,13 +21,14 @@ from typing import Any, Literal, NamedTuple, TypeAlias, cast
 
 import numpy as np
 import torch
+import torch.distributed as dist
 from vllm.config import VllmConfig, get_layers_from_vllm_config
 from vllm.config.cache import CacheConfig
 from vllm.distributed.kv_transfer import (
     get_kv_transfer_group,
     has_kv_transfer_group,
 )
-from vllm.distributed.parallel_state import get_pp_group
+from vllm.distributed.parallel_state import get_dp_group, get_pp_group
 from vllm.model_executor.layers.attention import Attention
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.model_executor.model_loader import get_model_loader
@@ -3124,6 +3125,12 @@ class RBLNModelRunner(KVConnectorModelRunnerMixin):
                 self.drafter.dummy_run()
 
         mega_cache.save(self.model_config.model, sig)
+
+        # NOTE(RBLN): the sampler warm-up above is not collective, so ranks leave
+        # warm-up hundreds of ms apart. Without this, the first forward's own
+        # all-reduce absorbs that skew and bills it to the first request's prefill.
+        if self.parallel_config.data_parallel_size > 1:
+            dist.barrier(group=get_dp_group().cpu_group)
 
     def _process_kv_cache_copy_ops(
         self,
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -21,14 +21,13 @@ from typing import Any, Literal, NamedTuple, TypeAlias, cast
 
 import numpy as np
 import torch
-import torch.distributed as dist
 from vllm.config import VllmConfig, get_layers_from_vllm_config
 from vllm.config.cache import CacheConfig
 from vllm.distributed.kv_transfer import (
     get_kv_transfer_group,
     has_kv_transfer_group,
 )
-from vllm.distributed.parallel_state import get_dp_group, get_pp_group
+from vllm.distributed.parallel_state import get_pp_group
 from vllm.model_executor.layers.attention import Attention
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.model_executor.model_loader import get_model_loader
@@ -3125,12 +3124,6 @@ class RBLNModelRunner(KVConnectorModelRunnerMixin):
                 self.drafter.dummy_run()
 
         mega_cache.save(self.model_config.model, sig)
-
-        # NOTE(RBLN): the sampler warm-up above is not collective, so ranks leave
-        # warm-up hundreds of ms apart. Without this, the first forward's own
-        # all-reduce absorbs that skew and bills it to the first request's prefill.
-        if self.parallel_config.data_parallel_size > 1:
-            dist.barrier(group=get_dp_group().cpu_group)
 
     def _process_kv_cache_copy_ops(
         self,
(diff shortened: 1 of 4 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3125,6 +3126,12 @@ def warmup_model(self) -> None:
 
         mega_cache.save(self.model_config.model, sig)
 
+        # NOTE(RBLN): the sampler warm-up above is not collective, so ranks leave
+        # warm-up hundreds of ms apart. Without this, the first forward's own
+        # all-reduce absorbs that skew and bills it to the first request's prefill.
+        if self.parallel_config.data_parallel_size > 1:
+            dist.barrier(group=get_dp_group().cpu_group)
+
```
rebel-cjlee: How about the end of `compile_or_warm_up_model` instead? `warmup_model` returns into NIXL registration, CPU affinity and `set_random_seed`, all per-rank and none of them collective, so on the NIXL path the ranks can drift apart again before the first request.

```suggestion
```

Then in the worker, after the `try/except/finally` rather than inside `finally`, so a rank that fails warm-up raises instead of joining the barrier:

```python
        # inside compile_or_warm_up_model
        ...

        set_random_seed(self.model_config.seed)

        # Measured before the wait: a rank's compile time must not carry its
        # peers' skew.
        elapsed = time.perf_counter() - st

        # NOTE(RBLN): nothing left between here and the first request is
        # collective, so skew surviving this point lands in the first forward's
        # DP all-reduce and is billed to that request's prefill.
        if self.parallel_config.data_parallel_size > 1:
            logger.info("Warm-up done; waiting for the other DP ranks.")
            dist.barrier(group=get_dp_group().cpu_group)
            logger.info("All DP ranks left warm-up.")

        return CompilationTimes(language_model=elapsed, encoder=0.0)
```
The two log lines are there so a peer that died in warm-up reads as a wait rather than a silent boot hang.

Moving it also simplifies the test. `test_rbln_worker.py`'s `_worker` helper already records order in `calls`, so it can assert `["warmup", "barrier"]` and pin the position, which the current test cannot. The real-runner fixture and the `maybe_use_device` mark go away with it.

Worth deciding whether the barrier should run on the skip path too. At the end of the method it also fires when warm-up is skipped, under `enforce_eager`, `VLLM_RBLN_COMPILE_MODEL=0` or `VLLM_RBLN_ENABLE_WARM_UP=0`. All three are global config, so the ranks still skip together and nothing hangs, but it is a behavior change either way. Inside the `else` if you would rather it did not.

rebel-yskim: 1. "How about the end of `compile_or_warm_up_model` instead?"
  -> Done. `finalize_kv_cache_registration` → `_register_kv_caches_impl` really is per-rank, so the ranks could separate again after `warmup_model()`.

2. "after the `try/except/finally` rather than inside `finally`, so a rank that fails warm-up raises"
  -> Kept that property, but from inside the `else` — a failing rank never reaches the barrier there either.

3. "Measured before the wait: a rank's compile time must not carry its peers' skew."
  -> `CompilationTimes.language_model` is reduced with `max()` in `Executor.initialize_from_config`, so the barrier leaves that value unchanged, and `@instrument(span_name="Warmup (NPU)")` wraps the whole
  method anyway.

4. "The two log lines are there so a peer that died in warm-up reads as a wait"
  -> Both in.

5. "it can assert `["warmup", "barrier"]` and pin the position... The real-runner fixture and the `maybe_use_device` mark go away with it."
  -> Moved to `TestCompileOrWarmUpModel`, asserts `["warmup", "barrier:dp"]`. `TestWarmupRendezvous` deleted; `_worker` takes `data_parallel_size`. Two of the four tests are mine, not yours — say if they over-cover.

6. "Inside the `else` if you would rather it did not [fire on the skip path]."
  -> Inside the `else`. Nothing compiles on the skip path, so there is no skew to absorb.

Could you please review about no.3?

rebel-cjlee: Your change also looks fine to me.

rebel-cjlee (approved):

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -1084,6 +1085,16 @@ def compile_or_warm_up_model(self) -> CompilationTimes:
...
                     finalize_kv_cache_registrations(get_kv_transfer_group())
 
+                # NOTE(RBLN): the sampler warm-up and the deferred KV-cache
+                # registration above are per-rank, so ranks reach this point
+                # hundreds of ms apart. Nothing left before the first request is
+                # collective, so that skew would otherwise land in the first
+                # forward's DP all-reduce and be billed to the prefill it runs.
+                if self.parallel_config.data_parallel_size > 1:
+                    logger.info("Warm-up done; waiting for the other DP ranks.")
+                    dist.barrier(group=get_dp_group().cpu_group)
```
rebel-jonghewk, line 1095: why scope this to DP? can it be the same situation for other model parallelisms?

rebel-yskim: TP and PP ranks are workers of one executor, and `Executor.initialize_from_config` calls `collective_rpc("compile_or_warm_up_model")` without `non_block`, so `MultiprocExecutor.collective_rpc`
  dequeues one response per worker before returning — that gather already is the rendezvous, and no skew on those axes survives into the first forward.

If there are some skewing cases in other parallelisms, we need to deep dive more 😢

rebel-jonghewk: Okay. Thanks for the explanation.

rebel-jonghewk (approved):

PR #958, opened 2026-08-20
fix(scheduler): reject a max_model_len the KV pool can never serve

### LINKED ISSUE #957: Prompts above the usable KV block pool are admitted and then hang the engine
## Summary

The usable KV cache pool is `(num_blocks - 1) * block_size` tokens: `BlockPool`
permanently reserves one block as its null block, so a request can never be given more
than the rest.

A prompt longer than that but shorter than `max_model_len` is accepted by admission and
then **never scheduled**: the scheduler retries its allocation on every step, forever,
with no error and no timeout. FCFS means every request queued behind it stops as well, so
a single such request freezes the engine process serving it.

The failure is silent. `/v1/completions` never returns and a stock build logs nothing
about it. `/metrics` keeps answering, with `vllm:num_requests_waiting` at 1 for reason
`capacity` and `vllm:kv_cache_usage_perc` at `0.0` because nothing was ever allocated.

Verified on both:

- vllm 0.22.0 + vllm-rbln 0.11.1a5.dev12+g1e49543d2 (field incident and live server demo)
- vllm 0.24.0 + vllm-rbln `dev` @1b138645 (unit reproduction, unmodified tree)

Field incident: MiniMax-M2.5 served with data parallel size 4 and expert parallelism,
`--num-gpu-blocks-override 13 --block-size 8192 --max-model-len 106496 --max-num-seqs 2`.
Multi-turn coding-agent sessions grow past 90k prompt tokens, and the round whose prompt
lands in `(98,304, 106,496)` stalls the engine.

## Root cause

Line numbers are vllm 0.24.0 and vllm-rbln `dev` @1b138645.

1. `vllm/v1/core/block_pool.py:191` — `BlockPool` permanently pops one null block from
   the free queue, so only `num_blocks - 1` blocks are ever allocatable.
2. `vllm/v1/engine/input_processor.py:404` — admission compares the prompt against
   `max_model_len` alone. Nothing compares `max_model_len` against the usable pool.
3. `vllm_rbln/v1/core/rbln_scheduler.py:730` — a new prefill is allocated with
   `full_sequence_must_fit=True`, so it reserves blocks for the whole sequence even under
   chunked prefill (intended design).
4. `vllm/v1/core/kv_cache_manager.py:386` — allocation returns `None` when the demand
   exceeds the free blocks, and free can never exceed `num_blocks - 1`.
5. `vllm_rbln/v1/core/rbln_scheduler.py:733` — the scheduler breaks out of the waiting
   loop and retries on the next step. Nothing detects that the condition is permanent,
   so the request never fails and never runs — 42.7 retries/s on the live server below.

## Affected configurations

The invariant is `num_blocks >= ceil(max_model_len / block_size) + 1`. Current standard
configurations all violate it by exactly one block:

| num_blocks | block_size | usable tokens | max_model_len | status |
|---|---|---|---|---|
| 13 | 8192 | 98,304 | 106,496 | **violated — field incident** |
| 14 | 8192 | 106,496 | 114,688 | violated above 106,496 tokens |
| 16 | 8192 | 122,880 | 131,072 | violated above 122,880 tokens |

`max_model_len` is exactly `num_blocks * block_size` in each, so all three boot without
complaint and stall on the first prompt above their usable capacity.

## Reproduction

Scheduler level, no device needed: build an `RBLNScheduler` with `block_size` 8192,
`num_blocks` 13 and `max_model_len` 106,496 set on `ModelConfig` (not only on
`SchedulerConfig` — the scheduler reads it from the former), add a request of any length
in `(98,304, 106,496)` — past the 12 usable blocks, still short enough to be admitted —
and call `schedule()` in a loop.

The request is never scheduled, on any step. A small request added behind it is never
scheduled either. With `num_blocks` 14 — one more than the prompt needs — the same
request is scheduled on the first step.

Live server (vllm 0.22.0 + vllm-rbln 0.11.1a5.dev12+g1e49543d2 with a logging-only patch
added to surface the cause, MiniMax-M2.5 at data parallel size 4, 20 blocks x 8192,
`max_model_len` 163,840, usable 155,648 tokens):

| request | result |
|---|---|
| 150,000 tokens (19 blocks) | HTTP 200 after 154 s |
| 160,000 tokens (20 blocks) | no response; `num_requests_waiting` 1 (`capacity`), `kv_cache_usage_perc` 0.0 |
| 1,000 tokens, queued behind it | no response — head-of-line blocked |

Aborting the client releases the queue and the server returns to a clean state.

## Proposed fix

Enforce the capacity invariant in `RBLNScheduler.__init__`, turning a silent hang into a
startup failure:

```
max_model_len (106496) exceeds the usable KV cache capacity (98304 tokens =
(13 blocks - 1 reserved null block) x 8192 tokens per block). Prompts above the usable
capacity would wait forever. Raise the KV cache block count (e.g.
--num-gpu-blocks-override) or lower --max-model-len.
```

A fix with regression tests is ready to submit. It also carries two test-side fixes the
invariant exposed: `create_rbln_scheduler` in `tests/native/v1/core/utils.py` passes
`max_model_len` to `SchedulerConfig` only, so every test in that suite silently runs at
opt-125m's native 2048; and several test configurations set `max_model_len` to
`num_blocks * block_size`, the same exact-fit shape as the configurations above.

A general starvation detector would be a useful safety net beyond this one fix: if the
waiting head fails allocation for a fixed number of consecutive steps with nothing
running and the pool at its maximum, the request should be failed rather than retried
forever.

rebel-hyungminkim:
### 🚀 Summary of Changes

The usable KV cache pool is `(num_blocks - 1) * block_size` tokens, because `BlockPool`
permanently reserves one block as its null block. Nothing checked `max_model_len` against
that, so a prompt in the gap was admitted and then failed allocation on every scheduling
attempt — waiting forever without erroring and head-of-line-blocking the engine. #957 has
the full analysis and reproduction.

`RBLNScheduler` now validates the invariant at construction and refuses the configuration:

```
max_model_len (106496) exceeds the usable KV cache capacity (98304 tokens =
(13 blocks - 1 reserved null block) x 8192 tokens per block). Prompts above the usable
capacity would wait forever. Raise the KV cache block count (e.g.
--num-gpu-blocks-override) or lower --max-model-len.
```

Two test-side fixes come with it, both exposed by the invariant:

- `create_rbln_scheduler` passed `max_model_len` to `SchedulerConfig` only, which the
  scheduler does not read, so every test in `tests/native/v1/core` silently ran at
  opt-125m's native 2048 whatever it asked for. It now reaches `ModelConfig`, and the
  default stays within the usable pool.
- 15 existing test configurations violated the invariant, 11 of them by setting
  `max_model_len` to exactly `num_blocks * block_size` — the same shape as the
  configurations in #957. Two further tests relied on the clamp: one needed a
  `max_model_len` above its prompt, and the full-prompt-must-fit test now creates pool
  pressure with a running decode instead of an oversized prompt, since a prompt longer
  than the pool can no longer be admitted at all.

Affected modules: Scheduler (`vllm_rbln/v1/core/rbln_scheduler.py`), native core tests.

---

* Resolves #957

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. `pytest tests/native/v1/core` — 195 passed, 2 skipped (192 passed before, plus the
   three new cases).
2. `pytest tests/native` — 87 failed, 1234 passed, 2 errors. The failing set is identical
   to the `dev` baseline; all of it predates this branch and lives in
   `tests/native/v1/worker`.
3. Remove the `_validate_kv_cache_capacity()` call and re-run
   `pytest tests/native/v1/core/test_rbln_scheduler.py::TestCapacityInvariant` —
   `test_max_model_len_beyond_usable_pool_is_rejected` fails, the other two pass.
4. Boundary check: `TestCapacityInvariant` also asserts that `max_model_len` equal to the
   usable capacity is accepted and that the longest admissible prompt still schedules, so
   the guard cannot go off by one in the other direction.

Verified against vllm 0.24.0+cpu on `dev` @1b138645.

---

### 📋 Checklist

* [ ] Relevant documentation has been updated (no user-facing docs cover KV block sizing)

---

### 💬 Notes

Two points worth a reviewer's attention:

- `create_rbln_scheduler` now defaults `max_model_len` to
  `min(max_num_batched_tokens, (num_blocks - 1) * block_size)` so the harness builds a
  valid configuration by default. The alternative is to leave the default alone and set
  `max_model_len` explicitly in the six tests that would otherwise fail; that keeps each
  test's intent visible in its own body at the cost of a larger diff. Say which you
  prefer and I will switch.
- The invariant assumes a single KV cache group at the scheduler's block size. A hybrid
  layout draws from the same pool for every group and so needs at least this much, which
  keeps the check conservative: it can miss a violation, never invent one. The docstring
  says so.

### DIFF
--- a/vllm_rbln/v1/core/rbln_scheduler.py
+++ b/vllm_rbln/v1/core/rbln_scheduler.py
@@ -69,6 +69,8 @@ class RBLNScheduler(Scheduler):
     ) -> None:
         super().__init__(*args, **kwargs)
 
+        self._validate_kv_cache_capacity()
+
         # Replace the upstream KVCacheManager with RBLNKVCacheManager
         # when sub-block prefix caching is enabled.
         # Sub-block size equals the prefill chunk size (max_num_batched_tokens)
@@ -127,6 +129,35 @@ class RBLNScheduler(Scheduler):
         # ceil(demand / pp) to spread decodes across microbatches. pp == 1 makes
         # the soft cap a no-op. See v1/core/utils.py.
         self._pp_size = self.vllm_config.parallel_config.pipeline_parallel_size
+
+    def _validate_kv_cache_capacity(self) -> None:
+        """Reject a max_model_len the KV cache pool can never serve.
+
+        BlockPool permanently reserves one null block, so only
+        (num_blocks - 1) blocks are allocatable. Admission checks a prompt
+        against max_model_len alone while a new prefill reserves the whole
+        sequence up front, so a prompt above the usable capacity is admitted
+        and then fails allocation on every step: it waits forever without
+        erroring and head-of-line-blocks the requests behind it. Upstream's
+        startup check sizes the pool against cdiv(max_model_len, block_size)
+        blocks, one more than are usable, so it does not cover this.
+
+        The bound assumes a single KV cache group at the scheduler's block
+        size. A hybrid layout draws from the same pool for every group and so
+        needs at least this much, which keeps the check conservative: it can
+        miss a violation, never invent one.
+        """
+        num_blocks = self.kv_cache_config.num_blocks
+        usable_tokens = (num_blocks - 1) * self.block_size
+        if self.max_model_len > usable_tokens:
+            raise ValueError(
+                f"max_model_len ({self.max_model_len}) exceeds the usable KV "
+                f"cache capacity ({usable_tokens} tokens = ({num_blocks} blocks "
+                f"- 1 reserved null block) x {self.block_size} tokens per "
+                "block). Prompts above the usable capacity would wait forever. "
+                "Raise the KV cache block count (e.g. --num-gpu-blocks-override) "
+                "or lower --max-model-len."
+            )
 
     def _decode_demand(self) -> int:
         """Total decode demand for this step's soft (ceil(demand/pp)) cap.
--- a/tests/native/v1/core/test_pd_disaggregation.py
+++ b/tests/native/v1/core/test_pd_disaggregation.py
@@ -28,7 +28,8 @@ from vllm_rbln.v1.core.utils import step_is_prefill
 
 BLOCK_SIZE = 16
 PROMPT_LEN = 2 * BLOCK_SIZE
-MAX_LEN = BLOCK_SIZE * 100
+# 99, not 100: one block of the pool is BlockPool's reserved null block.
+MAX_LEN = BLOCK_SIZE * 99
 
 
 def _scheduler(matched_tokens: int, *, is_async: bool = True, **kwargs):
--- a/tests/native/v1/core/test_rbln_scheduler.py
+++ b/tests/native/v1/core/test_rbln_scheduler.py
@@ -240,8 +240,9 @@ class TestScheduleBasic:
         assert len(sched.waiting) == 2
 
     def test_chunked_prefill_across_steps(self):
-        # A prompt longer than max_num_batched_tokens is chunked across steps.
-        sched = create_rbln_scheduler(max_num_batched_tokens=256)
+        # A prompt longer than max_num_batched_tokens is chunked across steps;
+        # max_model_len only has to clear the prompt itself.
+        sched = create_rbln_scheduler(max_num_batched_tokens=256, max_model_len=1024)
         req = create_requests(1, num_tokens=500)[0]
         sched.add_request(req)
         out = sched.schedule()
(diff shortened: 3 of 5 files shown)


vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -128,6 +130,35 @@ def __init__(
...
+        sequence up front, so a prompt above the usable capacity is admitted
+        and then fails allocation on every step: it waits forever without
+        erroring and head-of-line-blocks the requests behind it. Upstream's
+        startup check sizes the pool against cdiv(max_model_len, block_size)
+        blocks, one more than are usable, so it does not cover this.
+
+        The bound assumes a single KV cache group at the scheduler's block
+        size. A hybrid layout draws from the same pool for every group and so
+        needs at least this much, which keeps the check conservative: it can
+        miss a violation, never invent one.
```
rebel-jinhwan, line 148: Could we harden this validation for hybrid attention models? IIUC, it still cannot find the same issue when swa is added

PR #959, opened 2026-08-20
docs(agent): adjust the agent rules after first use

rebel-cjlee:
### 🚀 Summary of Changes

Follow-up to #936. Some adjustments from using the rules and from what the review surfaced.

- Patching now says when a patch is right, not only when it is wrong.
- Dropped the rule that the test diff should stay smaller than the code change.
- Added `What not to publish`. The repository is public, so throughput, latency, memory and accuracy form internal runs stay out of code, commits and PR descriptions.
- PR descriptions have to use the template.

---

* Resolves #
* Related to #936 

---

### ✅ Type of Change

---

### 🧪 How to Test

N/A

---

---

### 📋 Checklist

---

### 💬 Notes

@rebel-jonghewk one open question. The rule as written says a perf change gives the direction only, with no numbers. I left it that way rather than deciding on my own, but would a relative figure like "30% faster" be fine? Direction alone may be too weak to justify a perf change in review, though the internal channel is there for the details, so it works either way. Let me know and I will update the rule.

---

### DIFF
--- a/.claude/skills/finishing-a-change/SKILL.md
+++ b/.claude/skills/finishing-a-change/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: finishing-a-change
-description: Use before claiming a change is done, fixed, ready, or passing, and before writing a commit or PR — runs the checks, reviews the diff against the repository rules, and reports what was not done.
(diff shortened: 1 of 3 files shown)


rebel-jonghewk (approved):
LGTM

PR #960, opened 2026-08-20
fix(moe): add the replicated shared expert outside the TP all_reduce

rebel-kblee:
### 🚀 Summary of Changes

fix to support tp > 1, dp > 1 in deepseek

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/patches/deepseek_v2.py
+++ b/vllm_rbln/patches/deepseek_v2.py
@@ -38,14 +38,20 @@ def patched_deepseek_v2_moe_forward(
     if hidden_states.dtype != torch.float16:
         final_hidden_states *= self.routed_scaling_factor
 
+    shared_output = None
     if self.shared_experts is not None:
         shared_output = self.shared_experts(hidden_states)
         if hidden_states.dtype == torch.float16:
             shared_output *= 1.0 / self.routed_scaling_factor
-        final_hidden_states = final_hidden_states + shared_output
+        if not self.is_sequence_parallel:
+            final_hidden_states = final_hidden_states + shared_output
+            shared_output = None
 
     if self.tp_size > 1:
         final_hidden_states = tensor_model_parallel_all_reduce(final_hidden_states)
+
+    if shared_output is not None:
+        final_hidden_states = final_hidden_states + shared_output
     # FIXME(RBLN) - DO NOT reshape
     # return final_hidden_states.view(orig_shape)
     return final_hidden_states


vllm_rbln/patches/deepseek_v2.py
```diff
@@ -38,14 +38,20 @@ def patched_deepseek_v2_moe_forward(
...
-        final_hidden_states = final_hidden_states + shared_output
+        if not self.is_sequence_parallel:
+            final_hidden_states = final_hidden_states + shared_output
+            shared_output = None
 
     if self.tp_size > 1:
         final_hidden_states = tensor_model_parallel_all_reduce(final_hidden_states)
+
+    if shared_output is not None:
+        final_hidden_states = final_hidden_states + shared_output
```
rebel-cjlee, line 54: Just curious, I think we don't typically use sequence parallelism, so it looks the same as before. Did we have any problems with our previous TP setup?

rebel-cjlee (approved):

vllm_rbln/patches/deepseek_v2.py
rebel-kblee: When TP>1, DP>1 with EP -> backend set to True
(https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/config/parallel.py#L642-L656)
Also, deepseek v2 model set this config in shared_experts
(https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/model_executor/models/deepseek_v2.py#L266)
Therefore, as we ultimately need to distinguish between the cases `TP`, `DP` and `TP+DP `in order to decide whether to use `all_reduce` or `all_gather`, this adjustment is necessary.

rebel-kblee: @rebel-cjlee Would it be all right if I asked you to double-check this?

rebel-cjlee: Checked this against v0.24.0 and you're right, the fix is needed.

PR #962, opened 2026-08-21
refactor(optimum): move local_block_table allocation logic to scheduler

rebel-eunji:
### 🚀 Summary of Changes

This PR moves cache slot allocation and lifecycle management from individual Optimum models to the scheduler.

The cache slot identifies a per-request row in fixed-size, per-sequence on-device caches, such as sliding-window KV caches and linear-attention state caches. Since the scheduler is the source of truth for request admission, completion, and preemption, it is now responsible for:

- Allocating the lowest available cache slot when a request is admitted
- Keeping the slot pinned throughout the request lifecycle
- Releasing the slot when the request finishes or is preempted
- Passing the request-to-slot mapping to the model runner through `RBLNSchedulerOutput`

The model runner converts this mapping into `ModelInputForRBLN.cache_slot_ids`, allowing each model to consume scheduler-assigned cache slots without reconstructing request lifecycle state locally.

This refactoring also removes the now-unnecessary `optimum_attention` manager/strategy abstraction and its associated bookkeeping, padding utilities, and finished-request handling. Gemma3 retains only the model-generated state required during decoding—its padded cache length and attention mask—through a small model-specific state manager.

The terminology has also been clarified by renaming the model input from a “local block table” ID to a “cache slot” ID. The compiled Optimum-RBLN graph argument names remain unchanged at the runtime boundary for compatibility.

This change affects only the Optimum path. The native path is unchanged.

---

### Key Changes

- Moved cache slot allocation and ownership to `RBLNOptimumScheduler`
- Added `cache_slot_id_dict` to `RBLNSchedulerOutput`
- Added `cache_slot_ids` to `ModelInputForRBLN`
- Released cache slots on request completion and preemption
- Updated sliding-window, EXAONE-4.5, Gemma3/Gemma4, and Qwen3.5 models to consume scheduler-assigned slots
- Added shared cache slot padding logic for decoder batches
- Replaced Gemma3’s generic attention manager with a focused per-request hybrid-attention state manager
- Removed the unused `optimum_attention` manager/strategy package and related utility code
- Removed obsolete model-side request lifecycle reconstruction and bookkeeping

---

### Why This Change?

Previously, each model allocated and tracked cache slots independently. This required models to reconstruct the request lifecycle from running and finished request ID lists and required additional runner-side cleanup hooks.

It could also leave a stale slot assigned to a preempted request until the request resumed, potentially exhausting the slot pool while admitting another request.

Moving ownership to the scheduler makes the lifecycle explicit, avoids duplicated bookkeeping, ensures slots are released immediately, and significantly simplifies the model-side implementation.

---

### ✅ Type of Change

- [x] 🧬 Core engine changes (`core`)

---

### 🧪 Testing

Added and updated unit tests covering:

- Cache slot allocation using the lowest available ID
- Stable cache slot assignment across decode steps
- Cache slot reuse after request completion
- Immediate cache slot release on preemption
- Correct cache slot propagation from the scheduler to model inputs
- Gemma3 hybrid-attention state handling across prefill and decode
- Qwen3.5 linear-attention state placement using scheduler-assigned slots
- Correct request ordering and decoder batch padding behavior

---

### 📋 Checklist

- [x] PR title follows the Conventional Commits format
- [x] The test method and expected behavior are described
- [x] Relevant unit tests have been added or updated
- [x] The native execution path remains unaffected


tests/optimum/v1/models/optimum/test_qwen3_5_linear_state.py
```diff
@@ -214,84 +157,14 @@ def fake_prefill(**kw):
...
             inputs_embeds=torch.zeros(1, 1, 1),
             position_embed=torch.zeros(2, 1, 1, 1, 1),
+            cache_slot_ids=torch.tensor([1], dtype=torch.int16),
         )
         obj.forward(model_input)
 
-        assert passed["batch_idx"] == 1  # lowest free row (0 taken by A)
-        assert obj.attention_manager.get(False, 2, ["B"], []) == [1]  # recorded
-
-    def test_forward_lifecycle_reuses_freed_row_then_decodes_in_order(self):
```
rebel-seinpark, line 229: @rebel-eunji just question) why these scenarios are removed?

tests/optimum/v1/models/optimum/test_gemma3_hybrid_state.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+    # Each mask row gets its new token position enabled...
+    assert torch.equal(
+        recorded["attention_mask"],
+        torch.tensor([[1, 1, 1, 1, 0, 0], [1, 1, 0, 0, 1, 0]]),
+    )
+    # ...and is written back for the next step.
+    table = obj.attention_manager.table
+    assert torch.equal(table["B"].attention_mask, torch.tensor([[1, 1, 1, 1, 0, 0]]))
+    assert torch.equal(table["A"].attention_mask, torch.tensor([[1, 1, 0, 0, 1, 0]]))
```
rebel-seinpark, line 141: Please add small follow-up tests if useful.

```
# tests/optimum/v1/models/optimum/test_decoder_mixin.py

import torch

from vllm_rbln.model_executor.models.optimum.model_base import (
    RBLNOptimumDecoderMixin,
)


def _pad(cache_slot_ids: list[int], padded_batch_size: int) -> torch.Tensor:
    obj = RBLNOptimumDecoderMixin.__new__(RBLNOptimumDecoderMixin)
    return obj.pad_cache_slot_ids(
        torch.tensor(cache_slot_ids, dtype=torch.int16), padded_batch_size
    )


class TestPadCacheSlotIds:
    def test_full_batch_preserves_order_shape_dtype(self):
        # n == padded_batch_size: no padding row, ids kept in running order.
        out = _pad([2, 0], 2)
        assert out.shape == (2, 1)
        assert out.dtype == torch.int16
        assert out[0, 0] == 2 and out[1, 0] == 0

    def test_padding_row_does_not_alias_a_real_slot(self):
        # One request pinned to slot 0, batch padded to 4. The padding rows must
        # NOT reuse slot 0 (the naive "pad with 0" would corrupt the live row).
        out = _pad([0], 4)
        assert out[0, 0] == 0  # real request untouched
        pad_values = {int(out[i, 0]) for i in range(1, 4)}
        assert 0 not in pad_values  # never aliases the live slot

    def test_sparse_slot_passed_through_unclamped(self):
        # Slot id 5 with padded_batch_size(=bucket size) 2 (the per-seq cache is
        # max_num_seqs-wide, so 5 is a valid row): the id must pass through
        # unchanged — pad_cache_slot_ids must not clamp or remap it — and the
        # padding row must still avoid it.
        out = _pad([5], 2)
        assert out[0, 0] == 5
        assert int(out[1, 0]) != 5

    def test_pad_value_is_lowest_free_id(self):
        # Reals occupy {0, 2} in a batch of 4; the lowest id no request owns is
        # 1, so both padding rows take 1.
        out = _pad([0, 2], 4)
        assert out[0, 0] == 0 and out[1, 0] == 2
        assert out[2, 0] == 1 and out[3, 0] == 1

```

tests/optimum/v1/models/optimum/test_qwen3_5_linear_state.py
rebel-eunji: the attention_manager is not used for allocation of cache_slot_id anymore.
`test_forward_decode_pairs_each_request_with_its_own_row` almost replaces it.
And I added new test `test_cache_slot_id_lifecycle`.

tests/optimum/v1/models/optimum/test_gemma3_hybrid_state.py
rebel-eunji: 1. test_full_batch_preserves_order_shape_dtype is duplicated with other tests. It is verified in other tests. (includes checking dtype)
2. test_pad_value_is_lowest_free_id is almost similar with test_padding_row_does_not_alias_a_real_slot scenario.

I think adding `test_padding_row_does_not_alias_a_real_slot` is enough.

rebel-seinpark (approved):
LGTM!

rebel-jonghewk (approved):
lgtm

PR #119, opened 2025-10-20
update: Modify code to be compatible with v0.10.2

rebel-jiwoopark:
### 🚀 Summary of Changes

Modify code to be compatible with v0.10.2

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #97 

---

### ✅ Type of Change

* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [x] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`bug-fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [x] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `python offline_inference_basic.py`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [x] PR title follows Conventional Commits format
* [x] This PR is linked to an existing issue
* [x] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #975, opened 2026-08-24
feature(specdec): DFlash parallel drafting on RBLN

rebel-chanheo:
## Description

DFlash speculative decoding on the vLLM-native path. Every speculative token is drafted in **one non-causal forward** -- the sampled token followed by `num_speculative_tokens` mask tokens -- and each draft layer's context K/V is projected from the target's hidden states rather than replayed through the drafter.

Upstream's proposer cannot drive this platform: it opens a second DP all-reduce over a step the runner has already synchronized, enters a forward context carrying neither the padded token count nor RBLN's KV cache bases, and needs Triton and CUDA ops that are absent here.

## Design

**Causality is per model, not per process.** `attention_config.use_non_causal`, which upstream sets on the draft config only, is read where each model builds its attention, so a non-causal drafter and a causal target coexist. Absent the field the env flag decides, leaving target models, EAGLE3, medusa and ngram unchanged. `initialize_attn_backend` builds the draft metadata builders under the draft config so impl and builder resolve the same way.

**The spec-decode token reservation is given back in `Platform.check_and_update_config`.** `_set_max_num_scheduled_tokens` holds slots back for drafters that append draft tokens to the target batch; this one runs its own graph over its own batch, so the reservation only moves the prefill chunk off the KV block boundary the paged prefill kernel cannot straddle. The hook runs after that reservation and its validator, so the budget is simply put back. Any other budget is misaligned whatever it is, so it is refused rather than ignored.

**The drafting lookahead follows upstream 0.24.** `effective_lookahead_tokens` is zeroed on `load_kv_async and use_eagle` -- the P/D disaggregation edge -- rather than for every request on its first chunk, so `allocate_slots` reserves the page a crossing draft block moves into. Under an async P/D load upstream still zeroes it, and `use_eagle()` covers dflash, so the redirect target is checked at runtime for that path.

**A draft block never straddles two KV pages.** The kernel scatters the whole block at one dynamic offset, so a row whose context ends in a page's last `num_speculative_tokens` offsets is redirected to the next page start and its drafts dropped; model-input positions stay true, so RoPE is unaffected. When that page is unallocated the step returns before the forward, which is what keeps the scatter out of the pool's reserved null block.

**The projection is compiled; the cache write is not.** Eager dispatch pulls the projection's inputs to the host and `rms_norm` faults moving the norm weight. A compiled stateful store would give the cache input its own physical-view configuration, and alternating that with the drafter's forward rematerialises every cache view per proposal.

**The drafter's window lives in `attn_masks`, not `swa_attn_masks`.** `DFlashQwen3Attention` never declares `layer_types`, so vLLM gives every draft layer a `FullAttentionSpec` and the full-attention kernel, which never reads the SWA fields. Per-layer `seq_lens` and block tables are not an alternative: a second dynamic index on a partition the full layer already indexed is one more than the compiler allows.

**`num_tokens_across_dp` and `num_padded_tokens` travel together.** `RBLNDPMetadata` requires both to be None off the DP path, and the MoE tokens mask routes a DP=1 step there too.

The runner's EAGLE-only guards now accept either proposer through `DRAFT_MODEL_PROPOSERS`.

## Refused configurations

All at construction or `load_model`, before anything reaches a device:

| Refused | Why it cannot fail loudly on its own |
| --- | --- |
| `--max-num-seqs > 1` | acceptance collapses below the no-speculation baseline; output stays coherent |
| `enforce_eager`, `VLLM_RBLN_COMPILE_MODEL=0` | the context K/V write goes through an attention op that exists only as a compiled kernel |
| `VLLM_RBLN_USE_DEVICE_TENSOR=0` | the KV cache is allocated on `meta`, which accepts the host write and discards it |
| a drafter carrying fused MoE | `dummy_run` is a no-op, so an idle DP rank would strand the busy ranks in the draft collective |
| `dflash_causal` (tree drafting) | the causal kernels take no mask, which is the only place the block geometry is expressed |
| drafter/target RoPE style mismatch | the context K is projected from the target's hidden states; acceptance collapses silently |
| an explicit `max_num_scheduled_tokens` | the prefill chunk lands off the KV block boundary |

## Tests

43 tests, all CPU: `test_dflash.py` (29) draft mask geometry, page-crossing arithmetic, redirect-target allocation, context-write contiguity, the platform refusals and the MoE guard; `test_platform.py` (11) the token budget; `test_rbln_scheduler.py` (3) the drafting lookahead on the waiting, synchronous-remote and async-P/D paths.

## Limits

- Single sequence only. The cause of the acceptance collapse on a wider batch is unidentified; what is ruled out is in the guard's history.
- Tree drafting unsupported.


tests/native/v1/spec_decode/test_dflash.py
```diff
@@ -0,0 +1,189 @@
...
+"""Tests for RBLNDFlashProposer's draft-block geometry and cache write.
+
+Every case here pins something a run measured rather than something the code
+merely does, because each was a regression that cost a full 26-instance
+measurement to find:
+
+  - the mask's query positions must name the block's real slots; naming the
+    slots one block further on admitted the previous step's rejected draft K/V
+  - the block is open within itself, not causal; closing it measured 3.25
+    tokens/step against 3.62
```
rebel-scottlee: **Blocker (repo policy).** This docstring includes the size of an internal measurement run and two absolute tokens/step figures. AGENTS.md ("What not to publish") keeps absolute measurements from internal runs out of this public repo — code, comments, and tests included; only relative statements are allowed.

Suggest rewording the two bullets relatively, e.g. the causality bullet becomes "closing the block within itself reduced acceptance compared with the open block". Please also drop the run-size mention in the intro sentence.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,710 @@
...
+        crossing = (seq_lens % self.block_size) + num_query_per_req > self.block_size
+        if bool(crossing.any()):
+            next_page = (seq_lens // self.block_size + 1) * self.block_size
+            table_width = cad.block_table_tensor.shape[-1]
+            if int((next_page // self.block_size).max()) >= table_width:
+                # At the context ceiling there is no next page to redirect to.
+                self._dropped_rows = torch.ones_like(crossing)
+                return self.positions.new_zeros(num_query_total)
+            seq_lens = torch.where(crossing, next_page, seq_lens)
+        self._dropped_rows = crossing
```
rebel-scottlee: **Blocking.** The comment above relies on "allocate_slots already gave this request the lookahead block that crossing implies", but that invariant does not hold on this scheduler's waiting path.

Upstream reserves `K+1` lookahead tokens for DFlash (vLLM 0.24 `scheduler.py:249`), but `rbln_scheduler.py:731` zeroes `effective_lookahead_tokens` when `request.num_computed_tokens == 0` — and at that point the field is still 0 even when the request has prefix/sub-block cache hits, because it is only assigned after allocation (`rbln_scheduler.py:819/:842`).

Concrete path with defaults on (`enable_prefix_caching`, `VLLM_RBLN_SUB_BLOCK_CACHE`) and a single request: a prompt whose cache hit lets the remaining prefill finish in one waiting-path chunk, ending in the last `K` offsets of a KV block. `crossing` is true, `next_page // block_size` is still inside the table, so the redirect proceeds — but that block-table entry was never allocated and holds 0, which is the shared null block (`block_pool.py` reserves it and warns it needs special care). Both the slot-mapping gather in `_fill_first_pass_inputs` and this redirect then target the null block, and the compiled query graph scatters the block's K/V there before any result is returned. Dropping the row via `_dropped_rows` discards the draft ids but does not prevent the write.

Suggested fixes, in order of preference: (a) give the DFlash waiting path its real lookahead (restrict the `num_computed_tokens == 0` zeroing, which exists for P/D disaggregation, to that configuration); (b) if that must stay, verify the redirect target is an allocated entry and drop the row without running its write; (c) reject the prefix-cache + DFlash combination at startup. I have not reproduced downstream corruption on device — this is from reading the code — but the write into the null block itself is a cache-contract violation regardless.

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -69,6 +73,34 @@ def __init__(
...
+            #
+            # It has to be restored here rather than on the config, which a
+            # validator rejects: `max_num_scheduled_tokens` is required to carry
+            # the reservation. The equality test leaves an explicit
+            # user-supplied budget alone.
+            logger.info(
+                "Restoring the DFlash target token budget: %d -> %d.",
+                self.max_num_scheduled_tokens,
+                self.scheduler_config.max_num_batched_tokens,
+            )
```
rebel-scottlee: **Non-blocking.** The comment says "The equality test leaves an explicit user-supplied budget alone", but `max_num_scheduled_tokens` is a user-settable field, and a user who explicitly sets it to the boundary value `max_num_batched_tokens - delta` (which the upstream validator accepts) is indistinguishable from the auto-computed value here — their setting gets silently restored to the full budget.

The value alone cannot carry provenance. If you want to honor explicit settings, capture whether the field was user-supplied before `VllmConfig._set_max_num_scheduled_tokens` fills it in (e.g. in the platform's config hook), or soften the comment to say the boundary value is treated as auto-computed. Also worth a small test either way — this hunk currently has none.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,710 @@
...
+                self._proj_states[:bucket], self._proj_positions[:, :bucket]
+            )
+            key_slices.append(k[:, :, :take])
+            value_slices.append(v[:, :, :take])
+            offset += take
+        if len(key_slices) == 1:
+            keys, values = key_slices[0], value_slices[0]
+        else:
+            keys = torch.cat(key_slices, dim=2)
+            values = torch.cat(value_slices, dim=2)
```
rebel-scottlee: **Non-blocking (latent).** The multi-slice branch of this loop cannot work as written: `_project_context_kv` is compiled with `use_static_output=True`, so every call reuses the same output buffers. If the loop ever ran twice, the views collected in `key_slices`/`value_slices` would alias the same storage and the second call would clobber the first before the `torch.cat`.

Today this is unreachable — `num_context` is bounded by the scheduled tokens, which the scheduler caps at `max_num_batched_tokens == self.max_num_tokens` — so the `cat` branch is dead code that is also wrong. Suggest asserting `num_context <= self.max_num_tokens` and collapsing to a single call; if multi-slice is ever really needed, each slice must be cloned (and the input buffers grown to match).

tests/native/v1/spec_decode/test_dflash.py
```diff
@@ -0,0 +1,189 @@
...
+        assert not crossing[: BLOCK_SIZE - QUERY_LEN + 1].any()
+
+    def test_a_block_start_never_crosses(self):
+        assert not self._crossing([0, BLOCK_SIZE, 4 * BLOCK_SIZE]).any()
+
+    def test_redirect_lands_on_the_next_page_start(self):
+        lens = torch.tensor([BLOCK_SIZE - 3], dtype=torch.int64)
+        redirected = (lens // BLOCK_SIZE + 1) * BLOCK_SIZE
+        assert int(redirected[0]) == BLOCK_SIZE
+        assert not self._crossing([int(redirected[0])]).any()
```
rebel-scottlee, line 147: **Non-blocking (test coverage).** `TestPageCrossing` re-implements the crossing and redirect formulas inline instead of calling `_run_query_pass` (same pattern in `TestContextWriteContiguity`), so a regression in the real caller — including the unallocated-next-block case from the redirect comment thread — would not fail any test here.

Suggest driving `_run_query_pass` (and `_fill_first_pass_inputs`) through a stub metadata/compiled-call seam and pinning: a crossing request whose next block is allocated, and one whose next block is not. The scheduler budget-restore hunk would also benefit from a focused test distinguishing the auto-computed and explicit budget cases.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,710 @@
...
+        self,
+        target_token_ids: torch.Tensor,
+        target_positions: torch.Tensor,
+        target_hidden_states: torch.Tensor,
+        next_token_ids: torch.Tensor,
+        token_indices_to_sample: torch.Tensor | None,
+        common_attn_metadata: CommonAttentionMetadata,
+        mm_embed_inputs: tuple[list[torch.Tensor], torch.Tensor] | None = None,
+        num_rejected_tokens: torch.Tensor | None = None,
+    ) -> torch.Tensor:
```
rebel-scottlee: **Nit.** `propose` is annotated `-> torch.Tensor`, but the dropped-rows path (lines 148-151) returns `list[list[int]]`. The runner accepts both (`_draft_token_ids: list[list[int]] | torch.Tensor | None`), so only the annotation needs widening.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,710 @@
...
+            sliding = self._sliding_layer_names(attn_group)
+            full_metadata = copy(attn_metadata)
+            full_metadata.attn_masks = (
+                self._draft_block_mask(
+                    cad.seq_lens, num_reqs, num_reqs_padded, max_seq_len, None
+                )
+                .to(mask_dtype)
+                .contiguous()
+                .to(mask_device)
+            )
```
rebel-scottlee: **Nit (per-step host cost).** The builder's non-causal decode mask is built on the host at `max_model_len` width every step, and this code only reads its shape, dtype and device before replacing it with `_draft_block_mask`. With long contexts this is a per-step allocation (plus a `_staged` buffer held at that shape) that never reaches the kernel. Reading width/dtype from the builder's attributes instead would skip the throwaway mask. Fine to defer.

rebel-scottlee (changes requested):
Thanks — the single-graph design is solid and the platform constraints are unusually well documented in comments. I reviewed this against the currently supported envelope (MiniMax M2.5/M2.7 target, `max_num_seqs=1`, single in-flight request).

Two items block merge: an AGENTS.md policy violation in the new test docstring, and a KV-allocation path where a crossing redirect can write draft K/V into the shared null block — reachable with a single request through the default prefix/sub-block cache. The rest are non-blocking notes.

tests/native/v1/spec_decode/test_dflash.py
rebel-chanheo: Fixed in 49e33e5. The causality bullet is now relative and the run size is gone. The max_num_seqs guard's docstring and comment carried the same figures, so those went too.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Fixed in bf044bf. The step now checks the redirect target before using it and returns early when the slot is unallocated, so the query graph — and its scatter — never runs. With --max-num-seqs 1 there is one row, so giving up the step costs exactly the drafts of the row that would have been dropped anyway. Five cases pin it: allocated target, unfilled target, past the table, a non-crossing row that must not veto the step, and a batch  where only crossing rows are consulted.
On (a): restricting the zeroing to P/D would close the common path but leave P/D itself open (#415 measured 2P2D), so the check has to live here regardless. Worth doing separately as a performance fix — with the lookahead present, a crossing row could be redirected instead of dropped — but it touches a shared scheduler path and changes the block budget, so I'd rather measure that on its own.

tests/native/v1/spec_decode/test_dflash.py
rebel-chanheo: The crossing case you named is now covered — allocated and unallocated redirect targets, in TestRedirectTarget (bf044bf) — and the budget hunk has its own tests (49e33e5). I left the broader stub seam for _run_query_pass out of this PR; happy to take it as a follow-up if you'd rather have it now.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Good catch — collapsed to a single call with an assert on the bound it relied on (49e33e5). You're right that the branch was both unreachable and wrong.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Done rather than deferred (49e33e5). Reading width and dtype from the builder's attributes alone would not have helped, since the builder still builds the mask, so build() takes a build_attn_masks=False opt-out and the draft pass uses it. The non-causal check now reads builder.is_causal directly instead of inferring it from the mask's presence. Measured at roughly 2% of a decode step, constant in context length since the width is always max_model_len.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Widened to torch.Tensor | list[list[int]] (49e33e5).

vllm_rbln/v1/core/rbln_scheduler.py
rebel-chanheo: You're right that the value carries no provenance. The comment now says the reserved boundary is treated as auto-computed and restored either way. The decision moved into restored_dflash_token_budget() so it is testable, with eight cases covering the reserved value, the max_num_seqs scaling, other budgets, the other methods, and no speculation (49e33e5).

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -64,6 +64,26 @@
...
+
+    Falls back to the env flag when the field is absent, so target models,
+    EAGLE3, medusa and ngram are unchanged.
+    """
+    if not envs.VLLM_RBLN_FLASH_CAUSAL_ATTN:
+        return False
+    attention_config = getattr(vllm_config, "attention_config", None)
+    return not getattr(attention_config, "use_non_causal", False)
+
+
```
rebel-minkyu: Since `attention_config` and `use_non_causal` are always present, it seems we can remove this method and simply use `envs.VLLM_RBLN_FLASH_CAUSAL_ATTN and not vllm_config.attention_config.use_non_causal` from the caller sides.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-minkyu:
```suggestion
            if not self.is_causal and build_attn_masks:
```
for consistency with the name and the decode usecase.

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -216,7 +236,14 @@ def build(
...
         batch_pad: int,
         is_prefill: bool,
+        build_attn_masks: bool = True,
     ) -> RBLNFlashAttentionMetadata:
+        """`build_attn_masks=False` skips the non-causal mask.
+
+        The mask is built on the host at `max_model_len` width every step and
+        staged to the device. A caller that replaces it -- the DFlash draft pass
+        builds its own block mask -- otherwise pays for one it throws away.
+        """
```
rebel-minkyu: Since this docstring only describes just one argument, I think it would be better to either expand it to describe the whole method properly, or just remove it entirely.

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -216,7 +236,14 @@ def build(
         positions: torch.Tensor,
         batch_pad: int,
         is_prefill: bool,
+        build_attn_masks: bool = True,
```
rebel-minkyu:
```suggestion
        skip_attn_masks: bool = False,
```
Here and the relevant call sites. Considering it's an opt-out option, I think the suggested way seems more natural.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+                per_layer[layer_name] = full_metadata
+            if not sliding:
+                continue
+            # The window is expressed in the mask alone. Giving the sliding
+            # layers their own `seq_lens` and block table would put a second
+            # dynamic index on partitions the full-attention layer already
+            # indexed, and the compiler allows only one per partition
+            # (ControlAnalysis.cpp, "Input index and constBuf key should be
+            # consitent for a batch index and partition id"). It arrives as a
+            # segfault, not a diagnostic.
```
rebel-minkyu: This comment seems a bit unsuitable for general users. From a high-level perspective, it's unclear exactly which approach was unavailable and what was chosen as the alternative. Rather, It'd be better to state why masks for swa is set via `attn_masks` rather then `swa_attn_masks` here.

Just to clarify, is this comment intended to explain the reasoning behind the following process? Based on the upstream model definition, FullAttentionSpec is used even in the SWA layer, which utilizes the full attention kernel. Therefore, swa_attn_masks is not used, and we need to pass the SWA mask to attn_masks instead.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+                    num_reqs_padded,
+                    max_seq_len,
+                    self.sliding_window,
+                )
+                .to(mask_dtype)
+                .contiguous()
+                .to(mask_device)
+            )
+            for layer_name in sliding:
+                per_layer[layer_name] = window_metadata
```
rebel-minkyu, line 543: Fundamentally, if we patch the upstream codes to allow setting swa type for the swa layers, is it still not possible to use swa kernels, as we originally do with the usual hybrid models? Is it because of the context KV writing?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -142,6 +143,11 @@
     reorder_input_batch,
 )
 
+# Proposers that load a draft model and run it through the RBLN compile path:
+# they own draft attention groups, a warmup, and weights the runner has to make
+# contiguous. The n-gram and suffix proposers do none of that.
+DRAFT_MODEL_PROPOSERS = (RBLNEagleProposer, RBLNDFlashProposer)
```
rebel-minkyu: How about moving it to `spec_decode/__init__.py`?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -304,6 +311,12 @@ def __init__(
                 self.drafter = SuffixDecodingProposer(self.vllm_config)
             elif self.speculative_config.method == "medusa":
                 self.drafter = RBLNMedusaProposer(self.vllm_config, self.device)
+            elif self.speculative_config.method == "dflash":
+                self.drafter = RBLNDFlashProposer(self.vllm_config, self.device, self)
+                # Upstream turns this on unconditionally for DFlash: the drafter
+                # consumes the target's aux states through
+                # `precompute_and_store_context_kv`, not through a projection.
```
rebel-minkyu: Is this comment right? It seems the target's aux states are used in the projection like eagle3.

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -60,7 +60,47 @@ class RBLNSchedulerOutput(SchedulerOutput):
...
+    if speculative_config is None or speculative_config.method != "dflash":
+        return None
+    reserved = (
+        scheduler_config.max_num_batched_tokens
+        - speculative_config.max_num_new_slots_for_drafting
+        * scheduler_config.max_num_seqs
+    )
+    if max_num_scheduled_tokens != reserved:
+        return None
+    return scheduler_config.max_num_batched_tokens
```
rebel-minkyu: 1. If my understanding is correct, without the restore, `num_computed_tokens` would no longer be be aligned with `block_size`, which could lead to a block boundary crossing. While this might be fine on GPUs, it prevents proper KV writes on RBLN, which is why this adjustment is being made here. Is that right?
If so, under the `max_num_scheduled_tokens != reserved` condition, it wouldn't function correctly anyway. Therefore, shouldn't we raise a ValueError to block it, rather than just silently bypassing the restore?
2. Overall, this logic feels more like config postprocessing/validation. Then, how about applying it inside `Platform.check_and_update_config`, which is the extension point provided by the upstream? If then, we'd be able to remove the remaining diffs in this file.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+            # acceptance leaves behind -- a symptom, not the cause). The drafts
+            # on the prefill step, which is not batch-padded, are bit-identical
+            # between the two configurations, so the divergence begins with the
+            # first padded decode step.
+            raise NotImplementedError(
+                "DFlash speculative decoding requires --max-num-seqs 1; got "
+                f"{scheduler_config.max_num_seqs}. A wider decode batch "
+                "silently collapses acceptance to below the no-speculation "
+                "baseline."
+            )
```
rebel-minkyu, line 73: It feels a bit unnecessary to create a dedicated method just to throw a NotImplementedError for a simple condition. It seems it will get dropped in the follow-up PR, but it would be great if we could just inline it without the unnecessary debugging comments starting form this PR.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+            self.vllm_config.speculative_config.enforce_eager
+            or not envs.VLLM_RBLN_COMPILE_MODEL
+        ):
+            # The attention ops are pattern stubs the compiler replaces, so an
+            # eager context write would silently write nothing.
+            raise NotImplementedError(
+                "The DFlash drafter cannot run eager on RBLN: its context K/V "
+                "write goes through an attention op that only exists as a "
+                "compiled kernel."
+            )
```
rebel-minkyu: It'd be better to raise Error early in `__init__`.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+
+    def dummy_run(self, *args, **kwargs) -> None:
+        """Skip the drafter warmup.
+
+        Upstream's signature takes CUDA-graph state this platform does not
+        have, and the runner calls the `RBLNEagleProposer` one, so neither can
+        drive the other. The region compiles on the first real step, as
+        EAGLE's chained region does.
+        """
+        return None
```
rebel-minkyu: Could you elaborate more on why dummy_run is empty? Is it right that on-the-fly compilation is intended?

Also, if we leave the dummy run empty, I guess there can be a hang when using DP, caused by the absence of collective sync in the empty rank.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+            self._context_positions_buffer.device
+        )
+
+        # Query slots, block-table lookup on the host.
+        q_blocks = block_table[rows[:, None], query_pos // self.block_size]
+        self._slot_mapping_buffer[:num_query_total] = (
+            (q_blocks * self.block_size + query_pos % self.block_size)
+            .reshape(-1)
+            .to(self._slot_mapping_buffer.device)
+        )
```
rebel-minkyu: Is initializing `_slot_mapping_buffer` necessary? I believe slot_mapping is not being used in RBLN attentions.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+        the last `num_speculative_tokens` offsets of a page there is no next
+        page to move into. Restricting the zeroing to P/D disaggregation -- what
+        it exists for -- would close that path but leave the P/D configuration
+        itself open, so the check belongs here too.
+        """
+        pages = (next_page // self.block_size).to(torch.int64)
+        if int(pages.max()) >= block_table.shape[-1]:
+            return True
+        rows = torch.arange(pages.shape[0])
+        return bool((block_table.cpu()[rows, pages][crossing] == 0).any())
```
rebel-minkyu: Can't page allocation be assured in the scheduler? I've checked the upstream code and [this logic](https://github.com/vllm-project/vllm/blob/v0.24.0/vllm/v1/core/sched/scheduler.py#L849) is missed in our implementation. And if we add it, we might be able to utilize `effective_lookahead_tokens` properly and assure page allocated from the scheduler. If that works, we can remove this method entirely.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+            cache = layer.self_attn.attn.kv_cache
+            for token_start, count, block, offset in runs:
+                token_slice = slice(token_start, token_start + count)
+                cache_slice = slice(offset, offset + count)
+                for head in range(num_kv_heads):
+                    destinations.append(cache[0, block, head, 0, cache_slice, :])
+                    destinations.append(cache[1, block, head, 0, cache_slice, :])
+                    sources.append(keys[layer_index, head, token_slice, :])
+                    sources.append(values[layer_index, head, token_slice, :])
+        torch._foreach_copy_(destinations, sources)
```
rebel-minkyu, line 685: Could you confirm that it would work with DEVICE_TENSOR=0 as well? We might need a guard if it's not.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+            block_starts,
+            num_reqs,
+            num_reqs,
+        )
+        with set_forward_context(
+            per_layer,
+            self.vllm_config,
+            num_tokens=num_query_total,
+            num_tokens_across_dp=self._draft_dp_status(num_reqs, num_query_per_req),
+            num_padded_tokens=num_query_total,
```
rebel-minkyu: Could you check if it's safe to always set `num_padded_tokens` even when DP=1?

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+    attach_kv_cache_bindings,
+    build_kv_cache_forward_context_kwargs,
+)
+from vllm_rbln.v1.spec_decode.eagle import RBLNEagleProposer
+
+
+class RBLNDFlashProposer(DFlashProposer):
+    # `initialize_attn_backend` swaps this for the draft config, and assigning to
+    # an inherited attribute leaves mypy unable to infer its type.
+    vllm_config: VllmConfig
```
rebel-minkyu, line 53: This comment doesn't seem to be necessary.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+            num_reqs, self.num_speculative_tokens
+        )
+        if bool(self._dropped_rows.any()):
+            # An empty list, never zeros: the scheduler reads zeros as real
+            # token ids. The target-only step still advances the request to the
+            # aligned page, where speculation resumes.
+            rows = draft_ids.tolist()
+            return [
+                [] if bool(self._dropped_rows[i]) else rows[i] for i in range(num_reqs)
+            ]
```
rebel-minkyu, line 210: Since we safely redirect KV slots to the next block, avoiding the crossing of block boundary, it seems those written KVs are valid. Then, can't we just use those results after adjusting the mask accordingly rather than dropping them?

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+                attn_metadata,
+                self.runner.kv_caches,
+                self.runner.kv_cache_bases,
+                self.runner.kv_cache_view_infos,
+            )
+            if builder.is_causal:
+                raise RuntimeError(
+                    "the draft pass needs an explicit mask; the drafter must be "
+                    "built non-causal (attention_config.use_non_causal)"
+                )
```
rebel-minkyu: It'd be better to raise this Error early in the `__init__`.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+        """Draft `num_speculative_tokens` tokens in one forward.
+
+        Upstream's `propose` cannot drive this platform. It coordinates the
+        draft batch across DP ranks with its own all-reduce, but the runner has
+        already synchronized this step (`runner.dp_status`) and a second
+        collective leaves the ranks one call apart -- the gloo read error that
+        surfaces is the symptom. It also enters `vllm.forward_context`, which
+        carries neither the padded token count nor the KV cache bases that RBLN
+        attention layers resolve their caches from.
+        """
```
rebel-minkyu, line 158: I think it's a bit hard to understand what this docstring tries to deliver. Could you make it more concise? If it's a debugging artifact, it'd be better to remove it.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+        builder stays causal and emits no mask -- for a kernel that requires
+        one. Both sides have to read the same config.
+        """
+        draft_vllm_config = self._create_draft_vllm_config()
+        with set_current_vllm_config(draft_vllm_config):
+            original, self.vllm_config = self.vllm_config, draft_vllm_config
+            try:
+                super().initialize_attn_backend(kv_cache_config, kernel_block_sizes)
+            finally:
+                self.vllm_config = original
```
rebel-minkyu, line 227: Could you explain what this method is trying to do? If it is to check `causal` config, would it be possible to just add an explicit handling for that specific condition without this method?

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,769 @@
...
+
+    def _fill_first_pass_inputs(
+        self,
+        next_token_ids: torch.Tensor,
+        target_positions: torch.Tensor,
+        cad: CommonAttentionMetadata,
+        num_rejected_tokens: torch.Tensor | None,
+        num_query_per_req: int,
+    ) -> tuple[int, torch.Tensor, torch.Tensor]:
+        """What `copy_and_expand_dflash_inputs_kernel` does, in torch.
```
rebel-minkyu:
```suggestion
        """What `copy_and_expand_dflash_inputs_kernel` does, in upstream.
```

rebel-minkyu (changes requested):

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Dropped.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Investigation notes dropped;

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Cut to the two reasons upstream's propose can't run here. The gloo symptom was a debugging artifact and is gone.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: It builds the draft metadata builders under the draft config. Impl and builder each read causality from the config they're constructed under, so both have to see the same one; upstream uses the target's, which leaves the impl non-causal and the builder causal. Docstring now says that. Scoping it to this call is what keeps set_current_vllm_config off the target's path — if you'd rather pass the config down explicitly instead, I'll take that.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Applied as "What upstream's copy_and_expand_dflash_inputs_kernel does, in torch."

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Your reading is right. swa_attn_masks is only built when kv_cache_spec carries a sliding_window, and only sliding_window_attention_naive_decode reads it. The drafter never declares layer_types, so every draft layer gets a FullAttentionSpec and the full-attention kernel, which never reaches those fields — hence attn_masks. Comment rewritten in that order, with the compiler constraint kept as the reason per-layer seq_lens isn't an alternative.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-chanheo: Took the remove option. The cost note now sits where the flag is read.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-chanheo: It wasn't — DFlash does go through a projection, same as eagle3 (line 1866 fuses combine_hidden_states for both). Fixed.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-chanheo: You're right that both are always present — attention_config has a default_factory and use_non_causal is a real field defaulting to False, so the getattr guards bought nothing. Inlined at both call sites; the design note stays once, on the builder. TestResolveIsCausal went with it — those three were the only causality tests that ran without reaching the gated test model, so tell me if you want coverage re-added at the builder level.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-chanheo: Renamed, and the call site now passes skip_attn_masks=True.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-chanheo: Good catch — the prefill branch was ignoring the flag entirely while only decode honored it, so the name was promising more than it did. Both branches now check it. Inert today (DFlash only builds with is_prefill=False), but the API no longer lies.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-chanheo: Moved, with the comment. Checked the import order both ways — spec_decode.eagle first and spec_decode first — since dflash imports eagle; neither cycles.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Moved to __init__, ahead of super().__init__() since it only reads the config.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Split it, because the one check was doing two jobs. The user-facing refusal moved to __init__ as NotImplementedError on self.dflash_causal — that's the config that makes _create_draft_vllm_config emit use_non_causal=False. What stays at the build site is the plumbing invariant it actually caught: initialize_attn_backend builds under the draft config, and a causal builder there means that scoping came undone — now an assert. One behavior change worth flagging: dflash_causal=True with VLLM_RBLN_FLASH_CAUSAL_ATTN off used to run (non-causal by accident); it is now refused. Say the word if you'd rather keep the exact old condition.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Confirmed — no RBLN backend reads CommonAttentionMetadata.slot_mapping; every other RBLN call site passes torch.tensor(0)  # dummy (eagle.py:343/454, rbln_model_runner.py:956), and only forward_context.slot_mapping, a per-layer dict, is ever read. So the per-step host block-table gather that filled it was dead work. Dropped the fill and the buffer use; the metadata now passes the same dummy as the rest.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Not safe. RBLNDPMetadata.make asserts it is None off the DP path (forward_context.py:66), and set_forward_context enters make when dp_size > 1 or VLLM_RBLN_USE_MOE_TOKENS_MASK, which defaults True — so DP=1 trips it every proposal. Now derived from _draft_dp_status, which is already None at dp_size == 1, so the two travel together.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Doesn't work, and fails silently: with it off the cache lands on device="meta" (rbln_model_runner.py:2710), and _foreach_copy_ into a meta slice discards the write without an error. Guarded in __init__ next to the eager refusal.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Empty because upstream's signature takes CUDA-graph state we have no equivalent for and the runner calls the EAGLE one; the region compiles on the first real step, as EAGLE's chained region does.
On the hang you're right that it needs a condition, and eagle.py:373 already states it: fused MoE is the only reader of the padded token dimension, so a draft without it runs no collective. Ours is dense, but that was unstated — load_model now refuses a drafter containing MoERunner.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: You're right that it's missing, and it's an upstream sync gap rather than a design choice: rbln_scheduler.py still carries upstream's v0.13 form (0 if request.num_computed_tokens == 0), ported at #119 for v0.10.2, while v0.24 narrowed it to load_kv_async and self.use_eagle. Adopted upstream's condition, so the waiting path now gets its real lookahead and allocate_slots reserves the page a crossing needs.
Not removable entirely, though. use_eagle() is True for dflash, so upstream still zeroes it under an async P/D load — and that is a configuration we measure. _page_unallocated stays, now guarding only that path, and its docstring says so. Three tests pin the three cases.

vllm_rbln/v1/core/rbln_scheduler.py
rebel-chanheo: ② Moved to check_and_update_config. The reason the PR gave for keeping it on the scheduler doesn't hold: _set_max_num_scheduled_tokens and its validator run at config/vllm.py:1231, this hook at :1350, so the validator is already satisfied — and Scheduler.__init__ is the field's only reader.
The whole rbln_scheduler.py diff is gone.
① Also right — a ValueError now. Under DFlash the upstream validator already rejects max_num_scheduled_tokens == max_num_batched_tokens, so the value is either auto-computed or one that cannot align the prefill chunk; there is nothing to fall back to.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Implementable, but not worth it yet. The mask is built from the redirected seq_lens, so key_pos <= key_end covers the gap between the true context end and the page boundary — stale K/V. Using the drafts means excluding that gap in the full mask and computing the sliding-window distance from true positions while the block writes at redirected slots: two position spaces in one mask. A row only crosses in a page's last k offsets, so the payoff is about k / block_size of steps. Happy to take it as a follow-up if that turns out to matter.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Your read is right, and there are two reasons. swa_attn_masks only exists when kv_cache_spec carries a sliding_window, and it feeds sliding_window_attention_naive_decode through local_block_tables = block_tables[..., :1] — one window-sized block per request with a rolling cache_offsets. _write_context_kv writes every layer into one paged geometry addressed by block_table[req, pos // block_size], so declaring the SWA layers needs a second write path with those rolling offsets. It also splits the drafter into two KV-cache groups, which re-raises the one-dynamic-index-per-partition limit the mask approach exists to avoid. Worth doing separately if we want the SWA kernel's cache savings.

vllm_rbln/v1/spec_decode/dflash.py
rebel-minkyu: Don't we just override `dummy_run`s regardless of the upstream's signatures? And doesn't Eagle do warm-ups for the chained forwards as well? If not, we'd better fix Eagle to properly perform warm-ups, and I believe we prefer warm-ups of all the possible graphs, and avoid on-the-fly compilation as much as we can. Could you check it again?

vllm_rbln/platform.py
```diff
@@ -60,6 +60,43 @@ def bypass_backend(graph_module: torch.fx.GraphModule, example_inputs):
...
+
+    `VllmConfig._set_max_num_scheduled_tokens` holds batch slots back for
+    drafters that append their draft tokens to the target batch. This one runs
+    its own graph over its own batch, so the reservation buys nothing and only
+    moves the prefill chunk off the KV block boundary, which the paged prefill
+    kernel cannot straddle.
+
+    This hook runs after that reservation and its validator, so the budget can
+    simply be put back. Any other value cannot be made to work -- the chunk
+    would be misaligned whatever it is -- so it is refused rather than ignored.
```
rebel-minkyu:
```suggestion
```

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,764 @@
...
+        )
+
+        per_layer = self._build_draft_attn_metadata(
+            query_cad,
+            block_starts,
+            num_reqs,
+            num_reqs,
+        )
+        # Both travel together: `RBLNDPMetadata.make` requires them off the DP
+        # path to be None, and the MoE tokens mask routes a DP=1 step there too.
```
rebel-minkyu:
```suggestion

```

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,764 @@
...
+        """
+        if self.vllm_config.parallel_config.data_parallel_size == 1:
+            return None
+        dp_status = self.runner.dp_status
+        assert dp_status is not None, (
+            "dp_status is not saved from _determine_batch_padding"
+        )
+        _, num_reqs_across_dp, _ = dp_status
+        assert int(num_reqs_across_dp[self.dp_rank]) == num_reqs
+        return num_reqs_across_dp * tokens_per_req
```
rebel-minkyu: It seems we should rebase dp_status related implementation here. `DPStatus` has five attributes now.

While rebasing, would it be possible to remove `draft_dp_status` completely and just reuse `determine_draft_batch_execution_and_padding`? It seems we can make `determine_draft_batch_execution_and_padding` general for both eagle and dflash with the following modification. (need verification through)

[Here](https://github.com/RBLN-SW/vllm-rbln/blob/feat/dflash-clean/vllm_rbln/v1/worker/dp_utils.py#L320-L321) to:
```
tokens_per_rank = tuple(reqs * query_len for reqs in status.num_reqs)
```
Could you review this approach?

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,764 @@
...
+        draft without it runs no collective of its own -- which is what lets
+        `dummy_run` do nothing. With one, an idle rank that skipped its draft
+        would leave the busy ranks blocked inside that collective.
+        """
+        if any(isinstance(module, MoERunner) for module in self.model.modules()):
+            raise NotImplementedError(
+                "The DFlash drafter cannot contain fused MoE on RBLN: its "
+                "warmup and idle-step draft are no-ops, which would strand the "
+                "busy ranks in the draft collective."
+            )
```
rebel-minkyu: It seems this method would become redundant once the dummy_run is properly incorporated.

vllm_rbln/v1/spec_decode/dflash.py
rebel-minkyu: Since the scheduler now applies `limit_lookahead_tokens` properly, the pages are guaranteed to be allocated in advance and the `block_table == 0` case can no longer occur here. Even under PD, `load_kv_async` case schedules nothing and an actual step comes with `num_computed_tokens > 0` and the full lookahead_tokens. Therefore, we can just drop the unallocation condition from the give-up decision and keep it only as an assertion to check a regression. Then, the method no longer matches its name, and what remains is a single comparison, so we can remove this method entirely and inline the page ceiling check at the call site. Could you check this again please?

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Done — _draft_dp_status is gone and both proposers use determine_draft_batch_execution_and_padding.

Your tokens_per_rank change is right, but tokens_across_dp on the next line needs it too — that is the value the caller consumes, so leaving it as status.num_reqs silently drops the 1 + num_speculative_tokens factor:

tokens_per_rank = tuple(reqs * query_len for reqs in status.num_reqs)
any_prefill = False
tokens_across_dp = torch.tensor(tokens_per_rank, dtype=torch.int32)

The query_len == 1 assert is dropped. Verified on DP4.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Correct. The refusal existed because an empty dummy_run meant an idle rank never drafted. Now the idle guard is EAGLE's — is_idle and not draft_has_moe — so _check_no_moe is gone and draft_has_moe is set in load_model, same as EAGLE.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: You're right, and my earlier answer was wrong. Checked with rbln_backend: stage= on a fresh cache, DP4:

- EAGLE3 k=5 — 9 graphs, all stage=warmup, none at stage=runtime. Three are drafter graphs ([1,512], [1,6], [1,1]), so the chained pass is warmed up.
- DFlash k=7 (before) — 6 at warmup, all target; 3 at stage=runtime, all drafter.

The signature isn't a reason either — EAGLE overrides under the same constraint. dummy_run now compiles both projection buckets and the query pass; re-checked and stage=runtime is empty.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Dropped. The line went with _draft_dp_status — see the reply on dflash.py:207.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Done. The unallocation case is an assertion now, and the page-ceiling comparison is inlined:

pages = (next_page // self.block_size).to(torch.int64)
if int(pages.max()) >= cad.block_table_tensor.shape[-1]:
     ...give up...
rows = torch.arange(pages.shape[0])
assert not bool((cad.block_table_tensor.cpu()[rows, pages][crossing] == 0).any())

_page_unallocated is removed.

vllm_rbln/platform.py
rebel-chanheo: Dropped, and the claim was wrong. The hook runs after the validator in the API server only — the config is validated again in every worker, where max_num_batched_tokens < max_num_scheduled_tokens + delta rejects the restored budget and no worker starts. The docstring now states the policy; the worker case is handled in the code.

vllm_rbln/v1/spec_decode/dflash.py
```diff
@@ -0,0 +1,774 @@
...
+        per_layer = self._build_draft_attn_metadata(
+            query_cad,
+            block_starts,
+            num_reqs,
+            num_reqs,
+        )
+        # `RBLNDPMetadata.make` requires this to be None off the DP path.
+        _, num_tokens_across_dp = self._determine_batch_execution_and_padding(
+            num_reqs, num_query_total, False, first_pass=False
+        )
```
rebel-minkyu, line 758: Currently, DFlash seems to assume the drafter has no MoE and ignore MoE padding across DP ranks. It's same in the `dummy_run`, where `_determine_batch_execution_and_padding` is not being used at all. However, it seems that MoE is not typically used in DFlash models, so I think we don't need to worry about supporting it for now.

However, since it is currently not supported and can cause a hang issue when a drafter is MoE, I think it would be good to have at least a `NotImplementedError` guard using `draft_has_moe`.

vllm_rbln/platform.py
```diff
@@ -60,6 +60,53 @@ def bypass_backend(graph_module: torch.fx.GraphModule, example_inputs):
...
+    # budget alone therefore fails that check and no worker starts. Give the
+    # validator its headroom instead: the prefill chunk follows
+    # `max_num_scheduled_tokens`, which still lands on `budget`.
+    if getattr(scheduler_config, "_rbln_dflash_budget_restored", False):
+        return
+    delta = budget - reserved
+    logger.info("Restoring the DFlash target token budget: %d -> %d.", reserved, budget)
+    scheduler_config.max_num_scheduled_tokens = budget
+    scheduler_config.max_num_batched_tokens = budget + delta
+    scheduler_config._rbln_dflash_budget_restored = True
```
rebel-minkyu: It seems the comments explain why `scheduler_config.max_num_batched_tokens = budget + delta` is newly added, but setting `max_num_batched_tokens` to a value not aligned with block_size looks quite risky. Would it be okay?

rebel-minkyu: If we need to fix it, how about adding a patch to fix the upstream code fundamentally? It seems [the slot reservation itself from the upstream](https://github.com/vllm-project/vllm/blob/v0.24.0/vllm/config/speculative.py#L1084-L1086) is not necessary for DFlash, because DFlash write context KV in advance, and doesn't require a placeholder for the model input buffer. Then, we won't need this postprocessings for the budget adjustment at all.

vllm_rbln/v1/spec_decode/dflash.py
rebel-chanheo: Agreed — the description listed this as refused, but no guard existed.

Still a gap after 56ae39c9: _run_query_pass keeps only num_tokens_across_dp and drops the BatchDescriptor, so the forward runs at this rank's own num_query_total. The idle-skip's draft_has_moe covered the collective, not the shape.

Added _require_dense_drafter, called from load_model before the buffers and compiles. The idle skip is now keyed off idleness alone, and draft_has_moe is a class-level False.

vllm_rbln/platform.py
rebel-chanheo: Both taken; the second is the fix — done that way.

max_num_batched_tokens is also sub_block_size (rbln_scheduler.py:78) and InputLayout.query_len_padded on prefill, so budget + delta failed block_size % sub_block_size == 0 and left the compiled prefill length unaligned — the very thing the hook existed to prevent.

max_num_new_slots_for_drafting has two consumers, both in _set_max_num_scheduled_tokens, and the registry lands before it (EngineArgs.__post_init__ vs __post_init__:1231), where check_and_update_config at :1350 was always too late. Answering 0 for dflash drops restore_dflash_token_budget, the mutation and the sentinel.

The explicit-budget refusal stays, and @rebel-scottlee's objection to it is gone: the auto-computed value is now exactly max_num_batched_tokens, so an inequality is a user-supplied budget.

rebel-minkyu (approved):

rebel-scottlee (approved):
LGTM

PR #976, opened 2026-08-24
fix(kv): backport LMCache mp MultiConnector lose-path

rebel-jinhwan:
### 🚀 Summary of Changes

Backport [vllm-project/vllm#47505](https://github.com/vllm-project/vllm/pull/47505) onto `LMCacheMPConnectorUpstream.update_state_after_alloc`.

A non-chosen `MultiConnector` sub-connector is called with `num_external_tokens=0`, but upstream decides whether to load from `needs_retrieve()` alone (block counts). It then enters `WAITING_FOR_LOAD` and leaks lookup locks — nothing outside `end_session` releases them.

The only behavioural change vs upstream is this assignment:

```python
# Exact patch vs upstream — the rest of this method is a copy.
# upstream:  condition = tracker.needs_retrieve()
condition = tracker.needs_retrieve() and num_external_tokens > 0
```

Native path only (`vllm_rbln/patches/`). Delete the module once vllm#47505 merges and the vllm-rbln pin picks it up (`TODO(vllm-pr-47505)`).

---

* Related to [vllm-project/vllm#47505](https://github.com/vllm-project/vllm/pull/47505)

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `uv run --no-sync pytest tests/native/patches/test_lmcache_mp_connector.py --noconftest -q`
2. Expected: 2 passed. A losing connector (`num_external_tokens=0`) goes `READY` and frees all lookup locks; a chosen connector (`num_external_tokens>0`) stays `WAITING_FOR_LOAD` and frees only the vLLM-hit range.
3. Edge: `--noconftest` is required in a venv whose vLLM does not vendor `lmcache_mp_connector`. The full native suite's `pytest_configure` applies every patch and will `AttributeError` on that missing module.

---

### 📋 Checklist

---

### 💬 Notes

- No in-tree vllm-rbln issue yet; this tracks the upstream PR.
- `register_patch` has no `condition`. Images / venvs without the vendored connector will fail at `apply_registered_patches()`.
- AI assistance was used.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/native/patches/test_lmcache_mp_connector.py
+++ b/tests/native/patches/test_lmcache_mp_connector.py
@@ -1,0 +1,107 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ``lmcache`` is an optional runtime dependency of vllm-rbln, so the patch
+# target module cannot be imported for real here. Stub it in sys.modules
+# before importing the patch so `patched_update_state_after_alloc`'s internal
+# `from vllm...lmcache_mp_connector import ...` resolves to the stub.
+import enum
+import sys
+import types
+from types import SimpleNamespace
+from typing import Any
+
+_TARGET_MODULE = "vllm.distributed.kv_transfer.kv_connector.v1.lmcache_mp_connector"
+
+
+class _RequestState(enum.Enum):
+    PREFETCHING = "prefetching"
+    WAITING_FOR_LOAD = "waiting_for_load"
+    READY = "ready"
+
+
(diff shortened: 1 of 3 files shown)


tests/native/patches/test_lmcache_mp_connector.py
rebel-cjlee, line 1: I'd just drop this test file rather than fix it. Since `lmcache` isn't in CI we can't import the real module, so everything here is all stubs, and the test ends up checking that a copy of upstream behaves like a copy of upstream. The part that actually worries me, whether the patch lands on the class the factory really resolves, can't be covered by stubs anyway. Dropping the file also gets rid of the `sys.modules[..] = stub` that's installed at import and never restored, which would shadow the real module for the rest of the session once `lmcache` is installed.

How we test patch modules in general is still an open question, and I don't think this PR has to answer it.

vllm_rbln/patches/lmcache_mp_connector.py
rebel-cjlee, line 1: Could we enforce the removal trigger here, the way PR #923 does? A TODO on its own tends to stick around, and this one is a full copy of the upstream method, so it silently reverts any unrelated change upstream makes to it.

A version assert is awkward since vllm#47505 isn't released yet. Two things come to mind. One is asserting on the version we expect the fix to land in. The other is inspecting the upstream source and checking whether the guard is already there. The source check is exact, while a guessed version fires on that release whether or not upstream actually merged, so it reads more as a reminder than a trigger. (I think a reminder is probably good enough here)

What do you think? Or is there a better way to do this?

rebel-cjlee (approved):

PR #977, opened 2026-08-24
other(ci): reinstall rebel-compiler at REBEL_COMPILER_VERSION when set

rebel-junamsong:
## What

When compiler CI triggers this pipeline with `REBEL_COMPILER_VERSION` set, reinstall
`rebel-compiler` at that version over the locked env, in **every** test suite. No-op for
normal vllm-rbln PRs.

## How

- New `.buildkite/scripts/rebel-compiler-override.sh` (shared): if `REBEL_COMPILER_VERSION`
  is set, `uv pip install rebel-compiler==$VERSION` from `pypi.rebellions.in/simple` (creds
  from the `UV_INDEX_REBELLIONS_*` secrets); announces it via a log group + a build annotation.
- Both suites call it right after `uv sync`, in their `&sync`: the vllm-rbln lanes
  (`vllm-rbln/pipeline-pr.yml`) and the optimum tests (`optimum-pr/pipeline.yml`).

Kept in a script (not inline YAML) so obedients does not interpolate the runtime vars at
pipeline-upload time.

Pairs with rebellions-sw/rebel_compiler#13073, which sets `REBEL_COMPILER_VERSION` when it
triggers this pipeline.


rebel-thkim (approved):
👍

rebel-cjlee (approved):

rebel-seinpark (approved):

PR #984, opened 2026-08-24
other(optimum): path-gate PR steps and refresh smoke suite

rebel-seinpark:
### 🚀 Summary of Changes

Reduces the HF-hub load that the **optimum-rbln** PR pipeline puts on every push, in two parts. No native-path code is touched.

**1. Path-gate the optimum PR steps.** The pipeline uploaded every step on every PR, so a change touching only native-path code still spun up all the optimum NPU model-loading suites. Each step now carries `if_changed`, and `.buildkite/pr.yml` uploads the pipeline with `--git-diff-base origin/$TARGET_BRANCH` (mirroring the existing vllm-rbln uploader) so `if_changed` is honored. A step runs only when the PR touches a path that can affect the optimum path.

The `include` stays broad (`vllm_rbln/**`); the narrowing lives in a deliberately conservative `exclude` of native-path-only paths (`patches/**`, `compilation/**`, `v1/worker/rbln_*.py`, `v1/spec_decode/**`, `tests/native/**`, the vllm-rbln pipeline dir). This is the safe direction: over-including only wastes a run, whereas a restrictive `include` would silently skip the suite on a shared-code change — e.g. a regression in `platform.py` or `envs.py`, which both paths use.

**2. Smoke suite.**
- `MultimodalSmoke` now reads the chat-template processor with `local_files_only=True`. `setUpClass`'s `LLM(...)` has already pulled the model (processor included) into the HF cache, so the second load need not re-issue an etag HEAD per file against the hub — a redundant round that contributes to HF rate-limit hits in CI.
- Qwen3.5-VL moves from a commented-out placeholder in `test_rsd1.py` to a live `TestQwen35VL` in `test_rsd8.py` (`NUM_DEVICES=8`).

**Model path affected:** optimum-rbln only.

---

* Related to #

---

### ✅ Type of Change

* [x] ❓ Other (`other`): CI path-gating + optimum smoke suite update

---

### 🧪 How to Test

1. On this PR, the `upload optimum-pr` step uploads the pipeline with `--git-diff-base origin/dev`. Verify the optimum steps **run** here (the PR touches `tests/optimum/**` and `.buildkite/**`).
2. Open a PR that touches only a native-path file (e.g. `vllm_rbln/patches/**` or `tests/native/**`) and verify the optimum steps are **skipped**.
3. The `smoke rsd-8` step now runs `TestQwen35VL` (compile + generate on 8 devices) and must produce at least one token.

Verified locally: `pre-commit` (ruff / mypy / license) passes on all changed files; the pipeline YAML parses and every step resolves the `if_changed` anchor. The NPU smoke suites themselves were **not** run locally — they require an RBLN NPU.

---

N/A

---

### 📋 Checklist

---

### 💬 Notes

- A prefetch → `HF_HUB_OFFLINE=1` variant of the smoke steps was prototyped and dropped: its payoff hinges on whether the CI `HF_HOME` cache is persistent across builds (unknown), and each smoke class loads its model roughly once, so the reduction was marginal. The always-valid piece (`local_files_only`) is what remains.
- The `exclude` list is intentionally conservative; it can be tightened once the gating behavior is observed on a few PRs.
- Needs a linked development issue before merge (checklist item above is unchecked).

### DIFF
--- a/.buildkite/pipelines/optimum-pr/pipeline.yml
+++ b/.buildkite/pipelines/optimum-pr/pipeline.yml
@@ -19,6 +19,27 @@ _secrets: &secrets
   HF_TOKEN: HF_TOKEN
   HF_HOME: HF_HOME
 
+# Run a step only when the PR touches a path that can affect the optimum path.
+# Honored solely when the pipeline is uploaded with --git-diff-base (see
+# .buildkite/pr.yml); without that flag every step runs, as it did before.
+# The exclude list is deliberately conservative -- it drops only paths that are
+# native-path-only, since over-including merely wastes a run while
+# over-excluding would silently skip a suite that should have caught a break.
+_if_changed: &if_changed
+  include:
+    - ".buildkite/**"
+    - "tests/optimum/**"
+    - "vllm_rbln/**"
+    - "pyproject.toml"
+    - "uv.lock"
+  exclude:
+    - ".buildkite/pipelines/vllm-rbln/**"
+    - "tests/native/**"
+    - "vllm_rbln/patches/**"
+    - "vllm_rbln/compilation/**"
+    - "vllm_rbln/v1/worker/rbln_*.py"
+    - "vllm_rbln/v1/spec_decode/**"
+
 _sync: &sync "echo '--- :package: uv sync' && uv sync --locked --python 3.12 --extra test --extra dev --extra runtime && bash .buildkite/scripts/rebel-compiler-override.sh && echo '+++ :pytest: pytest'"
 
 steps:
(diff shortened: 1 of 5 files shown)


.buildkite/pipelines/optimum-pr/pipeline.yml
```diff
@@ -19,13 +19,35 @@ _secrets: &secrets
...
+# native-path-only, since over-including merely wastes a run while
+# over-excluding would silently skip a suite that should have caught a break.
+_if_changed: &if_changed
+  include:
+    - ".buildkite/**"
+    - "tests/optimum/**"
+    - "vllm_rbln/**"
+    - "pyproject.toml"
+    - "uv.lock"
+  exclude:
```
rebel-junamsong, line 35: @rebel-eunji Could you please review the following exclude list?

rebel-junamsong (approved):
LGTM

.buildkite/pipelines/optimum-pr/pipeline.yml
rebel-eunji: It looks fine.

rebel-eunji (approved):

PR #986, opened 2026-08-25
refactor(ci): build the optimum suites' environment once per build

rebel-wansoo:
The optimum PR pipeline kept dying in `uv sync` with

```
Timeout (300s) when waiting for lock on `/mnt/uv_cache/sdists-v9/editable/e5311780509d7016`
```

`e5311780509d7016` is a hash of the editable install *path*. Every job checks out
to the same container path, so they all hash to one cache entry, and uv holds
that lock for the whole build rather than only while writing the result
(astral-sh/uv#15380, open). Ten jobs start within two minutes of each other, so
the project builds serialised across the group. Jobs that survived paid for it
too — a passing job in build 338 reported `Prepared 1 package in 4m 23s` for the
single editable entry, next to `Installed 187 packages in 18.61s` for everything
else.

Rather than work around the lock, this removes the contention: `/mnt/cache` is
shared across a build's jobs, so **one step builds the environment there and the
ten suites only read it**. No suite runs `uv sync` any more, and only one job
builds the project, so there is nothing left to contend for.

Two flags on that sync are load-bearing rather than incidental:

* `--no-editable` — an editable install points at the builder's checkout, and no
  other job's pod has that tree. A wheel carries the code itself.
* `--compile-bytecode` — otherwise every suite writes `.pyc` into the shared
  `site-packages` the first time it imports, and they import concurrently.

`UV_PROJECT_ENVIRONMENT` points `uv sync` and `uv run` at the shared path;
`VIRTUAL_ENV` does the same for the `uv pip` call inside
`rebel-compiler-override.sh`, which would otherwise land in the image's
`/opt/venv`.

### DIFF
--- a/.buildkite/pipelines/optimum-pr/pipeline.yml
+++ b/.buildkite/pipelines/optimum-pr/pipeline.yml
@@ -21,11 +21,12 @@ _secrets: &secrets
 
 # Every job hashes to the same editable entry in the shared NFS cache and uv holds that
 # lock for the whole build, so the project install skips the cache (astral-sh/uv#15380).
-# uv rejects --no-cache under this pipeline's symlink link mode, hence the copy override.
+# Both overrides are load-bearing: uv rejects --no-cache under this pipeline's symlink
+# link mode, and uv pip otherwise targets the image's VIRTUAL_ENV over the project .venv.
 _sync: &sync |
   echo '--- :package: uv sync'
   uv sync --locked --python 3.12 --extra test --extra dev --extra runtime --no-install-project
-  UV_LINK_MODE=copy uv pip install -e . --no-deps --no-cache
+  UV_LINK_MODE=copy uv pip install --python .venv/bin/python -e . --no-deps --no-cache
   bash .buildkite/scripts/rebel-compiler-override.sh
   echo '+++ :pytest: pytest'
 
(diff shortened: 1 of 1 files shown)


.buildkite/pipelines/optimum-pr/pipeline.yml
```diff
@@ -19,9 +21,18 @@ _secrets: &secrets
...
 
 steps:
+  - label: ":package: shared venv"
+    key: "venv"
+    image: "${DEVTOOLS_DOCKER_IMAGE}"
+    secrets: *secrets
+    command:
+      - |
+        uv sync --locked --python 3.12 --extra test --extra dev --extra runtime --no-editable --compile-bytecode
+        bash .buildkite/scripts/rebel-compiler-override.sh
```
rebel-seinpark, line 58:
```suggestion
    command:
      - |
        echo '--- :package: uv sync'
        uv sync --locked --python 3.12 --extra test --extra dev --extra runtime --no-editable --compile-bytecode
        echo '--- :hammer: rebel-compiler-override'
        bash .buildkite/scripts/rebel-compiler-override.sh
```
just add marker plz

rebel-seinpark (approved):
LGTM! Thank you!

PR #989, opened 2026-08-25
model: support deepseek v3.2 with fp8

rebel-kblee:
### 🚀 Summary of Changes

support deepseek v3.2
- sparse attention mechanism (indexer, mla)
- e2e check (done)
- kv cache fp8 support (done)

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/attention/backends/mla/flashattn_mla.py
```diff
@@ -180,7 +185,7 @@ def _v_up_proj(self, x: torch.Tensor, W_UV: torch.Tensor) -> torch.Tensor:
             [batch, seq_len, num_heads * v_head_dim]
         """
         b_size, num_heads, seq_len, _ = x.size()
-        x = torch.matmul(x, W_UV)
+        x = torch.matmul(x, W_UV).to(torch.bfloat16)
```
rebel-cjlee: It now casts to `torch.bfloat16` unconditionally, but it's the shared exit for both the sparse and dense path. In case of `enforce-eager`, `platform.py` forces the dtype to `fp16`, so a `bf16` tensor reaches an fp16 `o_proj` and it dies on a dtype mismatch. Is it intended?

vllm_rbln/patches/mla.py
```diff
@@ -55,6 +56,40 @@ def patched_get_mla_prefill_backend(vllm_config) -> type[_RBLNNoOpMLAPrefillBack
...
+        "(512 fp8 kv_c + fp16 scale + k_pe), quantized/packed in-kernel by the "
+        "sparse_attn_deepseek_mla fp8 kernel. The compiler routes fp8 vs bf16 by "
+        "the cache channel width (768 != 576), so allocate head_size=768 int8. "
+        "cache_dtype_str stays 'auto' so the page size is the plain "
+        "head_size*itemsize (768B), not the upstream fp8_ds_mla 656B layout."
+    ),
+)
+def patched_mla_get_kv_cache_spec(self: MLAAttention, vllm_config):
+    cache_dtype = vllm_config.cache_config.cache_dtype
+    if cache_dtype and cache_dtype.startswith("fp8"):
```
rebel-cjlee: It only looks at `cache_dtype`, so every MLA layer gets the 768B int8 spec. With `supported_kv_cache_dtypes` now accepting fp8, DeepSeek R1 with `--kv-cache-dtype fp8` used to be rejected cleanly and now feeds a cache the dense kernels can't read. Does It need to gate on `self.use_sparse`?

vllm_rbln/patches/models_utils.py
```diff
@@ -23,10 +23,81 @@
...
+    from vllm.config import get_current_vllm_config
+
+    try:
+        cache_dtype = get_current_vllm_config().cache_config.cache_dtype
+    except Exception:
+        return False
+    return bool(cache_dtype) and cache_dtype.startswith("fp8")
+
+
+def rbln_num_attn_module(model_config) -> int:
```
rebel-cjlee: `rbln_num_attn_module` is a plain helper rather than a patch too, and having it in `/patches` makes `rbln_model_runner.py:2641` reach in there. Could it move to `v1/worker/utils.py` as well? `get_kv_cache_names` already takes this value as a parameter, so producer and consumer would sit together.

vllm_rbln/patches/models_utils.py
```diff
@@ -23,10 +23,81 @@
...
 
+@register_patch(
+    target="vllm.model_executor.models.utils.extract_layer_index",
+    reason=(
+        "DeepSeek-V3.2 sparse attention adds another KV-cache module "
+        "(the lightning indexer) per decoder layer. "
+    ),
+    apply_immediately=True,
+)
+def rbln_extract_layer_index(layer_name: str, num_attn_module: int = 1) -> int:
```
rebel-cjlee: This reads more like out own implementation than a patch, so how about moving it to `v1/worker/utils.py`? `get_kv_cache_names` already live there, and since that file mirrors `vllm.v1.worker.utils`, it could just be `extract_layer_index` without the `rbln_` prefix. After the move, `kv_cache_bindings.py:161` and `v1/worker/utils.py:872` are a one line import swap, and `patches/models_utils.py` keeps just the `vllm.v1.worker.utils` registration.

vllm_rbln/patches/models_utils.py
```diff
@@ -23,10 +23,81 @@
...
+)(rbln_extract_layer_index)
+
+
+def _dsa_indexer_cache_is_fp8() -> bool:
+    from vllm.config import get_current_vllm_config
+
+    try:
+        cache_dtype = get_current_vllm_config().cache_config.cache_dtype
+    except Exception:
+        return False
```
rebel-cjlee: When `get_current_vllm_config()` fails, this answers "not fp8" rather than "couldn't read the config". That global only lives inside a `set_current_vllm_config` block, and the dynamic KV resize path runs outside one, so it raises. Under fp8 that gives 2 modules instead of 3, which no longer matches the layer indices baked in at model build time, and `bind_kv_cache` dies with a bare `NotImplementedError`. All three callers already hold a `cache_config`, so taking it as an argument would drop the try/except and give the same answer everywhere.

-> `def rbln_num_attn_module(module_config, cache_config) -> int:`

vllm_rbln/patches/sparse_attn_indexer.py
rebel-cjlee, line 1: I think this belongs in `patches/deepseek_v2.py`. All three targets live in `vllm.model_executor.models.deepseek_v2`, and every other file under `patches/` is named after the module it patches. On top of that, `vllm.model_executor.layers.sparse_attn_indexer` is a real upstream module that this file never touches, so the name points somewhere it shouldn't.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-cjlee, line 1: This only drops a blank line, and the file is unrelated to the rest of the PR. Could we keep it out of the diff?

vllm_rbln/v1/attention/kv_cache_bindings.py
```diff
@@ -158,7 +158,9 @@ def _get_ordered_layer_names(
 ) -> list[str]:
     index_to_names: dict[int, list[str]] = defaultdict(list)
     for layer_name in kv_caches:
-        layer_index = extract_layer_index(layer_name, num_attn_module)
+        layer_index = model_executor_utils.extract_layer_index(
+            layer_name, num_attn_module
+        )
```
rebel-cjlee: If `extract_layer_index` moves to `v1/worker/utils.py` (vllm_rbln) as suggested above, this doesn't need the module lookup. Swapping the import from `vllm` to `vllm_rbln` should be enough.

vllm_rbln/v1/worker/utils.py
```diff
@@ -868,7 +868,9 @@ def get_kv_cache_names(
     """
     index2name: dict[int, list[str]] = defaultdict(list)
     for layer_name in kv_caches:
-        index2name[extract_layer_index(layer_name, num_attn_module)].append(layer_name)
+        index2name[
+            model_executor_utils.extract_layer_index(layer_name, num_attn_module)
+        ].append(layer_name)
```
rebel-cjlee: The same with `kv_cache_bindings.py`

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -2638,9 +2638,9 @@ def initialize_kv_cache_tensors(
             kv_cache_view_infos,
         )
 
-        num_attn_module = (
-            2 if self.model_config.hf_config.model_type == "longcat_flash" else 1
-        )
+        from vllm_rbln.patches.models_utils import rbln_num_attn_module
+
+        num_attn_module = rbln_num_attn_module(self.model_config)
```
rebel-cjlee: Once the `num_attn_module` helper lives in `v1/worker/utils.py`, this can just add the name to the existing import on line 139. The function local import from `patches/` goes away with it.

vllm_rbln/v1/attention/backends/mla/flashattn_mla.py
rebel-kblee: This was a workaround that was removed before it was fixed by the compiler.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-cjlee:
```suggestion
self.kv_cache_names = get_kv_cache_names(kv_caches, num_attn_module)
self.kv_caches = [kv_caches[name] for name in self.kv_cache_names]
forward_context = self.compilation_config.static_forward_context
for layer_name, kv_cache in kv_caches.items():
    forward_context[layer_name].kv_cache = kv_cache
```

Changing this should let us drop the remaining patch in `models_utils.py`. What `bind_kv_caches` gives us fits in these 4 lines, and `get_kv_cache_names` was already running the same ordering loop a second time anyway. We could also pull the 4 lines into our own `bind_kv_cache`, but this is the only place we call it, so I don't think it's worth splitting out.

One follow up: rbln_worker.py:967, 974, 1040 and 1042 all explain themselves in terms of upstream bind_kv_cache. The behaviour stays the same since the new loop parks the same views, but those comments would point at a call that no longer exists, so they need rewording. The `assert len(self.kv_cache_names) == len(self.kv_caches)` on 2657 also becomes redundant once both come from the same list.

vllm_rbln/v1/worker/utils.py
```diff
@@ -57,6 +56,53 @@
...
+def num_attn_module(model_config) -> int:
+    hf_config = model_config.hf_config
+    if getattr(hf_config, "model_type", None) == "longcat_flash":
+        return 2
+    text_config = getattr(model_config, "hf_text_config", hf_config)
+    # A DSA model puts the lightning-indexer key cache next to MLA:
+    # 2 modules, or 3 when the indexer cache is fp8 (a companion fp16 scale cache).
+    if hasattr(text_config, "index_topk") or hasattr(hf_config, "index_topk"):
+        return 3 if _dsa_indexer_cache_is_fp8() else 2
+    return 1
```
rebel-cjlee: The `except Exception` in `_dsa_indexer_cache_is_fp8` answers "not fp8" when it really just failed to read the config. `_reallocate_kv_cache` runs outside `set_current_vllm_config`, so an fp8 run can end up with 2 modules instead of 3. If `num_attn_module` took `cache_config` as an argument, `_dsa_indexer_cache_is_fp8` could go away entirely. Could you take a look? All three callers already hold one, and `deepseek_v2.py` even computes the same thing on line 151 before reading it back from the global on 156.

vllm_rbln/patches/models_utils.py
```diff
@@ -25,6 +25,18 @@
...
+rbln_num_attn_module = num_attn_module
+rbln_extract_layer_index = register_patch(
+    target="vllm.v1.worker.utils.extract_layer_index",
+    reason=(
+        "vllm.v1.worker.utils snapshots extract_layer_index at import time, so "
+        "bind_kv_cache misses the patch on the defining module."
+    ),
+    key="vllm_rbln.patches.models_utils.rbln_extract_layer_index.v1_worker_utils",
+    apply_immediately=True,
+)(extract_layer_index)
```
rebel-cjlee: Please take a look at the comment in `rbln_model_runner.py`.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-kblee: Oh, I understand and agree. It's a good idea. I’ve checked the review and will give it a go. Thank you.

vllm_rbln/v1/worker/utils.py
rebel-kblee: Thanks. I agree with the content of your review. I'll apply this review.

rebel-cjlee (approved):

PR #993, opened 2026-08-25
model: support axk2 with fp8

rebel-kblee:
### 🚀 Summary of Changes

same as deepseek v3.2 with fp8 support pr (https://github.com/RBLN-SW/vllm-rbln/pull/989)
difference
- reference vllm repo is separated from upstream (https://github.com/SKT-AI/vllm)
- I have selected and included only the necessary parts, and have verified that they function correctly
-  Once SKT has incorporated the changes into the upstream repository (= after the VLLM version has been bumped), we can simply remove the relevant code.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/patches/metrics.py.bak
rebel-cjlee, line 1: Looks like this slipped in by accident. It's byte identical to `patches/metrics.py`, which is unchanged from `dev`.

vllm_rbln/patches/axk2/ORIGIN
rebel-cjlee, line 1: Just checking this file is intentional? It has everything a vendoring note needs, and both sha256s check out on my end.

rebel-cjlee (approved):
The structure looks good to me. Freezing the SKT sources and keeping everything that moves in `model.py` is the right call here.

Two things.
Is there a plan for when this gets deleted? `_upstream_has_axk2()` is already there, so `register_model()` could raise instead of registering once upstream ships axk2. A version bump would then fail loudly and remind us to drop the directory, rather than our frozen copy quietly shadowing the real one. ORIGIN says `register_model()` self disables today, but it registers unconditionally, so that line needs a touch up either way.

`_rbln_axk2_moe_forward` has the same body as `patched_deepseek_v2_moe_forward` in `deepseek_v2.py`, down to the line. Worth merging them so the two can't drift apart.

vllm_rbln/patches/metrics.py.bak
rebel-kblee: Yes... it's my mistake. thank you for check this.

vllm_rbln/patches/axk2/ORIGIN
rebel-kblee: I’m not entirely sure about the ORIGIN code either. It’s true that it works perfectly well even without it. However, when utilising code from other repositories, I’m not sure how best to organise and manage the revision and meta information, so I’ve set it to ORIGIN for the time being, but I’m still a bit unsure about it...

PR #996, opened 2026-08-26
fix(test): make optimum OpenAI entrypoint tests resilient to HF Hub flakiness

rebel-seinpark:
## Background

`tests/optimum/entrypoints/openai` fails intermittently in CI on HuggingFace
Hub connection issues, even though the tests themselves are fine.

## Cause

`RemoteOpenAIServer` starts a `vllm serve` subprocess that touches the Hub at
startup -- it fetches the tokenizer and re-validates every already-cached file
with a per-file etag HEAD. That path is exposed to transient connection errors
and Hub rate limits (429/5xx), and it is the remaining source of flakiness (the
pre-download was already retried, but only for weights).

## Change

This vllm-rbln build already has an offline startup path: when `HF_HUB_OFFLINE`
is set, `EngineArgs` resolves the model through `get_model_path() ->
snapshot_download(local_files_only=True)`. Opt into it.

- **Warm-up (parent, online, retried):** pull the model's **complete snapshot**
  with `huggingface_hub.snapshot_download`, with exponential backoff on transient
  network errors.
- **Server (subprocess, offline):** launch with `HF_HUB_OFFLINE=1`, so startup
  issues **no Hub request at all** -- no download, no etag revalidation -- and
  cannot fail on a flaky link or a rate limit.

Why the *complete* snapshot: the offline `snapshot_download(local_files_only=
True)` inside `get_model_path` demands every repo file (down to `.gitattributes`
/ `LICENSE.md` / `README.md`) and raises `IncompleteSnapshotError` otherwise -- a
weights-only warm-up is not enough.

The retry set is `(requests.exceptions.RequestException, LocalEntryNotFoundError)`:
`RequestException` covers connection resets and 429/5xx (`HfHubHTTPError`);
`LocalEntryNotFoundError` is how hf_hub reports a dropped connection when the file
is not yet cached (an `OSError`, not a `RequestException`), so it must be listed
explicitly. A plain `OSError` (disk full, permission) is left out so a real local
bug fails fast.

## Also

Raise the server-ready ceiling from 240s to 600s so a slow or overloaded CI
runner does not time out during startup on otherwise healthy runs.

## Scope

optimum path only; touches only the test helper
`tests/optimum/entrypoints/openai/utils.py`. No production code changes.

## Testing

- `uvx pre-commit run --files tests/optimum/entrypoints/openai/utils.py` passes.
- Verified the previously failing resolution locally: after the full warm-up,
  `HF_HUB_OFFLINE=1` `get_model_path("facebook/opt-125m")` returns the local
  snapshot (the `.gitattributes` / `LICENSE.md` / `README.md` that CI reported
  missing are now present), instead of raising `IncompleteSnapshotError`.
- The full suite still needs an NPU CI run to confirm end-to-end (my local box
  falls back to the CPU backend, an unrelated KV-cache limit).

### DIFF
--- a/tests/optimum/entrypoints/openai/utils.py
+++ b/tests/optimum/entrypoints/openai/utils.py
@@ -46,14 +46,7 @@ def _warm_hf_cache_with_retry(
     model: str, revision: str | None, *, retries: int = 5, base_delay: float = 2.0
 ) -> None:
     """Download the model's complete snapshot into the local HF cache, retrying
-    on transient connection errors.
-
-    The server is launched with ``HF_HUB_OFFLINE=1`` (see ``_start_server``), the
-    offline path this vllm-rbln build is built around: EngineArgs then resolves
-    the model through ``get_model_path() -> snapshot_download(local_files_only=
-    True)``, which demands the *complete* snapshot and raises
-    ``IncompleteSnapshotError`` if any repo file is missing -- a weights-only
-    cache is not enough."""
+    on transient connection errors."""
     for attempt in range(1, retries + 1):
         try:
             huggingface_hub.snapshot_download(model, revision=revision)
--- a/tests/optimum/entrypoints/openai/utils.py
+++ b/tests/optimum/entrypoints/openai/utils.py
@@ -22,18 +22,53 @@ from typing import Any
 
 import anthropic
 import httpx
+import huggingface_hub
 import openai
 import requests
-from vllm.engine.arg_utils import AsyncEngineArgs
+from huggingface_hub.errors import LocalEntryNotFoundError
 
 # from tests.models.utils import TextTextLogprobs
 from vllm.entrypoints.cli.serve import ServeSubcommand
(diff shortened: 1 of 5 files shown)


tests/optimum/entrypoints/openai/utils.py
```diff
@@ -22,20 +22,55 @@
...
+) -> None:
+    """Download the model's complete snapshot into the local HF cache, retrying
+    on transient connection errors.
+
+    The server is launched with ``HF_HUB_OFFLINE=1`` (see ``_start_server``), the
+    offline path this vllm-rbln build is built around: EngineArgs then resolves
+    the model through ``get_model_path() -> snapshot_download(local_files_only=
+    True)``, which demands the *complete* snapshot and raises
+    ``IncompleteSnapshotError`` if any repo file is missing -- a weights-only
+    cache is not enough."""
```
rebel-eunji: Could you remove redundant comments generated by AI?

rebel-eunji (approved):

PR #997, opened 2026-08-26
fix(kv-cache): port RBLN SWA external-token allocation to vllm 0.24 hooks

rebel-jonghewk:
### Summary

`RBLNSlidingWindowManager` overrode `allocate_new_computed_blocks`, but vllm 0.24.0 (upstream 588db18362) split that hook into `add_local_computed_blocks` + `allocate_external_computed_blocks`. Since the 0.24.0 bump (#829) the override has been dead code, so on the KV-connector D side the base implementation allocates `cdiv(num_external_tokens, block_size)` blocks for the SWA group while P holds a single block. Any remote prefill longer than one block then kills the D engine in nixl `_apply_prefix_caching` (`assert num_local_blocks <= len(remote_group)`).

This PR overrides the two new hooks and adds a test that goes through the real `KVCacheCoordinator`, so the hook names are verified against the pinned vllm instead of by direct method call.

### Details

- **Root cause (ICR-15)** — gpt-oss-120b pdd-dp2ep + `RblnNixlConnector`, block_size 4096, 6077 external tokens. The D-side dump shows `local_block_ids=[[1, 2], [3, 4]]` vs `remote.block_ids=[[2], [1, 3]]`: group 0 (SWA) got 2 blocks on D because `SingleTypeKVCacheManager.allocate_external_computed_blocks` ran instead of the RBLN override. 81-token requests passed because both sides were 1 block.

- **Fix** — `add_local_computed_blocks` keeps the "no prefix hits" assertion and seeds `num_cached_block`. `allocate_external_computed_blocks` allocates exactly one block when there are external tokens and none yet, matching `allocate_new_blocks`.

- **Test** — `TestCoordinatorExternalTokens` builds a hybrid (RBLN SWA + Full) `KVCacheCoordinator` and calls `allocate_new_computed_blocks` with 6077 external tokens; SWA group must hold 1 block, Full group `cdiv(6077, 4096) == 2`. It fails on `dev` with `assert 2 == 1` and passes with this change. Direct-call tests are renamed to the new hooks.

### Next Steps

- Re-run the ICR-15 case (`pdd-dp2ep-throughput`, `kv_connector=nixl`, long_context input) once merged.
- Consider a CI guard that fails when an override name in `vllm_rbln` no longer exists on the pinned vllm base class.

---

* Related to https://rbln.atlassian.net/browse/ICR-15

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `pytest tests/native/v1/test_kv_cache.py` against vllm 0.24.0+cpu — 16 passed.
2. `git stash -- vllm_rbln/v1/kv_cache.py` and rerun `-k Coordinator` — fails with `assert 2 == 1` (reproduces ICR-15).
3. E2E: prefill/decode `vllm serve openai/gpt-oss-120b --block-size 4096 --data-parallel-size 2 --enable-expert-parallel --kv-transfer-config '{"kv_connector":"RblnNixlConnector","kv_role":"kv_both","kv_buffer_device":"rbln"}' --no-enable-prefix-caching --num-gpu-blocks-override 48`, send a >4096-token prompt; D must not raise `AssertionError` in `_apply_prefix_caching`.

---

### 📋 Checklist

---

### 💬 Notes

Draft: unit-tested only; E2E on the ICR-15 case is pending.

### DIFF
--- a/tests/native/v1/test_kv_cache.py
+++ b/tests/native/v1/test_kv_cache.py
@@ -126,11 +126,13 @@ class TestRBLNSlidingWindowManager:
         assert len(m.req_to_blocks["r"]) == 1
         assert pool.get_num_free_blocks() == free0 - 1
 
-    def test_allocate_external_noop_when_present(self):
+    def test_allocate_external_rejects_preexisting_blocks(self):
+        # add_local_computed_blocks runs first and leaves the list empty; a
+        # block already present means the coordinator sequence broke.
         m = _manager(_never_pool())
         m.req_to_blocks["r"] = [object()]
-        m.allocate_external_computed_blocks("r", 0, 5)
-        assert len(m.req_to_blocks["r"]) == 1
+        with pytest.raises(AssertionError):
+            m.allocate_external_computed_blocks("r", 0, 5)
 
     def test_find_longest_cache_hit_returns_empty_per_group(self):
         # Prefix caching disabled: one empty list per kv_cache_group_id.
--- a/tests/native/v1/test_kv_cache.py
+++ b/tests/native/v1/test_kv_cache.py
@@ -100,33 +100,37 @@ class TestRBLNSlidingWindowManager:
         m.req_to_blocks["r"] = [object()]
         assert m.allocate_new_blocks("r", 10, 10) == []
 
-    def test_allocate_computed_fresh_no_external_allocates_nothing(self):
+    def test_add_local_computed_fresh_marks_cached_without_blocks(self):
         m = _manager(_never_pool())
-        m.allocate_new_computed_blocks("r", [], 0, 0)
+        m.add_local_computed_blocks("r", [], 0, 5)
         assert m.num_cached_block["r"] == 0
         assert m.req_to_blocks["r"] == []
 
-    def test_allocate_computed_fresh_with_external_allocates_one(self):
-        pool = _pool()
-        free0 = pool.get_num_free_blocks()
-        m = _manager(pool)
-        m.allocate_new_computed_blocks("r", [], 0, 5)
-        assert len(m.req_to_blocks["r"]) == 1
-        assert pool.get_num_free_blocks() == free0 - 1
-
-    def test_allocate_computed_already_cached_is_noop(self):
-        # Idempotent fast path: a request already in num_cached_block draws no
-        # block (external tokens are ignored once cached).
-        m = _manager(_never_pool())
-        m.num_cached_block["r"] = 0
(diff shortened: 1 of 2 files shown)


vllm_rbln/v1/kv_cache.py
```diff
@@ -83,24 +96,12 @@ def allocate_new_computed_blocks(
...
-
-        # Sentinel for the base-class fast path; 0 because RBLN neither skips
-        # nor pulls from prefix cache.
-        self.num_cached_block[request_id] = 0
-
-        if num_external_computed_tokens > 0:
-            new_blocks = self.block_pool.get_new_blocks(1)
-            req_blocks.extend(new_blocks)
+        if req_blocks:
+            return
```
rebel-cjlee: Can this go back to `assert len(req_blocks) == 0`? The coordinator runs `add_local_computed_blocks` for every group first, and that already asserts the list is empty, so a non empty one here means something broke. Returning hides it, and so would dropping the check.

tests/native/v1/test_kv_cache.py
```diff
@@ -100,33 +100,37 @@ def test_allocate_new_blocks_noop_when_present(self):
...
-
-    def test_allocate_computed_rejects_prefix_hit_blocks(self):
-        # A prefix-cache hit (non-empty new_computed_blocks) is unsupported.
+    def test_allocate_external_noop_when_present(self):
         m = _manager(_never_pool())
-        with pytest.raises(AssertionError):
-            m.allocate_new_computed_blocks("r", [object()], 0, 0)
+        m.req_to_blocks["r"] = [object()]
+        m.allocate_external_computed_blocks("r", 0, 5)
+        assert len(m.req_to_blocks["r"]) == 1
```
rebel-cjlee: If that goes back to an assert, this one goes with it. It hand builds a state `add_local_computed_blocks` already rejects, so nothing can reach the branch it's pinning.

vllm_rbln/v1/kv_cache.py
rebel-jonghewk: Agreed, restored `assert len(req_blocks) == 0` in 646f99599f489619aff8d72e9af3a458b6ff8940.

tests/native/v1/test_kv_cache.py
rebel-jonghewk: Replaced with `test_allocate_external_rejects_preexisting_blocks`, which pins the assert instead (`pytest.raises(AssertionError)`). Same commit.

rebel-cjlee (approved):

PR #999, opened 2026-08-26
perf(compile): dispatch compiled graphs without Dynamo's cache lookup

rebel-cjlee:
### 🚀 Summary of Changes

Every call into a compiled graph pays a fixed cost unrelated to the graph: the `torch.compile` wrapper, the frame eval, and  cache walk, and a guard check over **every tensor that enters the graph**. RBLN needs every KV cache to enter as an input, so each step re-checks dtype/device/shape/stride on tensors that are allocated once and never replaced. `keep_tensor_guards_unsafe` already dropped the value guards; this removes what is left.

`Dispatcher` owns the mapping instead. The key is the staged inputs' shapes and dtypes plus context the runner already knows (prefill vs decode, DP pad width, KV-connector presence). The first call per key goes through Dynamo and its transformed bytecode is recorded as a clone of the target (a clone rather than a `__code__` swap, which would be visible to every other holder of the function and would mutate the code object a bound method's class shares.

**Limitations**

- **Requires `fullgraph=True`.** A graph break leaves the rest of the frame in a resume function that only runs compiled under the frame eval, and a recompile-limit bail-out adds no cache entry, so `_register` would bind the previous call's graph to the new key. `compile()` rejects the pair.
- **Key completeness is a precondition, not a check.** A registered key is served `_value_key` raises on anything it cannot key and `TestKeyDiscrimination` pins the situations the runner produces, but nothing verifies this at runtime. A `verify` mode calling `entry.guard_manager.check(...)` on the warm path would close it (designed, not implemented).
- **Assumes serial calls.`** Identifying the recorded graph relies on Dynamo's cache list being MRU-ordered; a second concurrent caller of the same target would need this serialized.
- **The sampler is compiled but not dispatched** (one lookup per step remains).

---

* Resolves #
* Related to #

---

### ✅ Type of Change

---

### 🧪 How to Test
```
uv run --no-sync pytest tests/native/compilation -x
uv run --no-sync pytest tests/native -m model_compile --model-compile
```
Unit lane covers the key, the clone, and the Dynamo contract (MRU order, entry.code/compile_id) against real Dynamo. The model-compile lane covers output correctness against HF logprobs, all three dispatch sites (runner / eagle / medusa), and DP4 e2e. Device counts differ per test (1 / 2 / 4) and RBLNWorker._init_device_env asserts on a mismatch.

---

---

### 📋 Checklist

---

### 💬 Notes

Worth putting behind an env var so it can be flipped without editing the source? It would buy A/B measurement without a rebuild (every number here came from flipping the flag by hand) and a way to confirm or rule out the dispatcher in the field, where a key that is too coarse surfaces as wrong output rather than an error.

A single global flag gating the use_direct_dispatch argument leaves the call sites that don't opt in (compute_logits, the sampler) untouched, which seems right. It would belong in RBLN_NON_COMPILE_ENV (it changes graph selection, not the compiled artifacts), so it must not invalidate the mega-cache key. Options welcome.

---

### DIFF
--- a/vllm_rbln/compilation/dispatch.py
+++ b/vllm_rbln/compilation/dispatch.py
@@ -20,6 +20,7 @@ import torch
 from torch._C._profiler import _RecordFunctionFast
 from vllm.distributed.kv_transfer import has_kv_transfer_group
 from vllm.forward_context import get_forward_context, is_forward_context_available
+from vllm.sequence import IntermediateTensors
 
 from vllm_rbln.logger import init_logger
 
@@ -42,14 +43,19 @@ _COPIED_SLOTS = (
 
 
 def _value_key(value: Any) -> Any:
-    """The part of an argument that decides which graph it belongs to."""
+    """The part of an argument that decides which graph it belongs to.
+
+    A tensor contributes the members Dynamo's TENSOR_MATCH compares, less
+    requires_grad, which the step loop's inference_mode pins.
+    """
     if isinstance(value, torch.Tensor):
-        # torch.Size is a tuple, so this stays hashable and allocation-free.
-        return value.shape, value.dtype
+        return value.shape, value.dtype, value.device, value.stride()
     if value is None or isinstance(value, (bool, int, float, str, bytes, torch.dtype)):
-        return value
-    if hasattr(value, "items"):  # IntermediateTensors and friends
-        return tuple((k, t.shape, t.dtype) for k, t in value.items())
+        # Bare, True would fold into 1 and 1 into 1.0: Python hashes them equal
+        # where Dynamo specialises them apart.
+        return type(value), value
+    if isinstance(value, IntermediateTensors):
+        return tuple((k, _value_key(t)) for k, t in value.items())
     raise TypeError(
         f"Dispatcher cannot key a {type(value).__name__} argument. Dynamo may "
         "guard on its value, and a key that does not carry it would serve the "
--- a/vllm_rbln/compilation/dispatch.py
+++ b/vllm_rbln/compilation/dispatch.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 from types import CodeType, FunctionType, MethodType
 from typing import Any
 
@@ -35,7 +36,8 @@ _COPIED_SLOTS = (
     "__module__",
     "__doc__",
     "__annotations__",
-    "__type_params__",
+    # A function carries type params only from 3.12, as does the test's enumeration.
+    *(("__type_params__",) if sys.version_info >= (3, 12) else ()),
 )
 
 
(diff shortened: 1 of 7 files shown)


rebel-daeyang (approved):
LGTM

vllm_rbln/compilation/dispatch.py
```diff
@@ -0,0 +1,200 @@
...
+
+def _value_key(value: Any) -> Any:
+    """The part of an argument that decides which graph it belongs to."""
+    if isinstance(value, torch.Tensor):
+        # torch.Size is a tuple, so this stays hashable and allocation-free.
+        return value.shape, value.dtype
+    if value is None or isinstance(value, (bool, int, float, str, bytes, torch.dtype)):
+        return value
+    if hasattr(value, "items"):  # IntermediateTensors and friends
+        return tuple((k, t.shape, t.dtype) for k, t in value.items())
```
rebel-jonghewk: Are you keeping `device` and `stride` out on purpose?

rebel-jonghewk: feels like it should be included.

vllm_rbln/compilation/dispatch.py
```diff
@@ -0,0 +1,200 @@
...
+    raise TypeError(
+        f"Dispatcher cannot key a {type(value).__name__} argument. Dynamo may "
+        "guard on its value, and a key that does not carry it would serve the "
+        "graph recorded for a different value."
+    )
+
+
+def _forward_context_key() -> tuple[Any, ...]:
```
rebel-jonghewk, line 64: Nit: 

Can you pull this method out of dispatch.py? This function pulls three vllm symbols into dispatch.py. 

May be define sth like this in forward_context.py 
```
def dispatch_context_key() -> tuple[bool | None, int | None]:
    if not is_forward_context_available():
        return (None, None)
    ctx = get_forward_context()
    max_pads = None if ctx.dp_metadata is None else ctx.dp_metadata.max_pads_across_dp
    attn_metadata = ctx.attn_metadata
    if isinstance(attn_metadata, dict):
        attn_metadata = next(iter(attn_metadata.values()), None)
    return (
        None if attn_metadata is None else attn_metadata.is_prefill,
        None if max_pads is None else max_pads.shape[0],
    )

```
and can be injected to Dispatcher? 
```
class Dispatcher
    def __init__(
        self,
        target: Any,
        compiled: Any,
        context_key: Callable[[], Hashable] = tuple,
    ) -> None:
```

```
return cast(CompiledTarget, Dispatcher(target, compiled, dispatch_context_key))
```

vllm_rbln/compilation/dispatch.py
rebel-cjlee: I checked, and it can't actually break today. Every tensor that gets here is an `InputStager` buffer, so it's always contiguous and always on the stager's device. But that's true becau;se of `input_stager.py`, not because this module checks it, and the docstring here says a key coarser than the guards is wrong rather than reported. So I widened it to `(shape, dtype, device, stride())`.

Two more while I was in there. The scalar branch returned the value bare, so `True` and `1` gave the same key even though Dynamo treats th;em as different types. And `hasattr(value, "items")` let a plain dict through, so that's `isinstance(value, IntermediateTensor)` now, and it calls `_value_key` recursively so the members can't drift apart again. Worth doing anyway, since intermediate tensors are the only argument `stage()` passes through as is.

vllm_rbln/compilation/dispatch.py
rebel-cjlee: `_forward_context_key` reads `dp_metadata.max_pads_across_dp`, and we define that field ourselves in `forward_context.py`, so the coupling you're pointing at is real. But only one of the three key members lives t here. `is_prefill` comes from `RBLNFlashAttentionMetadata` and `has_kv_transfer_group()` is upstream vLLM, so moving it only helps one out of three, and the KV connector part would look odder there than it does here. I'd also rather keep the whole key in one file, and right now `dispatch.py` depending on `forward_context.py` is the normal direction anyway.

On the constructor parameter, I'd skip it for now. If someone forgets to pass it, the `tuple` default silently drops the context from the key, which is the one bug this class can't catch.

rebel-jonghewk: has_kv_transfer_group() is constant for the process (your own test_dispatch.py says so), so it adds nothing to the key. (Maybe we can drop it?)

And what's left  is_prefill and max_pads.shape[0], both read off get_forward_context(). That makes the whole  key forward-context state, which forward_context.py should own.

On the default: agreed that `= tuple` would silently drop the context — so make it required, and fold the bool into it: `compile(..., dispatch_key: Callable[[], Hashable] | None = None)`. None = off; runner/eagle pass forward_context.dispatch_context_key; medusa passes a named no-context key so the "runs outside the forward context" decision is stated at the call site rather than defaulted. dispatch.py then imports nothing from vllm.*, and the tests stop monkeypatching a module global.

That said, none of this changes behavior — it's about where the key lives, not what's in it — so I'm fine landing this as-is and revisiting the split in a follow-up. Dropping the dead has_kv_transfer_group() member is the one bit I'd still do here.

rebel-jonghewk (approved):

PR #1002, opened 2026-08-26
fix(optimum): detect a compiled model dir by rbln_config.json

rebel-eunji:
### 🚀 Summary of Changes

`os.path.exists(model_path)` accepted any local directory, so a plain HuggingFace checkpoint -- which is what `model_path` points at under HF_HUB_OFFLINE=1 -- was taken for a pre-compiled model.

Gate on the top-level `rbln_config.json` instead. That is optimum-rbln's own marker for a compiled directory (`RBLNBaseModel._load_compiled_model_dir`), it always sits at the top level, and a plain HF checkpoint never has one.

Globbing for top-level `*.rbln` would not work: models whose config sets `_allow_no_compile_cfgs` (e.g. PaliGemma, which vLLM supports) compile no graph of their own and keep every `.rbln` inside submodule subdirectories.

Applying the same predicate to `cached_model_path` also stops a cache dir left half-written by an interrupted compile from counting as a hit.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/model_executor/models/optimum/model_base.py
+++ b/vllm_rbln/model_executor/models/optimum/model_base.py
@@ -36,18 +36,13 @@ from vllm_rbln import envs
 from vllm_rbln.logger import init_logger
 from vllm_rbln.utils.optimum.block_size import get_attn_block_size
 from vllm_rbln.utils.optimum.bucket import select_bucket_size
+from vllm_rbln.utils.optimum.paths import is_compiled_dir
 from vllm_rbln.utils.optimum.registry import get_rbln_model_info
 
 from .base import ModelInputForRBLN, PartialPrefixInfo
 from .compilation import RBLNCompileSpec
 
 logger = init_logger(__name__)
-
-
-def _is_compiled_dir(path: str | None) -> bool:
-    if path is None:
-        return False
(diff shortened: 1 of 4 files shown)


vllm_rbln/model_executor/models/optimum/model_base.py
```diff
@@ -44,6 +44,12 @@
 logger = init_logger(__name__)
 
 
+def _is_compiled_dir(path: str | None) -> bool:
+    if path is None:
+        return False
+    return bool(path) and os.path.isfile(os.path.join(path, "rbln_config.json"))
```
rebel-seinpark:
```suggestion
    if not path:
        return False
    return os.path.isfile(os.path.join(path, "rbln_config.json"))
```

rebel-eunji: I applied it and refactored.

rebel-seinpark (approved):

rebel-cjlee (approved):

PR #1006, opened 2026-08-26
fix(eagle3): support a pipeline-parallel MiniMax-M2 target

rebel-chanheo:
## What

Makes EAGLE3 work when a MiniMax-M2 target runs pipeline-parallel, which is what a P/D deployment needs: the drafter lives on the last stage, and the aux hidden states it consumes come from layers spread across every stage.

`--speculative-config` has to be on the prefill role too, not just decode. The drafter keeps its own KV cache and the only thing that fills its prompt region is a prefill forward — `llm_base_proposer.py`'s comment sits in the `num_speculative_tokens == 0` branch, i.e. the draft forward runs even when zero draft tokens are requested, precisely to keep that cache in sync. Since `validate_same_kv_cache_group` places the draft's attention layers in `kv_cache_config.kv_cache_groups`, that cache is a full member of what NIXL registers and transfers. Upstream `vllm-project/vllm#39266` puts `--speculative-config` on its prefill role for the same reason.

## Scope

One (target, draft) pair is claimed: a **MiniMax-M2** target with an `Eagle3MiniMaxM2ForCausalLM` draft, which vLLM maps to `llama_eagle3`'s `Eagle3LlamaForCausalLM`. Anything else is rejected at startup rather than left to fail mid-compile: `RblnPlatform._validate_eagle3_pp_config` raises for a pipeline-parallel EAGLE3 target outside `EAGLE3_PP_TARGET_ARCHS`. A draft whose `eagle_config` sets `use_aux_hidden_state=false` is exempt, and has to be — nothing is captured anywhere then, `aux_hidden_state_layers` stays empty, and upstream's unpatched forward is harmless under the split. Rejecting that would block a configuration that works.

The halves of the fix have different reach, and only one of them is architecture-bound.

**Naming the EAGLE head** (`patches/llama_eagle3.py`) applies to every draft that resolves to `llama_eagle3.Eagle3LlamaForCausalLM`, which is most EAGLE3 checkpoints — the stock `yuhuili/EAGLE3-LLaMA3.1-Instruct-8B` (`LlamaForCausalLMEagle3`) included. The `deepseek_eagle3` and `qwen3_eagle3` heads carry the identical defect and are not patched.

**Carrying the aux states across stages** is split by what is actually architecture-bound. The slot naming, which indices a stage receives versus captures, the handoff placeholder, and the single reader for the draft's `use_aux_hidden_state` flag live in `v1/spec_decode/eagle3_pp.py` and are architecture-independent. What stays in `patches/minimax_m2.py` is the `forward` body alone: the captured tensors have to be graph outputs of the forward that produced them, so stashing them on the module and merging outside is the same class of thing that broke the graph in *Two things that surprised us*. Adding the next architecture is that body, not another copy of the mechanism.

The handoff placeholder is rebound where the aux layers become known — `set_aux_hidden_state_layers` in the runner's `load_model` — rather than by patching a model's `__init__`. `ForCausalLM.__init__` copies the inner model's callable onto itself and the runner calls that copy, so the installer binds it there and reads the key set at call time. At PP=1 a stage receives nothing, so it is a no-op.

The guard raises rather than warns because of how the unguarded case fails. The requested capture indices are global — `(2, num_layers // 2, num_layers - 3)` off the PPMissingLayer-padded list, so `(2, 16, 29)` for Llama-3.1-8B — while a stage matches them against its own `enumerate`, which restarts at zero. At PP=2 stage 1 owns 16..31 and therefore harvests global 18 and 32 and never sees 29, and stage 0, which did capture 2 and 16 correctly, drops them at the handoff. Two tensors reach an `fc` sized for three, and the shape mismatch surfaces mid-compile. Where the counts happen to line up the same confusion is silent: the drafter is fed the wrong layers, compiles, passes a short probe, and only loses acceptance rate.

Fifteen model files in vLLM share the defective pattern — `_maybe_add_hidden_state([], 0, ...)` plus `idx + 1` off a stage-local `enumerate`, then a non-last-rank return that drops the list — and `llama.py` is one of them. Fixing the index half once in `EagleModelMixin` is possible and is deliberately not done: it corrects which layer each capture is, but the captures still never cross a stage boundary, so no additional target starts working. On the case above it turns two wrong tensors into one right one, which still cannot fill three, and the guard rejects that configuration before it runs. Upstream #50514 fixes both halves in the mixin and has each model opt in; this branch aims to be what gets deleted when that lands.

## Why these shapes

Two facts drive most of the diff.

**A capture index is global, and an aux tensor has to survive the handoff.** Index `i` means "input to layer `i`", so a stage's incoming hidden state carries the same index as the previous stage's final capture — taking both would duplicate a boundary layer. Received indices are at or before `start_layer` and captured ones past it, so received-then-captured is already ascending, which is the order the drafter's `fc` expects its concatenated blocks in. Getting that order wrong corrupts output rather than failing.

**The EAGLE head has to be named past the target's full depth.** `pipeline_adjusted_layer_index` (added for MTP/nextn in #980) already routes a layer named past the target's layer count to `(end - start) + offset`, which is where its KV cache sits in the compacted per-rank list — and an EAGLE3 head belongs to that same class. Upstream names it with `start_layer_id=get_num_layers(parallel_config)`, the count on *this* rank, so it lands below the threshold and is treated as a target layer. With PP=1 the per-rank count already equals the full depth, which is why this is invisible until a rank that does not start at zero — the rank the drafter lives on. Renaming also makes the name identical on a prefill and a decode instance, which is what NIXL matches KV regions by.

## What is deliberately not patched

An earlier revision of this branch also patched `Eagle3LlamaForCausalLM.__init__`/`load_weights`/`compute_logits` to keep the draft->target scatter index a named buffer. That defect is real — the index is input-independent, so it const-folds into an anonymous constant that weight-free apply cannot resolve by name, and the out-of-bounds host-op write segfaults in KV warmup. But `RBLNEagleProposer` already avoids it (#831): it stays in draft-vocab space and maps ids after the argmax, so upstream's `compute_logits` is never the path that runs. Those three patches have been dropped and the module docstring records where the real fix lives.

They survived the rebase because they patch the model while #831 fixed the proposer, so git had nothing to conflict on and the suite stayed green. A patch that is merely harmless is still a patch: it asserts something about upstream that has stopped being true. **After a rebase, "no conflict" is not evidence that a patch is still needed** — that has to be checked separately.

## Two things that surprised us

**Keying a dict by a tensor index inside the layer loop breaks the graph at that statement.** The backend then receives the residual-stream add on its own, as two no-op bf16 casts, and rejects it while *registering* the module rather than while compiling it — so the error names neither a layer nor an op. Factoring the capture into a closure does the same thing, because dynamo lifts the captured variables into arguments instead of inlining. Upstream appends to a list, and this is plausibly why. `vllm_rbln/compilation/backends.py`'s own log is what showed it: its `caller=` field names the code that produced each graph, and a split shows up as anonymous input names (`l_x_0_`) instead of real ones.

**The draft and the target are named against different origins,** and #980 had already solved this for MTP/nextn before we got here. Recognising that the EAGLE3 head is the same class of layer replaced what had been a separate index rule on our side, so this branch renames the head — and moves `target_layer_count` with the name — and leaves `pipeline_adjusted_layer_index` untouched. Deriving the head's index by arithmetic on its own name does not work: subtracting the rank's start goes negative off stage 0, and subtracting the rank's layer count is right *only* on stage 0.

## Verification

Four PP4 runs on the device, one per configuration worth separating. Every one of them registers 15/16/16/16 KV layers, raises no aux-slot `KeyError`, and moves the spec-decode counters consistently (`draft tokens = proposing steps x num_spec`): M2.5 with the drafter's own `(1, 30, 58)`; M2.5 with `(2, 31, 59)`; then M2.7 and M2.1, which share MiniMax-M2.5's depth, hidden size and vocabulary, so they re-run the same split against different weights. Their acceptance counts are not evidence of anything — the only drafter available is M2.5's — but the handoff, the KV registration and the counters are.

**The `(2, 31, 59)` run is the one that earns its place.** 31 is stage 2's `start_layer` on the `[15, 16, 16, 15]` split, so it is both stage 1's last capture and stage 2's incoming hidden state — the index that an off-by-one in the split double-counts or drops. The drafter is byte-identical to M2.5's (same `model.safetensors` md5); only the aux ids in its config differ, so the run moves one variable. Its greedy output is identical to the `(1, 30, 58)` run's. Speculative decoding is lossless, so the target's output cannot depend on which layers the drafter reads: had the boundary been mishandled, the drafter would have been fed a different hidden state, proposed differently, and the verified tokens would have diverged. None of the other three aux sets touches a boundary at all.

The startup guard was exercised through the CLI rather than only through its own unit test: a Qwen3-1.7B target with an EAGLE3 draft at PP=2 is rejected before any compilation begins.

A 1P1D probe passes with the transfer path and the local path producing the same output. That probe predates the refactor above, and the one thing it uniquely covers — NIXL matching KV regions by name — is unaffected by it: `get_total_num_hidden_layers()` and `hf_text_config.num_hidden_layers` return the same 62 for this target, which has no arch-config convertor of its own, so the head's name is unchanged.

The layer naming was checked directly rather than inferred: on the 62-layer PP4 split the prefill instance registers 15/16/16/**16** KV layers across its stages — 63 in total, one more than the target's depth — and the decode instance registers 63 per rank. Stage 3 owns 15 target layers and registers 16, which is the head landing after them rather than sorting in among them. Both instances therefore agree on the name, which is what NIXL matches regions by.

## Tests

`tests/native/patches/test_minimax_m2.py` drives all four stages sequentially in one process with stub layers that stamp their own global index into what they return, so a stage-local index shows up as a wrong number rather than as a shape that still happens to fit. Bands come from the real `get_pp_indices`, and the handoff placeholder goes in through `install_aux_handoff_slots` on an outer object that owns the copy, which is the shape the runner calls.

The aux sets it runs are parametrized, and one of them puts every index exactly on a PP4 band boundary. A boundary index is both one stage's last capture and the next stage's incoming hidden state, so an off-by-one in the split double-counts all three at once — and compiles, and passes a short probe. Widening the capture range by one fails exactly that case and leaves the other eleven green.

`tests/native/v1/spec_decode/test_eagle3_pp.py` covers the guard and the flag reader: a supported target passes, an unsupported one raises, an unsupported one with aux off passes, and no non-EAGLE3 method is gated. It also asserts every allowlisted architecture actually has a non-upstream `forward`, so the allowlist cannot claim a target the patch does not cover.

`tests/native/patches/test_llama_eagle3.py` pins the head's name to the target's full depth on every rank, keeps `target_layer_count` equal to it, and checks against `pipeline_adjusted_layer_index` that the named head lands after every target layer on each of the four ranks rather than sorting in among them.

Config objects and stubs only — no checkpoint, no device.

## Rebased onto dev

`#980` landed the MTP/nextn KV index rule while this was in progress. That rule covers the EAGLE3 head once the head is named past the target's full depth, so the index handling this branch previously carried in `patches/attention.py` is gone and `pipeline_adjusted_layer_index` is used as-is.

### DIFF
--- a/tests/native/patches/test_llama_eagle3.py
+++ b/tests/native/patches/test_llama_eagle3.py
@@ -1,0 +1,131 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""The EAGLE3 head is named past the target's full depth, on every rank.
+
+Config objects only -- no checkpoint, no device.
+"""
+
+from __future__ import annotations
+
+import pytest
+from vllm.model_executor.models.llama_eagle3 import LlamaModel as Eagle3LlamaModel
+
+import vllm_rbln.patches.llama_eagle3 as patch_module
+
+# The suite truncates models via VLLM_RBLN_NUM_HIDDEN_LAYERS and the patched
+# get_pp_indices honors it; the split this reasons about is the real 62-layer one.
+from vllm_rbln.patches.distributed_utils import (
+    original_get_pp_indices as get_pp_indices,
+)
+from vllm_rbln.v1.worker.utils import pipeline_adjusted_layer_index
+
+NUM_LAYERS = 62
+PP_SIZE = 4
+
+
+def _vllm_config(per_rank_layers: int):
+    """A config shaped like the one the head's __init__ actually reads.
+
+    `model_config` is the target's and `speculative_config.draft_model_config` the
+    draft's -- the two the upstream __init__ mixes, and the reason a per-rank count
+    can pass for the full depth.
+    """
+    from types import SimpleNamespace
+
+    return SimpleNamespace(
+        model_config=SimpleNamespace(
+            get_total_num_hidden_layers=lambda: NUM_LAYERS,
+            get_num_layers=lambda parallel_config: per_rank_layers,
+            hf_text_config=SimpleNamespace(num_hidden_layers=NUM_LAYERS),
+        ),
+        speculative_config=SimpleNamespace(
+            draft_model_config=SimpleNamespace(
+                hf_config=SimpleNamespace(num_hidden_layers=1),
+            ),
+        ),
+    )
+
+
+@pytest.fixture
+def captured(monkeypatch):
+    """Record what the patch hands upstream instead of running upstream."""
+    seen = {}
+
+    def stub_init(model, *, vllm_config, start_layer_id=0, prefix=""):
+        # Upstream builds the layers here, so anything the layers read has to be
+        # set on the config before this point.
+        seen["start_layer_id"] = start_layer_id
+        seen["target_layer_count"] = getattr(
+            vllm_config.speculative_config.draft_model_config.hf_config,
+            "target_layer_count",
+            None,
+        )
+
+    monkeypatch.setattr(patch_module, "_orig_model_init", stub_init)
+    return seen
+
+
+# 15 is stage 3's own layer count on the 62-layer PP4 split, which is what upstream
+# passes; 62 is the full depth. Only the first rank makes the two agree.
+@pytest.mark.parametrize("per_rank_layers", [15, 16, NUM_LAYERS])
+def test_the_head_is_named_past_the_full_depth(per_rank_layers, captured):
+    patch_module.patched_eagle3_llama_model_init(
+        object(), vllm_config=_vllm_config(per_rank_layers), prefix=""
+    )
+
+    assert captured["start_layer_id"] == NUM_LAYERS
+
+
+@pytest.mark.parametrize("per_rank_layers", [15, 16, NUM_LAYERS])
+def test_target_layer_count_stays_paired_with_the_name(per_rank_layers, captured):
+    # `llama.py` recovers a draft-relative index as `extract_layer_index(prefix) -
+    # target_layer_count`, i.e. off the name, not off the loop counter. Raising the
+    # name alone would give stage 3 `62 - 15 = 47` and trip the layer_types assert.
+    patch_module.patched_eagle3_llama_model_init(
+        object(), vllm_config=_vllm_config(per_rank_layers), prefix=""
+    )
+
+    assert captured["target_layer_count"] == captured["start_layer_id"]
+
+
+@pytest.mark.parametrize("rank", range(PP_SIZE))
+def test_the_named_head_lands_after_every_target_layer(rank):
+    # The point of the rename: RBLN's index rule routes a name at or past the full
+    # depth to the end of the rank's compacted KV list. Upstream's per-rank name
+    # would sort in among the target layers instead.
+    from types import SimpleNamespace
+
+    start, end = get_pp_indices(NUM_LAYERS, rank, PP_SIZE)
+    model_config = SimpleNamespace(
+        get_layers_start_end_indices=lambda parallel_config: (start, end),
+        get_total_num_hidden_layers=lambda: NUM_LAYERS,
+    )
+
+    head = pipeline_adjusted_layer_index(
+        f"model.layers.{NUM_LAYERS}.self_attn.attn", model_config, None, 1
+    )
+    targets = [
+        pipeline_adjusted_layer_index(
+            f"model.layers.{i}.self_attn.attn", model_config, None, 1
+        )
+        for i in range(start, end)
+    ]
+
+    assert targets == list(range(end - start))
+    assert head == end - start
+
+
+def test_the_patch_is_the_one_installed():
+    assert Eagle3LlamaModel.__init__ is patch_module.patched_eagle3_llama_model_init
--- a/tests/native/patches/test_llama_eagle3.py
+++ b/tests/native/patches/test_llama_eagle3.py
@@ -48,7 +48,6 @@ def _vllm_config(per_rank_layers: int):
         model_config=SimpleNamespace(
             get_total_num_hidden_layers=lambda: NUM_LAYERS,
             get_num_layers=lambda parallel_config: per_rank_layers,
-            hf_text_config=SimpleNamespace(num_hidden_layers=NUM_LAYERS),
         ),
         speculative_config=SimpleNamespace(
             draft_model_config=SimpleNamespace(
--- a/vllm_rbln/patches/llama_eagle3.py
+++ b/vllm_rbln/patches/llama_eagle3.py
@@ -14,7 +14,8 @@
 """Name the EAGLE3 head's layers past the target's full depth.
 
 Upstream names them past the count on *this* pipeline rank, which is only the
-total when PP=1. See the patch's own reason string for what that breaks.
+total when PP=1, and pairs that count with `target_layer_count`. Both move here.
+See the patch's own reason string for what that breaks.
 
 The draft->target vocab scatter is deliberately NOT patched here. Upstream builds
 that index inside ``compute_logits`` from two input-independent operands, so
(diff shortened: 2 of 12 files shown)


vllm_rbln/patches/llama_eagle3.py
```diff
@@ -0,0 +1,75 @@
...
+        "TODO(vllm-project/vllm#50514): delete once that lands and is released."
+    ),
+)
+def patched_eagle3_llama_model_init(
+    self, *, vllm_config: VllmConfig, start_layer_id: int = 0, prefix: str = ""
+) -> None:
+    _orig_model_init(
+        self,
+        vllm_config=vllm_config,
+        start_layer_id=vllm_config.model_config.hf_text_config.num_hidden_layers,
```
rebel-cjlee: Is there a reason to prefer `hf_text_config.num_hidden_layers`?

`pipeline_adjusted_layer_index` compares the head's index against `get_total_num_hidden_layers()`, and the two only agree for the base convertor. Same expression upstream vllm#50514 uses, so this becomes a pure revert later.

```suggestion
        start_layer_id=vllm_config.model_config.get_total_num_hidden_layers(),
```

vllm_rbln/patches/llama_eagle3.py
```diff
@@ -0,0 +1,75 @@
...
+        "Nobody binds the last entry, and KV registration fails with "
+        "`EnsureSyncedOnPhysicalView ... has no physical_view`.\n"
+        "\n"
+        "Naming past the full depth also makes the head's name identical on a "
+        "prefill and a decode instance, which is what NIXL matches KV regions by, "
+        "and keeps it clear of every target band for any split. The caller's "
+        "`target_layer_count` is left alone: it pairs with the `layer_idx` the "
+        "decoder layer is constructed with, which is the loop counter and not the "
+        "index the name carries, and the only reader (`llama.py:193`) sits behind a "
+        "`layer_types` walrus that this draft config leaves unset. "
```
rebel-cjlee: This says the `layer_idx` is the loop counter, but the one `llama.py:193` reads comes from `extract_layer_index(prefix)` at `llama.py:141`, so it carries `start_layer_id`. Upstream keeps `start_layer_id` and `target_layer_count` equal because both come from `target_layer_num` in `Eagle3LlamaForCausalLM.__init__`, so raising one of them here splits the pair. On the 62-layer PP4 split, stage 3 would give `(0 + 62) - 15 = 47` and trip the `assert effective_layer_idx< len(layer_types)` just below, for a one-layer draft.

Unreachable for this checkpoint since `layer_types` is unset, as you note. Still, assigning `target_layer_count` the same value before calling `_orig_model_init` is two lines and keeps the pair consistent for a draft config that does set it. It would have to be before the call: the layers that read it are built inside `LlamaModel.__init__`, and `self.config` there is the same object the caller assigns to. What do you think?

rebel-chanheo: Correct, and the reason string was wrong. Two layer_idx: the one the decoder layer is constructed with (llama_eagle3.py:173) is the loop counter; the one llama.py:193 subtracts from comes from extract_layer_index(prefix) (llama.py:141), i.e. the name — which is what this patch raises. Raising one of the pair does split it.

Set to the same value before _orig_model_init, since the layers that read it are built inside LlamaModel.__init__ off the same config object. Reason string now carries 62 - 15 = 47 and the assert, and states that unset layer_types is why the value must be set here rather than relied on.

PP4 on device: no layer_types / effective_layer_idx in logs across 4 runs, KV registration unchanged at 15/16/16/16.

vllm_rbln/patches/llama_eagle3.py
rebel-chanheo: Taken. The threshold is get_total_num_hidden_layers() (v1/worker/utils.py:95), which goes through model_arch_config; the two expressions agree only for the base convertor (MTP convertors and MossAudioModelArchConfigConvertor override it, models/config.py:537 assigns it). Naming the head against one and testing it against the other is the defect this patch exists to avoid.

Applied as a variable rather than inline, since target_layer_count now reuses the value:
target_depth = vllm_config.model_config.get_total_num_hidden_layers()

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-cjlee: This is overwritten at line 306 by `eagle3_aux_hidden_states_enabled(...)`, and nothing reads it in between.

```suggestion
```

tests/native/patches/test_llama_eagle3.py
```diff
@@ -0,0 +1,131 @@
...
+    can pass for the full depth.
+    """
+    from types import SimpleNamespace
+
+    return SimpleNamespace(
+        model_config=SimpleNamespace(
+            get_total_num_hidden_layers=lambda: NUM_LAYERS,
+            get_num_layers=lambda parallel_config: per_rank_layers,
+            hf_text_config=SimpleNamespace(num_hidden_layers=NUM_LAYERS),
+        ),
```
rebel-cjlee, line 51: Nothing reads `hf_test_config` any more, here or in the patch. Dropping it keeps the fake to what the code under test uses, and a revert to `hf_text_config.num_hidden_layers` then fails instead of passing quietly.

```suggestion
        model_config=SimpleNamespace(
            get_total_num_hidden_layers=lambda: NUM_LAYERS,
            get_num_layers=lambda parallel_config: per_rank_layers,
        ),
```

rebel-cjlee (approved):

vllm_rbln/patches/llama_eagle3.py
```diff
@@ -0,0 +1,90 @@
...
+        "`llama.py:193` subtracts it from `extract_layer_index(prefix)` -- the index "
+        "the *name* carries, not the loop counter the decoder layer is constructed "
+        "with. Raising one alone would make stage 3 of the 62-layer PP4 split "
+        "compute `62 - 15 = 47` and trip the `effective_layer_idx < len(layer_types)` "
+        "assert just below, for a one-layer draft. Unreachable while this draft "
+        "config leaves `layer_types` unset, which is why it has to be set here "
+        "rather than relied on. It has to be set before the original runs: the "
+        "layers that read it are built inside `LlamaModel.__init__`, off the same "
+        "config object.\n"
+        "TODO(vllm-project/vllm#50514): delete once that lands and is released."
```
rebel-cjlee, line 73: Is there a way to explicitly raise an error when the upstream PR is merged? This would remind us without relying on our memory.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-chanheo: Applied in e2d4ebb9. Only the self.drafter annotation sits between the two assignments, and both readers run later.

tests/native/patches/test_llama_eagle3.py
rebel-chanheo: Applied in fd7f3053. Confirmed: reverting the patch to hf_text_config.num_hidden_layers now fails with AttributeError instead of passing.

vllm_rbln/patches/llama_eagle3.py
rebel-chanheo: multi_connector.py:36 is that mechanism, but it works because #46865 was merged and the removal release was known. #50514 has been open since 2026-07-31 with no milestone and is in none of 0.25–0.28, so any bound I pick now fires falsely first — and the response to that is to raise the number, which is the reflex we're trying to avoid.

Meanwhile grep "TODO(vllm" lists all five sites, and a pin bump has to re-verify these anyway. I'll add the multi_connector assert once #50514 merges and the release is known; filing an issue.

Checked while looking: across 0.24 → 0.28 the four bodies these patches wrap or replace are byte-identical and all three defects remain, despite the files changing by 57/153/456 lines.

PR #1009, opened 2026-08-27
fix(scheduler): skip the sub-block match on a preemption resume with a connector

rebel-jinhwan:
## Problem

`VLLM_RBLN_SUB_BLOCK_CACHE` + an LMCache connector: under KV pressure, a preempted request's resume kills EngineCore with

```
AssertionError: Preempted request ... has num_computed_tokens 48640 but expected 48128
```

Root cause (full analysis in rebellions-sw/lmcache-rbln#142): the scheduler queries `get_num_new_matched_tokens(request, local)` **before** the sub-block match. The connector records that `local` as the request's resume point (LMCache's `LoadSpec.vllm_cached_tokens`). When the sub-block match then wins the arbitration, the scheduler resumes at `local + sub_block`, but the connector's record stays at `local`. LMCache's preemption-resume assertion (`vllm_v1_adapter.py`) then fires and takes down all DP ranks in lockstep.

## Fix

Skip the sub-block match for a request resuming from preemption while a KV connector is configured. The scheduler then commits `local + external`, which is exactly what the connector recorded, for every case of the assertion (`hit <= local`, `hit > local`, full hit with the last-token adjustment).

Why this shape:

- LMCache's assertion only fires on the resume path, and the `LoadSpec` it checks is the one written by the resume-pass lookup, so the initial-pass sub-block match is harmless and stays.
- The connector is queried exactly once. An earlier revision re-queried `get_num_new_matched_tokens(request, local + sub_block)`; that is outside the KV connector contract (a second call is only allowed after `None`), and the LMCache MP connector accumulates `num_stored_tokens` on every call, so the re-query would silently drop STORE ops there (thanks @rebel-cjlee).
- The MP connector already returns 0 for `PREEMPTED` requests, so "block-granular resume" is what LMCache assumes anyway.

Cost: recomputing under one block of tokens on a preemption resume.

## Known limits (pre-existing, not introduced here)

The same assertion also misfires, independent of sub-block caching, when `min_retrieve_tokens > 0` skips a load (`LoadSpec.lmcache_cached_tokens` is still recorded as the full hit) or when `max_tokens_per_load` caps the offer at a chunk boundary that `local` does not sit on. Both need the LMCache-side assertion to trust `request.num_computed_tokens`; that is a separate upstream change.

## Test

`tests/native/v1/core/test_sub_block_prefix_caching.py::TestSubBlockVersusKVConnector::test_resume_from_preemption_skips_the_sub_block_match` — a resuming request that would otherwise take a sub-block match emits no copy op and schedules its whole prompt. Fails without the change.

Fixes rebellions-sw/lmcache-rbln#142


vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -576,8 +576,23 @@ def schedule(self, throttle_prefills: bool = False) -> RBLNSchedulerOutput:
...
+                        # and asserts on preemption resume. The lookup is cached,
+                        # so this costs no extra remote traffic.
+                        self.connector.get_num_new_matched_tokens(
+                            request,
+                            num_new_local_computed_tokens + num_sub_block_tokens,
+                        )
                     if num_sub_block_tokens > 0 and num_external_computed_tokens > 0:
-                        # Cancel the KV connector match in favor of the sub-block match
+                        # Cancel the KV connector match in favor of the sub-block
+                        # match.
```
rebel-cjlee: Looks like unnecessary diff

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -576,8 +576,23 @@ def schedule(self, throttle_prefills: bool = False) -> RBLNSchedulerOutput:
...
+                        # (external == 0 also covers hit <= local).
+                        # update_state_after_alloc(..., 0) can't tell "lost the
+                        # arbitration" from "no hit", so re-query with the
+                        # corrected count or the recorded resume point goes stale
+                        # and asserts on preemption resume. The lookup is cached,
+                        # so this costs no extra remote traffic.
+                        self.connector.get_num_new_matched_tokens(
+                            request,
+                            num_new_local_computed_tokens + num_sub_block_tokens,
+                        )
```
rebel-cjlee: The re-query means `get_num_new_matched_tokens` is now called twice for the same request in one pass. Is that safe for every connector we run?

The one I'd check first is LMCache's MP connector. It initializes a counter inside that hook, but with `+=`:
```python
tracker.increase_num_stored_tokens(ret)  # lmcache_mp_connector:951 (lmcache 0.5.1)
    self.nuom_stored_tokens += n                 # lmcache_mp_connector:250
```

That value means "the first N tokens are already in LMCache, don't store them again". Calling the hook twice looks like it would double it, and `GetStoreMetadata` then computes `min_available_tokens - num_stored_tokens` (line 358), which is bounded by the request length. So, the request might never produce a STORE op at all. Silently, since it just returns `None`.

It also looks reachable rather than a corner  case: `ret` is only chunk-aligned (line 948), so `external` can be smaller than a block and lose the arbitration to a sub-block match.

I may be misreading it, so could you confirm?

vllm_rbln/v1/core/rbln_scheduler.py
rebel-jinhwan: https://github.com/RBLN-SW/vllm-rbln/pull/1009/commits/859ba233b0121b67dcf5d3eef65fe61b4e864661

vllm_rbln/v1/core/rbln_scheduler.py
rebel-jinhwan: Good point — I had only considered LMCache's in-process mode (the legacy connector) and hadn't verified the MP connector. You're right that the re-query double-counts num_stored_tokens there, and it's outside the get_num_new_matched_tokens contract anyway.

I've reworked the PR: instead of re-querying, the scheduler now skips the sub-block match for a request resuming from preemption while a connector is configured. The connector is queried exactly once, and the scheduler commits local + external, which is what the connector recorded. PTAL, thanks!

https://github.com/RBLN-SW/vllm-rbln/pull/1009/commits/5d59d51f1e54908290899eeb4d3455c47bda3d2c

vllm_rbln/v1/core/rbln_scheduler.py
```diff
@@ -572,11 +572,24 @@ def schedule(self, throttle_prefills: bool = False) -> RBLNSchedulerOutput:
...
-                    sub_block_match, num_sub_block_tokens = self._try_sub_block_match(
-                        request,
-                        num_new_local_computed_tokens,
-                        num_external_computed_tokens,
+                    # Not on a preemption resume with a connector: the connector
+                    # was queried above with the block-aligned local count and
+                    # recorded it as the resume point (LMCache asserts on it),
+                    # and the KV connector contract allows no second query to
+                    # move it. Recomputing under one block is cheaper than
+                    # tripping that assertion.
```
rebel-cjlee: Could you compress this to a line or two? The constraint is worth documenting, but the resume point and the "no second query" rule are really one fact.

rebel-cjlee (approved):

PR #1011, opened 2026-08-27
feature(nixl): transfer KV across asymmetric TP and PP, and keep MLA correct

rebel-ykchoi:
#### Summary

Upstream's NIXL connector pairs KV regions by position — local region `i` against
remote region `i`, with a scalar `rank_offset` absorbing the difference. That holds
only when both sides cut the cache the same way. On RBLN a region is one chiplet
area, so region `i` names different KV heads as soon as the two sides' tensor-parallel
degrees differ, and a pipelined peer publishes only the layers of its own stage.
Positional pairing then moves the wrong bytes **with every length check still
passing** — the transfer completes, the descriptors line up, and the decode reads
another rank's heads.

This PR pairs by content instead, on two axes that compose: which KV heads an area
holds, and which of our layers a peer's stage covers. Both are derived from what the
peer advertises in the handshake, so any combination of prefill/decode parallel
degrees works within the bounds the worker's class docstring lists.

Two latent defects surfaced while building it and are fixed here: sliding-window
guards that keyed on a feature flag instead of the model, and per-shard handshake
state that accumulated across a retry.

##### A. Pairing by content

**1. The head axis.** A shard owns `total_kv_heads // tp_size` heads, which the
compiler cuts into one piece per chiplet area — except that a shard owning fewer
heads than the device has chiplets gets replicas instead, with the replication axis
innermost. Remote descriptors are emitted in **local** region order, each addressed
into whichever peer slice covers our head range, so `_compute_desc_ids`' positional
pairing keeps holding downstream. A peer with more TP ranks (fan-in) contributes
only the areas whose heads that rank owns; a peer cut finer than us has our region
read in as many pieces as it spreads those heads over.

**2. The layer axis.** A pipeline stage is matched to our layers by **name**, not by
position, so uneven splits and either-side pipelines pair correctly. A stage holding
more layers than our band is trimmed — every metadata field describing layers moves
together, or regions-per-layer comes out wrong for anything dividing one by the other.

**3. Per-peer transfer state.** A peer serving part of what a whole-engine handle
covers gets its own local descriptor list, since one handle would describe every
peer's band at once and NIXL rejects the pairing on length. Host staging has no
per-area list to narrow, so it borrows upstream's own head-band split, and the
borrowed handle is shared — cleanup releases the entry, not the handle.

**4. Metadata.** Neither axis is carried by upstream's agent metadata, so a subclass
carries the registered layer names and the chiplet geometry. A private schema version
folded into the compatibility hash keeps peers speaking different schemas apart;
an upstream consumer ignores the extra fields.

##### B. MLA

MLA is the head axis degenerated: the latent is replicated across ranks, so there is
no band to match and positional pairing is already correct. What it needs is to be
kept **out** of head-band matching, whose bands come from a configured KV-head count
the latent does not have — bands computed there would be plausible and wrong. The
two sides' chiplet geometry is required to agree, since each derives the block stride
from its own device buffers.

Host staging needs the same distinction one level down: its block copy splits K/V
before indexing blocks, and an MLA cache has no K/V level to split. The copy moves
out of the model runner so that branch is reachable from a test, and the MLA flag
rather than the rank decides it — SSM/conv and cross-layer pools are 3D as well.

##### C. Two fixes

**5. Sliding-window guards keyed on the model, not the flag.** The guards said a
sliding window bars model parallelism across P/D, but keyed on the view-optimization
env flag, so the same model passed with the flag off. They key on the window itself
now, and the head axis gets the refusal the layer axis already had.

**6. Per-shard state published once.** Upstream clears a remote engine only after a
handshake it finished, so one that raises partway leaves whatever it already wrote.
The shard list is now built locally and assigned at the end: a retry starts from
empty instead of appending its shards a second time and reading every block twice.

##### D. Bounds enforced early rather than left to fail late

A prefill resumes on a chunk boundary, so a fetch from a peer is trimmed to one
rather than reported unrounded. Regions per layer are checked on **every** D2D peer,
pipelined or not — that figure is geometry (K/V times the chiplet count) and depends
on no parallel size, so a mismatch means the two sides cannot be paired at all.
Upstream would notice only at transfer time, as a descriptor-count assert.

##### E. Module split

`worker.py` and `scheduler.py` each held both what the transfer direction does not
decide and the pull direction itself. They are split the way upstream's `v1/nixl/`
package splits its own, which also stops our `worker.py` from being read as the
compatibility shim upstream ships under that name. **Commit 2 is a move only** —
concatenating each pair reproduces the file it came from.

#### Testing

- **Unit** — head-band matching, replicas and fan-in area partition, the layer
  overlap and stage trimming, the per-shard registration and descriptor ids, and the
  handshake fan-out (`test_rbln_nixl_handshake.py`); the read path and its shard
  selection (`test_rbln_nixl_pull_worker.py`); registration, guards and the host
  buffer (`test_rbln_nixl_base_worker.py`); the chunk-boundary fetch and the save
  path (`test_rbln_nixl_base_scheduler.py`); the metadata extension and its
  compatibility hash (`test_rbln_nixl_metadata.py`); the host-staging block copy and
  the flag that reaches it (`tests/native/v1/worker/test_utils.py`,
  `test_rbln_model_runner_kv_cache.py`).
- Each refusal listed in the worker's class docstring has a case, since nothing else
  would notice if one stopped firing. Several cases enter through the upstream method
  that calls our override rather than calling it directly — an override reached only
  by name goes silently dead when upstream re-routes the call.
- `pytest tests/native` and `tests/torch_compile/unit` green; `pre-commit` clean on
  every changed file.

#### Risk / compatibility

- Equal TP with neither side pipelined takes upstream's path unchanged — the
  head-matching routing predicate is false there, so none of the new code runs.
- Host staging never head-matches; the per-area narrowing is D2D only.
- The module split changes no behaviour.

---

* Resolves rebellions-sw/fsw-inference#466 — TP support and the MiniMax
  prefill=PP / decode=TP·EP pairing this PR implements.
* Related to rebellions-sw/fsw-inference#492 — MLA is kept out of head-band
  matching here and its chiplet geometry is checked
* Related to https://github.com/rebellions-sw/nixl-rbln/pull/28

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

2. `uv run --no-sync pytest tests/native/v1/worker/test_utils.py tests/native/v1/worker/test_rbln_model_runner_kv_cache.py`
3. Expected: all green. Every topology bound the worker's class docstring lists has a
   case that fails if the guard stops firing.
4. Edge cases covered: a peer that replicates a head across areas; a peer cut finer
   than this rank (fan-in) and one cut coarser; a stage holding more layers than this
   rank's band; a repeated handshake; a handshake that raises partway.

---

### 💬 Notes


vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
```diff
@@ -0,0 +1,1855 @@
...
+        axis innermost (``slice_id = area // (areas // slices)``).
+        """
+        assert total_kv_heads % tp_size == 0, (
+            f"KV heads {total_kv_heads} not divisible by TP size {tp_size}"
+        )
+        heads_per_rank = total_kv_heads // tp_size
+        assert slices > 0 and heads_per_rank % slices == 0, (
+            f"{heads_per_rank} heads not divisible into {slices} slices"
+        )
+        assert areas % slices == 0, (
```
rebel-jongwonlee: ### `_slice_head_bounds` validates peer-supplied values with `assert`

The three guards here are `assert`s (L635, L639, L642). But this helper is called
with **values that arrived over the handshake**, not only with our own config —
three of the seven call sites pass the peer's:

| call site | function | peer-supplied args |
|---|---|---|
| L676 | `_build_head_matched_remote` | `remote_tp_size`, `nixl_agent_meta.kv_areas` / `.kv_slices` |
| L855 | `_peer_head_split` | same |
| L1747 | **`_validate_head_matched_handshake`** | same |

The last one is the handshake-validation path itself — the function whose job is to
reject a peer we cannot pair with is asserting on that peer's numbers.

**Why `raise` may fit better here**

1. **It crosses a process boundary.** `assert` reads as an internal invariant
   ("cannot happen if the code is right"); these are external inputs that *can* be
   wrong because the other engine was configured differently.

2. **It is a documented user-facing refusal.** The class docstring already lists it:
   *"each side's TP degree must divide the model's KV heads; below that upstream
   replicates one head across ranks, which no head band names."* A model with 8 KV
   heads brought up at TP=16 is accepted by vLLM (it replicates heads) and lands
   here as `AssertionError: KV heads 8 not divisible by TP size 16`, which does not
   say which engine is misconfigured or what to change.

3. **It differs from the neighbouring code.** `_head_split` (L865) rejects the same
   class of failure with a `RuntimeError` naming both sides: *"this rank's slice
   spans {per_slice_l} KV heads and the peer's {per_slice_r}, which does not divide
   it; the two sides must cut heads at commensurate granularities."* The other 26
   refusals in this file are `raise` as well, so these three look like the only
   exceptions — I could not tell whether that is deliberate.

**Secondary, not my main argument:** under `-O` the guards disappear,
`heads_per_rank` floors to 0, `per_slice_l` becomes 0, and
`head = base_l + (area_l // replicas_l) * 0` is 0 for every rank and every area, so
every region resolves to the peer's region 0 with matching lengths and descriptor
counts — precisely the "wrong bytes, every length check still passing" failure this
PR sets out to eliminate. I have no evidence anyone runs this with `-O`, so I'd
treat it as a robustness point rather than a live bug.

**Suggestion:** promote at least the `total_kv_heads % tp_size` check to a
`RuntimeError` naming the local and remote TP sizes, in the style of `_head_split`'s
message. The other two could stay as `assert` if they are genuinely internal
invariants — though `slices` also arrives from `nixl_rbln` at runtime rather than
from this code, so it may deserve the same treatment.

rebel-ykchoi: Agreed on all three, and they are now `RuntimeError`s.

**Why all three rather than just the first.** You flagged `slices` as secondary,
but it decided the question: it does not come from this code either — it arrives
from `nixl_rbln`'s registration at runtime, and the peer's copy arrives over the
handshake. Promoting only the TP check would have left the next reader asking why
the three split, so the helper now has no `assert` left. Its inputs are all
external.

**Naming both sides.** `_head_split` can name both because it receives both;
`_slice_head_bounds` sees one side's numbers at a time and cannot tell whose they
are. Callers now pass a `side` label, so the message says which engine to change:

```
RBLN NIXL: the peer tensor-parallel size 16 does not divide the model's 8 KV
heads; upstream then replicates one head across ranks and a head band would be
a fraction of a head, which no descriptor names.
```

Scope tag is `RBLN NIXL:` rather than `D2D:` — `_validate_head_matched_handshake`
reaches this helper on host staging too, through a pipelined peer.

**On the `-O` point.** I traced it and it does not hold, so I did not use it as
justification. With the asserts stripped, a degenerate geometry does not reach the
descriptor build: `_validate_head_matched_handshake` compares `local_len *
per_slice_r` against `remote_len * per_slice_l`, and a zero `per_slice` on either
side makes that comparison fail — then divides by zero formatting the message. So
the failure surfaces, just as a confusing error rather than a clear one. Your main
argument stands on its own without it.

**Tests.** Each refusal has a case, including a peer that advertises zero slices
(the guard is what keeps the modulo beside it from dividing by zero). Two more go
through `_validate_head_matched_handshake` and assert the message names the right
side — one for a bad peer geometry, one for a bad local one — since a call site
wearing the wrong label would send an operator to the wrong engine.

rebel-jongwonlee: Confirmed on the branch: no `assert` left in the helper, all seven call sites pass `side`, and the peer-labelled ones do carry `remote_tp_size` / `nixl_agent_meta.kv_areas` / `.kv_slices`. The rendered message matches your example. Nothing blocking.

Two things I did not ask for and think are better than what I suggested:

- **`slices <= 0` in the second guard.** A peer that advertises `slices=0` would have hit `ZeroDivisionError` on the modulo beside it, and that value arrives over the handshake like the rest. The new parametrised case covers it.
- **Promoting all three.** I had left the last two as "assert if genuinely internal", and the reason you give — `slices` comes from `nixl_rbln`'s runtime registration, not from this code — settles it. That is a correction to my read, not a concession to it.

**One optional follow-up, not blocking.** `side: str` accepts any string, and `side` feeds only the message, so a typo (`side="peeer"`) or a synonym (`side="remote"`) would raise the right refusal with the wrong label, pass the new tests — they pass `side` directly rather than exercising a call site — and surface only as log text. Since the message exists to tell an operator which engine to change, a wrong label sends them to the other one.

```python
from typing import Literal

def _slice_head_bounds(..., *, side: Literal["local", "peer"]) -> tuple[int, int]:
```

mypy then rejects it at the call site:

```
error: Argument "side" to "_slice_head_bounds" has incompatible type
  "Literal['peeer']"; expected "Literal['local', 'peer']"  [arg-type]
```

I checked this repo runs mypy over `vllm_rbln/` (`tools/pre_commit/mypy.py` `FILES`, plus the `mypy-3.10/3.11/3.12` hooks), so it would be enforced rather than decorative. Worth it mainly because the call sites will grow — today's seven are all correct.

rebel-ykchoi: Taken — `side: Literal["local", "peer"]` in `b245cccc`.

I ran your suggestion as a pair rather than taking it on the argument, because
"the type catches it" only means something next to "nothing else does":

| mutation | mypy |
|---|---|
| `Literal` in place, `side="peeer"` at a call site | `error: ... incompatible type "Literal['peeer']"; expected "Literal['local', 'peer']"  [arg-type]` |
| `Literal` in place, `side="remote"` | same, on the synonym |
| **`Literal` reverted to `str`, same typo** | **passes** |

The third row is the one that made it worth doing. I also confirmed the
enforcement you cited rather than assuming it: `tools/pre_commit/mypy.py` has
`FILES = ["vllm_rbln/"]`, and the `mypy-local` / `mypy-3.10|3.11|3.12` hooks run
it, so this is checked rather than documented.

Your read of the test gap is right, and narrower than I had it. The parametrised
cases pass `side` directly; the two that go through a real call site
(`_validate_head_matched_handshake`) cover 2 of the 7. I had been closing that by
adding a case per call site — one for the peer label, then one for the local
label a round later. The type closes the remaining five at once, which is the
better shape: the list of call sites will grow, and a test per call site would
not.

One thing the type does not reach, for the record: the parametrised test data
itself. `side` there is inferred as `Any`, and the assertion is
`side in str(e.value)` against a message that interpolates whatever it is given,
so a typo in the parametrize list would pass. That is test data rather than a
call site, so I left it; noting it in case it matters to you.

rebel-jongwonlee (approved):
Verified on the branch. `5b1ada46` is the two-line signature change; the seven call
sites already passed literals from `c2fc28e8`, so what this commit adds is mypy
holding them there. CI is green including the pre-commit mypy hooks, and the two
label tests (`test_a_peer_geometry_is_reported_as_the_peer_s` /
`test_our_own_geometry_is_reported_as_ours`) carry the negative assertion, which is
what actually pins a call site rather than the helper. Nothing blocking — approving.

**On the `-O` trace.** You're right and I was wrong about the ordinary shape of it:
with one side degenerate the comparison does fail, and the message then divides by
zero, so it surfaces confusingly rather than silently. Thanks for tracing it instead
of taking my word.

One correction back, since you phrased it as "either side": it doesn't hold when
*both* are zero. `local_len * 0 != remote_len * 0` is `0 != 0`, so the check passes.
8 KV heads with local TP16 -> peer TP32 gets there — `tp_ratio != 1`, so
`_is_head_matched_peer` is true — and then `_head_split(0, 0)` returns 1 via
`per_slice_l <= per_slice_r`, and `head = base_l + (area_l // replicas_l) * 0` is 0
for every rank and area. So under `-O` that one shape still reaches the descriptor
build with every region pointing at the peer's region 0.

I'm not asking for anything here — it's `-O`-only, it's the secondary argument I'd
already set aside, and the `RuntimeError`s make it unreachable in any normal run.
Recording it only because you went and checked, and the "either side" version of the
claim is the one that would get quoted later.

**On the parametrize gap.** Agreed, and I'd go further: it can't be closed the way
the call sites were. `tools/pre_commit/mypy.py` has `FILES = ["vllm_rbln/"]`, so
`tests/` is outside mypy's scope entirely — annotating the test parameter, or
hoisting the cases into a typed module-level list, would both be decorative there.
Leaving it is the right call, not just an acceptable one.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
rebel-ykchoi: You're right, and "either side" was the wrong quantifier. I ran your counterexample:

```
8 KV heads, local TP16 -> peer TP32, asserts stripped
  heads_per_rank      local 0  / peer 0
  per_slice           local 0  / peer 0
  local_len*per_slice_r != remote_len*per_slice_l   ->  0 != 0  ->  False   (no raise)
  _head_split(0, 0)   ->  per_slice_l <= per_slice_r  ->  returns 1   (no ZeroDivisionError)
```

So the byte invariant catches it only when *exactly one* side is degenerate. With
both at zero it passes, `_head_split` short-circuits before the modulo, and every
region resolves to the peer's region 0 — the shape reaching the descriptor build
under `-O` after all. My correction was right about the asymmetric case and wrong
as stated.

Correcting it here rather than editing the earlier comment, so the thread keeps
what was claimed and what it turned into.

Nothing to change in the code: the three `RuntimeError`s make all of it
unreachable in a normal run, which is the argument the commit stands on.

PR #1014, opened 2026-08-27
fix(spec-decode): pad the sampler batch axis to avoid mid-decode recompile

rebel-jonghewk:
### 🚀 Summary of Changes

The sampler graphs are `dynamic=False` and sized from the request batch, but the batch reached them at the per-step `num_reqs`. The rejection-sample op sizes its buffers from `len(num_draft_tokens)`, and `bonus_logits = logits[bonus_logits_indices]` carries one row per request — so as requests finish and the batch shrinks it presents a row count warm-up never compiled, and Dynamo retraces **on the serving path**, inside the decode loop. `compilation/backends.py` already flags this by logging `stage=runtime` at warning level.

The model does not have the problem: the runner pads its batch to a decode bucket, so the forward always runs one shape. Only the sampler side was left open. The spec-length axis was closed the same way earlier, by padding the rejection-sample op to a config-fixed `num_speculative_tokens`; this closes the batch axis to match.

**Padded at the sampler boundary, not at the producer.** `SpecDecodeMetadata` is upstream's type and its documented per-field shape is `[batch_size]` for the live batch. Its other readers are sized by that same live count:

- `eagle_prepare_inputs_padded` differences `cu_num_draft_tokens` against `valid_sampled_tokens_count` and `query_start_loc[1:]`, both `num_reqs`-sized.
- `_preprocess` forwards `logits_indices` as `token_indices` into the compiled model on the prefill branch, which is reachable because `_calc_spec_decode_metadata` is gated on `use_spec_decode` alone.
- `kv_sharing_fast_prefill` takes `logits_indices.size(0)` as `num_logits_indices`.

Padding in the producer would have redefined the contract for all three. `_sample` already receives the metadata and already owns the pad-in/slice-out pattern in its sibling branch, so this pads a copy there; `logits_indices` then needs no padding at all, because by that point the gather has already run. Both branches now end in a single `_depad_sampler_output`.

**Why padding rather than warm-up.** Padding keeps the compiled graph count at one, so it costs no extra device memory. Compiling each reachable batch would add a graph per shape — the same constraint that currently blocks the equivalent fix for the drafter.

That makes `_warmup_sampler_decode_batches`' sweep dead: with the axis pinned, the smaller sizes it walked cannot occur again, and leaving them would keep graphs resident for nothing. It is collapsed to the one size that will be seen, and both sites now name the same bound, `bucketing_manager.max_batch_size`. `decode_batch_buckets[-1]` is the same number — all three managers end there, and `ManualBucketingManager` raises when the last manual bucket is not `max_batch_size` — but it reaches the bound through the bucket list, so the size the pad pins to and the size warm-up compiles now read as one thing.

Model path: **vLLM-native**.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

```
uv run --no-sync pytest tests/native/v1/worker/test_rbln_model_runner.py -q \
  -k "TestPadDepad or TestSamplePadding or rejection_sampler_warmup"
```

Six host-side tests, no device. They cover the padded request axis at `num_reqs < max_batch_size`, that the torch rejection-sampler branch is left on the live batch, that `sum(num_draft_tokens)` / `max_spec_len` / `target_logits_indices` / `logits_indices` are untouched (the packed layout the op requires), that the depad slices logprobs by `num_reqs * output_width` rather than `num_reqs`, that warm-up issues one rejection-sample call at the bound instead of sweeping, and — the one that would have caught the producer-side version — that a padded metadata reaching `eagle_prepare_inputs_padded` raises instead of silently mis-shaping. Against a dev runner every one of them fails except the torch-branch test, which covers the path this change deliberately leaves alone.

End to end, on a spec-decode run whose batch shrinks mid-run (variable-length prompts, batch > 1, ngram is enough):

1. `grep -c "stage=runtime" <log>` — non-zero before, with `caller=…rbln_sampler.py(greedy_sample)` and `graph inputs=l_logits_:bf16[N,vocab]` for several `N` below the batch size; 0 after.
2. With `VLLM_RBLN_METRICS=True`, `E2E (DECODE)` mean and p99 drop while the median is unchanged — this removes a tail, not per-step cost. Acceptance rate unchanged.

---

### 📋 Checklist

---

### 💬 Notes

Reviewer decisions worth confirming:

- **`_pad_rows` reused rather than a new helper.** For a 1-D tensor `t.shape[1:]` is empty, so its `expand` is the same expression a delta-based helper would need, and it is already a no-op at the bucket. Its `n > bucket` truncation branch is unreachable here — `input_batch.num_reqs` is scheduler-bounded by `max_num_seqs // pp`, which is exactly the bucket the pad targets.
- **`dataclasses.replace` on `SpecDecodeMetadata`** re-runs `__post_init__`, which recomputes `max_spec_len` from the padded `num_draft_tokens`; appending zeros leaves it unchanged. Asserted in the tests rather than left implicit.
- **`_pad_rows` / `_pad_sampling_metadata` / `_depad_sampler_output` / `_pad_spec_decode_metadata`** now form a ~70-line cluster with no runner state whose whole subject is sampler input shapes. It reads like it wants its own module under `v1/sample/`, but that is a separate change and this file is already 3.5k lines — raising it rather than folding it in.

Still draft: the issue link is missing, and correctness is verified only at unit level. No golden-output comparison has been run, so nothing yet proves the padded rows cannot influence the sampled tokens; matching acceptance rate is suggestive, not sufficient.


tests/native/v1/worker/test_rbln_model_runner.py
```diff
@@ -209,6 +227,110 @@ def test_depad_sampler_output_trims_to_num_reqs(self):
...
+        ),
+        max_num_reqs=8,
+        rejection_sampler=SimpleNamespace(
+            impl=SimpleNamespace(rejection_sample=rejection_sample)
+        ),
+    )
+
+    runner._warmup_sampler_decode_batches()
+
+    assert len(rejection_sample.call_args.args[1]) == 4
```
rebel-cjlee, line 346: This assert passes on dev too. `call_args` is the last call, and the old loop's last iteration already ran at `batch_size == 4`, so `args[1]` was already length 4. I ran the five new tests against a dev runner and three fail, this one passes. What changed is the call count, so add `assert rejection_sampler.call_count == 1` above it.

```suggestion
    assert rejection_sampler.call_count == 1
    assert len(rejection_sample.call_args.args[1]) == 4
```

rebel-kblee (approved):
it looks good to me 👍

rebel-cjlee (approved):
The approach looks right to me and I left a few inline notes.

1. The test command points at `test_sampler_request_axis_pad.py`, which does not exist.
2. The fifth test you describe, padded metadata reaching `eagle_prepare_inputs_padded`, is not in the diff. It is the only evidence you give for padding at the boundary rather than the producer, so I would add it or drop the sentence. I checked and it really does raise.
3. `decode_batch_buckets[-1], which a manual bucketing config need not carry up to it` is not true. All three managers always end at `max_batch_size`, `ManualBucketingManager` raises otherwise. Also the code names `bucketing_manager.max_batch_size`, not `max_num_reqs`.
4. In the Notes, `scheduler-bounded by max_num_seqs` should be `max_num_seqs // pp.

The code is fine for 3 and 4, only the wording is off.

PR #1018, opened 2026-08-28
feature(nixl): transfer KV in the push direction

rebel-ykchoi:
#### Summary

The connector served only pull transfers, where the consumer reads a finished
request's KV out of the producer. Upstream offers push as a sibling, where the
producer writes into the consumer, and that is the direction it is extending.

Pairing regions describes what two peers hold, so the handshake, the descriptor
construction and the combinations a handshake refuses are already direction-free.
What a direction owns is issuing the transfer and deciding when a request is done.
Both fit in a module beside the read path's, and the shared class keeps everything
neither direction decides.

Writing turns two of the read path's assumptions around, and adds a pairing that
has to be refused. It also reaches two request states the read path never sees.

##### A. The direction split

`push_worker.py` and `push_scheduler.py` sit beside `pull_worker.py` and
`pull_scheduler.py`; the base worker and base scheduler keep what the direction does
not decide. The connector is selected by name, so a deployment picks a direction the
way it already picks a buffer device.

##### B. Two assumptions the write direction inverts

**1. A replica is a valid source, not a valid sole destination.** Reading from a
chiplet replica is safe because any copy answers with the same bytes. Writing one is
not: the peer's other chiplets keep serving what was there before, so a slice's
descriptors fan out over **every** copy of it. The fan-out is decided from the peer's
advertised geometry, the same input the read path uses to pick one copy.

**2. A consumer settles on the first notification, not the last.** Under pull, a
request is done when every peer has been heard from. Under push, the consumer is
told by whichever producer rank finishes, and the count that notification carries
means the opposite of what it means on the read path — so it is read the other way
round rather than reused.

##### C. Directions must refuse each other at the handshake

Two peers running opposite directions move the same bytes by protocols that do not
meet, yet they describe identical geometry to one another: every length, every band,
every regions-per-layer check passes. This vLLM hashes nothing that separates them —
the connector name is not a factor in the compatibility hash — so a mismatched pair
would proceed and stall rather than refuse. The direction now joins the schema
version in the RBLN extension of that hash, so the pair is rejected at the handshake
where it is still legible.

##### D. Two states the read path never reaches

**3. A request the serving layer turned away.** A prompt past the context length or
a client that went away still reaches the connector: the engine submits it
pre-aborted so the hooks run and the producer releases the blocks it pinned. Upstream
registers an empty receive for it and then reads `remote_block_ids`, a field the
write path fills in the one call a rejected request never reaches. It is now filled
where upstream is about to need it. Without that, the consumer's engine core dies and
takes every other request on that instance with it.

**4. A consumer under a partial prefix-cache hit.** It registers fewer blocks than
the producer holds, because its cache covered the head of the prompt and it needs
only the suffix. Where a whole-engine handle describes the peer, upstream aligns the
two lists by keeping the head of the longer one — which lands the producer's head in
the slots the consumer kept for its tail. Lengths agree, no assert fires, and the
consumer decodes against KV belonging to other tokens. The per-shard route already
trimmed to the suffix; the delegated one now trims before handing the request over,
so upstream finds the lists equal and leaves them alone. Upstream trims its own write
path from 0.27.2, which the code names as the condition for dropping this.

#### Testing

- **Unit** — the write path's transfer issue and its per-shard route, the replica
  fan-out, the notification accounting and its inverted count, the rejected-request
  field, the delegated-route trim and its equal-block-size precondition, the writer
  thread's lifecycle and handle release on failure
  (`test_rbln_nixl_push_worker.py`); the direction in the compatibility hash and the
  metadata that carries it (`test_rbln_nixl_metadata.py`); the connector and factory
  registration (`test_rbln_nixl_connector.py`, `test_factory.py`); the shared pieces
  the direction reaches into (`test_rbln_nixl_base_worker.py`,
  `test_rbln_nixl_base_scheduler.py`, `test_rbln_nixl_handshake.py`).
- Cases that cover an override enter through the upstream method that calls it rather
  than calling it directly, so an override that goes dead when upstream re-routes a
  call is caught.
- Each refusal the write direction adds has a case that fails if the guard stops
  firing — including the opposite-direction handshake, which is invisible to every
  other check.
- `pytest tests/native` green; `pre-commit` clean on every changed file.

#### Risk / compatibility

- **The read path is untouched in behaviour.** The direction is a separate connector;
  a deployment that does not name it runs exactly what it ran before.
- **A mixed-direction pair now fails fast.** Previously it would have handshaken and
  then stalled. This is a behaviour change only for a configuration that could not
  have worked.
- The delegated-route trim is scoped to the case upstream mis-aligns and carries the
  upstream version that removes the need for it.

---

* Resolves rebellions-sw/fsw-inference#493 — RBLN NIXL connector push mode.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

2. `uv run --no-sync pytest tests/native`
3. Expected: all green.
4. Edge cases covered: a consumer whose chiplets replicate a slice, so one write
   would leave the other copies stale; a completion notification arriving from one
   producer rank while others are still writing; a request rejected by the serving
   layer before it was ever scheduled; a consumer that registered only the suffix of
   a prompt after a partial prefix-cache hit; a peer running the opposite direction;
   a per-shard write whose peer disagrees on block size; a submission that fails
   after the handle exists.

---

### 📋 Checklist

---

### 💬 Notes

- This branches from `feat/nixl-region-pairing` rather than `dev`: the write
  direction is built on the region pairing that branch introduces, and reviewing it
  against `dev` would show that work a second time.

### DIFF
--- a/vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
+++ b/vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
@@ -14,7 +14,7 @@
 
 from collections import defaultdict
 from dataclasses import replace
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, ClassVar, Literal
 
 import msgspec
 import numpy as np
@@ -104,6 +104,12 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
 
     compat_hash: str | None
     xfer_handshake_metadata: NixlHandshakePayload | None
+
+    #: Whether this side originates the bytes into the peer's memory. Two things
+    #: turn on it: a chiplet replica is a valid source to read from but not a
+    #: valid sole destination to write to (`_head_matched_desc`), and a peer
+    #: moving bytes the other way must not pass the handshake (`rbln_compat_hash`).
+    _writes_into_peer: ClassVar[bool] = False
 
     def __init__(
         self, vllm_config: VllmConfig, engine_id: str, kv_cache_config: "KVCacheConfig"
@@ -548,9 +554,15 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
         peer_areas: list[int] | None = None,
         split: int = 1,
         region_ids: list[int] | None = None,
+        replica_fanout: int = 1,
     ) -> tuple[int, list[tuple[int, int, int]]]:
         if self._sw_ratio is None:
-            if registered_layer_names is None and peer_areas is None and split == 1:
+            if (
+                registered_layer_names is None
+                and peer_areas is None
+                and split == 1
+                and replica_fanout == 1
+            ):
                 # No SWA view opt, whole-engine peer: upstream's Full-only
                 # layout, one handle covering every region.
                 return super().register_local_xfer_handler(block_size)
@@ -562,8 +574,14 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
                 peer_areas=peer_areas,
                 split=split,
                 region_ids=region_ids,
+                replica_fanout=replica_fanout,
             )
-        assert registered_layer_names is None and peer_areas is None and split == 1, (
+        assert (
+            registered_layer_names is None
+            and peer_areas is None
+            and split == 1
+            and replica_fanout == 1
+        ), (
             "RBLN NIXL: SWA view-opt is not supported with pipeline "
             "parallelism or heterogeneous tensor parallelism"
         )
@@ -751,9 +769,13 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
                         num_blocks=num_blocks,
                     )
                 )
-        # The same count `_register_shard_local_xfer_handler` builds locally
-        # and `_shard_descs_per_block` records.
-        assert len(out) == len(logical_pairs) * len(areas_iter) * num_blocks * split
+        # Per block, a region becomes one descriptor per head piece per peer
+        # copy -- the same count `_register_shard_local_xfer_handler` builds
+        # locally and `_shard_descs_per_block` records.
+        fanout = replicas_r if self._writes_into_peer else 1
+        assert len(out) == (
+            len(logical_pairs) * len(areas_iter) * num_blocks * split * fanout
+        )
         return out
 
     def _fan_in_peer_areas(
@@ -832,7 +854,9 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
         sub_len = desc_len // split
 
         # Replicas of a slice hold identical bytes, so reading any one of them
-        # answers; take the first.
+        # answers -- but writing only one leaves the peer's other chiplets on
+        # stale KV. Reading takes the first; writing takes them all.
+        fanout = replicas_r if self._writes_into_peer else 1
 
         out: list[tuple[int, int, int]] = []
         pieces: list[tuple[int, int]] = []
@@ -846,23 +870,25 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
                     f"outside the peer's range (it owns heads {base_r}.."
                     f"{base_r + per_slice_r * slices_r - 1})."
                 )
-            remote_region = logical_r * areas_r + slice_r * replicas_r
-            page = remote_lens[remote_region]
-            if page % per_slice_r != 0:
-                raise RuntimeError(
-                    f"RBLN NIXL D2D: peer region {remote_region} block length "
-                    f"{page}B does not split into {per_slice_r} heads."
-                )
-            # Heads are contiguous inside a block, so skipping `head_within` of
-            # them is a plain byte offset.
-            head_offset = head_within * (page // per_slice_r)
-            if sub_len + head_offset > page:
-                raise RuntimeError(
-                    f"RBLN NIXL D2D: local region {region_id} piece {j} wants "
-                    f"{sub_len}B at +{head_offset}B, past the end of the peer's "
-                    f"{page}B region {remote_region}."
-                )
-            pieces.append((remote_bases[remote_region] + head_offset, page))
+            for k in range(fanout):
+                remote_region = logical_r * areas_r + slice_r * replicas_r + k
+                page = remote_lens[remote_region]
+                if page % per_slice_r != 0:
+                    raise RuntimeError(
+                        f"RBLN NIXL D2D: peer region {remote_region} block "
+                        f"length {page}B does not split into {per_slice_r} "
+                        "heads."
+                    )
+                # Heads are contiguous inside a block, so skipping
+                # `head_within` of them is a plain byte offset.
+                head_offset = head_within * (page // per_slice_r)
+                if sub_len + head_offset > page:
+                    raise RuntimeError(
+                        f"RBLN NIXL D2D: local region {region_id} piece {j} "
+                        f"wants {sub_len}B at +{head_offset}B, past the end of "
+                        f"the peer's {page}B region {remote_region}."
+                    )
+                pieces.append((remote_bases[remote_region] + head_offset, page))
 
         for block_id in range(num_blocks):
             for base, page in pieces:
@@ -894,6 +920,22 @@ class RblnNixlWorkerBase(NixlBaseConnectorWorker):
             side="peer",
         )
         return self._head_split(per_slice_l, per_slice_r)
+
+    def _peer_replica_fanout(
+        self, nixl_agent_meta: RblnNixlAgentMetadata, remote_tp_size: int
+    ) -> int:
+        """How many of the peer's copies of a slice one transfer must touch.
+
+        The compiler duplicates a KV head across chiplet areas when a shard
+        owns fewer heads than the device has chiplets. Reading is free to pick
+        one; writing has to fill them all, or the peer's other chiplets keep
+        serving what was there before.
+        """
+        if not self._writes_into_peer:
+            return 1
+        if not self._is_head_matched_peer(remote_tp_size):
+            return 1
+        return max(1, nixl_agent_meta.kv_areas // nixl_agent_meta.kv_slices)
 
     @staticmethod
     def _head_split(per_slice_l: int, per_slice_r: int) -> int:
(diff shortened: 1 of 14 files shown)


vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
rebel-jindol21: Isn't "fanout" effective at this point?

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/push_worker.py
```diff
@@ -0,0 +1,302 @@
...
+                    msg="Push WRITE submission failed; releasing handle",
+                    error=e,
+                    dst_engine_id=engine_id,
+                    remote_pp_rank=global_rank,
+                )
+                # Outbound only: there is no local metadata to invalidate, so
+                # release this peer's handle and let the remaining peers go.
+                if handle is not None:
+                    self.nixl_wrapper.release_xfer_handle(handle)
+                self.xfer_stats.record_failed_transfer()
```
rebel-jongwonlee, line 275: ### A per-peer write failure leaves the consumer waiting indefinitely

This `except` releases the one peer's handle and lets the loop continue to the
remaining peers. That is right for the producer — the other peers' bytes are still
worth sending — but combined with this PR's writer-count accounting it strands the
consumer:

1. The failed rank never issues its transfer, so its `notif_msg=notif_id` is never
   delivered. The consumer sees `writers - 1` notifications.
2. `_writer_still_pending` therefore returns `True` for every one of them, so none is
   forwarded to `super()._get_new_notifs()`.
3. Upstream seeds `_recving_transfers.setdefault(req_id, [])` only inside that
   forwarded path (upstream `push_worker.py:692`), so `_pop_done_transfers` never
   reports the request and it never reaches `done_recving`.
4. The producer is unaffected: only the succeeded handles go into
   `_sending_transfers`, they complete, and `get_finished` puts the request in
   `done_sending`. P frees its blocks; D does not.

**The escape hatch the upstream comment names does not cover this.** I followed it:
the lease sweep (upstream `base_worker.py:1955-1971`) walks `_reqs_to_send` and adds
to **`done_sending`** — it exists to stop a producer stranding blocks on a remote, and
there is no matching sweep over `_recving_metadata`. `_handle_failed_transfer`
(`base_worker.py:2050`) is the consumer-side equivalent, but it only runs where *we*
own the NIXL handle, which in push mode we never do. The scheduler has no timeout
either: the request sits in `WAITING_FOR_REMOTE_KVS` on the skipped-waiting queue
holding its blocks until the client aborts.

**This is a trade the PR makes, not a regression it introduces.** Before it, the first
notification settled the request and the consumer decoded from a half-written cache —
silently. Turning silent corruption into a detectable stall is clearly the right
direction. My question is only whether the stall is meant to be terminal, since
nothing in the connector appears to end it.

Two shapes that would close it, if either fits: give up the whole request on the first
peer failure (release the handles already issued and let the producer's own failure
path run), or have the writer emit the missing notification as an explicit failure
marker the consumer can turn into `_failed_recv_reqs`. The second keeps the
partial-progress behaviour this code clearly wants.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/push_worker.py
```diff
@@ -0,0 +1,302 @@
...
+
+        TODO(vllm>=0.27.2): delete -- upstream trims the write path itself there.
+        """
+        local = list(local_block_ids)
+        for i, remote_group in enumerate(remote_block_ids):
+            num_remote = len(remote_group)
+            assert num_remote <= len(local[i]), (
+                f"group {i}: consumer registered {num_remote} blocks but this "
+                f"producer holds {len(local[i])}; the pair cannot be aligned"
+            )
```
rebel-jongwonlee: ### `_trim_to_consumer_blocks` asserts on a length the consumer supplied

`num_remote` is `len(meta.remote.block_ids[i])` — it reaches this process over the
handshake, not from our own config. If a consumer ever registers *more* blocks than
the producer holds, this fires as
`AssertionError: group 0: consumer registered 9 blocks but this producer holds 8`,
which is exactly the message an operator needs.

You already took this position one PR down: `c2fc28e8` turned all three
`_slice_head_bounds` guards into `RuntimeError`s because their inputs cross a
process boundary, and `ff790096` typed the `side` label on top of that. This is the
same class of input — a peer's advertised shape that cannot be paired with ours —
and it is the only guard of that kind left in the stack. A `raise RuntimeError`
naming the engine and request id would put it in the same shape as the rest.

**Secondary, and weaker than the argument above:** under `-O` this guard and the
`assert len(local_descs) == len(remote_descs)` at L241 both vanish. The trim below
is then skipped too (`9 < 8` is false, so the local list stays at 8) and, since a
group's descriptor count is proportional to its block count, an 8-vs-9 pair reaches
`make_prepped_xfer`. So it does still surface — as a NIXL error from two hunks up
rather than as the message here. I have no evidence anyone runs this with `-O`, so
I would treat that as a footnote, not a reason on its own.

**Explicitly not this one, for contrast:** the `assert not both` in `start_load_kv`
(L78) reads correctly as an `assert` to me. It guards an ordering invariant *inside*
this process — `request_finished` runs after the step's metadata is built, so the
host copy always lands a step earlier — and the docstring's "Nothing states that" is
exactly what an `assert` is for. I mention it only so the suggestion above doesn't
read as "convert every assert".

rebel-jongwonlee (commented):
Read the whole change. The split into `base_worker` / `pull_worker` / `push_worker`
makes the direction-specific parts far easier to follow than the single `worker.py`
did, and "pairing does not care -- it describes what two peers hold" is a good line to
have drawn: it's what keeps the region-pairing work in #1011 from having to know about
direction at all.

Two inline notes. The rest looked routine to me: `connector.py` and `factory.py`
register the new connector by name and nothing else; `push_scheduler.py` is 51 lines
whose only substance is the `setdefault` for a pre-aborted request's
`remote_block_ids`; `_ensure_push_writer` mirrors the start upstream inlines in
`register_kv_caches`, which the D2D deferral never reaches, and both sites guard on
the thread being unset. Folding direction into the compat hash
(`RBLN_NIXL_CONNECTOR_VERSION` 2→3 plus `rbln_writes_into_peer`) is the right place for
it — upstream's eleven factors carry neither the connector name nor the direction, so
two engines configured opposite ways would otherwise pass every geometry check and
simply never move bytes.

One thing I'm deliberately leaving out: the read path has the same trimming problem
`_trim_to_consumer_blocks` solves here, and we'd rather handle that as its own patch
than widen this one.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/push_worker.py
rebel-ykchoi: Agreed, and done in f74f482f.

`_trim_to_consumer_blocks` now takes the engine id and the request id and raises

    RBLN NIXL: consumer <engine> registered 2 block(s) for request <req> in
    group 0, more than the 1 this producer holds; the peer's advertised length
    crosses the handshake, so the pair is refused rather than trimmed to it.

Same shape as the `_slice_head_bounds` guards, for the reason you name: the
length is the peer's, not ours. The test pins the identifiers rather than just
the exception type.

Your reading of L78 matches mine — `assert not both` is an ordering invariant
inside this process, so it stays an assert. And I took the `-O` note as the
footnote you marked it as; the guard is worth having on its own terms.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/push_worker.py
rebel-ykchoi: Your chain is right, and I confirmed each link independently — including the two escape hatches you checked: the lease sweep feeds `done_sending` only, and `_handle_failed_transfer` needs a handle we never own on the consumer side in push mode. So the stall is terminal today.

It is also not new information to this PR's design, which is why I would rather not fix it here: the same asymmetry — a push consumer settles on the peer's notification alone, with nothing of its own to fall back on — shows up in two more places than the failure path:

* an aborted request whose consumer registers an empty recv, where the producer strands its prefill blocks until the lease expires;
* a notification that lands on a consumer rank before it has registered.

Fixing them one at a time inside the writer loop would give three overlapping patches to the same accounting. So: **your option two**, a failure marker the writer emits and the consumer turns into `_failed_recv_reqs`, as its own change covering the whole completion path. Both ends already exist — the push path sends standalone notifications for its registration handshake, and `_failed_recv_reqs` is what feeds `done_recving` — so it is a matter of joining them, not of adding a mechanism. It keeps the partial-progress behaviour, and it is the only one of the two that gives the consumer something to report rather than something to wait for.

Option one (give up the whole request on the first peer failure) would close this instance in a few lines, so if you would rather not merge a known terminal stall, say so and I will take it here and leave the marker for the follow-up.

rebel-jongwonlee (approved):
`f74f482f` checked out. Both call sites carry the pair, the message names engine,
request and group, and the test pins the identifiers rather than the exception type
— that is the shape I was after. CI green on build 679. And yes, `assert not both`
at L78 stays an assert; you read the contrast the way I meant it.

**On the stall: take the deferral, and don't take option one.** Two reasons, one of
which is me correcting my own comment.

I wrote that either shape "would close it". That was wrong about option one. I went
through every `send_notif` call site — pull completions, heartbeats, PUSH_REG, and
the MLA fan-out notify — and there is no path anywhere that signals a failure to a
peer. So giving up the whole request on the first peer failure leaves the producer
releasing its blocks on the lease and saying nothing; the consumer still receives
nothing and still hangs. It buys a partial write avoided, not a stall closed. Taking
it here would spend a few lines and a change in failure semantics to leave the
problem exactly where it is.

Second, the stall is not this PR's to answer for. With `writers == 1` — equal TP,
the common shape — a submission failure sends no notification at all, the RBLN
filter is never even reached, and plain upstream strands the consumer in precisely
the same way. What the counting changes is the multi-writer case, and there it
replaces a silent half-written decode with the stall that was already upstream's
behaviour everywhere else. Blocking on it would hold this PR to a bar upstream does
not meet.

I also checked your second instance and it holds: a notification arriving before
`_recving_metadata` is populated hits the `Unrecognized request` branch and is
dropped, with nothing to replay it once the metadata lands. Three patches to one
accounting would be the wrong shape. Option two as its own change is right, and your
point that both ends already exist — standalone notifications on the push path,
`_failed_recv_reqs` feeding `done_recving` — is what makes it a join rather than a
new mechanism.

Approving.

PR #1020, opened 2026-08-28
feature(specdec): enable DFlash speculative decoding with max_num_seqs > 1

rebel-scottlee:
### 🚀 Summary of Changes

This PR lifts the first limit listed in #975: `--max-num-seqs > 1` was refused because acceptance dropped to roughly a third of the single-sequence value and we didn't know why. The cause turned out to be a batch-size mismatch between the target and the drafter. This PR fixes it and removes the startup guard.

**What was wrong**

The runner pads target verification to a decode bucket: locally, the smallest compiled bucket that fits the live request count, or under DP, the bucket agreed across ranks. The DFlash drafter, on the other hand, ran its query pass at the raw live request count. `_run_query_pass` computed the step's `BatchDescriptor` but never used it; the draft metadata, the model inputs, and the sample indices were all sized by `num_reqs`.

So whenever the live count wasn't a bucket (one active request at `max_num_seqs=2`, three at `max_num_seqs=4`), or a DP peer pushed the target to a wider bucket, the target verified at one batch shape while the drafter drafted at another. The two shapes give slightly different BF16 results, and that is enough to flip near-tie draft choices and collapse acceptance. It also explains why #975 saw the drop with a single in-flight request. As a side effect, a non-bucket live count compiled a new drafter graph in the middle of serving, since warmup only runs `dummy_run` at bucket sizes.

**What this PR does**

The drafter now runs at the same decode bucket as the target.

- `_run_query_pass` uses the step's `BatchDescriptor`. The draft metadata is built with the bucket as its batch pad, the model runs at `num_reqs_padded` rows, and the forward context gets the padded query batch as its token dimension (still `None` off the DP path).
- Under DP, the drafter replays the target's decision from the already-published `dp_status` through `determine_batch_execution_and_padding`, so no extra collective is needed. A rank that is prefilling locally is left alone, since the target doesn't pad prefills to a decode bucket.
- Padded rows are filled with the mask token at position 0 instead of whatever the previous step left in the buffers. A bucket whose query block doesn't fit the token buffers is rejected rather than silently run at the live shape.
- `propose()` already keeps only the leading real rows, and warmup already covers every bucket with `dummy_run`, so neither changes. Every drafter shape is now compiled during warmup.
- `_require_dense_drafter` still rejects a fused-MoE drafter, but the message now gives the reason that still holds after this change: DP-idle ranks skip the draft, so a busy rank's expert collective would wait on peers that never join.

`_require_single_sequence` is removed. The batched verification path is correct now that the mask-bound bugs in the batched causal decode kernel (query positions 1–7 corrupted from the second verification on) are fixed on the compiler side.

`rbln_model_runner.py` is not touched.

---

* Related to #975 (resolves its first listed limit)

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit tests:
   ```
   pytest tests/vllm/v1/spec_decode/test_dflash.py
   ```
   `TestSchedulerCapacity` checks that the proposer initializes at `max_num_seqs` 1/2/4/16. `TestQueryPassBatch` covers the query-pass contract: a live count between buckets runs at the next bucket, a bucket-sized batch isn't padded, padded rows hold defined values, `propose()` returns one row per real request, DP pads the token dimension to the bucket, an oversized bucket is rejected, and the DP query batch matches the target's.
2. Single-request E2E (MiniMax-M2.5 + DFlash on RBLN, greedy) at `--max-num-seqs` 1, 2, and 4. Mean acceptance length should stay in the same range as capacity 1 instead of falling below the no-speculation baseline.
3. Three concurrent requests at `--max-num-seqs 4`, a live count that isn't a bucket. No drafter graph should be compiled during serving, and acceptance should stay in the same range as the bucket-sized cases.

---

### 📋 Checklist

---

### 💬 Notes

* This needs a rebel-compiler build that includes the batched-decode mask-bound fix (rebel-compiler #13226, merged into `dev`). `dev`'s runtime pin, `rebel-compiler~=0.11.3.dev499`, already contains it, so this PR carries no pin change of its own. Without that fix, `max_num_seqs > 1` output is wrong regardless of this PR.


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -173,6 +173,30 @@
...
+
+    ``max_batch_size`` is a safe upper bound on the number of buckets: integer
+    division strictly decreases every exponential candidate, so the manager will
+    stop at ``min_batch_size`` long before reaching the limit.
+    """
+    if strategy == "exponential" and speculative_method == "dflash":
+        return max(configured_limit, max_batch_size)
+    return configured_limit
+
+
```
rebel-minkyu: With the later commits in this PR, the drafter now runs at the same decode bucket as the target, so the target/drafter batch-size mismatch that this change was meant to fix no longer seems to exist. Given that, is the `_decode_batch_bucket_limit` still necessary? Isn't the acceptance collapse at MNS > 1 resolved by the compiler-side fixes?

rebel-minkyu (commented):
I guess this PR requires the corresponding compiler-side change to be merged to properly fix the acceptance collapse, so we better update the pinned rebel-compiler version as well after it gets merged. Is my understanding correct?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -461,11 +461,12 @@ def __init__(
...
+        max_decode_batch_size = decode_batch_size(
+            self.max_num_reqs, self.parallel_config.pipeline_parallel_size
+        )
         self.bucketing_manager = get_bucketing_manager(
             envs.VLLM_RBLN_DECODE_BATCH_BUCKET_STRATEGY,
-            max_batch_size=decode_batch_size(
-                self.max_num_reqs, self.parallel_config.pipeline_parallel_size
-            ),
+            max_batch_size=max_decode_batch_size,
             min_batch_size=envs.VLLM_RBLN_DECODE_BATCH_BUCKET_MIN,
```
rebel-minkyu, line 468: One minor comment. It seems we can just keep the original code without this diff. Thank you!

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-scottlee: Yes, with the later changes, the drafter now runs at the target’s decode bucket, so forcing the full bucket ladder is no longer needed for batch alignment. I removed _decode_batch_bucket_limit and restored the configured bucket limit.
As for the acceptance collapse, it is as I explained offline.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-scottlee: Reverted to the original call in 3600d2b.

rebel-minkyu (approved):
LGTM!

rebel-minkyu:
@rebel-cjlee This PR requires your approval too. Please take a look when you have time.

rebel-cjlee (approved):

PR #1023, opened 2026-08-28
perf(nixl): issue a pull-mode KV read behind the model submission

rebel-ykchoi:
### 🚀 Summary of Changes

Upstream calls `start_load_kv` before the forward, so a read's NIXL submissions — one
per producer shard, multiplied by a pipelined producer — run on the engine thread ahead
of the submission the model waits on. Under data parallelism that delay is not local:
the rank still submitting reaches the next shape barrier behind its peers and holds them
there. This is what the change removes, and it holds whether or not the engine thread
has other work after the dispatch — the barrier is a CPU collective on the step path,
so asynchronous scheduling does not absorb it.

This PR holds the read and issues it once a submission is in flight. The blocks stay
disjoint: a request whose KV is still arriving cannot be in the batch that step submits,
because the scheduler withholds it until the load is reported finished.

Three sites issue, since any can be a step's last chance. Which one carries a read
depends on the step it lands on. A read on a step with no forward of its own — which is
what a step is when the decode engine's `running` set is empty, so a busier one lands its
reads on steps that do carry a forward — is issued behind the following dummy step's
submission, or on entry to the next `execute_model` when no dummy step runs. A read on a
step that does run a forward is issued behind that submission. The entry site is not an
optimization: a request is listed for receive once and the list is cleared while the
metadata is built, so a read nobody issues strands its request for good.

Two things the argument rests on are now stated in the code rather than assumed. The
entry flush always precedes that step's own `start_load_kv`, so holding a second read is
asserted against; overwriting one would strand its request the same way. And a later step
comes at all only because the scheduler counts a request waiting on remote KV as
unfinished work — a dependency on upstream queue accounting that this PR does not own and
that a paused scheduler does not honour.

A flush replays the whole of `start_load_kv`, so the producer-side lease arming and the
heartbeats move with the read. Their expirations are absolute instants and their cadence
is decided where the metadata is packaged, so only their arrival shifts: blocks are held
marginally longer, never released early.

The hook is RBLN's, so under `MultiConnector` it lives on nested children and is reached
the way the deferred registration hook already is. Push is unchanged: its writer thread
already submits away from the engine thread.

Affects the vLLM-native path, pull direction only.

---

* Resolves rebellions-sw/fsw-inference#539

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run the unit tests covering the change:

   ```
   pytest tests/native/distributed/kv_connector/test_rbln_nixl_connector.py \
          tests/native/distributed/kv_connector/test_v1_utils.py \
          tests/native/v1/worker/test_rbln_model_runner.py
   ```

2. Verify output: all pass. What they pin —
   * `start_load_kv` keeps the read and does not reach upstream's issue path
   * holding a second read without an intervening flush is refused
   * a flush issues the held read exactly once, and a second flush issues nothing
   * the helper walks `MultiConnector` children and skips connectors without the hook
   * on the dummy step the flush follows the submission and precedes the draft
   * on `execute_model` entry the flush precedes that step's connector work

3. Edge cases tested: warm-up asks no connector; a drained group flushes nothing (it has
   no submission of its own, and the next `execute_model` picks the read up instead).

---

### 📋 Checklist

---

### 💬 Notes

Rebased onto `dev`: the asymmetric per-shard read path this needs has landed, so the
branch no longer carries a prerequisite.

The second commit answers review: it asserts the handoff invariant and records the
dependency on the scheduler's accounting, both described above.


rebel-huijong (approved):
I've confirmed that hiding the KV transfer launch behind device execution works as intended. Of the diff in this PR, only vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/connector.py and vllm_rbln/v1/worker/rbln_model_runner.py relate to that functionality, so those are the only files I looked at. Since the branch this gets merged into will be PR'd against dev again anyway, I'll review the rest at that point.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/connector.py
```diff
@@ -159,6 +167,29 @@ def __init__(
...
+
+        A request is listed for receive once, so a read nobody issues strands it
+        for good -- hence every site that can be a step's last chance flushes.
+        """
+        meta = self._deferred_load_meta
+        if meta is None:
+            return
+        self._deferred_load_meta = None
+        assert self.connector_worker is not None
+        self.connector_worker.start_load_kv(meta)
```
rebel-jongwonlee, line 202: ### Please state the dependency on `skipped_waiting`

The last piece of the three-site argument is that another `execute_model` is guaranteed
to come. It is, but only because `get_num_unfinished_requests()` counts
`len(self.skipped_waiting)` (`scheduler.py:2116`) — that is what keeps `has_requests()`
and therefore `has_work()` true, and stops the busy loop parking on the input queue
while a read is held.

**That is an upstream invariant this PR does not own, and it fails silently.** If it
ever stopped holding, the held read would never be issued, and the scheduler does not
re-list a request for receive — so the request strands for good. Precisely the failure
the third site exists to prevent.

There is already one case where it does not hold: `PauseState.PAUSED_ALL` forces that
count to 0. It resumes on the next `execute_model` after unpause, so that one is a
delay rather than a loss — but the fact that the edge exists at all shows how fine the
dependency is.

A sentence or two here or at the entry-flush site would be enough. Someone changing
upstream queue accounting later has no reason to trace their way to this file.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/connector.py
```diff
@@ -159,6 +167,29 @@ def __init__(
...
 
+    def start_load_kv(self, forward_context: "ForwardContext", **kwargs) -> None:
+        """Keep this step's read; `flush_deferred_load` issues it.
+
+        Safe because the scheduler withholds a request until its load is
+        reported finished, so the read and the forward touch disjoint blocks.
+        `clear_connector_metadata` rebinds that field to None, so the object
+        kept here survives the step.
+        """
+        self._deferred_load_meta = self._connector_metadata
```
rebel-jongwonlee, line 182: ### Guard the unconditional overwrite

```python
self._deferred_load_meta = self._connector_metadata
```

Today this cannot drop a read: the entry flush always runs before the step's own
`start_load_kv`. But that ordering is what the whole three-site argument rests on, and
overwriting a held meta is the same "strands its request for good" failure this
docstring warns about, reached by a different route.

`assert self._deferred_load_meta is None` pins it for free.

rebel-jongwonlee (commented):
Checked `4c431e66`. Both things you asked a reviewer to look at — a fourth place that
could be a step's last chance, and the disjointness argument — I traced and found
nothing to raise. `start_load_kv` is reachable from exactly two places (L1634, L1708),
each with a guaranteed flush; `model_executable` is called in exactly two places, each
flushed on the adjacent line. A receiving request is pushed to `skipped_waiting` and
contributes no tokens, and the blocks it receives into are the freshly allocated
uncached suffix, so they cannot be in the batch. An abort in the window does not
undercut it either: `delay_free_blocks=True` keeps the blocks reserved until
`finished_recving`.

Two questions and three requests below.

## Questions

**1. Was this judged with async scheduling on?**

The premise here is that the host is idle once the forward has submitted. But #951 is
in the base and is on by default when device-tensor and the RBLN sampler are both
enabled — which is exactly the configuration this path runs in. Under async scheduling
the engine thread does not sit idle after the dispatch; it runs straight on into
preparing the next step, so the read prep lands in front of the *next* dispatch rather
than in dead time.

That does not remove the win — this step's NPU starts sooner either way. But the
picture the description paints is the synchronous one, and I would like to know whether
that difference was accounted for.

**2. Is the wider deferral scope intended?**

`flush_deferred_load` replays the whole of `start_load_kv`, so the P-side bookkeeping
(`_reqs_to_process`, the `_reqs_to_send` lease arming) and `_send_heartbeats` move with
it. On the no-forward path that means they land *after* that step's `get_finished`, one
step late.

I could not find harm in it — the expirations are absolute values computed
scheduler-side, so only their arrival shifts, and the heartbeat rate is unchanged with
only its phase moving by a step. But both the naming and the description say "load", so
I would like to know whether this was deliberate or fell out of deferring at function
granularity. If deliberate, a sentence in the `flush_deferred_load` docstring would say
so.

## Request 3 (the other two are inline)

**The base note in the PR body is stale.** "Based on `feat/nixl-region-pairing` (#1011),
not `dev` … Please merge that one first" — both #1011 and #1018 have landed and this
branch is clean on `dev`. Someone reading this later would go looking for a
prerequisite that no longer exists.

rebel-jongwonlee (approved):
This is the response I hoped for on Q1 — re-measuring and finding the win comes from a
different mechanism than the description claimed is a better outcome than confirming it.
And the frontend-resolution explanation matches what I had found independently:
`Asynchronous scheduling is %s` logs at `config/vllm.py:1007`, while
`current_platform.check_and_update_config` runs at `:1350`, so that line reports
upstream's decision before RBLN gets to override it. Worth knowing generally — it is not
a reliable indicator of whether async actually ran.

The barrier framing holds up better than mine did: a CPU collective on the step path is
exactly what async scheduling cannot absorb, which is why the win survives where my
"host is idle after dispatch" premise does not.

On Q2 you found a better argument than the one I was reaching for. I had gotten as far as
"I cannot find harm in it"; `_reqs_to_process` existing precisely for this lag — async
pushing the P-side expiration and the D-side read apart — says the mechanism already
tolerates the gap rather than that the gap happens to be survivable. That belongs in the
docstring and it is there.

**`PAUSED_NEW` is a catch I missed.** `get_num_unfinished_requests` returns
`len(self.running)` there, and a request waiting on remote KV is in `skipped_waiting`, so
the count drops it in both pause states, not one. Naming both is right.

Verified the three requests:

* **R1** — `assert self._deferred_load_meta is None` with
  `test_holding_a_second_read_is_refused`, which fails without it.
* **R2** — both facts are in the `flush_deferred_load` docstring: the `skipped_waiting`
  dependency with both pause states, and that a flush replays the whole of
  `start_load_kv` with absolute deadlines.
* **R3** — base note gone, rebased, `mergeable: true`.

**One note on the reworded description, non-blocking.** It says a read landing on a step
with no forward of its own is "the usual case, since a request waiting on remote KV does
not share a step with one that is decoding". The second half is load-dependent rather
than structural: `schedule()` runs the RUNNING loop before the WAITING loop
(`scheduler.py`, "First, schedule the RUNNING requests", then line 625), so a
`WAITING_FOR_REMOTE_KVS` request contributing no tokens does not stop the requests
already decoding from contributing theirs. The step has no forward when D's `running` set
is empty — a first arrival, low load, or every request still waiting on its KV — which
may well be what your configuration does, but a busier D should land reads on
forward-carrying steps.

Since the description now leans on which site carries a read, it might be worth saying
"the case this configuration measures" rather than "the usual case", or naming the
condition (`running` empty). The three-site argument does not depend on which one
dominates, so this is wording rather than substance.

Approving.

rebel-cjlee (approved):

PR #1029, opened 2026-08-31
fix(deps): bound the RBLN dependencies to a patch line, and enforce it

rebel-junamsong:
## Problem

`optimum-rbln` and `torch-rbln` carried only a floor:

```toml
"torch-rbln>=0.4.0 ; platform_machine == 'x86_64'",
"optimum-rbln>=0.11.2",
```

A floor is not a pin for a package that gets published. A released wheel resolves against whatever is newest **at install time**, not at release time — so every vllm-rbln release installs today's optimum-rbln instead of the one its SDK version matrix documents. The deploy matrix check reports it for thirteen releases at once:

```
2024.08.30.1: vllm-rbln==0.0.6         brings optimum-rbln==0.11.2,      matrix documents 0.1.9
2025.07.31.0: vllm-rbln==0.8.2         brings optimum-rbln==0.9.1,       matrix documents 0.8.2
2025.10.31.0: vllm-rbln==0.9.2.post1   brings optimum-rbln==0.10.0.post2, matrix documents 0.9.2
2026.07.31.0: vllm-rbln==0.11.1.post1  brings optimum-rbln==0.11.2,      matrix documents 0.11.1
…
```

Those wheels are published and cannot be corrected. What can be corrected is that the next one carries a ceiling.

## Change

```diff
-    "torch-rbln>=0.4.0 ; platform_machine == 'x86_64'",
-    "optimum-rbln>=0.11.2",
+    "torch-rbln>=0.4.0,<0.4.1 ; platform_machine == 'x86_64'",
+    "optimum-rbln>=0.11.2,<0.11.3",
```

The cap is at the **patch line**, not the minor. A minor-level cap (`~=0.11.2` → `<0.12`) still admits 0.11.3, which is exactly the drift the matrix reports. `.postN` of the floor stays admitted, since it fixes the same line.

**No version changes:** both currently resolved versions (optimum-rbln `0.11.2`, torch-rbln `0.4.0`) already satisfy the new bounds.

## The guard — `tests/test_packaging_bounds.py`

Makes it a rule rather than a habit. Every requirement in `pyproject.toml` whose name contains `rbln` — across `dependencies`, `optional-dependencies` and `dependency-groups` — must:

1. declare a floor, and
2. be capped to that floor's patch line.

It asserts what the specifier **admits** (probing next-patch / next-minor / next-major) rather than which operators it spells, so `>=X.Y.Z,<X.Y.Z+1` and `~=X.Y.Z.0` both pass while an unbounded or minor-capped one fails — with the exact string to write:

```
AssertionError: `optimum-rbln>=0.11.2` in [project.dependencies] still admits 0.11.3
(next patch). An RBLN dependency must be capped to the patch line of its floor
(0.11.2) … write `>=0.11.2,<0.11.3`.
```

A separate test asserts the known RBLN requirements are still discovered, so a rename cannot silently empty the parametrisation and leave the guard green while checking nothing.

`rebel-compiler` is deliberately out of scope: it is selected per build rather than per release (the `runtime` extra), and is bounded by its own convention.

## Where it runs

An always-on gate in `.buildkite/pr.yml` next to `pre-commit`, before the heavy suites upload.

- **Always-on**, because an unbounded requirement is only visible in the pyproject of the PR that introduces it — after a release is published it cannot be corrected.
- **`--no-project`**, because it reads `pyproject.toml` and does not need the project installed, so it costs no shared-venv sync.

## Verification

| check | result |
|---|---|
| `pytest tests/test_packaging_bounds.py` | 5 passed |
| same test with the bounds reverted | 2 failed (both caps), with the actionable message above |
| `ruff check` / `ruff format --check` | clean |
| `mypy --python-version 3.10` / `3.13` | no issues |
| `uv lock --check` | up-to-date |

## Note on `uv.lock`

It carries the two updated specifiers only. A full `uv lock` in my environment also rewrites optimum-rbln's and torch-rbln's `source` from `pypi.org` to the internal nightly mirror — an artifact of local index resolution, unrelated to this change — so the lock was updated to mirror pyproject and left otherwise untouched. `uv lock --check` accepts the result.


rebel-eunji (approved):
lgtm 👍

PR #1031, opened 2026-08-31
fix(sampler): pad allowed_token_ids_mask with the rest of the batch axis

rebel-eunji:
### 🚀 Summary of Changes

This was missed in #1014 
The allowed_token_ids mask tensor also needs to be padded, just like the other tensors such as top_k / top_p.

---

* Related to #1014

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

```
uv run --no-sync pytest tests/native/v1/worker/test_rbln_model_runner.py -q \
  -k "TestPadDepad or TestSamplePadding"
```

The new `test_pad_sampling_metadata_pads_allowed_token_ids_mask` fails without the fix (the mask comes back at 2 rows instead of the bucket's 4) and passes with it; the surrounding pad/depad tests still pass (11 total).

End to end, the crash reproduces with the ngram spec-decode offline example plus `SamplingParams(..., allowed_token_ids=list(range(1000)))`, fewer prompts than `max_num_seqs`, and prompts repetitive enough for ngram to draft: before this fix the first drafted decode step raises `RuntimeError` from `masked_fill_` inside the rejection sampler's bonus path; after it, the run completes.

---

### 📋 Checklist

---

### 💬 Notes

Found by review of #1014, but the non-spec exposure predates it — #1014 only extended the reach to the rejection branch. Verified at unit level and by code reading of the upstream mask application; the end-to-end repro above has not been run on an NPU in this session.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -3497,9 +3497,6 @@ def _pad_sampling_metadata(md: SamplingMetadata, bucket: int) -> SamplingMetadat
         temperature=_pad_rows(md.temperature, bucket),
         top_p=_pad_rows(md.top_p, bucket),
         top_k=_pad_rows(md.top_k, bucket),
-        # Repeating the last request's mask row is safe: a real request always
-        # allows at least one token, and the padded rows' samples are discarded
-        # by _depad_sampler_output.
         allowed_token_ids_mask=_pad_rows(md.allowed_token_ids_mask, bucket),
     )
     if not md.no_penalties:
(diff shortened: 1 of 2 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3497,6 +3497,10 @@ def _pad_list(lst):
         temperature=_pad_rows(md.temperature, bucket),
         top_p=_pad_rows(md.top_p, bucket),
         top_k=_pad_rows(md.top_k, bucket),
+        # Repeating the last request's mask row is safe: a real request always
+        # allows at least one token, and the padded rows' samples are discarded
+        # by _depad_sampler_output.
```
rebel-eunji:
```suggestion
```

rebel-jonghewk (approved):
Thanks! lgtm. Can you also check `min_p`?

rebel-cjlee (approved):

PR #1032, opened 2026-08-31
fix(metrics): keep MODEL+SAMPLE meaningful under spec decode

rebel-yskim:
## Problem

With `VLLM_RBLN_METRICS`, the `MODEL + SAMPLE` section approximates device time by relying on where the step's first blocking device read happens: the model call and `_sample` are async dispatches, and the wait for the device lands inside `_bookkeeping_sync` (the `tolist()` of the sampled tokens), which is one of the three timed callables.

Spec decode breaks that premise. On the spec-decode path the first blocking read moves out of every timed callable:

- the spec-only logits gather (`logits = logits[logits_indices]`) in `execute_model`'s postprocess block syncs on the forward **before** `_sample` runs, and
- the drafter executes between `_sample` and `_bookkeeping_sync` (for MTP/EAGLE it runs a draft-model forward there).

`MODEL + SAMPLE` then collapses to dispatch cost while the E2E residual absorbs the whole device wait, so every mtp/ngram config in the perf CI reports a huge "host overhead" gap that is actually device time. Non-spec configs are unaffected and report normal gaps.

Verified with a per-region breakdown on a small config: enabling ngram spec decode moved the decode-step device wait from the `bookkeep` region into the `postprocess` region while `bookkeep` collapsed to dispatch cost, and the region times summed to E2E.

## Fix

Fold the runner's `postprocess` region into the graph sum: the spec-only gather is a hard sync on the target forward, so MODEL + SAMPLE keeps reading as the time the step was bound to the target graph and the sampler. The block is inline with no method to wrap, so the patch registers a `timed_region` wrapper over the module's `record_function_or_nullcontext` symbol; it adds `add_graph_time` only for that region name and passes every other region through untouched. A test pins the region literal to `execute_model`'s source, so renaming the region fails the suite instead of silently degrading the report.

The drafter is deliberately not folded: MODEL + SAMPLE stays the main-graph number, and the drafter's time -- under MTP/EAGLE a draft-model forward, plus any sampler wait its first read absorbs -- shows up in the E2E residual, which is therefore not pure host overhead in spec-decode configs. Verified on a spec-decode perf CI config: folding the drafter inflated the section's mean and tail with draft time, while excluding it leaves a tight distribution that tracks the target graph. The report format does not change.

Non-spec configs are unaffected: the postprocess block does no spec gather there and the draft region never runs.

### DIFF
--- a/vllm_rbln/patches/metrics.py
+++ b/vllm_rbln/patches/metrics.py
@@ -24,12 +24,19 @@ absorbs the wait. Async defers that read out of the pass, so there it is dispatc
 A pass is recorded only once its phase and its graph time are both known, which is
 what keeps those counts equal.
 
+Spec decode moves that first blocking read again -- into the spec-only logits gather
+in execute_model's postprocess block, and runs the drafter between _sample and
+_bookkeeping_sync. Neither is a wrappable method, so timed_region folds the runner's
+"postprocess" and "draft" profiler regions into the same graph sum; both are empty
+without spec decode.
+
 The whole feature lives in this module so the runner carries no metrics code at all. A
 range that is not a whole method -- the model call sits mid-way through execute_model --
 is reached by wrapping the callable the runner holds, and the phase, which is a local of
 execute_model, is read where the runner computes it.
 """
 
+import contextlib
 import functools
 import json
 import os
@@ -39,6 +46,7 @@ from dataclasses import dataclass, field
 from enum import Enum
 
 import numpy as np
+from vllm.v1.utils import record_function_or_nullcontext
 
 from vllm_rbln import envs
 from vllm_rbln.logger import init_logger
@@ -266,7 +274,9 @@ _shutdown = RBLNWorker.shutdown
 
 @functools.wraps(_execute_model)
 def execute_model(self, *args, **kwargs):
+    global _ACTIVE_CTX
     ctx = _ctx(self)
+    _ACTIVE_CTX = ctx
     ctx.start_pass()
     output = _execute_model(self, *args, **kwargs)
     # The engine calls sample_tokens() only if execute_model() returned None, so a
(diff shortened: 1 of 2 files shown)


rebel-cjlee (approved):
Nothing critical here for a debug tool, looks good to me.

Just Two small wording things:

1. `propose_draft_token_ids` is actually a real method, so "no method to wrap" only applies to postprocess. Easy to miss because the local closure right above has the same name.
2. This hangs entirely on the two region name strings matching the literals in the runner, so renaming a region would quietly put the metrics back where they were without failing anything.

vllm_rbln/patches/metrics.py
```diff
@@ -357,6 +367,30 @@ def shutdown(self):
     _shutdown(self)
 
 
+_ACTIVE_CTX: _PerformanceContext | None = None
+
+# The regions that hold device wait under spec decode (see the module docstring);
+# they belong to the graph sum, and are empty without spec decode.
+_GRAPH_REGIONS = frozenset(
+    {"rbln_model_runner: postprocess", "rbln_model_runner: draft"}
```
rebel-jonghewk: I think `draft region` is a wrappable method

rebel-yskim: You're right — the region wraps exactly one call to `propose_draft_token_ids`, which is a real method; I had mistaken the same-named local closure in `sample_tokens` for the definition. The drafter is now wrapped through the registry like the sampler, and only `postprocess` is still matched by region name.

[25a5f79d](https://github.com/RBLN-SW/vllm-rbln/pull/1032/changes/25a5f79d403559f06c6ef8ea459fc2448a4343a9)

rebel-jonghewk (approved):

rebel-jonghewk (approved):
lgtm

PR #1033, opened 2026-08-31
fix(sampler): size min_p to the logits rows the sampler receives

rebel-eunji:
## Summary

Fix `RBLNMinPLogitsProcessor.apply()` to align the size of `min_p` with the number of logits rows received by the sampler.

This issue affects only the vLLM-native execution path.

## Problem

A request using `min_p > 0` fails with a broadcasting `RuntimeError` on the first decode step where the number of active requests is smaller than the decode bucket size.

The mismatch occurs because:

1. Upstream `MinPLogitsProcessor.update_state()` sizes `min_p` according to the number of active requests.
2. The native runner hands the sampler logits padded to the decode bucket.
3. Upstream `apply()` performs the following in-place operation:

   ```python
   max_probabilities.mul_(self.min_p)
   ```

   Since the number of rows in `self.min_p` and `max_probabilities` can differ, the operation fails with a broadcasting error.

## Solution

Before calling upstream `apply()`, this PR zero-pads `min_p` to the logits row count and asserts that `min_p` is never longer than the logits.

This follows the same row-alignment contract already used for `temperature`, `top_k`, and `top_p` through `_pad_rows`.

Zero-padded `min_p` values are safe because `min_p = 0` is a no-op, and samples produced for padded rows are discarded by the runner.

The padded tensor replaces `self.min_p` the same way the existing dtype sync does, and `super().apply()` runs unchanged. `update_state()` re-slices `min_p` from the float32 buffer on the next batch change, so the padded copy does not outlive the bucket it was built for.

Prefill needs no handling. `_update_states()` removes every request the step did not schedule from the persistent batch, and the scheduler never mixes phases, so a prefill step runs with one request and one logits row. `min_p` is therefore never longer than the logits.

## Scope

- This is a pre-existing issue and is independent of #1014 and #1031.
- The non-speculative decoding path has always padded logits to the decode bucket size.
- Speculative decoding is not affected because `build_rbln_logitsprocs()` excludes MinP when `speculative_config` is enabled.
- The optimum path is unchanged. `RBLNInputBatch.refresh_metadata_rbln()` already sizes `min_p` to the bucket, so the new resizing branches are not used there.
- Existing optimum-path coverage is provided by:
  `tests/optimum/v1/sample/test_sampler_bucket.py::test_min_p_with_padded_bucket`

### Why the Optimum Path Is Unaffected

The optimum runner aligns the processor state with the bucket size before it reaches the processor.

`RBLNInputBatch.refresh_metadata_rbln()` passes `bucket_size`, rather than `num_reqs`, as `BatchUpdate.batch_size` to `get_and_reset()`. As a result, `MinPLogitsProcessor.update_state()` creates a bucket-sized `min_p` tensor from the start.

Padding slots use the default sampling parameters, so their `min_p` values are zero and therefore have no effect. The pooled logits passed to the RBLN sampler also contain `bucket_size` rows. Because the logits and `min_p` already have matching row counts when `apply()` runs, the resizing logic introduced by this PR is not exercised on the optimum path. This behavior is covered by `test_min_p_with_padded_bucket`.

The native runner does not have an equivalent state-sizing mechanism. Instead, it pads the tensors in `SamplingMetadata` afterward through `_pad_sampling_metadata()`. Because MinP’s per-request state is stored inside the processor object, that helper cannot resize it. The row alignment must therefore be handled directly in `RBLNMinPLogitsProcessor.apply()`.

## Related Work

- Related to #1014

## Type of Change

- [x] Bug fix

## Testing

Run the native logits processor tests:

```bash
uv run --no-sync pytest tests/native/v1/sample/test_rbln_logits_processor.py -q
```

One host-side test covers the public `update_state()` and `apply()` contract without requiring a device.

### 1. Decode: pad missing rows as no-ops

`test_decode_pad_rows_are_no_ops`

- Uses two active requests and four logits rows.
- Verifies that MinP masking is applied to the active rows.
- Verifies that padded rows remain unchanged.

Before this fix, the test fails with:

```text
RuntimeError: The size of tensor a (4) must match the size of tensor b (2)
```

The failure was reproduced against the unmodified processor before committing.

An end-to-end native-path request using:

```python
SamplingParams(temperature=1.0, min_p=0.5)
```

also fails before this change when the number of active requests is smaller than the decode bucket size, and completes successfully after the fix.

## Notes

- This issue was discovered while reviewing #1014. `_pad_sampling_metadata()` aligns tensors stored in `SamplingMetadata`, but the per-request MinP state lives inside the processor object, out of that helper's reach. Hence the fix sits in the processor.
- An earlier revision of this PR also sliced `min_p` to the leading rows for a supposed multi-request prefill step. That step does not exist on the native path (see Solution), so the second commit removes the branch, its test, and the restore of `self.min_p` that only that scenario needed.
- The decode-direction crash has not been reproduced on an NPU end to end in this session.


vllm_rbln/v1/sample/rbln_logits_processor.py
```diff
@@ -91,8 +91,18 @@ class RBLNMinPLogitsProcessor(MinPLogitsProcessor):
...
+
+        # update_state sizes min_p to the live request count, but on decode the
+        # native runner hands the sampler logits padded to the batch bucket.
+        # A min_p of 0 makes the padded rows no-ops.
+        pad = logits.shape[0] - self.min_p.shape[0]
+        assert pad >= 0
+        min_p = self.min_p
+        if pad:
+            min_p = torch.cat([min_p, min_p.new_zeros(pad, 1)])
+        self.min_p = min_p.to(logits.dtype)
```
rebel-jonghewk, line 101: @rebel-eunji Quick question before i look into this deeper. Are these operated on device tensor or cpu tensors?

rebel-eunji: The min_p tensor follows the device we set. Maybe, it works as rbln tensor in VLLM_RBLN_USE_VLLM_MODEL=1 and as cpu tensor in VLLM_RBLN_USE_VLLM_MODEL=0.

https://github.com/vllm-project/vllm/blob/v0.24.0/vllm/v1/sample/logits_processor/builtin.py

In the upstream code, min_p_device tensor is used if the device is not cpu.

rebel-jonghewk: I suspect there may be a noticeable slowdown when min_p is enabled, so it might be worth looking into the overhead. (This is out of scope for this PR, though.)

rebel-jonghewk (approved):

rebel-cjlee (approved):

vllm_rbln/v1/sample/rbln_logits_processor.py
rebel-eunji: Let's land the fix PR for the error first, and I'll keep digging into the perf issue.

PR #1034, opened 2026-08-31
docs(agent): keep agent-written changes small enough to review

rebel-cjlee:
### 🚀 Summary of Changes

When a coding agent writes a whole change by itself, the result is often too large to review. Recent examples in this repository, measured on `dev`:

- A `reason=` string of 1816 characters. The median in this repository is 194.
- A new file where more than half of the lines are comments and docstrings. The usual amount here is one such line for every 4 lines of code.
- A helper function with its own docstring that is called from only one place.

Each of these looks reasonable by itself. But when they happen together, a human reviewer cannot check every line. Checking every line is what a review needs.

`AGENTS.md` did not prevent this. The Comments section said that comments "may be long", and it did not say where the extra text should go instead. The Scope section set limits for helper functions and for abstractions, but it said nothing about new modules, about unrelated cleanup, or about who is responsible for the result.

This PR has two commits:
1. Set limits on comment length, and state that long explanations belong in the PR description.
2. State who is responsible for a change, and limit what it may add on top of the fix. This adds an accountability rule, a check before a new file or layer is created, a rule that keeps unrelated refactoring out, and a size report in `finishing-a-change`.

#### Why these numbers

Each limit is what the repository already does. Measured on `dev`, not chosen.

| Rule | Already inside it | Worst case |
| --- | --- | --- |
| Comment block 5 lines or fewer | 92%, or 96% ignoring the 13-line licence header | about 94 blocks at 6 or more |
| Docstring text 15 lines or fewer | 97% | 73 lines |
| `reason=` 400 chars or fewer | 81%, median 194 | 1816 chars in 30 lines |
| Comments and docstrings per file | median 0.25 | 0.61 |

These limits do not require a new style, They only stop the recent outliers.

Our rule says not to move the extra text into the commit message instead. The commit body does not last here. None of the last 40 commits on `dev` has a body beyond trailers, because the squash merge keeps only the subject line.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

---

### 🧪 How to Test

N/A

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/.claude/skills/finishing-a-change/SKILL.md
+++ b/.claude/skills/finishing-a-change/SKILL.md
@@ -50,7 +50,11 @@ Repository-specific checks:
 - No `test_*.py` under `tests/optimum/optimum_correctness/` — those are `fire` scripts.
 - No test that is skipped as a placeholder.
 
-## 4. Report what you did not do
+## 4. State the size
+
+Report the diff as source lines and test lines. If the test diff dwarfs the source, or the source dwarfs what your one-line summary describes, cut scope before opening the PR.
+
+## 5. Report what you did not do
 
 State it explicitly, every time:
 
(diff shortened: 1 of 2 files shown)


rebel-jonghewk (approved):
lgtm

PR #1036, opened 2026-08-31
fix(moe): widen the tokens mask when the routing bias is wider

rebel-mhkang:
### 🚀 Summary of Changes

#1027 built the MoE tokens mask in the routing weights' dtype. That is too narrow when the router has an fp32 `e_score_correction_bias` over bf16 scores (A.X-K2): the compiler folds the topk/gather/normalize/scatter routing chain into one fused routing op, and that op's output follows the wider of (scores, bias). The mask multiply then has an fp32 left operand and a bf16 right one, and the type checker refuses it:

```
data types float32 and bfloat16 do not match in BroadcastRel
```

`_routing_mask_dtype` mirrors that widening rule when choosing the mask dtype. A wider mask is always safe — torch promotes the product and the cast is inserted for us — and models with no bias, or a bias no wider than the weights, keep the narrow mask #1027 asked for.

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. `uv run --no-sync pytest tests/native/model_executor/layers/fused_moe/` → 30 passed. `TestRoutingMaskDtype` drives the real `RBLNMoERunner.forward` on the DP=1 path and asserts the dtype of the weights that reach the quant method; it is red with the helper reverted to #1027's rule.
2. Compile `skt/A.X-K2-NVFP4`: the `BroadcastRel` failure is gone and the build reaches codegen.

---

### 💬 Notes

The dtype rule is duplicated from the compiler's type relation for the fused routing op; the comment states the invariant so the coupling is findable. Fixing it on the compiler side instead — having the routing rewrite cast back to the dtype of the `scatter_elements` it replaces — would let this be reverted.

Affects the vLLM-native path only.

### DIFF
--- a/tests/native/model_executor/layers/fused_moe/test_moe_runner.py
+++ b/tests/native/model_executor/layers/fused_moe/test_moe_runner.py
@@ -23,12 +23,10 @@ from vllm.model_executor.layers.fused_moe.runner.moe_runner import MoERunner
 
 from vllm_rbln.model_executor.layers.fused_moe.runner.moe_runner import (
     RBLNMoERunner,
+    _routing_mask_dtype,
 )
 from vllm_rbln.model_executor.layers.fused_moe.runner.moe_runner import (
     _apply_grouped_topk_torch as grouped_topk,
-)
-from vllm_rbln.model_executor.layers.fused_moe.runner.moe_runner import (
-    _routing_mask_dtype,
 )
 
 
--- a/tests/native/model_executor/layers/fused_moe/test_moe_runner.py
+++ b/tests/native/model_executor/layers/fused_moe/test_moe_runner.py
@@ -17,14 +17,15 @@
 # expert without one), keep the top `topk_group` groups, then pick `top_k`
 # within them. Pure torch, hand-verifiable inputs.
 
+from types import SimpleNamespace
+
 import torch
 from vllm.model_executor.custom_op import maybe_get_oot_by_class
 from vllm.model_executor.layers.fused_moe.runner.moe_runner import MoERunner
 
-from vllm_rbln.model_executor.layers.fused_moe.runner.moe_runner import (
-    RBLNMoERunner,
-    _routing_mask_dtype,
-)
+import vllm_rbln.model_executor.layers.fused_moe.utils as fused_moe_utils
+from vllm_rbln.model_executor.layers.fused_moe.runner import moe_runner
+from vllm_rbln.model_executor.layers.fused_moe.runner.moe_runner import RBLNMoERunner
 from vllm_rbln.model_executor.layers.fused_moe.runner.moe_runner import (
     _apply_grouped_topk_torch as grouped_topk,
 )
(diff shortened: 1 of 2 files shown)


vllm_rbln/model_executor/layers/fused_moe/runner/moe_runner.py
```diff
@@ -28,6 +28,34 @@
...
+    always safe -- torch promotes the product and the relay frontend inserts the
+    cast -- and it is also what the mask was before it took a dtype at all.
+
+    Everything else keeps the narrow mask #1027 asked for: with no bias, or a
+    bias no wider than the weights, this is the weights' own dtype.
+    """
+    dtype = routing_weights.dtype
+    if e_score_correction_bias is None:
+        return dtype
+    if e_score_correction_bias.dtype.itemsize > dtype.itemsize:
```
rebel-kblee, line 50: This helper widens whenever the router has a bias attribute, even when the code never adds it. In the `scoring_func == "softmax"` non-grouped branch we never add the bias. The fused op stays bf16 there, but the mask becomes fp32

tests/native/model_executor/layers/fused_moe/test_moe_runner.py
```diff
@@ -227,6 +228,39 @@ def test_softmax_post_norm_sums_to_one_while_pre_norm_leaks_mass(self):
         assert pre.sum().item() < 1.0
 
 
+class TestRoutingMaskDtype:
```
rebel-kblee: Could this test code actually trigger the error we encountered? It seems the testing for the modified code in `moe_runner.py` below should be included.

vllm_rbln/model_executor/layers/fused_moe/runner/moe_runner.py
```diff
@@ -28,6 +28,34 @@
...
+    So the mask multiplied into those weights has to be that same wider dtype.
+    Narrower is the bug this exists to avoid: the fused multiply then has an fp32
+    left operand and a bf16 right one, and relay's type checker refuses it
+    ("data types float32 and bfloat16 do not match in BroadcastRel"). Wider is
+    always safe -- torch promotes the product and the relay frontend inserts the
+    cast -- and it is also what the mask was before it took a dtype at all.
+
+    Everything else keeps the narrow mask #1027 asked for: with no bias, or a
+    bias no wider than the weights, this is the weights' own dtype.
+    """
```
rebel-cjlee: I think it is too verbose. Also it's the first place in the repo that names a compiler pass and a rebel_compiler source path, and this repo is public. Can you check it?

vllm_rbln/model_executor/layers/fused_moe/runner/moe_runner.py
rebel-cjlee, line 50: `itemsize` can't tell fp16 from bf16. Is that a combination we don't need to worry about here?

rebel-mhkang: Confirmed, and thank you — the helper keys off the *attribute*, not off whether the routing chain actually consumed the bias. I drove the DP=1 `forward` with bf16 logits and an fp32 bias and captured the dtype the quant method receives:

| `scoring_func` | `num_expert_group` | `router_logits` reaching `apply()` |
| --- | --- | --- |
| sigmoid | 2 | fp32 |
| sigmoid | None | fp32 |
| softmax | 2 | fp32 |
| softmax | **None** | fp32 — bias never added |
| None | **None** | fp32 — bias never added |

So exactly as you say: in the non-grouped softmax branch (and the plain-topk `else` branch) the bias never enters the scores, the fused op stays bf16, and the mask widens anyway. It is not the reported failure mirrored — the multiply still typechecks, because torch promotes the product and the narrow side gets the cast — but it hands the quant method a needlessly wide fp32 `[E, t]`, which is the thing #1027 removed.

Reachability today: of the models that reach this runner (qwen2/qwen3-moe, gpt-oss, deepseek-v2, minimax-m2, A.X-K2), the ones with a bias are deepseek-v2 and A.X-K2 — both sigmoid + grouped, so the bias is consumed — and minimax-m2, which is non-grouped but whose RBLN `forward` feeds fp32 router logits, so the mask is fp32 either way. No configuration gets the spurious widening today, but it is a trap sitting one config away.

The part I would rather fix is underneath it: that branch drops the bias from routing altogether, while upstream's `FusedTopKBiasRouter` adds it for softmax too (`scores_for_choice = scores + e_score_correction_bias`). A model in that shape is mis-routed before any dtype question arises — ernie4.5-moe and longcat-flash are that shape upstream (bias, no `num_expert_group`). That is a routing change rather than a dtype one, so I did not fold it into this PR.

Which would you prefer?

1. Gate the widening on the branch that actually adds the bias (a local at the two call sites) — makes the helper honest, leaves the routing gap.
2. Add the bias in the non-grouped softmax branch to match upstream — makes the widening correct as written, but changes routing behaviour.
3. Both, in a follow-up.

My preference is to keep this PR as the dtype fix and open a follow-up issue for 2, since 1 becomes dead code the moment 2 lands.

tests/native/model_executor/layers/fused_moe/test_moe_runner.py
rebel-mhkang: You are right — those cases only exercised the helper, so they proved nothing about the multiply the bug is in. Replaced in d8abd9a9 with a test that builds an `RBLNMoERunner` and drives the real `forward` on the DP=1 path, asserting the dtype of the weights that reach `quant_method.apply` — the tensor the mask is actually multiplied into. Verified red with the helper reverted to #1027's weights-dtype-only rule:

```
E       assert torch.bfloat16 is torch.float32
```

What it cannot do is reproduce the `BroadcastRel` error itself, since no compilation happens in that lane; that stays the A.X-K2 compile check in the PR description. What it pins is the dtype the error is a consequence of.

vllm_rbln/model_executor/layers/fused_moe/runner/moe_runner.py
rebel-mhkang: Cut down in d8abd9a9: the docstring is now a short comment, and the pass name and the `rebel_compiler` source path are gone. I removed the same two names from the PR description as well, since that is just as public. What is left is the invariant (the compiler folds the routing chain and the fused op's output follows the wider of scores and bias, so the mask has to follow the same rule) with no reference to where the compiler implements it.

vllm_rbln/model_executor/layers/fused_moe/runner/moe_runner.py
rebel-mhkang: It never reaches that comparison here. Every model on this path creates the bias as `nn.Parameter(torch.empty(..., dtype=torch.float32))`, and the only narrowing cast upstream applies to it is ROCm-gated, so the bias side is always fp32 and only the routing-weights side varies (fp32 / bf16 / fp16). An fp16-vs-bf16 compare would need both operands 16-bit in different types, which nothing in the reachable set produces. And fp16 is forced on this path only for `enforce_eager`, where nothing is fused and the rule does not apply at all.

If you would rather not lean on that, `torch.promote_types(routing_weights.dtype, e_score_correction_bias.dtype)` is the exact rule in one line — it returns fp32 for the fp16/bf16 pair, which is what torch's own `add` would produce for those two anyway, so it agrees with the compiler by construction rather than by argument. Happy to push that instead if you prefer it.

rebel-cjlee: Make sense. Fine as is then.

rebel-mhkang: Went with option 1 in 0932cebc. `_routing_mask_dtype` now takes `scoring_func` and `use_grouped_topk` and only consults the bias when the routing actually adds it, so the non-grouped softmax and plain-topk branches keep the narrow mask. The condition sits in the helper rather than at the two call sites, so the "which branches add the bias" rule is stated once instead of three times.

Added a test for that branch: bf16 logits + fp32 bias with `scoring_func="softmax"` and no `num_expert_group` must reach the quant method as bf16. It is red without the gate (`assert torch.float32 is torch.bfloat16`). Full directory: 31 passed.

The routing gap itself — those branches never adding the bias, where upstream does — is untouched and I will open a follow-up issue for it.

rebel-kblee (approved):

rebel-cjlee (approved):

PR #1042, opened 2026-09-01
perf(sampler): cut eager ops around the n-gram rejection-sample op

rebel-eunik:
## 🚀 Summary of Changes

`RBLNRejectionSampler.forward` wrapped `rbln::rejection_sample` in a chain of eager ops. Those tensors live on the NPU, so this removes per-op dispatch.

### 1. Temperature and softmax moved behind `rejection_sample`

`forward` applied both itself; they now sit inside `rejection_sample` (argument `target_logits`), eager in the torch impl as before and folded into the compiled graph in the NPU impl — whose `apply_sampling_constraints()` is consequently a
no-op, kept for shape parity. 
Dropping the float32 widening also lets the graph softmax in the model dtype.

### 2. Output and bonus placement moved into the compiled graph

The `[B, K+1]` output was assembled by boolean-mask indexing and a scatter, so nine`nonzero` calls ran on data the op had just produced. The graph now returns it finished, using only fixed shapes so no index is computed and gathered

### ~~3. Target rows: gather → slice~~ (reverted)

~~Gathered logits hold **target** rows (a draft was proposed there) and **bonus** rows (the position after a request's last draft), interleaved per request:~~

```
num_draft_tokens      : [3, 0, 2, 0, 1]
logits_indices        : [0, 1, 2, 3, 103, 104, 105, 106, 206, 207, 208]
target_logits_indices : [0, 1, 2,       5, 6,           9]
bonus_logits_indices  : [         3, 4,       7, 8,          10]
```

~~The two arrays partition the 11 gathered rows, so the runner can permute `logits_indices` to put the target rows first:~~

```
logits_indices        : [0, 1, 2, 104, 105, 207, | 3, 103, 106, 206, 208]
                         └─── target: logits[:6] ───┘ └─── bonus ───┘
```

~~The sampler then slices instead of gathering. Only the order changes.~~

- ~~**ngram and suffix only.** `sample_hidden_states` is gathered in the same order and eagle/medusa read it to draft the next round, so a method that reads hidden states must not join the allowlist in `spec_logits_target_first()`.~~
- ~~**The index arrays stay interleaved.** `_get_logprobs_tensors()` scatters by them and indexes in sampled-token order; permuting them would silently return logprobs for the wrong rows. Only `logits_indices` moves.~~
- ~~**The predicate takes the config, not the sampler**, since `_calc_spec_decode_metadata()` runs on every pipeline-parallel rank.~~

### 4. Two smaller changes

- **Warm-up dtype (bug fix).** It built float32 logits while the real path feeds
  the model dtype, and the graph is guarded on it, so the first decode recompiled.
- **Spec-decode metadata built on the device**, their only consumer being the graph.

---

## 📌 Related Issues / Tickets

---

## ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

## 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

## 📸 Screenshots / Logs (if applicable)

---

## 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -346,16 +363,22 @@ def rejection_sample(
...
         bonus_token_ids: torch.Tensor,
         sampling_metadata: SamplingMetadata,
         synthetic_mode: bool = False,
         synthetic_conditional_rates: torch.Tensor | None = None,
     ) -> torch.Tensor:
+        target_logits = self.apply_sampling_constraints(
+            target_logits,
+            cu_num_draft_tokens,
+            sampling_metadata,
+        )
```
rebel-eunji, line 349: This doesn't seem to do anything. Is it just for consistency with the other implementations?

rebel-eunik: Yes, that's correct. I kept `apply_sampling_constraints` just for consistency. It only runs assertions on the conditions and doesn't perform any other operations.

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -454,138 +440,114 @@ def rejection_sample(
...
     draft_token_ids: torch.Tensor,
-    target_probs: torch.Tensor,
+    target_logits: torch.Tensor,
     cu_num_draft_tokens: torch.Tensor,
     top_k: torch.Tensor | None,
     top_p: torch.Tensor | None,
-) -> tuple[torch.Tensor, torch.Tensor]:
-    return torch.ops.rbln.rejection_sample(
+    temperature: torch.Tensor,
+    draft_per_batch: torch.Tensor,
```
rebel-eunji, line 486: Both draft_token_ids and draft_per_batch are required? I think draft_per_batch can be composed with draft_token_ids and cu_num_draft_tokens.

rebel-eunik: That's right. However I actually tried moving it into the graph, but the latency increased slightly on CR03.

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -454,138 +440,114 @@ def rejection_sample(
...
+    head = torch.where(
+        (positions == num_accepted.unsqueeze(1)) & partially_accepted.unsqueeze(1),
+        recovered_token_ids,
+        head,
+    )
+    out = torch.cat([head, out[:, max_spec_len:]], dim=1)
+
+    bonus = bonus_token_ids.squeeze(-1).to(dtype=out.dtype)
+    bonus_mask = all_accepted.unsqueeze(1) & (positions_k1 == counts.unsqueeze(1))
+    return torch.where(bonus_mask, bonus.unsqueeze(1), out)
```
rebel-eunji: squeeze and unsqueeze the bonus_token_ids tensor?

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -454,138 +440,114 @@ def rejection_sample(
...
         #                                       tokens (in [0, num_draft_tokens[i]]).
         # ------------------------------------------------------------------
-        recovered_token_ids, num_accepted = self._compiled_rejection_sample(
-            reshaped_draft_token_ids,
-            reshaped_target_probs,
-            cu_num_draft_tokens.to(device),
-            top_k,
-            top_p,
+        # NOTE(RBLN): [B*K] temperature for the divide inside the graph. Padding
+        # and greedy rows must carry 1.0 -- a 0 would divide by zero.
```
rebel-eunji: Some comments seem out of sync with the current implementation. Could you update them to match the code, and reword a bit for readability?
(and Keep the notes of npu primitive in another code space please.)

vllm_rbln/v1/sample/rbln_rejection_sampler.py
rebel-eunji: Fair enough. Then just please add comments for the logics and the role of tensors.

rebel-eunji: e.g. specifying the shape of tensor is a idea...maybe

rebel-daeyang: I'll take a deeper look at moving this into the graph in a follow-up PR. For now, a comment as @rebel-eunji suggested would be enough.

rebel-eunji (approved):

rebel-daeyang (approved):

rebel-jonghewk (approved):
LGTM!  @rebel-eunik Can we merge this first? I think @rebel-daeyang has a follow up PR.

rebel-cjlee (approved):

PR #1045, opened 2026-09-02
feature(nixl): derive the KV cache's split axis and pair only within it

rebel-ykchoi:
#### Summary

Every head band in this connector reads a chiplet area as a set of KV heads over a
whole block. The compiler can instead cut the cache on the in-block token axis, which
makes an area every head over part of a block — and nothing the connector can reach
reports which happened: nixl-rbln says which slice each area carries, never why it
was cut that way.

The counts alone cannot answer it either. `4 areas / 4 slices` is head tiling on a
GQA model and a context cut on a sparse-MLA one; the same pair of numbers means two
different things. What separates them is the head count the registration already
holds.

The failure profile is what makes this worth a guard. Equal TP pairs positionally —
area k with area k — so the bytes are right and a symmetric smoke test passes.
Unequal TP switches to head-band matching, and a band computed from the configured
KV-head count names a range no area holds: registration, handshake and descriptor
construction all succeed, and only the bytes are wrong.

##### A. The axis is derived at registration, not asked for

A head axis of extent one cannot be cut, so the compiler replicates instead and more
than one slice there has no other explanation. That is an exclusion rather than a
heuristic, and it needs only what the registration loop already has: the layer spec's
`num_kv_heads`, and the slice ids nixl-rbln returns.

The derivation is deliberately narrow. Above one head both axes explain the same
count and nothing here can tell them apart, so an axis that is not proven stays
`HEAD`. That leaves one case knowingly uncovered — a cut turned on by a compiler flag
for a model with more than one KV head per rank stays classified as `HEAD`, and
unequal TP against it is still silently wrong. The class docstring records it as
unsupported rather than pretending the code can see it.

Regions that disagree are refused at registration: one advertised geometry cannot
describe two.

##### B. The axis is advertised, because the receiver cannot recompute it

What a peer receives is one engine-wide slice count, while the derivation reads a
count per region. So the axis joins the geometry in the RBLN metadata extension and
`RBLN_NIXL_CONNECTOR_VERSION` moves 3 → 4. That constant folds into the compatibility
hash, so a peer on the older schema fails the handshake cleanly — **P and D have to
be deployed together.**

##### C. Refusals follow the axis rather than the model

A peer that cut a different axis is refused: the area counts can agree while an area
means something else on each side. A context cut across unequal TP is refused for the
reason above, and the message names the escape hatch.

Equal TP stays open. Area k pairs with area k, no band is consulted, and both sides
cut the same way — refusing it would turn away the one shape that does work.

The existing MLA refusal is left alone. Moving its gate onto the axis would drop it
for a full-MLA model, whose latent is replicated rather than cut and therefore
classifies as `HEAD`; the two guards answer different questions and both are needed.

#### Testing

- **Unit** — the derivation on the three geometries that separate it (`num_kv_heads=1`
  with one slice, `num_kv_heads=1` with several, `num_kv_heads=8` with the same
  several), an engine whose regions disagree, the advertisement round-trip, and each
  refusal (`test_rbln_nixl_base_worker.py`, `test_rbln_nixl_handshake.py`,
  `test_rbln_nixl_metadata.py`).
- The refusal is also reached through `_validate_remote_agent_handshake` rather than
  by calling the guard directly, so removing that call is caught.
- Cases are chosen so the input decides the answer: `num_kv_heads=8, g=4` and
  `num_kv_heads=1, g=4` are the same geometry with opposite verdicts, and the two
  `HEAD` cases start from `NON_HEAD` so the assertion cannot pass on the field's
  initial value.
- `pytest tests/native/distributed` green; `pre-commit` clean on every changed file.

#### Risk / compatibility

- **Behaviour is unchanged wherever the axis derives to `HEAD`**, which is every
  geometry observed so far except a sparse-MLA model's. The derivation adds no new
  refusal for head-tiled caches.
- **The schema version moves**, so a peer speaking version 3 no longer handshakes.
  This is the intended effect: without it the two would pair and the axis field would
  be read as absent.
- The new refusals fire only where the descriptors were already wrong, so the
  configurations they turn away could not have worked.

---

* Resolves rebellions-sw/fsw-inference#551 — registering and transferring a KV
  cache the compiler cut on the chiplet-wise context-parallel axis.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

2. `uv run --no-sync pytest tests/native/distributed`
3. Expected: all green.
4. Edge cases covered: a single-head region replicated across chiplets (no cut); a
   single-head region cut into several slices; a multi-head region over the same
   slice count, which must not be read as a cut; an engine whose regions disagree on
   the axis; a peer advertising the other axis; a context cut against a peer with a
   different TP degree; host staging, which has no chiplet areas and keeps the
   default.

---

### 📋 Checklist

---

### 💬 Notes

- This branches from `feat/nixl-push-direction` rather than `dev`: the guard is added
  beside the direction-free checks that branch introduces, and the write path is
  where a wrong descriptor writes into a peer rather than reading from one.
- The version bump takes 4. Any other branch holding 4 has to move before both land,
  or the two would present the same number and the handshake would read them as
  compatible.


vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
```diff
@@ -495,12 +504,31 @@ def _register_kv_caches_impl(self, kv_caches: dict[str, torch.Tensor]) -> None:
...
+        region_non_head = {
+            heads == 1 and len(set(xfer.slice_ids[r * areas : (r + 1) * areas])) > 1
+            for r, heads in enumerate(logical_kv_heads)
+        }
+        if len(region_non_head) > 1:
+            raise RuntimeError(
+                "RBLN NIXL (D2D): this engine's KV regions were not all cut on "
+                "the same axis, which one advertised geometry cannot describe. "
+                f"Head counts per logical region: {logical_kv_heads}."
+            )
```
rebel-jongwonlee, line 522: ### The refusal prints half of what decided it

The discriminator is two conditions — `heads == 1` **and** more than one distinct slice
id in that region's areas — but the message reports only the first:

```
f"Head counts per logical region: {logical_kv_heads}."
```

When the head counts are what differ, that reads well. When they are not, it does not:
an engine whose regions are all single-head, some cut and some replicated, raises with
`Head counts per logical region: [1, 1, 1, 1]` and nothing that says which regions
disagreed or why. That shape is not far-fetched — a model with two attention flavours
registers layers under different specs, and gpt-oss-120b is 18 full plus 18
sliding-window.

Carrying the per-region slice counts alongside would make it diagnosable, e.g.
`[(heads, len(set(slice_ids[r*areas:(r+1)*areas]))) for r, heads in ...]`, or just the
boolean each region contributed. Non-blocking — the guard itself is right.

rebel-jongwonlee (approved):
Read the whole change. The shape of the argument is what makes this one convincing: the
axis is *excluded* rather than guessed, and where exclusion cannot reach, the code says
so instead of picking. `an axis that is not proven stays HEAD` is the right default, and
recording the residual case in the class docstring rather than papering over it is the
right way to leave it.

Five things I checked independently rather than taking from the description:

- **The head count read is the layer's, not a group's.** The loop unwraps
  `UniformTypeKVCacheSpecs` before `num_kv_heads` is read, which matters because that
  wrapper's page size is the sum over the group — a wrapped read would have made the
  derivation answer for the wrong thing.
- **The `slice_ids` slicing matches the layout.** `[r*areas : (r+1)*areas]` assumes
  logical-region-major over areas, and the fixtures bear that out: 2 MLA regions over 4
  areas give `[0,1,2,3]*2`, 4 FA regions give `[0,1,2,3]*4`.
- **The equal-TP exemption is exactly the shape that works.** `_is_head_matched_peer` is
  `tp_ratio != 1`, so the refusal fires only where bands are actually consulted.
  Positional pairing needs no band and both sides cut the same way, so refusing it would
  have turned away the one working configuration.
- **Leaving `_check_mla_constraints` alone is right, and for the reason given.** A
  full-MLA latent is replicated rather than cut, so its areas share a slice id and it
  classifies as `HEAD` — folding the MLA gate onto the axis would have dropped that
  guard for precisely the models it exists for. Sparse-MLA is the one that reaches
  `NON_HEAD`. Two questions, two guards.
- **Version 4 folds into the hash,** so an older peer fails the handshake instead of
  pairing on a field it does not send, and the `HEAD` default keeps a v2/v3 blob meaning
  what it used to.

The tests are worth calling out. Two false-positive shapes are closed deliberately
rather than incidentally: the head case starts the field at `NON_HEAD` so the assertion
cannot pass with the derivation never running, and
`test_a_peer_on_another_axis_is_refused_through_the_entry_point` exists because deleting
the call from `_validate_remote_agent_handshake` kept the suite green. And the two
geometries that separate the derivation carry *identical* area and slice counts, so the
head count is genuinely the only input deciding the answer.

One non-blocking suggestion inline, on the refusal message.

**Deployment note, not a change request:** this is the second consecutive schema break —
#1018 took `RBLN_NIXL_CONNECTOR_VERSION` 2 → 3 and has landed, and this takes it 3 → 4.
Both require P and D to be deployed together, and anyone who has already rolled out 3
will need a second coordinated restart. Worth bundling into one release note rather than
leaving it in two PR bodies.

Approving.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
rebel-ykchoi: Fixed in a follow-up commit rather than an amend, so the approved commit stays as reviewed.

Both counts a region contributed are now what the message carries:

```
(head count, distinct slice ids) per logical region: [(1, 1), (1, 2), (1, 1), (1, 2)]
```

The pair is derived once per region and the boolean the guard tests reads that same pair, so the condition and its diagnostic cannot drift apart.

On the example: gpt-oss-120b declares 8 KV heads, so `heads == 1` holds for none of its regions and the two attention flavours never reach this shape — a spec mix and a head count of one are different things. The message was worth fixing on its own terms.

The branch is also rebased onto current `dev`, which is why the push is a force.

rebel-jongwonlee (approved):
Re-approved on `7b520062`.

One correction, since it changes what you are waiting on: **the rebase did not dismiss
my approval.** There is no `review_dismissed` event on the timeline, and the API still
counts it — `latestReviews` returns `rebel-jongwonlee APPROVED`, and my earlier
approvals on #1011 and #1018 survived force-pushes the same way, so this repo does not
appear to dismiss on new commits. `REVIEW_REQUIRED` is coming from the other requested
reviewers, not from me: eight were requested on 09-02/03 and only mine has landed.

Re-approving is still worth doing for a different reason — my approval sat on
`14a06058`, and `cb57f5c9` (the refusal-message fix) came after it, so the stamp did not
formally cover the follow-up. Now it does.

I checked the rebase rather than taking it on trust:

* The added lines are byte-identical to what I read — diffing `057c2da8..cb57f5c9`
  against `<new merge-base>..7b520062` comes back empty.
* None of the twelve commits crossed touch the connector:
  `git log 057c2da8..56db188d -- vllm_rbln/distributed/kv_transfer/` is empty.
* CI green on build 956.

Unrelated to this PR, but it rode in on that range: #1069 removed
`VLLM_RBLN_SORT_BATCH` and moved the decision to
`is_cr13() or VLLM_RBLN_BATCH_ATTN_OPT`, resolved once in `__init__`. Anything still
exporting the old variable is now setting something nothing reads — though for a
deployment that already sets `VLLM_RBLN_BATCH_ATTN_OPT=1`, the resolved behaviour is
unchanged. Flagging it for whoever maintains launch scripts rather than as anything this
PR needs to do.

rebel-cjlee (approved):

PR #1046, opened 2026-09-02
fix(metrics): fold the PP hand-off into the pass on non-last ranks

rebel-yskim:
### 🚀 Summary of Changes

On a non-last PP rank the metrics pass ends right after the asynchronous
dispatch, so `MODEL + SAMPLE` (and `E2E`) reported the dispatch cost only —
the stage's actual time stayed hidden from the report. The last rank never
had this problem because its pass contains the sampled-token read.

This change surfaces that hidden part. The hand-off is where a non-last
rank's stage time becomes observable: the send completes only after the
stage's compute has produced the tensors. So the worker's send becomes a
named seam (`RBLNWorker._send_handoff`), and the metrics patch keeps the pass
open until it and folds its wall into the graph sum — the same shape as the
existing spec-decode postprocess region. The metrics module stays a pure
observer: it times what the worker already does and never adds device work
of its own, so runs behave identically with the metric on or off.

Everything is registry-gated on `VLLM_RBLN_METRICS`; without it the worker
seam is the plain send it replaced.

**Model path**: vLLM-native only.

---

* Related to # 

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `uv run --no-sync pytest tests/native/patches/test_metrics.py` — the
   new `TestPassBoundary` hand-off tests fail without the change.
2. Run any `pp > 1` case with `VLLM_RBLN_METRICS=1` and read the per-rank
   report: on non-last ranks `MODEL + SAMPLE` now carries the pass through
   the hand-off instead of stopping at dispatch; the last rank's sections are
   unchanged, and `E2E − MODEL + SAMPLE` keeps its meaning (host overhead
   around the graphs).
3. Edge cases: a pass that does not return `IntermediateTensors` (last rank,
   `pp == 1`, empty steps) closes where it always did; a hand-off that raises
   leaves the pass open, like any raising pass.

---

### 📋 Checklist

---

### 💬 Notes

On a loaded pipeline the hand-off wall also contains the wait for the next
rank to arrive, so on non-last ranks the section reads as "what this rank's
step actually spent" rather than device time alone; with an idle downstream
the two coincide. Reviewers may also want to eye `_send_handoff`: it exists
so the metrics patch has a wrappable seam, the same reason the postprocess
profiler region exists.

### DIFF
--- a/vllm_rbln/patches/metrics.py
+++ b/vllm_rbln/patches/metrics.py
@@ -49,6 +49,7 @@ from dataclasses import dataclass, field
 from enum import Enum
 
 import numpy as np
+from vllm.sequence import IntermediateTensors
 from vllm.v1.utils import record_function_or_nullcontext
 
 from vllm_rbln import envs
@@ -282,9 +283,9 @@ def execute_model(self, *args, **kwargs):
     _ACTIVE_CTX = ctx
     ctx.start_pass()
     output = _execute_model(self, *args, **kwargs)
-    # The engine calls sample_tokens() only if execute_model() returned None, so a
-    # non-None return means nobody else will end this pass.
-    if output is not None:
+    # A None pass is closed by sample_tokens, an IntermediateTensors pass by
+    # send_handoff below; anything else nobody would end.
+    if output is not None and not isinstance(output, IntermediateTensors):
         ctx.end_pass()
     return output
 
@@ -357,6 +358,21 @@ def determine_batch_execution_and_padding(self, *args, **kwargs):
     # A no-op on the dummy run, which reaches this with no pass open.
     _ctx(self).mark_phase(phase)
     return result
+
+
+_send_handoff = RBLNWorker._send_handoff
+
+
+@functools.wraps(_send_handoff)
+def send_handoff(self, *args, **kwargs):
+    # The send waits on this stage's compute, so its wall carries the stage's time.
+    ctx = _ctx(self.model_runner)
+    start = time.perf_counter()
+    try:
+        return _send_handoff(self, *args, **kwargs)
+    finally:
+        ctx.add_graph_time(time.perf_counter() - start)
+        ctx.end_pass()
 
 
 @functools.wraps(_shutdown)
@@ -433,6 +449,12 @@ def _register_patches() -> None:
             determine_batch_execution_and_padding,
             "Reads the pass phase where it is decided. The padded token dimension is a "
             "local of execute_model and is not observable from any wrapped callable.",
+        ),
+        (
+            f"{_WORKER}._send_handoff",
+            send_handoff,
+            "Times the hand-off and closes its pass; left outside the pass, "
+            "the send hides a non-last rank's stage time.",
         ),
         (
             f"{_WORKER}.shutdown",
--- a/vllm_rbln/patches/metrics.py
+++ b/vllm_rbln/patches/metrics.py
@@ -366,13 +366,13 @@ _send_handoff = RBLNWorker._send_handoff
 @functools.wraps(_send_handoff)
 def send_handoff(self, *args, **kwargs):
     # The send waits on this stage's compute, so its wall carries the stage's time.
+    # No try/finally: a raising hand-off leaves the pass open, like any raising pass.
     ctx = _ctx(self.model_runner)
     start = time.perf_counter()
-    try:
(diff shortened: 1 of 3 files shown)


vllm_rbln/v1/worker/rbln_worker.py
rebel-cjlee, line 1: The method exists only so the metrics patch has something to wrap. I'm not sure we should change production code for a debug feature, and AGENTS.md also says we should not add a helper that is called only once.

Could the patch wrap `RBLNWorker.execute_model` instead? 
@rebel-jonghewk What do you think? If you think the current shape is fine, please feel free to ignore my opinion and keep it as it is.

rebel-jonghewk (approved):
lgtm

vllm_rbln/patches/metrics.py
```diff
@@ -359,6 +360,21 @@ def determine_batch_execution_and_padding(self, *args, **kwargs):
...
+@functools.wraps(_send_handoff)
+def send_handoff(self, *args, **kwargs):
+    # The send waits on this stage's compute, so its wall carries the stage's time.
+    ctx = _ctx(self.model_runner)
+    start = time.perf_counter()
+    try:
+        return _send_handoff(self, *args, **kwargs)
+    finally:
+        ctx.add_graph_time(time.perf_counter() - start)
+        ctx.end_pass()
```
rebel-cjlee: By this point the phase and the graph time are already set, so if the send raises, doesn't `end_pass()` pass the guard and record the crash as a normal sample? The PR description says the pass stays open, so maybe just drop the try/finally?

vllm_rbln/patches/metrics.py
```diff
@@ -359,6 +360,21 @@ def determine_batch_execution_and_padding(self, *args, **kwargs):
     return result
 
 
+_send_handoff = RBLNWorker._send_handoff
+
+
+@functools.wraps(_send_handoff)
+def send_handoff(self, *args, **kwargs):
+    # The send waits on this stage's compute, so its wall carries the stage's time.
```
rebel-cjlee, line 368: If the send blocks until the next rank posts its recv, then in steady state every rank's number converges to the bottleneck state, doesn't it? That would hide which stage is actually slow. Would a separate HANDOFF section work better here? 
It's just my opinion, not blocking. So, feel free to keep it as it is.

rebel-yskim: `Send` does not block until the receiver alone — it completes at max(`compute`, `receiver arrival`) + `transfer`. Also, this PR does not affect any operation: current dev already blocks state in `send_tensor_dict` (gloo ccl), and this PR only makes that already-spent time visible in the report. 

> Would a separate HANDOFF section work better here?

A separate HANDOFF section would undo the fix: MODEL + SAMPLE returns to dispatch cost only, while the HANDOFF number carries the same max(compute, receiver) mixture, so attribution doesn't improve.

vllm_rbln/patches/metrics.py
rebel-yskim: You're right. applied: [f96dff7](https://github.com/RBLN-SW/vllm-rbln/pull/1046/commits/f96dff734e13225af9ef254ad40a743292de65a5)

rebel-cjlee (approved):

vllm_rbln/v1/worker/rbln_worker.py
rebel-cjlee: @rebel-yskim If you think it is worth it, could you check with @rebel-jonghewk and decide what to do?
I'm fine with whatever you two land on.

rebel-yskim: @rebel-cjlee We've discussed: 
There are some fundamental reason why we cannot directly access device time (Device + Sample), and we indirectly get the time in vllm-rbln side. This PR still contains overhead in device time: max(device, wait_for_send)+ send, and if we use `RBLNWorker.execute_model`, it also adds wait_for_recv + recv (there could be an overhead in unfortunate conditions, sometimes...).

SInce `_send_handoff` does not touch any functional things, I think current changes would be necessary.
We still know that both design (this PR /  using`RBLNWorker.execute_model`) does not fix fundamental problems in tracking device time, so I'll keep it in mind to fix in better way. Thanks for the comment.

PR #1049, opened 2026-09-03
fix(spec-decode): keep the post-step draft fetch off the prefill path

rebel-wonsubkim:
`EngineCore.post_step` fetches the drafts from the worker that produced them so the scheduler can size and allocate the next verification step. It runs whenever spec decode is on and async scheduling is off — and on RBLN async scheduling is off for **every** PP run, because PP under async scheduling is not supported yet (`platform.py`).

Between **decode** steps that fetch is a true dependence: step N's drafts are what step N+1 verifies. Between **prefill** chunks there is none, and upstream's own consumer says so — `update_draft_token_ids` drops the value on arrival with `# Ignore draft tokens for prefill chunks`, because a chunk with no anchor token has nothing to verify. `post_step` only receives `model_executed`, so it cannot tell the two apart.

At `pipeline_parallel_size == 1` paying for it anyway is invisible: the engine has one batch in flight and already blocks each step. Under PP it is not. The fetch is a synchronous round-trip to `output_rank` — the last stage — so the reply cannot come until the chunk has traversed the whole pipeline, and it is issued right after `step_with_batch_queue` returns early to refill `batch_queue`. The engine never gets back to `schedule()`, `batch_queue` never reaches `pp_size`, and the pipeline runs about **one microbatch deep instead of `pipeline_parallel_size`** — every stage but one idle, waiting for a chunk that the engine is not allowed to schedule yet.

The cost is the round-trip, not the drafting: an ngram drafter, whose own work across a prefill-only run is negligible, regresses prefill TTFT several-fold on PP4 purely by being configured. Stage times are unchanged — only the interval between chunks moves. eagle3 cannot avoid this by configuration at all, since its drafter fills its KV in the prefill forward and therefore requires a speculative config on the prefill instance.

```
expected — the period is one stage, so pp_size chunks are in flight

  pp0  |[ c0 ][ c1 ][ c2 ][ c3 ][ c4 ]
  pp1  |      [ c0 ][ c1 ][ c2 ][ c3 ][ c4 ]
  pp2  |            [ c0 ][ c1 ][ c2 ][ c3 ][ c4 ]
  pp3  |                  [ c0 ][ c1 ][ c2 ][ c3 ][ c4 ]
        pp0 takes the next chunk as soon as it has handed one on


actual — the period is a whole traversal, so one chunk is in flight

  pp0  |[ c0 ]                  [ c1 ]                  [ c2 ]
  pp1  |      [ c0 ]                  [ c1 ]                  [ c2 ]
  pp2  |            [ c0 ]                  [ c1 ]                  [ c2 ]
  pp3  |                  [ c0 ]                  [ c1 ]
  eng  |>>>>>>>>>>>>>>>>>>>>>>>|                 |
        take_draft_token_ids() reply ───────────┘  c1 scheduled
        stage times are the same in both -- the next chunk just does not arrive
```

### The change

Skip the fetch while every running request is still mid-prefill:

```python
running = self.scheduler.running
if running and all(request.is_prefill_chunk for request in running):
    return
```

That is the condition the consumer already applies, moved ahead of the round-trip, so **not fetching and fetching-then-discarding are the same**. The step that schedules a request's last chunk already reports `is_prefill_chunk == False` (`_update_after_schedule` advances `num_computed_tokens` first), so the fetch resumes on exactly the step whose drafts the next one will verify. The decode path is untouched.

With the guard the pipeline fills to `pp_size` again and prefill TTFT returns to parity with a run that carries no speculative config at all — for both ngram and eagle3. Sanity generations (short greedy, and long-context needle retrieval) are unchanged.

Self-disabling: once PP is supported under async scheduling the `not async_scheduling` term is false and the guard is never reached, so this can stay or be removed.

### Tests

`tests/native/patches/test_engine_core.py` — 14 cases on a stub engine (no config, checkpoint or device):

- all-prefill skips; any non-prefill still fetches; empty `running` still fetches (`all([])` is True, so the emptiness check carries weight)
- the three upstream conditions stay untouched (no spec / async on / nothing ran)
- a `None` result is not forwarded
- **the regression in miniature** — a chunked-prefill run of 101 steps fetches **once** instead of on every step
- the guard is per-step, not sticky: a prefilling request does not suppress the fetch for a decode request that joins later

Measurements and the full write-up: rebellions-sw/fsw-inference#561

### DIFF
--- a/tests/native/patches/test_engine_core.py
+++ b/tests/native/patches/test_engine_core.py
@@ -1,0 +1,160 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""The post-step draft fetch stays off the prefill path.
+
+A stub engine only -- the guard reads `scheduler.running` and the three flags
+`post_step` already branches on, so no config, checkpoint or device is needed.
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+import pytest
+from vllm.v1.engine.core import EngineCore
+
+from vllm_rbln.patches.engine_core import patched_post_step
+
+
+class _Executor:
+    def __init__(self, drafts="drafts"):
+        self.drafts = drafts
+        self.calls = 0
+
+    def take_draft_token_ids(self):
+        self.calls += 1
+        return self.drafts
+
+
+class _Scheduler:
+    def __init__(self, prefill_flags):
+        self.running = [
+            SimpleNamespace(is_prefill_chunk=flag) for flag in prefill_flags
+        ]
+        self.updated = []
+
+    def update_draft_token_ids(self, draft_token_ids):
+        self.updated.append(draft_token_ids)
+
+
+def _engine(prefill_flags, *, spec=True, async_scheduling=False, drafts="drafts"):
+    return SimpleNamespace(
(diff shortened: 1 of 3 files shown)


rebel-cjlee (approved):

rebel-daehwan (approved):
I already commented above. LGTM!

PR #1050, opened 2026-09-03
feature(profiler): flush the RBLN profiler at stop_profile without torch

rebel-ywkim:
### 🚀 Summary of Changes

With `RBLN_PROFILER=1` alone, the RBLN device profiler wrote its trace only when the process exited cleanly. vLLM kills its workers during shutdown, so the trace was frequently lost. The workaround was to also turn on the torch profiler(`--profiler-config.profiler=torch`), whose stop path makes the RBLN trace come out — which forces a torch profiler into the loop just to get an RBLN trace.

This adds a `WorkerProfiler` implementation that starts and stops an RBLN profiler session through rebel-compiler's Python API:

- `_start()` → `rebel.profiler.start()`
- `_stop()` → `rebel.profiler.done()`, which writes the trace

`RBLNWorker.profile()` selects it when no torch profiler is configured but `RBLN_PROFILER=1` is set. `profiler_config.profiler == "torch"` keeps exactly the behaviour it has today — the torch-rbln kineto integration still owns the RBLN session there, so the two never both drive it.

Offline `LLM` only for now: `vllm serve` mounts `/start_profile` just when`profiler_config.profiler` is set, so on a server the RBLN profiler still needs `--profiler-config.profiler=torch` alongside `RBLN_PROFILER=1`.

`stop_profile` reaches the worker as a blocking `collective_rpc` with no timeout, so it returns only after every worker finished writing: the trace is on disk before teardown. `WorkerProfiler.shutdown()` additionally stops a still-running session, which covers a SIGTERM-initiated shutdown.

**Model path:** vLLM-native only (`v1/worker/rbln_worker.py`). The optimum path has its own `profile()` in `v1/worker/optimum_worker.py` and is untouched.

**Affected modules:** Worker.

---

* Resolves #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ❓ Other (`other`)

---

### 🧪 How to Test

**1. New path — no torch profiler.** Take
`examples/experimental/offline_inference_profile.py` and drop the
`profiler_config=` argument, then run:

```bash
RBLN_PROFILER=1 RBLN_PROFILER_DIR=/tmp/rbln-prof python <script>
```

Expected: `llm.stop_profile()` returns after the trace is written, and
`/tmp/rbln-prof/rbln_<timestamp>.pb` exists *before* the process exits. The
trace must cover only the `start_profile` / `stop_profile` window, not the
warm-up `generate()` the example runs first.

**2. Regression — the torch path must be unchanged.** Same example, unmodified
(`profiler_config={"profiler": "torch", ...}`):

```bash
RBLN_PROFILER=1 RBLN_PROFILER_DIR=/tmp/rbln-prof python \
  examples/experimental/offline_inference_profile.py
```

Expected: torch trace under `torch_profiler_dir` and the RBLN trace under
`RBLN_PROFILER_DIR`, exactly as on `dev`.

**3. Error path.** No `profiler_config`, `RBLN_PROFILER` unset.

Expected: `RuntimeError: Profiling is not enabled. Please set --profiler-config
to enable profiling, or RBLN_PROFILER=1 for the RBLN profiler alone. ...`

**4. Unit.**

```bash
uv run --no-sync pytest tests/native/v1/worker/test_rbln_worker.py -q
```

Expected: 82 passed. The one added test (`TestProfile`) pins the invariant that
is not visible in any single line: on the torch path `RBLNWorker` makes zero
calls into the RBLN profiler, so the two never both own the session. Confirmed
by injecting the violation (`elif` -> an unconditional branch), which turns the
assertion red.

Nothing else is asserted on this side, on purpose. Whether a trace is written,
where it lands, and that it lands at `stop_profile` rather than at process exit
are rebel-compiler behaviours that need a device; those were checked by hand on
CR13 — trace under `RBLN_PROFILER_DIR`, none in the CWD, written on the `done()`
call rather than at teardown.

---

### 📋 Checklist

---


vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -32,6 +32,7 @@
 
 import torch.distributed as dist
 import torch.nn as nn
+from rebel._C import profiler as rbln_profiler
```
rebel-jonghewk: @rebel-ywkim can we have another API instead of accessing via rebel._C ?

rebel-ywkim: I will re-expose functions under  `_C.profiler` as a public entry point in compiler so that here we can access via `rebel.profiler`. Please let me know If this direction looks okay.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -1244,6 +1257,9 @@ def profile(self, is_start: bool = True, profile_prefix: str | None = None):
                     logger.debug(
                         "Starting torch profiler with tarce name: %s", trace_name
                     )
+                elif profiler_type is None:
+                    self.profiler = RblnProfilerWrapper(self.profiler_config)
+                    logger.debug("Starting RBLN profiler on %s", rank_suffix)
```
rebel-cjlee: Two questions here.

First, could this take an explicit profiler type, like `profiler="rebel"`, instead of using `None`? Right now the branch runs because no profiler is set, so reading this line along it says "nobody asked for a profiler, so start the RBLN profiler". An explicit name would say what we mean. Then we add `"rebel"` to line 210 to accept new profiler type. One thing to check, upstream declares `ProfilerKind = Literal["torch", "cuda"]`, so we may need to patch ProfilerConfig as well (like `vllm_rbln/patches/profiler.py`).

Second, could the `is_activated()` check live inside this branch? Today the real switch for the RBLN profiler is `rbln_profiler.is_activated()`, but that check only appears in the guard 40 lines above. If the branch checks it here, it stands on its own and does not depend on the guard staying the way it is.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -1206,10 +1217,12 @@ def take_draft_token_ids(self) -> DraftTokenIds | None:
 
     def profile(self, is_start: bool = True, profile_prefix: str | None = None):
         # Check if profiling is enabled
-        if self.profiler_config is None or self.profiler_config.profiler is None:
+        if self.profiler_config is None or (
+            self.profiler_config.profiler is None and not rbln_profiler.is_activated()
+        ):
```
rebel-cjlee, line 1228: You are already changing this line, so maybe fix this too. `self.profiler_config is None` can never be true. Upstream declares it as `profiler_config: ProfilerConfig = Field(default_factory=ProfilerConfig)`, so it always has a value, and line 210 already uses it without a `None` check. I would drop the first half on the condition.

tests/native/v1/worker/test_rbln_worker.py
```diff
@@ -1175,3 +1176,31 @@ def test_materialize_runs_the_smallest_compiled_decode_bucket(self):
         )
         RBLNWorker._materialize_kv_cache(worker)
         assert ran == [(4, 1, False)]
+
+
+class TestProfile:
```
rebel-cjlee, line 1246: Good invariant to lock down. Can you add the path this PR added as well?

With `is_activated` returning true, call `profile(True)` and `profile(False)`, then check `calls == ["start", "done"]` and that `worker.profiler` is the RBLN wrapper. It needs no device. If you keep the current shape and use the default `ProfilerConfig`, the test also proves that `profiler_config` has a real default, so this new path can be reached at all. If you move to `profiler == "rebel"`, just set that instead.

vllm_rbln/v1/worker/rbln_worker.py
rebel-ywkim: Thank you for pointing this out, I will clean up unnecessary statements following upstream.

vllm_rbln/v1/worker/rbln_worker.py
rebel-ywkim: Thank you for the suggestion.
Actually I also tried something similar so that we can explicitly use a ProfilerKind like `"rebel"`, but It raised a question about where the patch has to land.

If I add ProfilerKind `"rebel"`, users would naturally expect a usage like `LLM(profiler_config={"profiler":"rebel"})`.
Also, as far as I understand conventional way of applying patch is to register it with @register_patch and then upstream trigger it on `load_general_plugins()` - on the offline  path, `EngineArgs.__post_init__` triggers it.

However, profiler_config is validated before EngineArgs() is even constructed (`LLM.__init__ -> _make_config -> ProfilerConfig()` ) so with the usual patch path I get a ValidationError.

What I think I can do instead is to apply patch earlier than the plugin path allows: I woud patch ProfilerConfig to accept `"rebel"` at the moment vllm.config.profiler is imported, by installing an import hook in `platform.py`. This may work, but one thought in mind is that whether it deviates too much from the `patches/` convention - it is not applied by register_ops() and it cannot live under patches/, instead it would be a small module imported from platform.py instead.

As you pointed out, I also agree that setting RBLNProfiler on `profiler_type is None` does not itself express our intention. so If the method above is not acceptable, I would like to ask, as an alternative whether restructuring the branch so that it reads `elif profiler_type is None and rebel.profiler.is_activated()` would be explicit for now.

I would appreciate your thoughts on this.

rebel-cjlee: Ah, now I understand the problem.

`elif profiler_type is None and rebel.profiler.is_activated()` looks good enough to me. But it still feels a bit unnatural that a profiler gets set even though the user never set `profiler_config`. Enabling it by an env var could be another way.

One question, Is `rebel.profiler.is_activated()` true exactly when `RBLN_PROFILER=1`? If yes, then I have no concern, because setting that env var is already the user asking for the profiler. So let's go with this.

rebel-ywkim: I checked a bit more and found that `rebel.profiler.is_activated()` is not exactly the same as `RBLN_PROFILER=1`.

`rebel.profiler.is_activated()` asks whether RBLN Profiler is activated and it gets activated on runtime instance creation when `RBLN_PROFILER=1` is set. so `rebel.profiler.is_activated()` being true is equivalent to `RBLN_PROFILER=1` only once worker's runtime has been created. 

That is, If runtime instance is not created yet(e.g. VLLM_RBLN_ENABLE_WARM_UP=0) and RBLNWorker.profile() is called before it, `rebel.profiler.is_activated()` will be false, even though the user set `RBLN_PROFILER=1`.

so I think checking the actual flag(`rebel.flags.RBLN_PROFILER` - this matches what the runtime sees) is more correct for the user's intention. 
RBLNWorker.profile() would then set RblnProfilerWrapper following the env var, and its `_start`/`_stop` just signal the process-wide RBLN profiler session: they are no-ops while no profiled runtime exists, and once the first runtime is created it joins the session and is recorded from then on, so stop_profile still flushes the start-stop window.

I also felt a bit unnatural that the user enables the RBLN Profiler through an env var rather than ProfilerConfig, but considering upstream's ProfilerKind is closed, I'd like to ask If we can go with this approach.

rebel-cjlee: @rebel-ywkim Thank you for looking into this so carefully. The direction you suggested sounds good to me!

vllm_rbln/v1/worker/rbln_worker.py
rebel-ywkim: Just want to double check, upstream `gpu_worker.py` also has the dead `self.profiler_config is None` clause, so removing it will add one more diff from the upstream file we track. Is that okay, or do we preper to keep it aligned with upstream?

rebel-cjlee: I missed that point. Then, I think we should just leave it as is for now, following the upstream changes.

tests/native/v1/worker/test_rbln_worker.py
```diff
@@ -1240,3 +1241,52 @@ def test_materialize_runs_the_smallest_compiled_decode_bucket(self):
...
+        assert calls == []
+
+    def test_rbln_profiler_starts_and_flushes_at_stop(self, make_worker, monkeypatch):
+        calls: list[str] = []
+        monkeypatch.setenv("RBLN_PROFILER", "1")
+        monkeypatch.setattr(
+            wm,
+            "rbln_profiler",
+            SimpleNamespace(
+                is_activated=lambda: False,
```
rebel-cjlee: Nothing calls `is_activated` anymore now that the branch reads the flag, so this stub can go.

rebel-cjlee (approved):

rebel-ywkim:
@rebel-cjlee As rebel-compiler version has updated, I am about to merge this and resolved one conflict, could you please once check it?

rebel-cjlee:
@rebel-ywkim It looks good to me.

PR #1053, opened 2026-09-03
refactor(optimum): pad the decode batch in the model runner

rebel-eunji:
## Summary

Move Optimum decoder input preparation and decode-batch padding from individual model implementations into `RBLNOptimumModelRunner`.

`ModelInputForRBLN` now provides tensors in the dtype, shape, and row layout expected by compiled graphs. As a result, model `forward` methods only select the decoder and invoke the graph; the runner drops the padding rows from the output.

### Graph-ready model inputs

- Build prefill and decode inputs directly in their final dtypes:
  - tokens: `int64`
  - positions: `int32`
  - block tables and cache slot IDs: `int16`
- Add `padded_batch_size` and optional `batch_rows` to `ModelInputForRBLN`.
- Use a 1-D block table for prefill and padded 2-D block tables for decode.

### Runner-side decode padding

- Pad tokens, positions, block tables, and cache slot IDs in `_prepare_decode`.
- Use the scheduler-provided `dummy_block` when available.
- Otherwise, select a block and cache slot not used by the running requests.
- Support both:
  - running-order placement with decoder batch bucketing
  - fixed-row placement for models with per-row device state
- Gather the running requests' rows of the decode output in `execute_model`: `[:num_reqs]` for running-order placement, `[batch_rows]` for fixed-row placement. Model forwards return the graph output as is.
  - Exception: Whisper prefill still selects its own row, because that loop pads the full-batch decoder graph inside the model.

### Simplified model interface

Add one layout-policy method to `RBLNOptimumDecoderMixin`:

- `decode_layout(cache_slot_ids, block_tables) -> (padded_batch_size, rows)`

By default the rows are a `slice` over `[0, num_reqs)` of the smallest decoder bucket, so the runner's gather stays a view. Qwen3.5 and Whisper override it to return the full decoder batch with a row tensor: cache-slot rows for Qwen3.5's recurrent state, block IDs for Whisper's decoder KV-cache rows.

The following model-side helpers are removed:

- `preprocess_for_decoder`
- `pad_decoder_items`
- `pad_cache_slot_ids`
- `available_blocks`

Decoder `forward` implementations now consume `ModelInputForRBLN` directly.

## Additional Fixes

- Rename the EXAONE-4.5 EC guard to override the actual `build_prefill_inputs_from_cache` hook.
- Rename Gemma3's `build_decode_inputs` to `build_decode_attention_inputs`.
- Fix test fixtures to use the scheduler's 1-D per-request block-table format.
- Drop the per-step `hidden_states.clone()` into `ExecuteModelState.sample_hidden_states`; `sample_tokens` never read it.

## Testing

```bash
uv run --no-sync pytest tests/optimum/v1/models/optimum/ tests/optimum/v1/worker/ -q
# 59 passed

uv run --no-sync pytest tests/optimum/v1/sample/ -q
# 138 passed

uvx pre-commit run --files <changed files>
# All hooks passed, including mypy-local
```

New runner tests cover:

- running-order batch padding
- scheduler-provided scratch blocks
- padding rows that do not alias active blocks or cache slots
- fixed-row placement for models with per-row state
- dropping the padding rows of the decode output, for both placements
- graph input ordering for Gemma3 and Qwen3.5

## Validation Notes

No NPU end-to-end test was performed. Device validation is recommended for:

- multiple decoder buckets
- Qwen2.5-VL and Qwen3.5 MRoPE decode
- Whisper fixed-row decode
- sliding-window and hybrid-cache models
- prefix caching
- EC consumer flows


rebel-eunji:
Sampler suite finished on d2590297:

```
uv run --no-sync pytest tests/optimum/v1/sample/ -v
312 passed, 2272 warnings in 3033.86s (0:50:33)
```

This suite drives prefill and decode steps through the runner (including `_prepare_decode` padding, bucket selection with `use_multiple_decoder`, and the padded-logits sampler path) with a mock decoder model.

vllm_rbln/model_executor/models/optimum/exaone4_5.py
```diff
@@ -168,20 +167,21 @@ def embed_multimodal(self, **kwargs: object) -> MultiModalEmbeddings | dict:
...
 
-    def build_prefill_inputs(
+    def build_prefill_inputs_from_cache(
         self,
         input_ids: torch.Tensor,
         cached_mm_outputs: list,
         *,
         cache_position: torch.Tensor | None = None,
         running_requests_ids: list[str] | None = None,
+        mrope_position_deltas: dict[str, float] | None = None,
```
rebel-seinpark, line 177: nit) The EC Helper calls this hook with the `input=model input` keyword included, but the exaone override does not have a `model_input` parameter. While this path likely won’t be reached at this point, it seems it will first throw a `TypeError` rather than the intended `NotImplemented` error.

rebel-eunji: That's right, though this will be gone in the following PR  https://github.com/RBLN-SW/vllm-rbln/pull/1063

rebel-seinpark (approved):
LGTM!

vllm_rbln/v1/worker/optimum_model_runner.py
```diff
@@ -714,59 +681,120 @@ def _prepare_prefill(
...
         block_tables = torch.stack(block_tables_list)
+        cache_slot_ids = torch.tensor(
+            [scheduler_output.cache_slot_id_dict[req_id] for req_id in req_ids],
+            dtype=torch.int16,
+        )
 
-        return input_tokens, input_positions, block_tables, running_request_ids
+        batch_rows = self.model.decode_batch_rows(cache_slot_ids, block_tables)
+        rows: torch.Tensor | slice
+        if batch_rows is None:
```
rebel-jonghewk: instead of decode_batch_rows is None meaning `rows = arange(num_reqs)` 

How about refactoring by changing `decode_batch_rows + decode_padded_batch_size` to 


```
    def decode_layout(
        self, num_reqs: int, cache_slot_ids: torch.Tensor, block_tables: torch.Tensor
    ) -> tuple[int, torch.Tensor]:

        if self.use_multiple_decoder:
            padded_batch_size = select_bucket_size(num_reqs, self.decoder_batch_sizes)
        else:
            padded_batch_size = self.decoder_batch_size
        return padded_batch_size, torch.arange(num_reqs)
```

vllm_rbln/model_executor/models/optimum/llava_next.py
```diff
@@ -91,38 +90,22 @@ def _forward(
...
-            cache_position=cache_position,
+            block_tables=model_input.block_tables,
+            input_ids=model_input.input_tokens,
+            inputs_embeds=model_input.inputs_embeds if is_prompt else None,
+            cache_position=model_input.input_positions,
         )
 
         if not is_prompt:
-            logits = logits[:request_nums]
+            logits = logits[: len(model_input.running_requests_ids)]
```
rebel-jonghewk: instead of having this logic here for every model can you bring this also to the runner?

rebel-eunji: https://github.com/RBLN-SW/vllm-rbln/pull/1053/changes/77afafce5a1eba358fd052d9ee1db8dae8d93d05

rebel-jonghewk (approved):
lgtm

vllm_rbln/v1/worker/optimum_model_runner.py
rebel-eunji: I applied it. But torch.arange triggers indexing (advanced). So I applied with slice(0, num_reqs) to reduce overhead. (of course some models requires indexing)

PR #1054, opened 2026-09-03
fix(worker): index the device env var as a pool and rename it to RBLN_VISIBLE_DEVICES

rebel-cjlee:
### 🚀 Summary of Changes

`RBLN_DEVICES` was read as a per-rank slot list. It is now read as a pool of NPUs, the same way `CUDA_VISIBLE_DEVICES` works: one entry per NPU, extra entries are left unused, and each rank picks its own from the pool. The variable is renamed to `RBLN_VISIBLE_DEVICES` to match the CUDA name.

Most of the logic is now upstream's. `Platform.device_id_to_physical_device_id()` already parses the pool, applies the data parallel mapping, and raises on a short pool. The only RBLN part left is that a rank takes `VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK` entries instead of one:

```
selected = [
    current_platform.device_id_to_physical_device_id(first + offset)
    for offset in range(num_devices)
]
```

This removes the two branches, the length `assert`, the `entry * rsd` multiplication, and the DP offset calculation.

#### Fixed bugs

| # | Bug | Before | After |
|---|---|---|---|
| 1 | rsd>1 used NPUs that were not given to the job | `RBLN_DEVICES=4` + rsd 4 -> NPU 16-19 | the same value is now an error; every NPU has to be listed |
| 2 | non-MoE DP could not start without the env var | every replica picked NPU 0, 3 of 4 failed to register | `0, 1, 2, 3`, engine builds |
| 3 | a pool larger than the world size was rejected | `0,1,2,3` + tp 2 -> `AssertionError` | uses `0, 1`, leaves 2 and 3 idle |
| 4 | an exported but empty value crashed | `RBLN_DEVICES=""` -> `AssertionError` / `ValueError` | same as unset |
| 5 | `--device-ids` was ignored | `--device-ids 4,5` + tp 2 -> NPU 0, 1 | NPU 4, 5 |
| 8 | user config was checked with `assert` | message did not say what to fix; a trailing comma crashed | `ValueError` naming the count and the pool |
| 9 | multi node DP used the global DP rank | node B asked for NPU 4-7 that it does not have | vLLM's node-local mapping is used |
| 10 | the two device budget helpers counted different units | test skipped or worker failed | both count NPUs; `prefill_context_parallel_size` added |

#### Migration

**1. rsd > 1 must list every NPU.** One entry is one NPU now, not a group. `RBLN_VISIBLE_DEVICES=4` with rsd 4 is an error, because a rank needs 4 entries and the pool has 1.

```
# to keep the NPUs the old code actually used (4 x 4 = 16)
before:  RBLN_DEVICES=4                     with rsd=4   -> ran on NPU 16-19
after:   RBLN_VISIBLE_DEVICES=16,17,18,19   with rsd=4

# to use NPU 4-7, which "=4" looked like it meant
after:   RBLN_VISIBLE_DEVICES=4,5,6,7       with rsd=4
```

**2. `RBLN_DEVICES` still works** but prints a deprecation warning. It is moved to `RBLN_VISIBLE_DEVICES` and unset at startup. Unsetting is required: the runtime accepts both names but prefers `RBLN_DEVIECS`, so one left in the environment would override the value each worker writes for itself and put every rank on the same NPUs.

**3. `VLLM_RBLN_NUM_RAY_NODES` is removed.** Please drop it from launch scripts.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)
* [x] ❓ Other (`other`): env var rename and removal

---

### 🧪 How to Test

Unit

```bash
uv run --no-sync pytest tests/native/v1/worker/test_rbln_worker.py::TestInitDeviceEnv \
  tests/native/test_platform.py::TestDeprecatedDeviceControlEnvVar \
  tests/native/test_envs.py tests/native/test_utils.py -q
```

On a 32-NPU host, one layer only (`VLLM_RBLN_NUM_HIDDEN_LAYERS=1`, `Felladrin/Llama-160M-Chat-v1`), reading: `Local rank: %d, Selected devices: %s`:

| Case | Expected |
|---|---|
| non-MoE, DP 4, env unset | `0 / 1 / 2 / 3`, engine builds |
| MoE, DP 4, env unset | `0 / 1 / 2 / 3` |
| `RBLN_VISIBLE_DEVICES=0,1,2,3`, tp 2 | `0 / 1`, engine builds |
| `--device-ids 4,5`, tp 2, env unset | `4 / 5` |
| `RBLN_DEVICES=2,3`, tp 2 | deprecation warning, `2 / 3` |
| `RBLN_VISIBLE_DEVICES=4`, rsd 4 | clear `ValueError` |
| `RBLN_VISIBLE_DEVICES=4,5,6,7`, rsd 4 | `4,5,6,7` |

---

Before, non-MoE DP 4 with no env var:

```
(EngineCore_DP0) Local rank: 0, Selected devices: 0
(EngineCore_DP1) Local rank: 0, Selected devices: 0
(EngineCore_DP2) Local rank: 0, Selected devices: 0
(EngineCore_DP3) Local rank: 0, Selected devices: 0
RuntimeError: rbln_register_device_id failed for rbln:0 on physical NPU(s) [0]
```

After:

```
(EngineCore_DP0) Local rank: 0, Selected devices: 0
(EngineCore_DP1) Local rank: 0, Selected devices: 1
(EngineCore_DP2) Local rank: 0, Selected devices: 2
(EngineCore_DP3) Local rank: 0, Selected devices: 3
=== engine built
```

rsd 4 with a pool of four, one run:

```
RBLN_VISIBLE_DEVICES=4,5,6,7   VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK=4

Local rank: 0, Selected devices: 4,5,6,7
=== engine built

| NPU |     Process     |     PID    |  CTX  |
| 4   | VLLM::EngineCor |  3126262   | 10001 | 
| 5   | VLLM::EngineCor |  3126262   | 10002 |
| 6   | VLLM::EngineCor |  3126262   | 10003 |
| 7   | VLLM::EngineCor |  3126262   | 10004 | 
```

Deprecated name:

```
WARNING [platform.py:236] RBLN_DEVICES is deprecated and will be removed in a future release. Please use RBLN_VISIBLE_DEVICES instead; this run uses RBLN_VISIBLE_DEVICES=2,3.
```

---

### 📋 Checklist

---

### 💬 Notes

- `device_control_env_var` is a shared `RblnPlatform` attribute and vLLM uses it for the DP device mapping on **both model paths**. The optimum path does not read `RBLN_DEVICES` ifself, but it is affected through that path. **The optimum suite was not run**.
- `set_assigned_physical_gpu_ids()` is called in the worker because `UniProcExecutor` never publishes the mapping. tp 1 without DP uses `uni`, so `--device-ids` would be ignored there otherwise.
- Multi node DP and Ray were **not tested**. Ray is not installed in my environment, and I confirmed with the team that no deployment uses the Ray executor at the moment.

---

### DIFF
--- a/.buildkite/pipelines/vllm-rbln/scripts/lane.py
+++ b/.buildkite/pipelines/vllm-rbln/scripts/lane.py
@@ -30,6 +30,7 @@ _RSD_ENV = "VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK"
 _PARALLEL_KEYS = (
     "tensor_parallel_size",
     "pipeline_parallel_size",
+    "prefill_context_parallel_size",
     "data_parallel_size",
 )
 
@@ -49,15 +50,18 @@ def host_chip() -> str | None:
 
 
 def device_count() -> int:
-    """NPUs this run may use. RBLN_DEVICES wins when set, since a job may be
-    given some of them."""
-    visible = [d for d in os.environ.get("RBLN_DEVICES", "").split(",") if d.strip()]
+    """NPUs this run may use. The visible list wins when set, since a job may be
+    given some of them. Both spellings are read because this runs before the
+    platform plugin folds the deprecated one into the other."""
+    visible_list = os.environ.get("RBLN_VISIBLE_DEVICES", "")
+    raw = visible_list or os.environ.get("RBLN_DEVICES", "")
+    visible = [d for d in raw.split(",") if d.strip()]
     return len(visible) or len(glob.glob("/dev/rbln*"))
 
 
 def devices_needed(serve: dict[str, Any], env: dict[str, str]) -> int:
-    """Mirrors RBLNWorker._init_device_env: DP ranks do not share, and every
-    (tp x pp) rank takes VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK devices."""
+    """Mirrors RBLNWorker._init_device_env: DP ranks do not share, and every rank
+    of vLLM's world size takes VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK devices."""
     needed = int(env.get(_RSD_ENV, 1))
     for key in _PARALLEL_KEYS:
         needed *= int(serve.get(key, 1))
--- a/examples/experimental/qwen3_data_parallel.py
+++ b/examples/experimental/qwen3_data_parallel.py
@@ -66,7 +66,7 @@ def main(
 
     # CUDA_VISIBLE_DEVICES for each DP rank is set automatically inside the
     # engine processes.
-    os.environ["RBLN_DEVICES"] = str(local_dp_rank + 2)
+    os.environ["RBLN_VISIBLE_DEVICES"] = str(local_dp_rank + 2)
 
     # Sample prompts.
     prompts = [
--- a/examples/optimum/ec_disagg/serve_ec_encoder.sh
+++ b/examples/optimum/ec_disagg/serve_ec_encoder.sh
@@ -41,7 +41,7 @@ launch_encoder() {
     local port=$((BASE_PORT + idx))
 
     echo "Starting encoder $idx (device=$device, port=$port, push→$LLM_HOST:$LLM_PULL_PORT)"
-    RBLN_DEVICES=$device vllm serve "$MODEL_ID" \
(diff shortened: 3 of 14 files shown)


rebel-jonghewk (approved):
LGTM

PR #1055, opened 2026-09-03
perf(mega-cache): stream the bundle to disk and mmap it on load

rebel-jonghewk:
### 🚀 Summary of Changes

Saving the mega-cache bundle materializes three RAM copies of every staged `.rbln` blob (flush read, torch serializer bytearray, `to_bytes()`), so each rank's RSS jumps by ~3x its bundle size at save. On an 8-rank DP run all ranks save at once, and that transient was ~47% of the observed compile-time host-memory peak. Loading read the whole bundle and torch sliced a copy per artifact, ~2x the bundle.

`save()` and `load()` are now thin wrappers over rebel's `write_bundle(dst)` / `read_bundle(src)` (rebellions-sw/rebel_compiler#13404). rebel owns the bundle format on both sides: staged blobs stream into the bundle file, and the bundle is mmapped on load so torch's per-artifact slices are zero-copy. The tmp/fsync/rename protocol, the ENOSPC handling, and the bundle keying are unchanged; existing bundles load as before.

Synthetic payload (process RSS): save transient ~3.0x payload → ~0.04x; load transient ~2x payload anonymous → 0 anonymous (file-backed pages only).

**Requires a rebel with `write_bundle`/`read_bundle`.** An older rebel raises `AttributeError` inside the existing broad excepts, so save logs "could not persist" and load logs "failed to load" and the run compiles cold — no crash, no cache. `TestConformance.test_rebel_mega_cache_api` fails until the pin covers #13404; merge this after that publishes.

---

* Depends on rebellions-sw/rebel_compiler#13404

---

### ✅ Type of Change

---

### 🧪 How to Test

1. Run `pytest tests/native/v1/worker/test_mega_cache.py`
2. `TestSaveLoad` stubs rebel's `write_bundle`/`read_bundle` and asserts the file that reaches disk: round trip, re-save, nothing-to-save leaves no bundle and no `.tmp`, ENOSPC / fsync failure keep the previous bundle, corrupt and unreadable bundles warn and recompile. `TestConformance` asserts rebel exposes both entry points.
3. The streaming and mmap behavior themselves are tested in rebel (`tests/python/test_rebel/test_mega_cache.py` in #13404): byte-identity with torch's serializer, chunked writes, torch receives a memoryview, M: entries outlive the mapping, empty/corrupt files return None.

Run here with a stubbed `rebel` package carrying #13404's `mega_cache.py` and without the native conftest (`import rebel` is broken on this box): 30 passed in the save/load/path classes; `TestSignatureVllmConfig`/`TestWarmupWiring` not runnable locally. CI should confirm the full file once the rebel pin moves.

---

### 📋 Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ dev = [
 ]
 
 runtime = [
-    "rebel-compiler~=0.11.3.dev96",
+    "rebel-compiler~=0.11.3.dev235",
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ dev = [
 ]
 
 runtime = [
-    "rebel-compiler~=0.11.3.dev235",
+    "rebel-compiler~=0.11.3.dev237",
 ]
 
 
--- a/tests/native/v1/worker/test_mega_cache.py
+++ b/tests/native/v1/worker/test_mega_cache.py
@@ -28,7 +28,6 @@ import re
 from types import SimpleNamespace
 
 import pytest
-import torch
 
 from tests.native.vllm_config import make_vllm_config
 from vllm_rbln.v1.worker import mega_cache, rbln_model_runner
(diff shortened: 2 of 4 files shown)


rebel-eunji (commented):
LGTM 👍

rebel-eunji (approved):

rebel-cjlee (approved):

PR #1058, opened 2026-09-04
fix(nixl): transfer KV where a speculative draft's geometry differs from the target's

rebel-ykchoi:
#### Summary

NIXL registration required every non-MLA region to be the same size, and a head band
was derived once from the model config. Both describe the target's layers alone. A
speculative draft carries its own `num_key_value_heads`, so the regions of its KV
layers are a different size — and upstream puts target and draft in **one** group and
hands out per-layer tensors that differ. Differing sizes are normal there; this
connector refused them, so a P/D pair with a draft could not start at all.

Getting past that refusal is worse than hitting it. A band computed from the target's
count names a byte offset that is inside a draft region rather than out of range, so
registration, handshake and descriptor construction all succeed and only the bytes
are wrong.

The same confusion of "the model's geometry" with "this region's geometry" has a
second, independent instance on the read path. Upstream's remote descriptor builder
walks the peer's region list and feeds that **position** into
`get_backend_aware_kv_block_len` and `_is_region_replicated`, both of which index our
own per-region arrays. The position is our region only when our band starts at our
region 0 — true for a pipeline stage, false for a consumer holding every layer while
the producer is pipelined. Equal region lengths hid it: the wrong index returned the
right value.

Why this was not reachable before: the two eagle3 configurations in this tree pair a
draft whose KV head count and head dim match the target's. Query heads may split
while the KV geometry does not, so neither touches this axis. The axis is not
"is there an eagle draft" but **"do the draft and the target disagree on KV heads"**.

##### A. Geometry is recorded per region, not per model

Registration keeps, for each region, the page it spans, the KV head count of the
layer it belongs to, and the chiplet slices it was cut into. Every consumer reads its
own region's values instead of one engine-wide number.

The per-layer head count is recovered rather than trusted: upstream floors a per-rank
share at one, so below one head per rank the product overstates the model *and*
satisfies the divisibility guard meant to refuse that layout. A product no model in
the engine has is that case, and such a region gets no head band at all.

##### B. The head split moves to cuts, where the head count cancels

A split used to be derived from head counts that now differ per region. Expressed as
a ratio of cut counts instead, the head count cancels, so one split still describes a
whole transfer even when its regions do not agree on how many heads they hold.

##### C. The compatibility hash gains the draft's identity

A prefill carrying a draft can no longer pair with a decode that does not: the two
register different numbers of KV layers, and nothing upstream hashes separates them.
Gated on the methods whose draft has attention layers of its own — upstream aliases
the target's config as an n-gram draft, and n-gram is commonly configured on the
decode engine alone, so hashing it unconditionally would refuse a working pair.

Model-level values only. The hash is compared across every shard, so a per-rank
quantity would differ between pipeline stages and no pair would ever match.

##### D. A pipelined peer's positions are translated to our region ids

The view is scoped to `_build_fa_remote`, the upstream loop that walks a peer's list
while reading ours. Doing it around `add_remote_agent` instead would also catch the
local dlists that method builds from **our** region ids, which must not be translated.

Untranslated, the two sides disagree on descriptor lengths, NIXL rejects the pair at
transfer setup, the blocks are marked invalid, and the request falls back to
recompute — which surfaces as a throughput change rather than as an error.

##### E. Combinations that one engine-wide value cannot describe are refused

Host staging registers whole logical buffers and delegates to an upstream assertion
that wants one size for every non-MLA tensor, so a draft with its own head count is
refused there by name instead of failing inside that assertion. The same applies to a
cross-layer-blocks cache and to an engine mixing MLA and non-MLA layers: no partial
support and no fallback.

#### Testing

- **Unit** — per-region registration over a draft layer with its own head count, the
  replicated-head case that must get no band, the head split over regions that
  disagree, the head-band arithmetic against a peer at another TP degree, the
  position-to-region translation on a pipelined peer, the hash gaining and not
  gaining the draft, and each refusal (`test_rbln_nixl_base_worker.py`,
  `test_rbln_nixl_handshake.py`, `test_rbln_nixl_metadata.py`).
- The translation is reached through upstream's own `_build_fa_remote` loop rather
  than by calling the override directly, so an identity translation is caught.
- Cases are chosen so the input decides the answer: the same geometry with equal and
  with unequal draft head counts gives opposite expected bands, and the descriptor
  assertions decode each address back to (region, block, head offset) so a plausible
  wrong offset cannot pass.
- `pytest tests/native/distributed` green; `pre-commit` clean on every changed file.

#### Risk / compatibility

- **Without a draft the behaviour is unchanged.** One model has one `num_kv_heads`
  and one page size, so every per-region quantity is the engine-wide one and every
  new refusal is unreachable.
- **The compatibility hash gains a factor**, so a prefill with a draft no longer
  handshakes with a decode without one. This is the intended effect: the two register
  different numbers of layers and would otherwise pair.
- The refusals fire only where registration already failed or the descriptors were
  already wrong, so the configurations they turn away could not have worked.

---

* Resolves rebellions-sw/fsw-inference#541 — registering and transferring a KV cache
  whose speculative draft has a geometry of its own.

---

### ✅ Type of Change

* [x] 🐛 Bug fix (`fix`)

---

### 🧪 How to Test

2. `uv run --no-sync pytest tests/native/distributed`
3. Expected: all green.
4. Edge cases covered: a draft layer whose KV head count differs from the target's; a
   draft whose count matches, which must keep the previous behaviour; a region whose
   head count is a replicated product no model has; a peer at another TP degree
   against regions that disagree on head count; a pipelined producer against a
   consumer holding every layer; a prefill with a draft against a decode without one;
   host staging with differing per-layer sizes, which is refused.

---

### 📋 Checklist

---

### 💬 Notes

- Two commits, kept apart because they fail differently: the first is refused at
  registration, the second succeeds and moves the wrong bytes. Reviewing them
  together is easier than reviewing either alone, since both come from reading one
  model's geometry as every region's.
- The hash change means **a P/D pair using a draft has to be deployed together.**


rebel-chanheo (approved):
LGTM!

rebel-jongwonlee (commented):
Re-read this on `1953664a`. The only change since my last pass is the one line where
#1045's `region_cut` fix flowed through, so everything below is about the PR as it stood.

The shape of it is right, and the `_head_split` move is the part I would call out:
writing the split as a ratio of cuts rather than heads-per-slice makes `T` cancel, so one
split serves regions that disagree on head count — and the inverted comparison is the
same predicate (`per_slice_l > per_slice_r` ⟺ `cuts_r > cuts_l`), not a behaviour change.

Four things I checked rather than took on trust:

* `use_eagle()` covers `"dflash"`, so #975 rides the hash gate rather than slipping past it.
* `_regions_viewed_as`' plain attribute is safe — `_handshake_initiation_executor` is
  `max_workers=1`.
* Replacing the cross-layer equal-size assert with a per-layer self-consistency check
  keeps it an `assert` for the right reason: local values, not a peer's.
* The axis derivation still reads the per-rank count, so #1045's derivation is undisturbed
  by the new model-wide recovery living beside it.

Two inline notes, neither blocking.

**One minor thing that does not fit inline.** `_regions_viewed_as` argues from two
handshakes never overlapping, which holds. The axis it does not mention is registration
against a handshake — different threads. It is fine today because D2D registration
finishes in worker init before any request arrives, but the docstring reads as if
handshake-vs-handshake were the whole question. One clause would cover it.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
```diff
@@ -733,48 +867,21 @@ def _build_head_matched_remote(
...
-            areas_r,
-            slices_r,
-            side="peer",
+        split = self._head_split(
+            self.transfer_topo.tp_size * slices_l, remote_tp_size * slices_r
         )
-        split = self._head_split(per_slice_l, per_slice_r)
 
         replicas_l = areas_l // slices_l
         replicas_r = areas_r // slices_r
```
rebel-jongwonlee, line 883: ### The `slices <= 0` guard now sits behind three divisions by `slices`

#1045 put `slices <= 0` into `_slice_head_bounds` deliberately — its test says so:
*"without the `slices <= 0` half of that guard the modulo beside it divides by zero."*

Making head counts per-region moved `_slice_head_bounds` into the per-region loop, and
the guard went with it. The order here is now

```python
split      = self._head_split(tp * slices_l, remote_tp * slices_r)   # cuts_r % cuts_l
replicas_l = areas_l // slices_l
replicas_r = areas_r // slices_r
...
for logical_l, logical_r in logical_pairs:
    self._slice_head_bounds(..., slices_l, side="local")             # guard, three divisions late
```

where #1045 had both `_slice_head_bounds` calls first, ahead of every division. A peer
advertising `kv_slices = 0` now reaches `areas_r // 0` and raises `ZeroDivisionError`
instead of the message naming which side sent what. `_peer_head_split` has the same
shape — the shard loop calls it before `add_remote_agent` validates.

It still refuses, so this is a diagnostic regression rather than a correctness one. A
cheap fix would be validating the two advertised slice counts once on entry, before
anything divides by them; that also gives `_head_split` a precondition it can rely on
instead of leaning on `cuts_r <= cuts_l` to short-circuit a zero.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
```diff
@@ -281,6 +326,59 @@ def finalize_kv_cache_registration(self) -> None:
...
+        # a speculative draft where there is one.
+        known = {self.model_config.get_total_num_kv_heads()}
+        speculative_config = self.vllm_config.speculative_config
+        if speculative_config is not None:
+            draft_model_config = speculative_config.draft_model_config
+            if draft_model_config is not None:
+                known.add(draft_model_config.get_total_num_kv_heads())
+        if total not in known:
+            return None
+        return total
```
rebel-jongwonlee: ### This can alias one model's floored product onto another's real count

The recovery rests on *"a product no model here has is that case"*. With two models in
`known`, a product one model cannot have may be one the other genuinely has.

Target 4 KV heads, draft 8, TP 8:

| | true | per-rank (floored) | `total = p × w` | in `known={4,8}`? | returned |
|---|---:|---:|---:|---|---:|
| **target layer** | 4 (replicated 2× over 8 ranks) | `max(1, 4//8)` = 1 | **8** | yes | **8** ❌ |
| draft layer | 8 | `max(1, 8//8)` = 1 | 8 | yes | 8 ✓ |

The target region gets an 8-head band when its heads are replicated and it should get
none: rank 3 claims head 3 and holds head 1. That is the plausible-wrong-offset this PR
exists to remove, reintroduced through the recovery.

The root of it is that `p` and `w` do not determine `t`: for `t >= w`, `p = t // w`
leaves `t` anywhere in `[p·w, p·w + w - 1]`; for `t < w`, `p` is 1 and says nothing.
Testing membership of `p × w` picks one point in that range and hopes.

Consistency over the candidates would close it:

```python
candidates = {k for k in known if max(1, k // self.world_size) == layer_spec.num_kv_heads}
if len(candidates) != 1:
    return None                      # cannot tell which model this layer belongs to
k = candidates.pop()
if k < self.world_size:
    return None                      # replicated -- no band
return k
```

On the case above (`p=1, w=8`) both 4 and 8 are consistent, so it returns `None` and
refuses rather than banding wrongly.

**Is there something better available?** This already takes the layer name. If a
drafter's layers are distinguishable by name, attributing each layer to its own model
removes the ambiguity instead of detecting it — worth more than either version of the
heuristic. I could not tell from here whether the naming supports that.

Reachability is narrow (`target_total < TP` **and** the draft's count equal to TP), so I
am not calling it blocking. But the guard exists precisely to catch replication, and this
is the one place where two models in the set makes it blind.

rebel-ykchoi: Confirmed, and fixed. Reproduced your table: with a draft of 8 at TP 8 the target's
replicated 4-head layer comes back as 8.

`_layer_kv_heads` now collects the declared counts a share is consistent with and takes
one only when exactly one is, so the ambiguous case refuses instead of banding. Test: a
target of 4 with a draft of 8 at TP 8 returns None.

On attributing each layer to its model instead: the set difference that would do it is
computed on the proposer, which registration cannot reach. And names alone do not
separate them -- in the eagle head I read, the draft's layers are numbered past the
target's depth rather than carrying a prefix of their own. So detecting the ambiguity is
where this lands, as you expected.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/base_worker.py
rebel-ykchoi: Confirmed. A peer advertising zero slices raises ZeroDivisionError out of
`_build_head_matched_remote` and out of `_peer_replica_fanout`; our own count does the
same in `_peer_head_split`, since `cuts_r <= cuts_l` is what was keeping the zero away
from the modulo.

The advertised count is now checked where the peer's metadata is decoded, which is ahead
of every reader of the field, and our own gets an assert where registration sets it -- a
peer's claim and our invariant are different things. The new test drives the handshake:
the existing one calls the helper directly, so it pins the refusal without pinning that
the refusal runs first.

rebel-jongwonlee (approved):
All three land, and two of them in a better shape than I proposed. Checked the delta
rather than the description: the three new commits touch 54 lines across
`base_worker.py` and two test files, nothing else riding along. CI green on build 984.

**The head-count ambiguity.** The candidate set is the fix, and keeping
`total < self.world_size` alongside it matters — one surviving candidate can still be a
replicated layer, so the two checks answer different questions. The docstring no longer
claims a product no model has, which was the sentence the old logic rested on. And the
test reproduces the case rather than restating it: target 4, draft 8, TP 8, with the
comment carrying the consequence (`rank 3 would claim head 3 while it holds head 1`)
instead of just the expected return.

**The zero slice count — this is better than what I asked for.** I suggested validating
both counts once on entry. Splitting them by origin is the right call: the peer's claim
raises where its metadata is decoded, our own gets an assert where registration sets it.
That is the same line #1011 drew and #1045 kept, applied to a field I had lumped
together. I confirmed the placement covers every reader — `_query_agent_meta` runs at
L1543 and L1592, ahead of `_peer_head_split` (L1609), `_peer_replica_fanout` (L1610) and
`_build_head_matched_remote`.

You also found a third divisor I missed: I named `_build_head_matched_remote` and
`_peer_head_split` and did not check `_peer_replica_fanout`.

And the test drives the handshake rather than the helper. That distinction is exactly
what #1045 turned up — a suite that stayed green with the entry-point call deleted — so
pinning that the refusal *runs first*, not merely that it exists, is the version worth
having.

**The region view's docstring.** Naming `finalize_kv_cache_registration` rather than the
warm-up hook is the useful form: it says which call moving would break the attribute.

**On attributing layers to their model:** thanks for actually looking. The proposer-side
set difference being out of reach from registration, and eagle numbering the draft's
layers past the target's depth rather than prefixing them, settles it — detecting the
ambiguity is the reachable answer, and the code now does that instead of guessing.

Also noting you caught an error in your own description on the re-read
(`_build_head_matched_remote` walks a peer's list too, so "the one place" needed the
upstream qualifier). Nothing further from me.

Approving.

rebel-cjlee (approved):

PR #1060, opened 2026-09-05
perf(pp): reuse intermediate tensors buffer

rebel-huijong:
### 🚀 Summary of Changes

Allocate intermediate tensors once per graph and reuse them instead of allocating new tensors on every step.

When using device tensors, the graph input tensor changed on every PP step because `GroupCoordinator.recv_tensor_dict` allocates a new empty tensor on each call. This caused severe performance degradation.

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Run the simple TTFT check below before and after this PR:

```
import os

os.environ["HF_HUB_OFFLINE"] = "1"
os.environ["RBLN_ROOT_IP"] = "127.0.0.1"
os.environ["RBLN_LOCAL_IP"] = "127.0.0.1"
os.environ["VLLM_DISABLE_COMPILE_CACHE"] = "1"
os.environ["VLLM_RBLN_AUTO_PORT"] = "1"
os.environ["VLLM_RBLN_DISABLE_COMPILE_CACHE"] = "1"
os.environ["VLLM_RBLN_COMPILE_STRICT_MODE"] = "1"
os.environ["VLLM_RBLN_USE_VLLM_MODEL"] = "1"

import numpy as np
from vllm import LLM, SamplingParams

def main():
    np.random.seed(42)

    llm = LLM(
        model="MiniMaxAI/MiniMax-M2.7",
        max_num_seqs=4,
        block_size=8192,
        enable_chunked_prefill=True,
        max_num_batched_tokens=512,
        pipeline_parallel_size=4,
        enable_prefix_caching=False,
        disable_log_stats=False,
        kv_cache_dtype="fp8",
    )

    prompt = np.random.randint(1000, 151642, (1, 50*1024)).tolist()
    _ = llm.generate(prompt, SamplingParams(temperature=0.0, max_tokens=1, ignore_eos=True))
    outputs = llm.generate(prompt, SamplingParams(temperature=0.0, max_tokens=1, ignore_eos=True))
    ttft = outputs[0].metrics.first_token_ts - outputs[0].metrics.scheduled_ts
    print(f"TTFT: {ttft}")

if __name__ == "__main__":
    main()
```

---

### 📋 Checklist

---

### 💬 Notes

For the simplest fix, I implemented `recv_intermediate_tensors` in the model runner. However, as you can see in the model worker, receiving PP intermediate tensors is originally the model worker's responsibility, the model runner just takes the intermediate tensors as execute-model input and runs the forward pass. Given that, it might be a better direction to patch `GroupCoordinator.recv_tensor_dict` to accept buffers as an argument instead. What do you think @rebel-cjlee? I'll update the implementation based on feedback.

I also removed the `forward_pass` variable the name was confusing since it seems to mean running the forward pass, but "pass" could also read as skipping the forward and it was only used once anyway.

---

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -29,7 +29,7 @@ from vllm.distributed.kv_transfer import (
     get_kv_transfer_group,
     has_kv_transfer_group,
 )
-from vllm.distributed.parallel_state import get_pp_group
+from vllm.distributed.parallel_state import TensorMetadata, get_pp_group
 from vllm.model_executor.layers.attention import Attention
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.model_executor.model_loader import get_model_loader
@@ -419,10 +419,8 @@ class RBLNModelRunner(KVConnectorModelRunnerMixin):
         self.seq_lens = torch.zeros(self.max_num_tokens, dtype=torch.int32)
         self.seq_lens_np = self.seq_lens.numpy()
         self.discard_request_mask = torch.zeros(self.max_num_reqs, dtype=torch.bool)
+        self.intermediate_tensors_dict: dict[tuple[int, int], IntermediateTensors] = {}
         self.input_stager = InputStager(self.device)
-
-        # None in the first PP rank. The rest are after load_model
-        self.intermediate_tensors: IntermediateTensors | None = None
 
         # OPTIMIZATION: Cache the tensors rather than creating them every step.
         # Keep in int64 to avoid overflow with long context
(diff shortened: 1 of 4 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -419,11 +419,9 @@ def __init__(
         self.seq_lens = torch.zeros(self.max_num_tokens, dtype=torch.int32)
         self.seq_lens_np = self.seq_lens.numpy()
         self.discard_request_mask = torch.zeros(self.max_num_reqs, dtype=torch.bool)
+        self.intermediate_tensors_dict: dict[tuple[int, int], IntermediateTensors] = {}
```
rebel-cjlee, line 422: ~Since we already have a buffer in `input_stager`~, I think it might be better to merge them. What are your thoughts on this?

rebel-jonghewk (approved):
LGTM. 

Just quick question. 
What happens if the recv moves inside the model runner?
Seems like most of this function is spent recovering num_reqs_padded / query_len, and both are already locals of execute_model.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-huijong: Ah, are you referring to the hidden states in the input stager?

rebel-huijong: Looking at it, the input stager's hidden_state is the drafter's input, so it's a different thing. As we discussed verbally, let's merge this as-is for now and look for improvements as we work on further performance later.

rebel-cjlee (approved):

PR #1061, opened 2026-09-06
perf(spec-decode): sample and draft-bookkeep on host tensors under device tensors

rebel-daeyang:
### 🚀 Summary of Changes

**Problem.** Under `VLLM_RBLN_USE_DEVICE_TENSOR=1`, torch-rbln runs every eager op that has a non-fp16/bf16 input (int, bool, fp32) through a CPU fallback. The glue between the compiled graphs of a speculative step — token ids, indices, masks, the fp32 softmax of rejection sampling — is made of exactly those ops, so each one round-trips the host. Rejection sampling and the EAGLE drafter's bookkeeping both ran slower than on the host-tensor path.

**Solution.** Keep the target and draft graphs, the KV cache and the hidden states on the device; move the glue to the host.

- The EAGLE drafter keeps its bookkeeping (input ids, positions, backup ids, d2t) on the host. The input stager still stages the graph inputs onto the device, so the draft graph is unchanged. Draft ids come back with one small D2H per pass.
- A speculative step brings the target logits to the host once, in `execute_model`, where the int64 gather was already a fallback. The rejection sampler, its compiled op and the sampling metadata it reads work on host tensors; the op still executes on the NPU, the runtime just stages its inputs and outputs across the boundary. The bonus token comes from the eager vLLM `Sampler`, since the runner's RBLN sampler is compiled for device tensors
- The `min_tokens` logits processor follows the logits device, so it serves both the device path (a step without drafts) and the host path (a speculative step).

No new option: this is how device-tensor mode behaves. The host-tensor path (`VLLM_RBLN_USE_DEVICE_TENSOR=0`) is unchanged.

**Result** (MiniMax-M2.7 + Eagle3, dp4, profiled): rejection sampling per decode step matches the host-tensor path, the drafter's host time drops, tpot improves over device-tensor mode, and the output tokens are identical.

---

* Related to # (torch-rbln CPU fallback for non-fp16/bf16 eager ops)

---

### ✅ Type of Change

---

### 🧪 How to Test

1. MiniMax-M2.7 + Eagle3 spec decode, dp4: output tokens match device-tensor mode, tpot improves.
2. `tests/native` and `tests/torch_compile/unit` pass with device tensors on and off.

---

### 📋 Checklist

---

### 💬 Notes

- The stager's device-to-device copy of the hidden states per draft pass is pre-existing and not touched here.
- An opt-in variant behind an env var, with a host-tensor twin of the RBLN sampler for the bonus token, measured the same; it is left out to keep the diff small.

### DIFF
--- a/vllm_rbln/v1/worker/rbln_model_runner.py
+++ b/vllm_rbln/v1/worker/rbln_model_runner.py
@@ -1093,8 +1093,9 @@ class RBLNModelRunner(KVConnectorModelRunnerMixin):
 
         # Compute the draft token ids.
         # draft_token_indices:      [  1,   2,   3, 105, 106, 208]
+        # Stays on the host: the rejection sampler runs there (see `_sample`).
         draft_token_ids = self.input_ids[logits_indices]
-        draft_token_ids = draft_token_ids[target_logits_indices + 1].to(self.device)
+        draft_token_ids = draft_token_ids[target_logits_indices + 1]
 
         return SpecDecodeMetadata(
             draft_token_ids=draft_token_ids,
@@ -1315,6 +1316,9 @@ class RBLNModelRunner(KVConnectorModelRunnerMixin):
                     spec_decode_metadata, bucket
                 )
                 sampling_metadata = _pad_sampling_metadata(sampling_metadata, bucket)
+            # The logits came to the host in execute_model; the input batch keeps
+            # its sampling tensors on the device, so bring the few read here along.
+            sampling_metadata = _sampling_metadata_to_host(sampling_metadata)
             out = self.rejection_sampler(
                 spec_decode_metadata,
                 None,  # draft_probs
@@ -1732,7 +1736,12 @@ class RBLNModelRunner(KVConnectorModelRunnerMixin):
             sample_hidden_states = hidden_states
             assert self.use_wrapped_compute_logits
             if not self.is_prefill and spec_decode_metadata is not None:
-                logits = logits[logits_indices]
+                # NOTE(RBLN): a speculative step samples on the host, and this is
+                # its one D2H. torch-rbln runs int/bool/fp32 eager ops through a
+                # CPU fallback, so the int64 gather below and everything the
+                # rejection sampler does would otherwise round-trip the host op
+                # by op. The step without drafts keeps the device sampler.
+                logits = logits.cpu()[logits_indices]
 
         self.execute_model_state = ExecuteModelState(
             scheduler_output,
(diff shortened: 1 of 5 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3277,20 +3284,20 @@ def _warmup_sampler_decode_batches(self) -> None:
...
             num_spec,
             dtype=torch.int32,
-            device=self.device,
-        )
-        bonus_token_ids = torch.zeros(
-            batch_size, 1, dtype=torch.int64, device=self.device
+            device=device,
         )
+        bonus_token_ids = torch.zeros(batch_size, 1, dtype=torch.int64, device=device)
         dummy_sampling_metadata = SamplingMetadata(
```
rebel-jonghewk, line 3301: @rebel-daeyang  As far as I can trace, the compiled rbln_rejection_sample now takes cpu tensors under `USE_DEVICE_TENSOR=1`. 

Was that the intent? I read the PR as "keep the compiled graphs on device tensors, move only the eager glue to the host", but this turns the rejection op itself into a host-tensor graph, so device-tensor mode now runs one sampler graph in host mode.

rebel-daeyang: Yes, intended - the description contradicted itself, fixed.

`target_probs` is the output of a softmax over the host logits, and the op's outputs feed straight into eager torch ops, so I kept the rejection sampler's inputs on the host.
Moving them to the device now would just carry the same bytes across one step earlier.
Once the surrounding ops are optimized to run on device tensors, I'll move the inputs onto the device.

rebel-jonghewk: f2f: keep input as device tensor for compiled models.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3469,6 +3476,28 @@ def _pad_spec_decode_metadata(
...
+        "frequency_penalties",
+        "presence_penalties",
+        "repetition_penalties",
+        "prompt_token_ids",
+    )
+    moved = {
+        name: t.cpu()
+        for name in fields
+        if (t := getattr(md, name)) is not None and t.device.type != "cpu"
+    }
```
rebel-minkyu, line 3496: Upstream already keeps host copies of these sampling metadata, like `temperature_cpu`, `top_p_cpu_tensor`. Would it be possible to reuse them instead of copying them on every step?

PR #803, opened 2026-07-21
feature: Support automatic prefix caching for multimodal models

rebel-eunji:
### 🚀 Summary of Changes

This re-created pr #753 .

1. Support for APC
This PR enables APC (Automatic Prefix Caching) for multimodal models that use simple full attention.

2. Skip generating multimodal embeddings on prefix cache hits
When a prefix cache hit occurs, multimodal embedding generation is skipped, avoiding redundant computation.

3. Fix the logic of validating the model is using sliding window
The configuration of multimodal models usually have sliding window related parameters in the nested text_config
- https://huggingface.co/google/gemma-3-4b-it/blob/main/config.json
- https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B/blob/main/config.json

4. Refactor runner and Qwen-VL codes
- qwen_vl.py file was so messy because there are APC, EC, and special handling for multimodal(because of deepstack). So I've split the logics of qwen3_vl to qwen3_vl.py
- I reordered the methods of runner and qwen_vl codes to know the workflow easily

#### Qwen-VL family: model flow

**Class hierarchy**

```mermaid
classDiagram
    class RBLNOptimumQwenVLForConditionalGeneration {
      <<abstract>>
      owns shared prefill/decode flow
    }
    RBLNOptimumQwenVLForConditionalGeneration <|-- RBLNOptimumQwen2VLForConditionalGeneration
    RBLNOptimumQwenVLForConditionalGeneration <|-- RBLNOptimumQwen2_5_VLForConditionalGeneration
    RBLNOptimumQwen2_5_VLForConditionalGeneration <|-- RBLNOptimumQwen3VLForConditionalGeneration
    RBLNOptimumQwen3VLForConditionalGeneration <|-- RBLNOptimumQwen3VLMoeForConditionalGeneration
```

The base owns the whole prefill/decode flow; subclasses only fill in per-model hooks (`_create_*`, `_add_model_specific_args`) and Qwen3's deepstack.

**Prefill flow**

`build_prefill_forward_inputs` (owned by the base mixin) dispatches on whether `partial_prefix` is set; the EC consumer enters through `build_prefill_inputs_from_cache`. All four entry points converge on the same building blocks.

```mermaid
flowchart TD
    RUN["runner: execute_model"] --> K{"prefill path?"}
    K -->|"non-EC"| BPF["build_prefill_forward_inputs<br/>(base dispatch)"]
    K -->|"EC consumer"| BPC["build_prefill_inputs_from_cache"]

[...]

PR #1063, opened 2026-09-06
refactor(optimum): move mm encoding, the encoder cache, and MRoPE positions into the runner

rebel-eunji:
### 🚀 Summary of Changes

Move multimodal encoding, the encoder cache, embedding gathering, and MRoPE request state from the optimum model classes into `RBLNOptimumModelRunner`, mirroring upstream vLLM v0.24.0's `GPUModelRunner`.

The runner now:

1. Encodes the uncached items in the current prefill window (`_execute_mm_encoder`) and caches each one under its `mm_hash`.
2. Gathers the cached tensors overlapping the window into `mm_embeds` + `is_mm_embed` (`_gather_mm_embeddings`).
3. Keeps MRoPE positions and the decode delta per request on `CachedRequestState`.

The models only return one 2D tensor per item from `embed_multimodal`, scatter the gathered tensors, and build RBLN graph inputs. Qwen3-VL packs its deepstack layers into the item tensor along the hidden axis.

Every decoder prefill (normal, partial prefix-cache hit, resumed request, EC consumer) goes through the same `model.forward()`. The EC consumer's separate `prefill_decoder` path, `ec_disagg_helpers.py`, `PartialPrefixInfo`, the full/partial prefill builders, and the `mrope_position_deltas` side channel are removed.

Follow-ups included in this PR:

* The EC connector carries one tensor per `mm_hash`; the whole-prompt dict format (per-key registry, list flattening, non-tensor side channel) is gone. Producer and consumer must run the same version.
* `load_model` no longer asserts on `kv_block_adapter` for the EC producer, which has no KV cache.
* Prefix-cache KV copy failure falls back to a full prefill before `_preprocess` runs, so the rebuilt input still gets its multimodal embeddings.

**Model path:** optimum-rbln. `distributed/ec_transfer/` is shared but used by the optimum path only.

The optimum scheduler does not schedule encoder inputs, so the runner encodes every uncached item in the prefill window.

---

* Related to #803

---

### ✅ Type of Change

---

### 🧪 How to Test

1. Run `uvx pre-commit run --files $(git diff --name-only refactor/optimum-pad-in-runner...HEAD)`
2. Run
   ```bash
   uv run --no-sync pytest tests/optimum/v1/models/optimum tests/optimum/v1/worker tests/optimum/v1/distributed
   ```
   Expected at `0a9ceb5d`: `75 passed`.
3. Not yet run on an RBLN device: EC producer/consumer end to end, Qwen2/2.5/3-VL and EXAONE-4.5 inference, a prefix-cache boundary inside a media item, preempt/resume of a multimodal request.

---

### 📋 Checklist

---

### 💬 Notes

Review question: does the ownership boundary match upstream? Runner owns encoder execution, cache lookup, window slicing, and per-request positions. Models own the multimodal/MRoPE contracts and RBLN graph inputs.

`model_base.init_model()` still tells EC producer and consumer apart to decide which runtimes to load. That is about loading, not prefill, and is left as is.

---

<details>
<summary>Original description (kept for reference)</summary>

## Summary

This PR moves multimodal encoding, per-item encoder caching, embedding gathering, and MRoPE request state from the optimum model implementations into `RBLNOptimumModelRunner`.

The new flow follows the multimodal responsibilities of upstream vLLM v0.24.0:

1. The runner executes the multimodal encoder.
2. Encoder outputs are cached per multimodal item using `mm_hash`.
3. The runner gathers only the embeddings that overlap the current prefill window.
4. The model scatters the gathered embeddings and prepares RBLN-specific graph inputs.
5. MRoPE positions are computed once per request and sliced or extended by the runner.

As a result, normal prefill, partial prefix-cache hits, resumed requests, and EC consumers now share the same model `forward()` path.

**Model path:** `optimum-rbln`

---

## Motivation

The existing optimum multimodal path had accumulated several model-specific mechanisms that duplicate responsibilities handled by `GPUModelRunner` in upstream vLLM:

- A separate EC-consumer prefill path that bypassed `forward()`
- Full and partial variants of each prefill builder
- Model-side slicing of multimodal embeddings through `PartialPrefixInfo`
- A runner-owned `mrope_position_deltas` dictionary passed through model methods
- Whole-prompt multimodal dictionaries whose representation differed from the upstream `SupportsMultiModal` contract

These paths handled similar operations in different places, making prefix caching, EC disaggregation, and model-specific behavior difficult to reason about together.

This PR restores a single ownership boundary:

| Responsibility | Owner |
|---|---|
| Select multimodal items for the current prefill window | Runner |
| Execute the multimodal encoder | Runner |
| Cache encoder outputs by `mm_hash` | Runner |
| Slice cached outputs for partial prefix hits | Runner |
| Maintain per-request MRoPE positions and delta | Runner |
| Return one embedding tensor per multimodal item | Model |
| Scatter multimodal embeddings into text embeddings | Model |
| Convert positions to RBLN graph-specific cos/sin inputs | Model |
| Pack Qwen3-VL deepstack inputs | Model |

---

## Upstream alignment

The implementation mirrors the following upstream vLLM v0.24.0 concepts:

- `SupportsMultiModal.embed_multimodal`
  - Returns one 2D tensor per multimodal item, in prompt order.
- `GPUModelRunner._execute_mm_encoder`
  - Runs the encoder and caches each output by the corresponding item hash.
- `GPUModelRunner._gather_mm_embeddings`
  - Selects items overlapping the scheduled token window, slices their outputs, and builds `is_mm_embed`.
- `SupportsMRoPE.get_mrope_input_positions`
  - Computes the prompt's MRoPE positions and decode delta once per request.
- `CachedRequestState.mrope_positions` and `mrope_position_delta`
  - Store MRoPE state for subsequent prefill and decode steps.

One intentional difference remains: `RBLNOptimumScheduler` does not currently provide an encoder budget through `scheduled_encoder_inputs`. The optimum runner therefore encodes uncached items that overlap the current prefill window.

---

## New execution flow

```mermaid
flowchart LR
    S["Request state<br/>tokens + mm_features"]
    P["Runner prepares<br/>current token window"]
    E["_execute_mm_encoder<br/>encode uncached items"]
    C["encoder_cache[mm_hash]"]
    G["_gather_mm_embeddings<br/>slice to current window"]
    M["Model input preparation<br/>scatter + RBLN graph extras"]
    F["model.forward()"]

    S --> P --> E --> C --> G --> M --> F
```

All decoder-producing prefill paths now use this flow:

- Normal multimodal prefill
- Partial prefix-cache hit
- Resumed or preempted request
- EC consumer prefill

The EC producer runs only the encoder/cache portion and publishes each item through the connector.

---

### 1. Runner-owned encoder execution and cache

`RBLNOptimumModelRunner` now:

- Runs `model.embed_multimodal()` for uncached items in the current prefill window
- Validates that the encoder returns one output per input item
- Stores each output in `encoder_cache[mm_hash]`
- Publishes each item independently to the EC connector
- Frees entries through the scheduler's `free_encoder_mm_hashes`

The EC producer and consumer therefore use the same per-item cache representation as the normal path.

The previous whole-prompt cache representation and `ec_disagg_helpers.py` prefill bypass are removed.

### 2. Runner-owned embedding gathering

`_gather_mm_embeddings()` reads the cached outputs for items overlapping the current prefill window and returns:

- `mm_embeds`: cached item tensors in prompt order
- `is_mm_embed`: positions where those tensors must replace text embeddings

If a prefix-cache boundary falls inside an image or video placeholder, the runner selects only the uncached embedding rows using `PlaceholderRange.get_embeds_indices_in_range()`.

This replaces:

- `PartialPrefixInfo`
- `_mm_embed_tail_starts`
- `_build_partial_mm_embeds`
- Full/partial model-side prefill builders

### 3. Upstream-compatible model contracts

The Qwen-VL and EXAONE-4.5 implementations now return one 2D tensor per image or video item from `embed_multimodal()`.

`embed_input_ids()` receives the gathered item tensors and the runner-generated `is_mm_embed` mask.

For Qwen3-VL, each item's deepstack layers are concatenated along the hidden dimension:

```text
[num_item_tokens, hidden * (1 + num_deepstack_layers)]
```

This allows the encoder cache and partial-window slicing to treat the main embeddings and deepstack outputs as one per-item value. The model separates them again when preparing the compiled prefill graph.

### 4. Runner-owned MRoPE state

Qwen-VL models now implement `SupportsMRoPE.get_mrope_input_positions()`.

When a request is added, the runner computes and stores:

- `CachedRequestState.mrope_positions`
- `CachedRequestState.mrope_position_delta`

For each model invocation:

- Prefill takes the relevant slice of the precomputed prompt positions.
- Decode positions continue from the stored delta.
- A resumed prefill may combine the remaining prompt positions with completion positions.

The model only converts these positions into the cos/sin `position_embed` required by the compiled RBLN graph.

This removes the mutable `mrope_position_deltas` side channel and avoids recomputing MRoPE over the full prompt after a partial prefix-cache hit.

### 5. Single model-forward path

The EC consumer no longer builds `prefill_decoder` kwargs or invokes the runtime directly.

After the connector loads the encoder outputs, the consumer follows the same sequence as normal prefill:

```text
gather cached embeddings
→ build_prefill_forward_inputs
→ model.forward()
```

The following methods and branches are removed:

- `build_prefill_inputs_from_cache`
- `_build_partial_prefill_inputs_from_cache`
- `_build_full_prefill_forward_inputs`
- `_build_partial_prefill_forward_inputs`
- `_run_prefill_with_cached_encoder`

---

## Behavior changes

- **Resumed multimodal requests correctly re-inject media embeddings.**  
  Multimodal features are now read from persistent request state instead of only from `scheduled_new_reqs`.

- **Partial prefix-cache hits are handled by the current token window.**  
  The runner slices both multimodal embeddings and MRoPE positions to the uncached range.

- **EC cache entries are stored per item.**  
  Requests containing multiple images or videos no longer store the whole prompt output under the first item's hash.

- **EC consumers use the regular model forward path.**  
  There is no separate prefill-decoder invocation.

- **Qwen models no longer send pixel tensors to the model during prefill.**  
  Their model inputs contain gathered encoder outputs and precomputed MRoPE positions.

- **Whisper remains on the raw multimodal-input path.**  
  `multi_modal_kwargs` is retained for models that still perform encoding inside `forward()`.

- **Text-only multimodal calls return an empty item list.**  
  `embed_multimodal()` consistently returns `[]` when there is no media input.

---

## Scope

Main implementation changes:

- `vllm_rbln/v1/worker/optimum_model_runner.py`
- `vllm_rbln/model_executor/models/optimum/model_base.py`
- `vllm_rbln/model_executor/models/optimum/qwen2_vl.py`
- `vllm_rbln/model_executor/models/optimum/qwen3_vl.py`
- `vllm_rbln/model_executor/models/optimum/exaone4_5.py`

Removed obsolete abstractions:

- `PartialPrefixInfo`
- `vllm_rbln/v1/worker/ec_disagg_helpers.py`
- Model-side cached-prefill and partial-prefill builders

Related historical context: #803

---

## Testing

Run formatting and static checks:

```bash
uvx pre-commit run --files $(git diff --name-only refactor/optimum-pad-in-runner...HEAD)
```

Run the affected optimum model and runner tests:

```bash
uv run --no-sync pytest \
  tests/optimum/v1/models/optimum \
  tests/optimum/v1/worker
```

Expected result at `37438f8e`:

```text
60 tests passed
```

New tests cover:

- Splitting concatenated Qwen encoder output into per-item tensors
- Qwen3-VL per-item deepstack packing and graph-input reconstruction
- Scattering gathered embeddings with `is_mm_embed`
- Full and partial prefill-window gathering
- Prefix boundaries inside multimodal placeholders
- Structural placeholder tokens represented by `is_embed`
- Encoding only items missing from the runner cache
- Publishing each encoder output independently to the EC connector
- Prompt, decode, and resumed-window MRoPE positions
- Qwen2.5-VL and Qwen3-VL differences in MRoPE arguments
- Zeroing position embeddings for padded decode rows
- Ensuring the model does not run the vision encoder during prefill

---

## Validation still required

The unit tests use fake model/runtime objects. The following should be validated on an RBLN device before merge:

- End-to-end EC producer/consumer execution
- Qwen2-VL and Qwen2.5-VL inference
- Qwen3-VL inference, including deepstack
- EXAONE-4.5 multimodal inference
- Partial prefix-cache hits whose boundary falls inside a media item
- Preemption and resume of multimodal requests
- The optimum-backed Hugging Face `get_rope_index` integration
- Batched decode position embeddings

---

## Type of change

- [ ] Release
- [ ] Feature
- [ ] Model support
- [ ] Core engine change
- [ ] Bug fix
- [ ] Performance improvement
- [x] Refactor or code cleanup
- [ ] Documentation
- [ ] Other

---

## Checklist

- [x] Test commands and expected results are documented
- [x] Unit tests cover the new runner-owned responsibilities
- [ ] End-to-end tests have been run on an RBLN device
- [ ] This PR is linked to a development issue
- [ ] Relevant documentation has been updated, if required

---

## Notes for reviewers

The main review question is whether the ownership boundary now matches upstream vLLM:

- The runner owns encoder execution, encoder-cache lookup, window slicing, and request-level positions.
- Models implement the multimodal and MRoPE contracts and prepare only RBLN graph-specific inputs.

`model_base.init_model()` still distinguishes EC producers and consumers when deciding which optimum runtimes to load. That distinction concerns runtime loading, not prefill execution, and is intentionally left unchanged.

</details>


vllm_rbln/distributed/ec_transfer/ec_connector/rbln_ec_nixl_connector.py
```diff
@@ -943,7 +922,7 @@ def _initiate_pull(
         )
 
         # Initiate async READ (llm pulls from encoder)
-        indices = list(range(num_descs))
+        indices = [0]
```
rebel-yskim, line 925: nit, remove and use `[0]` directly? ?

vllm_rbln/v1/worker/optimum_model_runner.py
rebel-yskim, line 316: in EC disagg cases, producer does not have kv block (they do not load), please check.

rebel-yskim (changes requested):

vllm_rbln/distributed/ec_transfer/ec_connector/rbln_ec_nixl_connector.py
rebel-eunji: Since indices is used for both arguments, leaving it as is seems fine for readability — it makes it easy to see that the local and remote handles are the same.
```
def make_prepped_xfer(
    xfer_type: str,                   # Transfer type ("READ" or "WRITE")
    local_xfer_side_handle: int,      # Handle ID for the local transfer side
    local_block_descs_ids: list[int], # IDs of the local KV block descriptors
    remote_xfer_side_handle: int,     # Handle ID for the remote (target) transfer side
    remote_block_descs_ids: list[int],# IDs of the remote KV block descriptors
    notif_msg: Optional[bytes] = None # (Optional) Notification message payload sent on completion
) -> int:                             # Returns the handle ID of the created transfer session
```

vllm_rbln/v1/worker/optimum_model_runner.py
rebel-eunji: That's right. Thanks.

rebel-yskim: I'm curious that - I think the assertion is appended in `refactor/optimum-pad-in-runner`.
If you merge `refactor/optimum-pad-in-runner` PR first, then EC disagg would not work. What about fix in `refactor/optimum-pad-in-runner`? (If you want to merge this PR first, just let me know)

rebel-eunji: It would be better. I'll fix in refactor/optimum-pad-in-runner branch.

vllm_rbln/v1/worker/optimum_model_runner.py
```diff
@@ -386,82 +376,65 @@ def execute_model(
...
                 scheduler_output
             )
-
             if not self.try_copy_prefix_cached_kv(model_input, scheduler_output):
                 model_input, _ = self._prepare_inputs(
                     scheduler_output, use_cached_prefix=False
                 )
+            ec_connector_output = None
+            if isinstance(self.model, RBLNOptimumMultimodalMixin):
+                model_input, ec_connector_output = self._preprocess(
```
rebel-yskim, line 402: I think vision encoder is called in here - is it right?
then `capture_reports` does not contain vision encoder time now.

rebel-eunji: 1. Yes.
2. I don't think encoding is included in forward(). Should we add it there?

rebel-yskim: I think `VLLM_RBLM_METRICS` should contain the vision encoder time. How do you think about it? (Dev contained it, but this PR ignores it)

rebel-eunji: I'll fix it when I refactor VLLM_RBLN_METRICS in optimum_model_runner. 
(Currently, metric code in type1 is not managed well....:( )

rebel-yskim (approved):
LGTM 👍

rebel-jonghewk (approved):

PR #1064, opened 2026-09-07
feature(config): introduce RBLNConfig on the vLLM-native path

### LINKED ISSUE #1047: proposal(config): move the `VLLM_RBLN_*` options that change model behavior into `RBLNConfig`
## Summary

```yaml
VLLM_RBLN_USE_VLLM_MODEL: "1"
VLLM_RBLN_BATCH_ATTN_OPT: "1"
VLLM_RBLN_SORT_BATCH: "1"
```

| --- | --- | --- |

```
```

## Scope

These stay in `envs.py`:

| Variable | Why it stays |
| --- | --- |

| Old env var | Field today |
| --- | --- |

## What it would look like

### Cache Key

## Plan, in steps

## Not in this proposal

## Current status

```bash
RBLNConfig:

                        (default: False)

...
```

rebel-cjlee:
### 🚀 Summary of Changes

RBLN options on the vLLM-native path get a real config. `RBLNConfig` holds 23 fields, each fields makes one `--rbln-<field>` flag, and `additional_config` is how a value gets in.

`envs.py` still works and is still the live source. Only two read sites moved, so this PR does not change what any option does. It adds the surface.

**One design choice needs a look.** On the native path `additional_config` *is* the `RBLNConfig`, not a dict with an `"rbln"` key inside it.
`check_and_update_config` replaces what it was given with the resolved object. The proposal assumed we could not do that, because `additional_config` holds the optimum path's keys. That is not true on the native path: every optimum key is written under `sync_vllm_and_optimum`, which the native path never calls. The one exception was `user_max_num_batched_tokens`, which is ours, so a path gate removes it.

Making it the object solves two problems the proposal left open:

- **It reaches the other processes.** It is a `VllmConfig` field, so it is pickled to the worker and to EngineCore. No copy-back, no re-resolving.
- **The cache key works without a workaround.** `VllmConfig.compute_hash()` calls `additional_config.compute_hash()` when it is not a dict, so `ignored_factors` takes effect. The proposal needed a strip-and-restore trick for this. That trick is gone.

`RBLNConfig.compute_hash()` leaves out the five fields that do not change the compiled artifact, so turning one of them on keeps the mega-cache bundle.

Two smaller decisions worth knowing:

- `RBLNConfig` uses `vllm.config.utils.config`, like all 27 config classes in `vllm/config/`. That converts the types that arrive as JSON. `{"use_w8a8": "false"}` used to be a true string and is now `False`, and an invalid `decode_batch_bucket_strategy` u;sed  to pass silently.
- `get_rbln_config()` raises instead of falling back to the environment. A child process inherits env vars but not `--rbln-*` values, so a fallback would be right when the option came from the environment and wrong when it came from the command line. That is the one failure this change has to avoid.

**Two limits this brings.** On the native path `additional_config` is ours alone, so a user cannot put an unrelated key there. And `--gdn-prefill-backend` stops working: upstream writes that flag into `additional_config` without a guard, so it arrives as a field we do not have. It fails with our own message. It is the only upstream flag that writes there, and RBLN does not use it.

---

* Related to #1047 

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

1. `uv run --no-sync pytest testse/native/test_rbln_config.py -x`

17 tests: flag generation, the priority order, type conversion, unknown-key rejection, and which fields change the hash.

2. `uv run --no-sync pytest test/native -q`

1799 passed, 25 skipped

3. `vllm serve --help=rblnconfig` shows 23 flags with `VLLM_RBLN_USE_VLLM_MODEL=1`, and no `RBLNConfig` group without it.

4. The two moved read sites, one per process type:
  - `VLLM_RBLN_SORT_BATCH=1 vllm serve ...` and `vllm serve --rbln-sort-batch ...` must behave the same. The value is read in the worker process.
  - same for `--rbln-sub-block-cache`, which is read in EngineCore.
  Each process logs its resolved config as `RBLN config: {...}`, so you can see the value arriave.

---

---

### 📋 Checklist

---

### 💬 Notes

Look at these three:

1. **We take over `additional_config` on the native path.** Upstream calls that field an "opaque config ... for out of tree config registration", so we give up something upstream offers.
2. **The env vars print no deprecation warning yet.** `envs.py` is still the live source, so calling it deprecated would not be true. The warning belongs with the read-site migration.
3. **The two moved read sites are there on purpose.** Without a reader there is no way to see whether the config reaches the worker. If you would rather have zero behavior change, I can move them to the next PR.

Not done here: the other 93 read sites, the `tests/native/` env harness, the docs, and removing the promoted names from `RBLN_COMPILE_ENV`.

---

### DIFF
--- a/tests/native/conftest.py
+++ b/tests/native/conftest.py
@@ -517,6 +517,35 @@ def _drop_envs_shadows():
 
 
 @pytest.fixture(autouse=True)
+def _reset_rbln_config(monkeypatch):
+    """`vllm_rbln.config` publishes the resolved config in a module global.
+
+    A worker or scheduler built by one test leaves it behind, so a test that
+    never published one would silently read another test's.
+    """
+    from vllm_rbln import config
+
+    monkeypatch.setattr(config, "_rbln_config", None)
+
+
+@pytest.fixture
+def rbln_config():
+    """Publish an `RBLNConfig` for code that reads it without an engine.
+
+    Only a process with an engine resolves one by itself, so a unit test has
+    to say what it wants to read.
+    """
+    from vllm_rbln.config import RBLNConfig, set_rbln_config
+
+    def _publish(**overrides) -> RBLNConfig:
+        config = RBLNConfig(**overrides)
+        set_rbln_config(config)
+        return config
+
+    return _publish
+
+
+@pytest.fixture(autouse=True)
 def _isolate_rbln_ctx_standalone():
     """Clear the one env var the code under test writes to the process env.
 
--- a/tests/native/test_rbln_config.py
+++ b/tests/native/test_rbln_config.py
@@ -1,0 +1,164 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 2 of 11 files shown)


rebel-jonghewk (approved):
lgtm

PR #1065, opened 2026-09-07
fix(model): support to deepseekv32 mtp

rebel-kblee:
### 🚀 Summary of Changes

Details regarding support for DeepSeek v3.2 MTP in the current VLLM upstream version.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/platform.py
```diff
@@ -423,6 +423,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
 
                 cls._validate_eagle3_pp_config(vllm_config)
 
+            # TODO(kblee): drop when upstream stops forcing v32 MTP eager.
```
rebel-cjlee: Can we make the drop condition checkable? Upstream removed this in [vllm#52861](https://github.com/vllm-project/vllm/pull/52861/changes#diff-d9f90bb1c6d4f572294cec8ce7251c46224d8f6ecf5918d7f8a9fb768bb65dbbL762-L765), and that is not in v0.28.0 yet, so the comment could say `TODO(vllm>=0.29.0): delete` and name the PR.

One more thing. The block only runs when `spec_config.enforce_eager` is already True, so when upstream stops forcing it this code just does nothing and nobody notices. `gpt_oss.py` and `multi_connector.py` add a module level assert for this, something like `assert VLLM_VERSION < (0, 29)`. Could we do something like that here?

rebel-cjlee: A small test would also help. `recognize` in `tests/native/test_platform.py` can set `hf_text_config.model_type` to `deepseek_v32` and check that `enforce_eager` ends up False.

rebel-cjlee (approved):

PR #1069, opened 2026-09-07
feature: remove VLLM_RBLN_SORT_BATCH and update the sorting logic

rebel-eunji:
### 🚀 Summary of Changes

VLLM_RBLN_SORT_BATCH is removed and...
1. In optimum_model_runner, reordering works only when `requires_batch_sort config` is True
2. In rbln_model_runner, reordering works in CR13 or when VLLM_RBLN_BATCH_ATTN_OPT=1

---

* Resolves #
* Related to https://github.com/RBLN-SW/optimum-rbln/pull/559

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/worker/optimum_model_runner.py
```diff
@@ -1104,12 +1118,13 @@ def _update_states(self, scheduler_output: "RBLNSchedulerOutput") -> None:
...
-        sequences per partition — the kernel processes the first valid_batch[p]
-        rows for partition p, which is only correct when rows are sorted long→short.
+        Enabled by `VLLM_RBLN_SORT_BATCH=1` or compiled-model metadata. Required
+        for the batched dynamic decode kernel (VLLM_RBLN_BATCH_ATTN_OPT) to
+        early-exit on shorter sequences per partition — the kernel processes the
+        first valid_batch[p] rows for partition p, which is only correct when rows
+        are sorted long→short.
         """
-        if not envs.VLLM_RBLN_SORT_BATCH:
+        if not (envs.VLLM_RBLN_SORT_BATCH or self.sort_batch_by_length):
```
rebel-jonghewk: @rebel-eunji Can we just use `self.sort_batch_by_length` for optimum model impl instead of using `VLLM_RBLN_SORT_BATCH` ?

rebel-eunji: Yes. I will change it and just remove VLLM_RBLN_SORT_BATCH env var clearly in this PR.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -480,6 +480,14 @@ def __init__(
             parallel_config.data_parallel_size > 1
             and envs.VLLM_RBLN_SPECIALIZE_MOE_DECODE
         )
+        # The batched dynamic decode kernel (REBEL CR13, or any device with
+        # VLLM_RBLN_BATCH_ATTN_OPT) processes the first valid_batch[p] rows of
+        # partition p and early-exits on the rest, which is only correct when
+        # rows are sorted by descending sequence length.
+        self.sort_batch_by_length = (
+            "cr13" in current_platform.get_device_name().lower()
+            or envs.VLLM_RBLN_BATCH_ATTN_OPT
```
rebel-eunji: flash_attention.py is still relying on envs.VLLM_RBLN_BATCH_ATTN_OPT, not get_rbln_config().batch_attn_opt.
So I just set it like this. @rebel-cjlee

rebel-jonghewk: @rebel-eunji  How about sth like this? Just don't feel good about adding 'cr', 'ca' here and there.

```
# vllm_rbln/platform.py
_BATCHED_DECODE_SOCS = frozenset({"RBLN-CR13"})

@classmethod
@functools.cache
def uses_batched_dynamic_decode(cls) -> bool:
    return (
        envs.VLLM_RBLN_BATCH_ATTN_OPT
        or cls.get_device_name().strip().upper() in _BATCHED_DECODE_SOCS
    )
```

rebel-jonghewk (commented):
Overall this looks good, with a few minor nits. (I'd been hoping to remove the sort batch for a while, so this is a welcome change.)

rebel-cjlee (approved):

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-thkim: Just a quick question: 
1. For Type 2 models like GPT-OSS in CA-12, doesn't sorting require that VLLM_RBLN_BATCH_ATTN_OPT be enabled?
2. Are requests processed in order of length, regardless of the order in which they were received?

rebel-jonghewk (approved):

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-eunji: 1. As @rebel-mhkang said, the users will use VLLM_RBLN_BATCH_ATTN_OPT well for now. We decided to use it until batch attn opt is enabled in ATOM in the future.
2. Not quite. The order in input_batch is not change the order of scheduling.

+) For more details...
The sort does not change which requests get served or when; the scheduler still admits them in its own order, and that part is untouched. It only permutes the rows of the worker's persistent batch, so that within one step the longest sequence sits in row 0. Every request in the batch is computed in the same forward pass either way, and each request's output is the same regardless of its row.

The sort runs in `_update_states` every step but returns early when the batch is already in descending order, so in steady-state decode it is a no-op. It actually reorders when a new or resumed request enters, when a finished one is condensed out, or when a request in chunked prefill grows past its decode neighbours.

PR #1076, opened 2026-09-09
other: support Python 3.14, on torch-rbln 0.5.0rc0

rebel-junamsong:
## Summary

Widen `requires-python` to `>=3.10,<3.15` and bump `torch-rbln` to
`0.5.0rc0`, the first build that ships cp314 wheels — previously the one
blocker on this range. The bound is `>=0.5.0rc0,<0.5.1`, the patch-line
form `tests/test_packaging_bounds.py` requires.

## What changed

| surface | change |
| --- | --- |
| `pyproject.toml` | `requires-python` → `>=3.10,<3.15`, the 3.14 classifier, `torch-rbln>=0.5.0rc0,<0.5.1` |
| `uv.lock` | re-locked; torch-rbln is the only version change, cp314 wheel URLs added |
| `tests/test_python_support.py` | new — holds `requires-python` to `uv.lock`: every locked package must ship a distribution installable on each admitted minor |
| `.buildkite/pr.yml` | the new test joins the dependency-bounds lane, renamed `:package: packaging guards` to cover both |
| `.github/workflows/dispatch_pkg_update.yaml` | `3.14` added to the `python-version` choices |
| `DEVELOPMENT.md` | stated range 3.10–3.13 → 3.10–3.14 |

## Why the test — a green `uv lock` does not mean 3.14 installs

`uv lock` resolves versions and `requires-python` metadata; it never checks
wheel ABI-tag coverage. On the commit before the torch-rbln bump the lock was
valid and the uv-lock hook green, yet:

```
$ uv sync --frozen --python 3.14 --dry-run
error: Distribution `torch-rbln==0.4.0` can't be installed because it doesn't
have a source distribution or wheel for the current platform
hint: You're using CPython 3.14 (`cp314`), but `torch-rbln` (v0.4.0) only has
wheels with the following Python ABI tags: `cp310`, `cp311`, `cp312`, `cp313`
```

Every CI lane syncs on 3.12, so without this test the first 3.14 user is the
one who finds such a gap — including any future auto-bump that drops a cp314
wheel. Reverting the torch-rbln bump turns
`test_lock_covers_every_admitted_minor[3.14]` red.

## Test

```
uv run --no-project --python 3.12 --with pytest --with packaging \
  python -m pytest tests/test_packaging_bounds.py tests/test_python_support.py -q
```

9 passed.

## Note for repo admins

The commit-status context `[OB] dependency bounds` is renamed to
`[OB] packaging guards`; if the old name is a required status check in branch
protection, update it.

### DIFF
--- a/.buildkite/pr.yml
+++ b/.buildkite/pr.yml
@@ -34,16 +34,7 @@ steps:
     image: "${DEVTOOLS_DOCKER_IMAGE}"
     secrets: *uv_secrets
     command:
-      - "uv run --no-project --python 3.12 --with pytest --with packaging python -m pytest tests/test_packaging_bounds.py -q"
-
-  - label: ":snake: python support"
-    notify:
-    - github_commit_status:
-        context: "[OB] python support"
-    image: "${DEVTOOLS_DOCKER_IMAGE}"
-    secrets: *uv_secrets
-    command:
-      - "uv run --no-project --python 3.12 --with pytest --with packaging python -m pytest tests/test_python_support.py -q"
+      - "uv run --no-project --python 3.12 --with pytest --with packaging python -m pytest tests/test_packaging_bounds.py tests/test_python_support.py -q"
 
   # All gates must pass before the heavy test pipelines are even uploaded.
   - wait: ~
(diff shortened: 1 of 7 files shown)


rebel-seinpark (approved):
👍

PR #1077, opened 2026-09-09
perf(sampler): trim the host work around the rejection-sample graph

rebel-daeyang:
### 🚀 Summary of Changes

Follow-up to #1042: cuts the rejection sampler's per-step host work under `VLLM_RBLN_USE_DEVICE_TENSOR=1`.

**Problem** -- #1042 folded `rbln::rejection_sample` and the `[B, K+1]` output composition into one compiled graph. The host still did work around that graph on every speculative step: it launched a second graph for the bonus token, asked for float32 logits it rarely read, copied already-packed rows into zeroed buffers, and allocated the graph's small inputs anew, so the runtime re-validated and re-bound them each step.

**Solution** -- Three changes.

1. **Skip what is not needed.** Ask the bonus sampler for its logits only when logprobs are requested; the float32 widening is a host round trip on the device. A step whose requests all carry K drafts is already packed and padded for the op, so its rows are handed over as they are. The per-batch draft view gets its own buffer: two graph inputs aliasing one allocation lost the accepted drafts past the first.
2. **Greedy bonus token inside the graph.** An all-greedy step without logprobs hands the graph the bonus rows instead of sampled ids, and the graph takes their argmax with `rbln::argmax`. One launch instead of two. Every other step -- random or mixed rows, logprobs, the torch impl -- passes the ids as before. Warm-up compiles both variants, with int32 ids as the sampler's ops return them, so the first non-greedy step no longer recompiles.
3. **Reuse the graph's small inputs.** The `[B*K]` draft ids, `[B, K]` per-batch view, temperature, cu and count tensors are kept per batch size and handed over at the same addresses every step. Rows past the packed length keep an earlier step's values; the op reads them only into slots its acceptance count clips. The runner hands the draft ids over on the host and each impl brings them where it samples. A static output buffer was tried too and measured at noise, so it is not included.

vLLM-native path only.

---

* Follow-up to #1042

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

MiniMax-M2.7 + Eagle3 spec decode, dp4: output tokens match device-tensor mode, `_sample` is about a fifth faster, tpot improves. Unit tests: `tests/torch_compile/unit/v1/sample/test_rbln_rejection_sampler.py`, `tests/native/v1/worker/test_rbln_model_runner.py`.

---

### 📋 Checklist

### DIFF
--- a/tests/native/v1/worker/test_rbln_model_runner.py
+++ b/tests/native/v1/worker/test_rbln_model_runner.py
@@ -355,8 +355,17 @@ def test_rejection_sampler_warmup_uses_per_stage_batch_bound(monkeypatch):
 
     runner._warmup_sampler_decode_batches()
 
-    assert rejection_sample.call_count == 1
-    assert len(rejection_sample.call_args.args[1]) == 4
+    # One call per bonus-token graph: the argmax-in-graph one an all-greedy
+    # step without logprobs takes, and the pre-sampled-ids one every other
+    # step takes. Both are warmed at the same batch bound.
+    assert rejection_sample.call_count == 2
+    for call in rejection_sample.call_args_list:
+        assert len(call.args[1]) == 4
+    variants = [
+        (call.args[6] is None, call.kwargs["bonus_logits"] is None)
+        for call in rejection_sample.call_args_list
+    ]
+    assert sorted(variants) == [(False, True), (True, False)]
 
 
 class TestPredicates:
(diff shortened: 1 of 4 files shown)


rebel-eunik (approved):
LGTM

rebel-cjlee (approved):

rebel-jonghewk (approved):

PR #1078, opened 2026-09-09
perf(spec-decode): keep the EAGLE drafter's bookkeeping on the host

rebel-daeyang:
### 🚀 Summary of Changes

Follow-up to #1042: cuts the EAGLE draft latency under `VLLM_RBLN_USE_DEVICE_TENSOR=1`.

**Problem** -- torch-rbln runs every eager op with an **int/bool input through a CPU fallback**. Under device tensors, the drafter's glue between its draft graphs -- token ids, positions, backup ids, the d2t mapping, the index arithmetic of `prepare_inputs_padded` -- round-tripped the host op by op. Each trip is a launch plus a D2H.

**Solution** -- **Allocate that bookkeeping on the host.** The device keeps only the attention metadata builders and the input stager, which stage the graph inputs onto it. The draft graph is unchanged. Draft ids come back with one D2H per pass; d2t sits on the host next to them. A step without drafts moves its device-sampled ids to the backup ids' device.

**This is how the target model already runs.** The runner keeps `input_ids`, `positions` and `query_start_loc` on the host, does its bookkeeping there in eager torch, and stages the inputs onto the device right before `execute_model`. The EAGLE drafter now follows the same pattern.

---

* Follow-up to #1042

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

MiniMax-M2.7 + Eagle3 spec decode, dp4: output tokens match device-tensor mode, tpot improves.

---

### 📋 Checklist

### DIFF
--- a/tests/native/v1/spec_decode/test_eagle.py
+++ b/tests/native/v1/spec_decode/test_eagle.py
@@ -113,7 +113,13 @@ def _echo_model_exec(hidden_size):
     return executable
 
 
-def _call_propose(proposer, target_hidden_states=None):
+# The last slot of each request in _call_propose's [0, 2, 4] query layout.
+_LAST_SLOTS = torch.tensor([1, 3], dtype=torch.int32)
+
+
+def _call_propose(
+    proposer, target_hidden_states=None, token_indices_to_sample=_LAST_SLOTS
+):
     return proposer.propose(
         target_token_ids=torch.arange(4, dtype=torch.int32),
         target_positions=torch.arange(4, dtype=torch.int64),
@@ -121,7 +127,7 @@ def _call_propose(proposer, target_hidden_states=None):
         if target_hidden_states is None
         else target_hidden_states,
         next_token_ids=torch.tensor([30, 31], dtype=torch.int32),
-        token_indices_to_sample=torch.tensor([1, 3], dtype=torch.int32),
+        token_indices_to_sample=token_indices_to_sample,
         common_attn_metadata=make_cad([0, 2, 4], [10, 11]),
     )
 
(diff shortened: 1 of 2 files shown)


rebel-minkyu (approved):
LGTM!

vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -60,7 +60,10 @@ def __init__(
         device: torch.device,
         runner: "RBLNModelRunner",
     ):
-        super().__init__(vllm_config, device, runner)
+        # Bookkeeping on the host: torch-rbln runs int/bool eager ops through a
+        # CPU fallback. The stager still stages the graph inputs onto `device`.
+        super().__init__(vllm_config, torch.device("cpu"), runner)
+        self.device = device
```
rebel-eunik: While there is a comment explaining the CPU fallback, it's still a bit unclear exactly what impact this has on the parent class, and what remains unaffected since `self.device` is overwritten immediately after.
Could you expand the comment to briefly clarify this?

rebel-jonghewk (approved):
Can you share the relative perf gain this PR gives?

vllm_rbln/v1/spec_decode/eagle.py
rebel-daeyang: Thanks for the feedback -- reflected it in c1cb5b69. Could you take another look?

tests/native/v1/spec_decode/test_eagle.py
rebel-eunik: It looks like `test_first_pass_shifts_tokens_and_inserts_next` and `test_first_pass_uses_explicit_token_indices_verbatim` are now verifying the exact same thing

rebel-daeyang: Right, I applied the feedback in 434498d

rebel-eunik (approved):
LGTM

rebel-cjlee (approved):

PR #1079, opened 2026-09-09
other(bump): sync with vLLM 0.26.0

rebel-cjlee:
### 🚀 Summary of Changes

This PR moves vllm-rbln from vLLM 0.24.0 to 0.26.0.

Most of the work is not new features. It is adapting to upstream API changes, plus deleting two patches that upstream now ships itself.

#### Patches we deleted (0.26.0 has the same fix)

- `patches/gpt_oss.py`: the `load_weights` backport (EP rank bug)
- `patches/multi_connector.py`: the module, its entry in `patches/__init__.py` and its test

#### Upstream API changes we followed

| Upstream PR | What upstream changed | What we changed |
| --- | --- | --- |
| [#47728](https://github.com/vllm-project/vllm/pull/47728) | Renamed `max_num_batched_tokens` to `max_in_flight_tokens`. The value is also different now: `max_concurrent_batches * max_num_batched_tokens`. `remove_skipped_blocks()` params changed too. | optimum KV cache coordinator / manager / scheduler, `RBLNSlidingWindowManager`, one test |
| [#47782](https://github.com/vllm-project/vllm/pull/47782) | `get_computed_blocks()` returns a third value, `shared_prefix_boundary`. | `RBLNScheduler.schedule()` stores it, the same way upstream does |
| [#46384](https://github.com/vllm-project/vllm/pull/46384) | Partial prefix cache hits. `find_longest_cache_hit()` now returns `(blocks, hit_length)`, `get_num_blocks_to_allocate()` has one more argument, and a partial tail block is now copied by upstream itself. | `RBLNSlidingWindowManager` signatures. Also: our scheduler no longer reports sub-block tokens as "computed" tokens, because upstream reads a count that is not block-aligned as its own partial hit. |
| [#40996](https://github.com/vllm-project/vllm/pull/40996) | `InputBatch` now requires `max_num_blocks_per_req`. `MultiGroupBlockTable` no longer computes it. | The model runner computes the value for the placeholder batch |
| [#45880](https://github.com/vllm-project/vllm/pull/45880) | The push scheduler now sets `remote_block_ids` for a rejected request. The NIXL worker keeps `pp_size`. | Deleted our `RblnNixlPushConnectorScheduler.request_finished()` workaround, which existed for exactly this bug |
| [#48717](https://github.com/vllm-project/vllm/pull/48717) | `_logical_to_remote_kernel_block_ids()` is gone. It is now `_logical_to_kernel_block_ids(ids, ratio)`. | Renamed both call sites in the RBLN NIXL workers. This one was a real crash: the per-shard read and write paths called a method that no longer exists |

#### Test changes

Some test fixtures build a worker with `object.__new__` and set only the attributes the pinned vLLM reads. They now also set `pp_size`, `_bidirectional_kv_xfer_enabled` and `_physical_blocks_per_logical_kv_block`.

We also added tests that call out overrides **through the real coordinator**, not directly. Calling them directly hid three broken signatures, because the test then follows our own signature instead of what the coordinator passes.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)
* [x] ❓ Other (`other`): dependency bump to vLLM 0.26.0

---

### 🧪 How to Test

1. Unit tests: `uv run --no-sync pytest test/native -q`
2. Compile lane:
  `uv run --no-sync pytest tests/native --model-compile -m model_compile`

Expected: all pass.

---

---

### 📋 Checklist

---

### 💬 Notes

- @rebel-huijong Please look closely at the sub-block cache change in `rbln_scheduler.py`. Upstream now has its own partial-block reuse, and ours works differently. Both counts still add up to the same total, so the number of blocks upstream allocates does not change.
- Not part of this PR: `examples/experimental/run_specdec_suffix_test.py` fails to compile when `RBLN_USE_CUSTOM_KERNEL=1`. With `0`, the same script runs fine. This will be fixed when examples are renewed.

---


rebel-eunji (approved):
LGTM 👍

PR #1080, opened 2026-09-09
fix(dp): give each data parallel rank its own NPUs when a rank owns several

rebel-cjlee:
### 🚀 Summary of Changes

Under data parallelism, DP ranks did not get their own NPUs.

A rank took its NPUs at `local_rank * VLLM_RBLN_NUM_DEVICES_PER_LOCAL_RANK`. This index does not include the DP rank, so every replica started at the same place. It also went through `device_id_to_physical_device_id()`, which prefers vLLM's DP mapping. That mapping has one device per rank, but an RBLN rank owns RSD devices, so the engine failed to start.

2 changes:
- The slot now spans the whole deployment: `data_parallel_rank_local * world_size + local_rank`.
- The pool is read with `visible_device_id_physical_device_id()`, which ignores that mapping by design.

TP2 / DP2 / RSD2 now gets 0-3, 4-7, 8-11, 12-15.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit tests:
  `uv run --no-sync pytest tests/native/v1/worker/test_rbln_worker.py -k TestInitDeviceEnv`
  28 pass
2. On ATOM, serve with TP2 / DP2 / RSD4

---

---

### 📋 Checklist

---

### 💬 Notes

- `--device-ids` is still ignored on the native path. vLLM slices it per rank before the worker sees it, so the full list cannot be recovered. Use `RBLN_VISIBLE_DEVICES` instead. Fixing that would need a patch on vLLM's DP slicer, which this PR does not do.

---

### DIFF
--- a/tests/native/compile/models.py
+++ b/tests/native/compile/models.py
@@ -33,6 +33,17 @@ _QWEN3_30B_A3B_BASE = CompileModelSpec(
     },
     OPT_ENVS,
 )
+_QWEN1_5_MOE_A2_7B_BASE = CompileModelSpec(
+    "Qwen/Qwen1.5-MoE-A2.7B",
+    {
+        "max_num_seqs": 1,
+        "max_model_len": 8192,
+        "block_size": 4096,
+        "enable_expert_parallel": True,
+    },
+    OPT_ENVS,
+    chips=ATOM,
+)
 _MINIMAX_BASE = CompileModelSpec(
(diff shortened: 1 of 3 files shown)


rebel-jonghewk (approved):
lgtm

PR #1085, opened 2026-09-10
fix(compilation): raise the dynamo recompile limits

rebel-cjlee:
### 🚀 Summary of Changes

This PR changes the dynamo recompile limits set by `_ensure_torch_dynamo_configured()`. It sets `recompile_limits = 256` and `accumulated_recompile_limits = 256` instead of `cache_size_limit = 64`.

Two things change here:

- The value. #929 raised these two limits on the optimum path. There the value is computed from the number of sampler buckets. The native path just sets a value that is large enough. 256 does not come from a measurement. We can change it later, once we know how many recompiles a real workload needs.
- The names. In torch 2.11, `cache_size_limit` is only an old name for `recompile_limits`. Using the new names keeps this file consistent with the optimum path code.

The test was updated for both names, so that the fixture saves and restores the same two values it changes.

---

* Resolves #
* Related to #929

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `uv run --no-sync pytest tests/native/compilation/test_compiler.py -q`.
2. All 24 tests pass.

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/vllm_rbln/compilation/compiler.py
+++ b/vllm_rbln/compilation/compiler.py
@@ -38,7 +38,8 @@ def _ensure_torch_dynamo_configured() -> None:
     # To prevent nn.modules parameters to be modmel input, set false.
     # If this flag is set, nn.modules parameters are treated as model input.
     torch._dynamo.config.inline_inbuilt_nn_modules = False
-    torch._dynamo.config.cache_size_limit = 64
+    torch._dynamo.config.recompile_limit = 256
+    torch._dynamo.config.accumulated_recompile_limit = 256
 
     _DYNAMO_CONFIGURED = True
 
--- a/tests/native/compilation/test_compiler.py
+++ b/tests/native/compilation/test_compiler.py
@@ -38,12 +38,17 @@ def _dynamo_isolation():
     cfg = torch._dynamo.config
     saved = (
         cfg.inline_inbuilt_nn_modules,
-        cfg.cache_size_limit,
+        cfg.recompile_limit,
+        cfg.accumulated_recompile_limit,
         compiler._DYNAMO_CONFIGURED,
     )
     yield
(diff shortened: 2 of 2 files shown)


vllm_rbln/compilation/compiler.py
```diff
@@ -38,7 +38,8 @@ def _ensure_torch_dynamo_configured() -> None:
     # To prevent nn.modules parameters to be modmel input, set false.
     # If this flag is set, nn.modules parameters are treated as model input.
     torch._dynamo.config.inline_inbuilt_nn_modules = False
-    torch._dynamo.config.cache_size_limit = 64
+    torch._dynamo.config.recompile_limit = 256
+    torch._dynamo.config.accumulated_recompile_limit = 256
```
rebel-eunji, line 42: 1. What is the reason of setting to 256? 
2. optimum_model_runner also call compile function in sampler.
3. I think setting here is not a good choice. For example, rbln_model_runner compiles both model and sampler. Then it sets twice here. I think it should be called and set only once.

rebel-cjlee: 1. just large number enough, no specific reason, no measurement
2. Good catch, I actually missed that part, For now, I think we can use `VLLM_RBLN_USE_VLLM_MODEL` to ensure it doesn't affect the optimum path.
3. However, managing this within the runner also feels a bit awkward to me. Furthermore, this configuration is only set during the initial call, not every time `compile(...)` is called. Let me know if there's something I might be missing here.

rebel-cjlee: For resolving 2 and 3, I am thinking of extracting the part of dynamo configuration setting from compile method and creating a new method. Then, runner can call it at the initialization time. How about it?

It's currently set to 256, but we may need to scale it later depending on the buckets.

PR #1086, opened 2026-09-10
fix(worker): a model with no resolvable language model still loads

rebel-junamsong:
### 🚀 Summary of Changes

Fixes a model-loading failure triggered by the batch-sort capability probe when serving transcription-only models such as Whisper.

`_should_sort_batch_by_length` checks whether either the top-level model or its language submodel requires batch sorting. Previously, it used the presence of `get_language_model` to determine whether the language submodel should be inspected.

However, models implementing upstream `SupportsMultiModal` inherit `get_language_model` even when they do not have a resolvable language model. For example, calling this method on `RBLNOptimumWhisperForConditionalGeneration` raises `NotImplementedError`, preventing the model from loading.

This PR changes the capability check to use the RBLN-specific model interface:

* Always check the top-level `rbln_config.requires_batch_sort`.
* Inspect the language submodel only when the model implements `RBLNOptimumMultimodalMixin`.
* Skip the language-submodel check for models such as Whisper that implement `SupportsMultiModal` directly.

This approach uses an explicit RBLN contract instead of relying on method availability or suppressing `NotImplementedError`. It also avoids hiding genuine implementation errors from RBLN multimodal models.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run:

   ```bash
   pytest tests/optimum/v1/worker/
   ```

2. Verify that the existing batch-sort tests correctly inspect both the top-level model configuration and the language submodel configuration for models implementing `RBLNOptimumMultimodalMixin`.

3. Verify that the Whisper-like regression test returns `False` without calling the unresolvable `get_language_model`, allowing model loading to continue.

Expected result: all tests pass.

---

Before this fix, serving Whisper failed during model loading:

```text
File "vllm_rbln/v1/worker/optimum_model_runner.py", line 316, in _should_sort_batch_by_length
    language_model = get_language_model()
File "vllm/model_executor/models/interfaces.py", line 210, in get_language_model
    raise NotImplementedError(
NotImplementedError: No language model found in RBLNOptimumWhisperForConditionalGeneration!
You should initialize it via `_mark_language_model`, and make sure `embed_input_ids` is implemented.
```

---

### 📋 Checklist

---

### 💬 Notes

The initial patch handled an unresolvable language model by catching `NotImplementedError`.

The implementation was revised to avoid exception-based capability detection. The runner now consults a nested language model only through `RBLNOptimumMultimodalMixin`, which represents the RBLN models for which that operation is part of the expected interface.

### DIFF
--- a/examples/experimental/offline_inference_basic.py
+++ b/examples/experimental/offline_inference_basic.py
@@ -54,7 +54,7 @@ def main():
 
     # Generate texts from the prompts. The output is a list of RequestOutput
     # objects that contain the prompt, generated text, and other information.
-    outputs = llm.generate(prompts, SamplingParams(temperature=0.8, min_p=0.1))
+    outputs = llm.generate(prompts, SamplingParams(temperature=0.0))
     # Print the outputs.
     for output in outputs:
         prompt = output.prompt
--- a/examples/experimental/offline_inference_basic.py
+++ b/examples/experimental/offline_inference_basic.py
@@ -54,7 +54,7 @@ def main():
 
     # Generate texts from the prompts. The output is a list of RequestOutput
     # objects that contain the prompt, generated text, and other information.
-    outputs = llm.generate(prompts, SamplingParams(temperature=0.0))
+    outputs = llm.generate(prompts, SamplingParams(temperature=0.8, min_p=0.1))
     # Print the outputs.
     for output in outputs:
         prompt = output.prompt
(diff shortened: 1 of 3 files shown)


examples/experimental/offline_inference_basic.py
```diff
@@ -54,7 +54,7 @@ def main():
 
     # Generate texts from the prompts. The output is a list of RequestOutput
     # objects that contain the prompt, generated text, and other information.
-    outputs = llm.generate(prompts, SamplingParams(temperature=0.0))
+    outputs = llm.generate(prompts, SamplingParams(temperature=0.8, min_p=0.1))
```
rebel-eunji:
```suggestion
    outputs = llm.generate(prompts, SamplingParams(temperature=0.0))
```

rebel-eunji (approved):

rebel-jonghewk (approved):

PR #1089, opened 2026-09-10
refactor(nixl): split the NIXL worker along the lifetimes it holds

rebel-ykchoi:
#### Summary

`base_worker.py` had grown into one 2,364-line module holding a single class.
What accumulated there is four lifetimes, not four topics: construction, local
registration once per process, the handshake once per peer, and issuing a
transfer once per request. Reading any one of them meant reading past the other
three, and the same fact had picked up a different derivation in each — the
transfer topology was narrowed from `Optional` by the same assert in fifteen
places (`git grep -c "transfer_topo is not None"` on the base), and a region's
head band was derived once per registration path.

This PR is four things of quite different risk, and the source split is the
smaller half:

| | what it is | how to review it |
| --- | --- | --- |
| 1 | the source split — `base_worker.py` into `state.py`, `registration.py`, `handshake.py`, `transfer.py`, leaving the class declaration and `__init__` | not line by line: **What moved, and what did not** gives three commands that settle it |
| 2 | a test-harness rewrite — build the real worker instead of assembling its attributes | read the fixtures and one descriptor assertion |
| 3 | a guard that a test may not patch a connector module by name | read `test_patch_targets.py`'s docstring |
| 4 | two fixes, one widened refusal, five untested axes, and one unreachable default removed | ~30 source lines, worked examples below |

Nine commits -- the split and its fixes, then three that review added:

| | |
| --- | --- |
| `refactor(nixl): split the NIXL worker along the lifetimes it holds` | items 1-3 above |
| `fix(nixl): register a per-shard handle for a peer that replicates its slices` | the fan-out guard |
| `fix(nixl): pair a pipelined peer's regions on what it published, not on the trim` | the pairing check's position |
| `test(nixl): pin five factors a planted mutation changed unseen` | five untested axes |
| `fix(nixl): refuse a Mamba/SSM cache group at registration` | the widened refusal |
| `other(nixl): read one key one way` | drops an unreachable default from the shard read path |
| `test(nixl): assert the lifetimes' names stay disjoint, rather than checking once` | turns verification command 2 into an assertion |
| `other(nixl): point the older guard at the one the split added` | makes the two guards' cross-reference run both ways |
| `test(nixl): assert no lifetime's member is read from a sibling or the base` | the second proposition: disjoint *reads*, not just names |

#### Where to start reading

`state.py` first: it carries the vocabulary the other three read against
(`_kv_areas` against `_kv_slices`, `_viewed_region_ids`, the region-id layout
`(layer * K/V) * areas + area`), the `topo` accessor that replaced those
fifteen asserts, and the two local descriptor-list builders. Then the class
declaration in `base_worker.py`. Then `handshake.py`'s `_nixl_handshake`, where
the fan-out and pairing fixes land, and `add_remote_agent`.

`handshake.py` and `transfer.py` keep their methods in alphabetical order, so a
later move between modules shows up as a move and not as a reordering; it does
mean an entry point sits at the bottom of its file. `state.py` leads with the
accessor and keeps the two descriptor-list builders together, and
`registration.py` keeps the new head-band derivation next to the
`_layer_kv_heads` it wraps.

#### The boundary rule

- **`registration.py`** — produces the region table and the local geometry, once
  per process: the transfer regions, the host buffer, the handshake metadata it
  publishes.
- **`handshake.py`** — per peer, once: query a producer, refuse a pairing we
  cannot serve, build the descriptor lists that address it.
- **`transfer.py`** — per request: descriptor ids and the notification id.
- **`state.py`** — everything that crosses those, which is a typing device as
  much as a design one (above).
- **`base_worker.py`** — the assembly point: the class declaration and
  `__init__`. Nothing else, so the rule has no exceptions.

#### What moved, and what did not

The split commit is a move. Three things about it a command settles, so you do
not have to take them from me:

```sh
P=vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl
SPLIT="$P/state.py $P/registration.py $P/handshake.py $P/transfer.py $P/base_worker.py"
REST="$P/pull_worker.py $P/push_worker.py $P/connector.py $P/metadata.py"
REST="$REST $P/base_scheduler.py $P/pull_scheduler.py $P/push_scheduler.py"
REST="$REST $P/__init__.py"

# 1. No method was renamed: the `def` lines are the same multiset before and
#    after, bar the two added below. This prints the first physical line of
#    each signature, so a parameter on a continuation line is not covered.
diff <(git grep -hoP '^    def \K.*' <base>         -- "$P/*.py" | sort) \
     <(git grep -hoP '^    def \K.*' <split-commit> -- "$P/*.py" | sort)

# 2. Nothing shadows a sibling's `super()` chain: across the five modules the
#    split produced, the bare names are disjoint. (Over the whole package they
#    are not, and should not be -- the direction subclasses and the schedulers
#    override on purpose.)
git grep -hoP '^    def \K\w+' <split-commit> -- $SPLIT | sort | uniq -d

# 3. The transfer direction is untouched: an empty diff for every other file.
git diff --stat <base>..HEAD -- $REST
```

Two names are added: the `topo` accessor on the shared base, and
`_logical_head_bands`.

That second one is the diagnosis and it is the part of the move worth reading.
`register_kv_caches` and `_register_kv_caches_impl` each derived a region's head
band their own way, because the two paths learn a layer's region count
differently — one while building the regions, the other from the table upstream
filled. Both call `_logical_head_bands` now.

Two error messages also lose `(see the class docstring's bounds)`, because this
PR deletes that docstring's topology matrix; it is more than twice the length
`AGENTS.md` allows a docstring, so it lives at the bottom of this page instead.

Four of the five behaviour commits are described under **The fixes**; the
fifth is the table row above. The three review added are not among them --
`git diff --name-only` on each shows test files only.

#### The fixes

**A fan-out peer took the delegating path** (`fix(nixl): register a per-shard
handle …`). A push into a peer whose chiplet areas hold copies of one head slice
has to reach every copy, so the remote descriptor list carries one entry per
copy. The guard deciding whether upstream's whole-engine handle describes the
peer did not consult that fan-out.

Reachable at: local TP 4, four KV heads over four chiplet areas, peer TP 2, on
the write direction — head-matched, one piece per head, no fan-in, no pipeline,
so `split == 1`, `partial` false, `fan_in` false, and `_peer_replica_fanout`
returns 2 (it returns 1 unless this side originates the bytes). The old
predicate saw four falses and delegated. Symptom is not an error: the remote
list had one descriptor per copy and the local list one per block, the lengths
agreed, and a write landed on the wrong blocks of the peer.

**A pipelined peer's regions were paired on the trimmed metadata**
(`fix(nixl): pair a pipelined peer's regions on what it published …`). A stage
holding more layers than our band is presented to upstream trimmed to that band,
and the trim slices the peer's arrays by *our* regions per layer. The refusal
divides the peer's region list by its layer count — so on the trimmed object the
quotient is ours by construction and it cannot fire. A peer expanding a logical
region into more chiplet areas than we do passed, with its first areas of one
layer paired against our K and V of that layer. It now runs before the trim.

Note that this **adds** a call site rather than moving one: the new one is in
`_nixl_handshake`. The call inside `_validate_remote_agent_handshake` sees the
metadata that reached `add_remote_agent`; on the pipelined equal-TP path that
object has been trimmed, which is the case the new call covers.

**A Mamba/SSM cache group is now refused at registration** (`fix(nixl): refuse a
Mamba/SSM cache group …`). This is the one user-visible removal in the PR, so
what it changes:

- Such a group's KV cache is one blocks-first region per layer, which the
  per-area region table this connector publishes cannot describe. Mixed full and
  sliding-window attention is a *different* thing and stays supported.
- It was refused only alongside pipeline parallelism, and the two asserts that
  name the layout sit in branches a hybrid engine at one pipeline stage never
  reaches. So it registered — and the region-count branch that doubles for a K/V
  split ran on a layout that has none, leaving `num_regions` at twice the region
  table. That derives to an `IndexError` on the shard path and out-of-range
  descriptor ids on the transfer one.
- On D2D that is what the refusal replaces. On host staging we delegate to
  upstream instead, and the head-band table this connector builds has no band
  for such a layer, so anything that pairs by head range asserts — but where
  that leaves an equal-TP host-bounce transfer has never been exercised, because
  no Mamba/SSM model has been run over this connector at all. The refusal covers
  both paths deliberately rather than only the one whose failure is derivable.
- Both registration paths pass through `register_kv_caches`, which is where it
  now sits; the pipeline-only refusal it subsumes is gone.

**Axes that ran at a dimension of one** (`test(nixl): pin five factors …`). Five
axes here were built at a dimension of one by the cases that touched them, so a
mutation planted in each survived the whole suite: the copies a piece
is written to, how pieces and copies nest on the two sides, the width of one
region against another's, the position a borrowed handle is taken from, and the
speculative config the compatibility hash folds. They are pinned at values
neither the default nor the read path can produce alone.

#### The test harness

The connector tests set instance attributes by hand and bypassed `__init__` with
`object.__new__`, so they pinned the current structure — down to asserting an
`__mro__` slice — and would have cost whatever this split saved. They build the
real worker now. `conftest.py::make_worker` puts one over a real `VllmConfig`, a
real `KVCacheConfig` and real tensors; `utils.py::KvGeometry` produces the cache
config, the tensors and what the RBLN adapter reports from one place, which is
what a hand-assembled worker cannot do — it cannot catch a case where those
three disagree.

For a new test: `make_worker`. `utils.py::build_worker` is the cheaper builder,
for cases with no geometry; it stubs upstream's `__init__`. The `object.__new__`
sites that remain are not yet converted, not a third intended style.

Descriptor assertions decode an address back to `(region, block, head offset)`
and compare the pairing as a **set**, because what a transfer means is which
local piece is read into which peer piece. Absolute order is pinned in exactly
one place -- the Full/SWA range layout, because the desc ids offset into it. The
pairing contract pins the two lists' orders *against each other* instead, which
is what makes the paired control meaningful: reversing both sides stays green,
reversing one goes red.

#### The patch-target guard

A `patch` aimed at `base_worker.rebel` keeps succeeding after `rebel` moves to
`registration.py`, and silently stops reaching the code under test — coverage
that reads as coverage while covering nothing. So `utils.py` carries
`patch_in_package` / `patched_in_package` / `setattr_in_package`, which apply a
name across the package's modules and refuse a name none of them binds, and
`test_patch_targets.py` fails the suite if a test in this directory names a
connector module again. The intent is that the package can be split again
without touching tests; the scope is this directory.

#### Testing

- `pytest tests/native/distributed/kv_connector/` green at each of the nine
  commits, so the series is bisectable.
- Planted errors in the geometry, the refusals and the compatibility hash each
  turn the suite red, bar one pre-existing refusal the sweep reports as a known
  gap. For the pairing contract a paired control matters:
  reversing the descriptor loop on **both** sides together stays green (the
  assertions do not mirror the loop) while reversing **one** side goes red. Where
  desc ids offset into the list instead, reversing both is red on purpose.
- Two sweeps, because a call being made and a replacement taking effect are
  different questions: deleting each `self.X(...)` statement in the package one at
  a time, and renaming each upstream override away so upstream's own method runs.
  No survivors.
- `pre-commit` clean on every changed file. Its type check runs
  `--follow-imports skip` with no vLLM present, so it sees upstream's base as
  `Any` and the narrowing this adds is invisible to it.

#### Risk / compatibility

- `git diff --stat <base>..HEAD` over the connector package shows an empty diff
  for every other file in the package — `pull_worker.py`, `push_worker.py`,
  `connector.py`, `metadata.py`, `__init__.py` and the three schedulers — which
  is the cheapest proof that the transfer direction is untouched.
- Apart from what **The fixes** names, the refusals, the descriptor arithmetic
  and the handshake payload are the same code in a different file.

---

* Resolves rebellions-sw/fsw-inference#587 — a NIXL connector worker whose one
  class holds four lifetimes, so a fact is derived where it is consumed.

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

2. `uvx pre-commit run --files $(git diff --name-only <base>)`
3. Expected: green. No test outside `tests/native/distributed/kv_connector/`
   is touched.
4. The three commands under **What moved, and what did not** settle the move's
   three structural claims without reading the diff.

---

### 📋 Checklist

---

### 📐 Supported and unsupported P/D topologies

`AGENTS.md` caps a docstring's prose at fifteen lines and sends the surplus here,
so the matrix that used to sit on `RblnNixlWorkerBase` lives in this description.
Most rows are the docstring's, moved and not re-derived -- I did not re-verify
them one by one. Two are not: the Mamba/SSM row is new in this PR, and the
regions-per-layer row is one I added. Finding the code for a row costs a grep and
some guessing: about half say "not supported" in the message, others raise with
wording of their own, and one is an assert, so
`git grep -nE "not supported|raise RuntimeError|assert " ` over the package is
the honest starting point.

**Supported, and the conditions on it**

- the two pipeline sizes must tile each other
- both sides pipelined requires equal TP, and a pipelined peer may not have MORE
  TP ranks
- D2D only: one of OUR chiplet areas must fit inside a single peer rank's band
  (heads per area <= total KV heads / peer TP)
- each side's TP degree must divide the model's KV heads; below that upstream
  replicates one head across ranks, which no head band names
- D2D only: the coarser side's heads per area must be a whole multiple of the
  finer side's, or a descriptor would carry part of a head
- per-layer KV geometry may differ; head bands are per region

**Not supported**

- Mamba/SSM KV-cache groups, at any parallel size — new in this PR, see
  **The fixes**
- D2D only: unequal TP with uneven region slices
- D2D only: unequal TP with MLA
- D2D only: a context-cut KV cache across unequal TP, or a peer that cut it on a
  different axis
- D2D only: a peer whose regions per layer differ from ours
- MLA and non-MLA KV caches in one engine — the K/V split is engine-wide
- cross-layer blocks where per-layer sizes differ
- with PP: cross-layer blocks
- unequal P/D block sizes with unequal TP or with PP; the equal-TP,
  non-pipelined case is upstream's and unaffected
- sliding-window attention with any model parallelism across P/D — unequal TP or
  PP on either side
- more than one KV-cache group on any per-shard transfer — that is, whenever a
  peer serves less than a whole engine

---

### 💬 Notes

Things where the reason is worth having handed to you rather than reconstructed.

- **The class declaration names `NixlBaseConnectorWorker` after the three
  mixins**, although `state.py` already derives from it. Deliberate: it is the
  one place a reader sees what this class actually is, and C3 puts it last
  either way.
- **`state.py` annotates attributes assigned elsewhere** — mostly in
  `__init__`, `compat_hash` in `registration.py`. Two places for one fact, which
  is normally the thing to avoid — but a mixin's only base is `state.py`, so an
  attribute declared in one lifetime is invisible to the others, and an untyped
  read is not an error. That is what makes the loss silent. Giving them defaults
  in `state.py` instead was considered and rejected: a default would hide a
  producer that stopped assigning.
- **`registration.py` keeps a blocks-first branch whose comment says
  "Unreachable".** It is pre-existing and its previous comment already called it
  inert; this PR only corrected *why*, after moving the refusal. Upstream sets
  that flag from an SSM group (now refused) or a 5-dim attention shape (which no
  RBLN backend reports).
- **Test files mirror the five modules the split produced, with one
  exception**:
  `test_rbln_nixl_base_worker.py` keeps the SWA-view cases. Their subject is a
  value `__init__` computes, even though the entry points it gates live
  elsewhere. `state.py` has no mirror of its own — its members' tests sit with
  the lifetime whose axis they answer.
- **The one warning the suite emits** is an `ImportWarning` about the installed
  optimum-rbln and rebel-compiler versions, which is environmental.


rebel-jongwonlee (approved):
Ran your three commands rather than reading the move, and then went further — this is a
big enough change that a claim I can check mechanically is worth more than one I read.

**The structural claims hold.** The `def` multiset differs by exactly the two names you
name; the bare names across the five split modules are disjoint; and
`git diff --stat <base>..HEAD` over the other eight files in the package is empty. The
split commit itself is `-2127` from `base_worker.py` against `+2254` across the four new
files, so the 146-line net is module headers. The `topo` accessor's effect is visible the
same way: `transfer_topo is not None` goes from 19 occurrences to 3.

**Bisectable, verified.** I ran `tests/native/distributed/kv_connector/` at all six
commits and the base: 354 → 376 → 377 → 379 → 384 → 384 → 384, green throughout. The
counts move where you say they do.

**The three behaviour fixes are pinned.** I reverted each source change on its own and
the suite went red each time, with exactly one failure named for the case:

| reverted | fails |
| --- | --- |
| `or fanout > 1` | `test_a_fanned_out_peer_alone_forces_the_per_shard_path` |
| the pre-trim `_check_d2d_region_pairing` call | `test_a_wider_peer_geometry_is_refused_before_the_trim` |
| the Mamba/SSM `raise` | `test_an_ssm_group_is_refused_before_any_registration` |

I also planted a test that patches `registration.rebel` by module name; the guard failed
the suite and reported the file, line and target. It does what the docstring says.

**One thing I checked that the description does not spell out.** `read one key one way`
turns a silent default into a `KeyError`, so I traced the key's lifetime rather than take
"unreachable" on trust: the write at `handshake.py:935` is the line before
`overlapping.append`, the read is behind a non-empty `_overlapping_ranks`, and
`_cleanup_remote_engine` drops `_shard_descs_per_block` and `_overlapping_ranks` four
lines apart. Nothing can reach the read without the key, and nothing can outlive it. The
default was dead.

The fan-out worked example checks out too: at local TP 4 with four KV heads over four
areas, `heads_per_rank` is 1 so `slices_l` can only be 1, `cuts_l = 4` against
`cuts_r = 2` gives `split == 1`, the peer has fewer ranks so `fan_in` is false, and
`_peer_replica_fanout` returns `areas_r // slices_r`. Four falses and a fanout of 4 — the
old predicate delegated, exactly as you describe.

**A caveat on my green.** I ran the suite against an existing venv with
`vllm==0.24.0+cpu` and a `PYTHONPATH` redirect onto the worktree, not through
`uv sync --locked`. Same vLLM the branch pins, but not the project's resolved
environment — so read my runs as confirming the tests, not the lockfile.

**One non-blocking note, on a name this PR is well placed to fix.** `registration.py` now
holds `_layer_kv_heads` and gains `_logical_head_bands`, and both deal in a count whose
unit is not in its name. Inside `_register_kv_caches_impl` the two units sit next to each
other and the distinction survives only in a local variable and a comment:

```python
logical_kv_heads: list[int | None] = []
# None here means something else than for `logical_kv_heads` above:
logical_total_kv_heads: list[int | None] = []
```

`total` then disappears at every boundary — `self._logical_region_kv_heads`,
`_region_kv_heads()`, `_layer_kv_heads()` all return the model-wide count while
`layer_spec.num_kv_heads`, the per-rank one, keeps the plainer name. That ambiguity is
what produced the aliasing bug in #1058, and it caught both of us again in the #1045
thread, where "gpt-oss declares 8 KV heads" was reasoning about the model-wide count
against code reading the per-rank one.

Carrying the qualifier through — `_layer_total_kv_heads`, `_region_total_kv_heads`,
`_logical_region_total_kv_heads` — would put the unit where the reader is, and this is the
cheapest moment for it: the symbols are moving anyway and the second derivation is being
collapsed into `_logical_head_bands` in the same commit. Entirely your call; it changes
nothing and I am not asking for it as a condition.

Approving.

rebel-ykchoi:
Taken, in `6cbe9eca`. The argument is the one `test_patch_targets.py` already
makes, and it does transfer — two risks came out of the split and only one of
them got a test.

**I could run it, so here is what the import says.** What the discovery returns,
and how many names each binds:

```sh
python -c "import sys; sys.path.insert(0,'.'); \
from tests.native.distributed.kv_connector.test_lifetime_mixins import \
    _lifetimes, _own_callables; \
b, m = _lifetimes(); \
print([(c.__name__, len(_own_callables(c))) for c in [b] + m])"
```

`RblnNixlWorkerState` (8), then `RblnNixlHandshakeMixin` (29),
`RblnNixlRegistrationMixin` (12), `RblnNixlTransferMixin` (3) — the three
mixins, with no fourth class picked up. Your counts off `ast` were right. One
correction that changes nothing: the mixins bind 44 names, not 47.

**Planted two collisions in the package rather than only in the helper**, since
a self-check passes the same way a rule matching nothing does:

| planted | result |
| --- | --- |
| `_layer_kv_heads` on the transfer mixin, a name registration holds | `_layer_kv_heads: RblnNixlRegistrationMixin and RblnNixlTransferMixin` |
| `_layer_overlap` on the transfer mixin, a name `state.py` holds | `_layer_overlap: RblnNixlTransferMixin over RblnNixlWorkerState` |

Restored to a zero-line diff after each.

**One test added to yours.** `vars` over `def` lines is the part your prose
argues and the code does not pin, so `test_the_check_sees_what_a_def_line_does_not`
fixes it: a decorated `def`, an assignment to a `staticmethod`, and a value that
is not callable, where a `def`-line reader sees the first, misses the second and
counts the third.

**It went in a file of its own.** `test_patch_targets.py` states its subject in
its docstring and its name; this is a different proposition about the same
split, and the two are easier to keep true apart than merged under a name that
fits one of them. Same directory, so the two guards still sit together, and
each docstring points at the other.

Suite is green with the three new tests.

rebel-jongwonlee:
Checked all four, and the correction is yours.

**The counts.** Reproduced on `6cbe9eca`: state 8, handshake 29, registration
12, transfer 3, and the discovery returns exactly those three with no fourth
class. 44 is right and 47 was mine — I built stand-in classes with `type()` and
subtracted three dunder entries per class where it adds four (`__dict__`,
`__doc__`, `__module__`, `__weakref__`), so 44 came out 47. The `ast` counts
underneath were the same as yours.

**The plants.** Both reproduce exactly:

```
_layer_kv_heads on transfer  ->  _layer_kv_heads: RblnNixlRegistrationMixin and RblnNixlTransferMixin
_layer_overlap  on transfer  ->  _layer_overlap: RblnNixlTransferMixin over RblnNixlWorkerState
```

and both names are real ones from the classes you named — `_layer_overlap` is
`state.py`'s, `_layer_kv_heads` is registration's — so the plants sit where they
claim to, one per message shape.

Planting in the package rather than only in the helper is the part I could not
do and the part that matters: my run proved `_collisions` works on synthetic
classes, which says nothing about whether the discovery wires up to the real
one. That is exactly what `assert len(mixins) >= 3` is worried about, and your
run is what closes it.

**The third test is better than the two I sent.** `vars` over `def` lines was
prose in my comment and a claim in a commit message; it is an assertion now.
Ran it: `as_property = 1` is excluded for not being callable, `as_staticmethod`
collides from both sides, `Base` binds nothing callable, so
`["as_staticmethod: L and R"]` is exactly right.

**The file split reads right** — the two propositions are easier to keep true
apart, and a name covering both would fit neither.

**One thing it does not do yet.** The note says "each docstring points at the
other", but only one of them does:

```
test_lifetime_mixins.py:29   "`test_patch_targets.py` guards the other half
                              of what the split created."
test_patch_targets.py        no reverse pointer
```

and the commit is `1 file changed, 151 insertions(+)`. The cross-reference is
what the two-file decision rests on, so the half that is missing is the half
that matters more: `test_patch_targets.py` is the file that was already there,
so it is the one a reader meets first, and from it there is no way to discover
that the other guard exists. One line in its docstring closes it.

Approval stands.

rebel-ykchoi:
Added to `test_patch_targets.py`'s docstring in `e1208933`, so the reference now runs both ways.

rebel-jongwonlee (approved):
The rebase is the one I traced, and it lost nothing.

Eight commits on `b3cce75b`, with #1058's five dropped now that dev carries them
squashed. The resulting tree is byte-identical to `e1208933`, which is what you
would expect since what was dropped is what dev already has. `mergeable: true`;
`blocked` is my stale approval, which this replaces.

## The three commands, re-run on the new base

```
(1) diff of `def` lines, base..split
    27a28  > _logical_head_bands(
    65a67  > topo(self) -> TransferTopology:
```

Exactly the two the description names and nothing else, so the move is still a
move.

```
(2) sibling name duplicates across the five modules -> none
(3) diff of the direction files, base..split -> empty
```

## The guards

Ran the disjointness logic over the name sets at this head: state 8, handshake
29, registration 12, transfer 3 — 44 mixin names, zero collisions. Both
self-checks pass, and the cross-reference now runs both ways, one hit in each
docstring.

## One thing worth stating

The rebase moved the base from vLLM 0.24 to 0.26, so the split is the same
operation on different content — `handshake.py` picks up upstream's rekeying of
`_remote_agents` to `(pp_rank, tp_rank)` and the new `add_remote_agent`
signature, among others. That is why command (1) mattered here rather than being
a formality, and it comes back clean.

What I could not redo: I have no 0.26 environment, so the bisect across the
series and the mutation runs from my first pass were measured on 0.24 and do not
carry over. This pass is static.

Re-approving.

rebel-ykchoi:
Rebased onto `dev` now that #1023 has landed, and added one commit after your
approval. Both are worth naming.

**The rebase.** `b3cce75b` → `7481ab83`, no conflicts, and the carry-over is
unchanged:

```
git diff b3cce75b..<old tip>   --shortstat   # 17 files, 5986(+), 3868(-)
git diff origin/dev..HEAD      --shortstat   # the same three numbers
```

The only file #1023 and this branch both touch is `test_rbln_nixl_connector.py`,
and their hunks meet only in the import block, which merged on its own.

**The new commit.** `test(nixl): assert no lifetime's member is read from a
sibling or the base` — the second half of the proposition you named on #1098.
You wrote there that `test_lifetime_mixins.py` asserts the names are disjoint
and that whether the *reads* are disjoint too is a second proposition. That
proposition is about this split rather than about chunk mode: the rule it
asserts is the one `state.py`'s own docstring states, the file is this PR's, and
its docstring already carries half of it ("`state.py` is where a name two
lifetimes need belongs"). So it belongs here, not on top of it.

It reuses `_lifetimes()`, and reads the source rather than the live classes,
because the defect is about *which class* a `self.` read sits in and an
inherited attribute is indistinguishable from an own one at runtime. The
assembled subclasses are deliberately not readers for this purpose — they
inherit every lifetime, so the same call is typed there. On this branch that
leaves the four `transfer -> pull_worker/push_worker` reads passing, which is
what you called downward and intended.

**What it does not have here, stated plainly: a failing case.** Nothing on this
branch violates the rule, so the assertion lands green, and what shows the check
sees anything at all is the control test — the same shape
`test_the_check_sees_a_collision` already uses for the name guard. The first
change that trips it is #1098, where `_shard_chunk_grid` gains a second reader;
reverting only that PR's move makes this assertion fail with exactly the two
lines it should.

The other reason to put it here rather than there: the mypy that would catch
this class cannot. The hook's environment has no vllm, so
`NixlBaseConnectorWorker` degrades to `Any` under `ignore_missing_imports` and
every read through `self` is legal — run the hook's own mypy over the package
and it reports `Success` on all of it, while a mypy that can see vllm reports
the two `attr-defined` this asserts against.

```
uv run --no-sync pytest tests/native/distributed -q     # 428 passed, 4 skipped
uvx pre-commit run --files <the one changed file>
```

rebel-cjlee (approved):

PR #1090, opened 2026-09-11
refactor(config): read the RBLN options from RBLNConfig instead of envs

### LINKED ISSUE #1047: proposal(config): move the `VLLM_RBLN_*` options that change model behavior into `RBLNConfig`

rebel-cjlee:
### 🚀 Summary of Changes

`RBLNConfig` landed in #1064 as a config object, but no code read it. The `--rbln-*` flags wrote into it, and every call site still read `envs.py`. This PR moves the read sites to the config.

The rule: a place that already has a `VllmConfig` binds `rbln_config: RBLNConfig = vllm_config.additional_config` and passes the values down. Only 4 places use `get_rbln_config()` (constructors and classmethods whose signature belongs to upstream vLLM).

The global `set_rbln_config()` is removed. The config now travels on `VllmConfig`, which already reaches every worker.

**4 options stay in `evns.py` and are no longer fields:** `compile_strict_mode`, `num_hidden_layers`, `use_dynamic_kv_cache`, `nixl_swa_view_opt`. They are bring-up or experimental knobs, so passing them through many function signatures is not worth it.

Model path: vLLM-native only. The optimum-rbln path still reads `envs.py` and is not changed.

---

* Related to #1047, #1064

---

### ✅ Type of Change

---

### 🧪 How to Test

1. `pytest tests/native -q` : 1867 passed, 26 skipped, 16 xfailed
2. `pytest tests/torch_compile/unit/v1/sample -q` : 70 passed
3. `pytest tests/native -m model_compile --model-compile` : whole-model compiles
4. With an empty `VLLM_CACHE_ROOT`, run the same model twice: once with `VLLM_RBLN_SPECIALIZE_MOE_DECODE=0`, once with `additional_config={"specialize_moe_decode": False}`. Both must log the same `mega-cache config_signature=` and the second must log `Loaded rbln mega-cache bundle`.

---

---

### 📋 Checklist

---

### 💬 Notes

- **Existing mega-cache bundles will miss once.** The bundle key changes, so every deployment recompiles one time. Old bundle directories are not deleted automatically.
- **4 flags are removed** with their fields: `--rbln-compile-strict-mode`, `--rbln-num-hidden-layers`, `--rbln-use-dynamic-kv-cache`, `--rbln-nixl-swa-view-opt`.
- **Messages still mention the variables, not the flags.** This is on purpose: the variables are what deployments use today. The wording will change when we add the deprecation warning.
- Still to do for #1047: **renaming**, docs and CI harness, a deprecation warning, then removing the variables.

---


vllm_rbln/config.py
rebel-eunji, line 317: How about just allowing only dict type of values and just raising an error if values is not a dict type?

vllm_rbln/config.py
```diff
@@ -224,50 +206,47 @@ def build_rbln_config(additional_config: Any = None) -> RBLNConfig:
 
     if shadowed:
         logger.warning_once(
-            "Ignoring the environment variables for %s; the CLI value wins.",
+            "Both the environment and additional_config set %s; RBLNConfig "
+            "takes the additional_config value.",
             ", ".join(shadowed),
         )
```
rebel-eunji, line 229: How about raising an error, not silent setting?

```python
overrides = _env_overrides()
conflicts = sorted(
    name for name in set(given) & set(overrides) if given[name] != overrides[name]
)
if conflicts:
    raise ValueError(
        f"{', '.join(conflicts)}: the environment and additional_config give "
        "different values. Unset the variable or drop the flag."
    )
overrides.update(given)
```

vllm_rbln/config.py
rebel-cjlee: Upstream declares this argument as `type=union_dict_and_str`. It gives back a dict when the value is a JSON object, and the raw string when it is not. So a non-dict is a normal input here, and the action just passes it along rather than changing what upstream does.

A string is still rejected, by `build_rbln_config()`, which also covers `LLM(additional_config=...)`, since that never goes through argparse.

It would be also good to raise here instead if you prefer the earlier failure.

vllm_rbln/config.py
rebel-cjlee: It actually warns, so it's not silent. But you're right that it's easy to miss in a long start-up log.

I kept it as a warning because the priority order is part of the design: `additional_config` > env > default. If we raise, that order can never be used. Upstream does the same thing and doesn't even warn. Also, our CI sets some envs today, so raising would stop anyone from adding the matching flags to those lanes. And this branch goes away in the future release along with the env vars.

vllm_rbln/v1/attention/ops/sliding_window_attention.py
```diff
@@ -113,8 +117,10 @@ def sliding_window_attention_v1(
     block_tables: torch.Tensor,
     window_size: int,
     sinks: torch.Tensor | None = None,
+    *,
+    compile_model: bool,
 ) -> torch.Tensor:
-    if envs.VLLM_RBLN_COMPILE_MODEL:
+    if compile_model:
```
rebel-eunji, line 123: nit: How about removing this branch and compile_model field later?

rebel-eunji (approved):
I think this branch gets us where we wanted to be. LGTM.

vllm_rbln/v1/attention/ops/sliding_window_attention.py
rebel-cjlee: Good point, I'm thinking of removing `VLLM_RBLN_COMPILE_MODEL` and using `enforce_eager`.

rebel-cjlee: WIP in `enable-enforce-eager` branch.

PR #1094, opened 2026-09-11
fix(worker): end a multiprocess worker whose model step raises so the engine fails fast

rebel-scottlee:
### 🚀 Summary of Changes

When a worker under the multiprocess executor raises inside the model step, the replica doesn't fail. It hangs until `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` (300 s) runs out, and `/health` keeps returning 200 the whole time. We ran into this with an RDS SSD failure that leaves the worker's NPU context aborted ("guilty"): every later submission is rejected, so that worker can never make progress again, but nothing tells the engine.

Three things in upstream combine to produce the hang:

* `WorkerProc.worker_busy_loop` catches every `Exception`, logs it and keeps looping, so the worker process stays alive. The executor only learns about a broken worker through the process sentinel, and that never fires.
* With a KV connector every rank replies to `execute_model` / `sample_tokens`. Under async scheduling the FAILURE reply to `execute_model` is parked in a `FutureWrapper` and only re-raised when the paired `sample_tokens` reply is `None`. A worker whose forward failed answers `sample_tokens` with `EMPTY_MODEL_RUNNER_OUTPUT` instead (`execute_model_state is None`), so the stored exception is never raised and the step is treated as "no tokens".
* The healthy ranks block inside the forward collective waiting for the failed rank, so the engine sits in `acquire_read` on their reply queue.

The single-worker (`uni`) layout doesn't have this problem. The exception goes straight into EngineCore, the engine logs a fatal error and the replica exits so the orchestrator can restart it. This PR gives TP/PP/EP/DP workers the same contract.

The change itself is small. `worker_fail_fast` in `vllm_rbln/v1/worker/utils.py` wraps `RBLNWorker.execute_model`, `sample_tokens`, `execute_dummy_batch` and `AsyncRBLNModelRunnerOutput.get_output`. If one of them raises in a worker whose executor is `MultiprocExecutor` or a subclass of it, the worker logs the exception and ends its own process with `os._exit(70)`. From there on it's plain vLLM: with the mp executor the worker monitor sees the sentinel, `executor.shutdown()` takes down the other ranks (death pipe, SIGTERM after 5 s, SIGKILL after 4 more), the response queues close, EngineCore dies, in-flight requests get 500 and the API server exits.

A few decisions worth calling out:

* There is no error classification. Any exception escaping the model step ends the worker. That's what the single-worker layout already does, and a worker that can't run a step isn't serving anyway. Recoverable RDS errors are handled below this layer: LMCache turns a failed read or write into a cache miss, and `kv_load_failure_policy` decides whether to recompute.
* The async output path is covered as well. Under async scheduling `sample_tokens` returns an `AsyncRBLNModelRunnerOutput` and vLLM calls `get_output()` later, on the worker's async output thread. `WorkerProc.async_output_busy_loop` has no handler around that call, so a failed device-to-host copy or logprob conversion would kill only that thread. The worker would stay alive, the reply for that step would never be enqueued, and every later step's output would pile up in a queue nobody drains. `os._exit` from the copy thread ends the whole process, which is the signal the executor actually acts on.
* The scope is decided by executor class, not by backend name. `RBLNWorker` and `RBLNModelRunner` compute `fail_fast = issubclass(Executor.get_class(vllm_config), MultiprocExecutor)` once in `__init__`, and the async output gets that bool. This covers `mp`, `RayExecutorV2` (what vLLM 0.24 picks for `ray` by default) and custom `MultiprocExecutor` subclasses passed as a class or a qualified name, which is where vLLM's `WorkerProc` busy loop lives. `uni` and `external_launcher` run in process, so the exception already reaches the engine and the guard just re-raises. Ray is still not a supported backend on RBLN. The Ray cases in the tests substitute the executor module to check the policy; actual Ray execution is not verified.
* `os._exit` rather than `SystemExit`. The normal teardown runs KV-connector and offload cleanup, which can block on the very device that just stopped accepting work. The driver reclaims the context when the process goes away. The exit runs from a `finally`, so a logging failure can't keep the worker alive. If logging blocks outright there is no exit deadline, and the docstring says so.
* Exit code 70 is `EX_SOFTWARE`, used here for an unhandled worker error. The original exception stays in the log line; the multiprocess executor detects process death without interpreting the exit code.
* `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST=1` brings the old behaviour back for debugging. It's read on each call, parsed like the other booleans in `vllm_rbln/envs.py` (a typo leaves fail-fast on), and registered as a non-compile variable so it can't change the compile-bundle key.

Everything is in the worker layer, plus one platform diagnostic. No model-specific code is touched. The platform commit is independent of the fail-fast mechanism: the old warning claimed a fallback to `mp` that never happened and fired on every single-process run, because vLLM resolves `world_size` 1 to `uni`. It now warns only for the unsupported backends and says the selection is kept.

---

* Related to rebellions-sw/lmcache-rbln#164 (defect ①: worker failure not detected, 300 s hang)

> There is no issue for this in this repository; the investigation and the reproduction matrix live in the lmcache-rbln issue above.

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit tests, no NPU needed when run with `--device-tensor 0` (the default setting uses device tensors):
   * `pytest -q --device-tensor 0 tests/native/v1/worker/test_rbln_worker.py -k WorkerFailFast`. The real `execute_model`, `sample_tokens` and `execute_dummy_batch` run against a model runner that raises, under `mp`, `uni`, `external_launcher`, a `MultiprocExecutor` subclass given as a class and as a qualname, `ray` (V2 default) and `ray` legacy. The `mp`-family cases exit with 70, the in-process ones re-raise, and the toggle disables the exit. Take any of the three decorators off and the matching cases fail. The Ray cases substitute the executor module, so no Ray runtime is involved.
   * `pytest -q --device-tensor 0 tests/native/v1/worker/test_utils.py -k WorkerFailFast`. The decorator on its own: any exception exits with 70, success is untouched, the toggle is read on each call, a misspelt toggle keeps fail-fast on, and a failing logger still can't keep the worker alive.
   * `pytest -q --device-tensor 0 tests/native/v1/worker/test_async_output.py -k failure`. Same policy on the output thread for the copy stage and the logprob stage, plus a subprocess test that lets a real output-thread failure run to completion and checks the child leaves with 70.
   * `pytest -q --device-tensor 0 tests/native/v1/worker/test_rbln_model_runner.py -k fail_fast`. An async output built by the real `RBLNModelRunner.sample_tokens` inherits the runner's policy: a failing copy exits under `mp` and re-raises under `uni`.
   * `pytest -q --device-tensor 0 tests/native/test_platform.py -k "executor_backend or single_worker"`. The executor selection is kept, the warning only fires for the unsupported backends, and a default single-worker config resolves to `uni` without a warning.
2. End to end: serve any model with `--tensor-parallel-size 2` and `--kv-transfer-config '{"kv_connector":"RBLNLMCacheConnectorV1","kv_role":"kv_both"}'`, then make one rank's model step raise. We used the rebel-compiler vSSD fault injector in `guilty` mode; aborting one rank's NPU context any other way gives the same shape.
3. Expected: the worker log shows `RBLN worker <pid>: RBLNWorker.execute_model raised ... Ending this worker process with exit code 70`, then `Worker proc VllmWorker-N died unexpectedly, shutting down executor`, `EngineCore encountered a fatal error`, HTTP 500 on in-flight requests and the API server exiting, all within about 10 s of the fault. Without this change the same setup answers `/health` 200 and returns the first 500 after ~300 s (`RPC call to execute_model timed out`).
4. Edge cases: single worker (`uni`) is unchanged (500, then exit 0 via EngineCore, the guard does not fire). `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST=1` restores the old looping behaviour.

---

Verified on udc-12 (8× RBLN-CR13, UMD 3.5.0~rc5, vllm 0.24.0, lmcache-rbln main) with the vSSD guilty injector. Time from the fault until all worker, engine and API processes had exited, and the API server's exit code:

| Layout | Guilty rank | Before | After |
|---|---|---|---|
| TP2 | rank 0 (×3) | hang (>180 s, `/health` 200) | 10.4 s, exit 0 |
| TP2 | rank 1 (×2) | 300 s | 10.3 s, exit 0 |
| PP2 | first stage | 300 s | 9.2 s, exit 0 |
| TP2PP2 | PP0_TP1 | 300 s | 10.5 s, exit 0 |
| MoE TP4+EP (Qwen3-30B-A3B, 2-layer slice) | TP1_EP1 | 300 s | 11.8 s, exit 0 |
| single worker | — | 500 → exit 0 | unchanged |

In every case the guilty worker exited on its first rejected step, in-flight requests got 500, streams were cut, and no NPU context was left behind. The remaining ~10 s is the executor's shutdown grace (`VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS`, 5 s) plus the 4 s SIGTERM wait. TP ranks blocked in a collective needed the SIGKILL; the PP stage went on SIGTERM.

That table is from the first commit. After the review changes (executor-class scoping, per-call toggle, renames) and the merge with `dev`, I ran the TP2 guilty-rank-1 case again on the final tree. Same result: the guilty worker exited with 70 on its first rejected step and all processes had exited 10.3 s after the fault, API exit 0, no context left. With `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST=1` the old behaviour came back: the worker kept looping on `WorkerProc hit an exception`, `/health` answered 200 the whole time, and the engine only died when `RPC call to sample_tokens timed out` fired 300 s later.

Representative log excerpt (TP2, guilty rank 1):

```
(Worker_TP1 pid=225074) [vllm-rbln] ERROR [utils.py] RBLN worker 225074: RBLNWorker.execute_model raised RuntimeError: SysError(125): [cmd_dispatcher] SubmitJob() failed, rc=125. ... Ending this worker process with exit code 70 ...
(EngineCore pid=221885) ERROR [multiproc_executor.py:284] Worker proc VllmWorker-1 died unexpectedly, shutting down executor.
(EngineCore pid=221885) INFO  [multiproc_executor.py:426] [shutdown] Executor: waiting for worker exit count=2
(EngineCore pid=221885) WARNING [multiproc_executor.py:438] [shutdown] Executor: workers still running after grace period; sending SIGTERM count=1
(EngineCore pid=221885) WARNING [multiproc_executor.py:448] [shutdown] Executor: workers still running after SIGTERM; sending SIGKILL count=1
(EngineCore pid=221885) ERROR [core.py:1233] EngineCore encountered a fatal error.
(APIServer pid=217721) INFO:     127.0.0.1:47926 - "POST /v1/completions HTTP/1.1" 500 Internal Server Error
```

The async output path is covered by the subprocess test rather than a device run: the injector rejects the forward submit, so every device trial failed in `execute_model` before `get_output()` ran.

---

### 📋 Checklist

---

### 💬 Notes

* Because nothing is classified, a software bug that raises in the model step of a multiprocess worker now also ends that worker, and with it the replica. lmcache-rbln#164 defect ③ (`Logical device rbln:N is not assigned` on TP/PP rank > 0) is an example: a TP+RDS server exits during warm-up until that is fixed in lmcache-rbln. We confirmed it by running without the test-only workaround. We think that beats a rank silently running without RDS, but reviewers should know.
* Shorter fail-over is available by lowering `VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS`; the fixed 4 s SIGTERM wait in `_ensure_worker_termination` is upstream.
* Other worker-side threads are still unguarded. The KV connector's own background threads, for instance, can lose a thread without ending the process. Separate follow-up.
* A structured "context aborted" signal from the runtime would let the engine react without the process exit. Until then process death is the only signal the executor acts on.


tests/test_device_fail_fast.py
```diff
@@ -0,0 +1,89 @@
...
+def test_success_path_is_untouched(exits):
+    assert _Worker("mp", None).execute_model() == "output"
+    assert exits == []
+
+
+def test_can_be_disabled_for_debugging(exits, monkeypatch):
+    monkeypatch.setenv("VLLM_RBLN_FAIL_FAST_ON_DEVICE_ERROR", "0")
+    with pytest.raises(RuntimeError):
+        _Worker("mp", RuntimeError("SubmitJob() failed")).execute_model()
+    assert exits == []
```
rebel-jinhwan: How about move to vllm_rbln/v1/worker/utils.py ?

rebel-scottlee: Sounds good. I’ll move the fail-fast helpers and decorator into vllm_rbln/v1/worker/utils.py and reuse the same helper for the async output path.

I’ll also move the relevant tests into the existing test_utils.py and add the async output regression test to test_async_output.py.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -1163,6 +1164,7 @@ def get_supported_tasks(self) -> tuple[SupportedTask, ...]:
         return self.model_runner.get_supported_tasks()
 
     @torch.inference_mode()
+    @fail_fast_on_device_error
```
rebel-cjlee: These three decorators are not covered by any test. `sample_tokens`(L1167), `execute_model`(L1179) and `execute_dummy_batch`(L1267) can all be removed, and no test fails:

```
$ uv run --no-sync pytest -q tests/native/v1/worker/test_utils.py
135 passed          # before

$ sed -i 's/^    @fail_fast_on_device_error$//' vllm_rbln/v1/worker/rbln_worker.py
$ uv run --no-sync pytest -q tests/native/v1/worker/test_utils.py
135 passed          # same result, but now no method is guarded
```

`TestFailFastOnDeviceError` tests the decorator function itself. It uses a local `def step(worker)` stub and never imports `RBLNWorker`, so it cannot tell us whether the decorator is applied.

`test_async_output.py` does check this for `get_output()`, but the worker part has nothing like that.

vllm_rbln/v1/worker/utils.py
```diff
@@ -47,6 +48,53 @@
...
+            exc,
+            WORKER_DEVICE_FAILURE_EXIT_CODE,
+            exc_info=exc,
+        )
+    finally:
+        # StreamHandler flushes each record; avoid shutdown's handler locks.
+        os._exit(WORKER_DEVICE_FAILURE_EXIT_CODE)
+
+
+def fail_fast_on_device_error(function: _F) -> _F:
```
rebel-cjlee: Two small things about the wrapper.

1. The env flag could be read when the decorator is applied, not inside `except`. Today the guard reads it at the moment the device has already failed. Reading it once at import is simpler, and when the flag is off there is no wrapper at all:
  ```
  def fail_fast_on_device_error(function: _F) -> _F:
    if not envs.VLLM_RBLN_FAIL_FAST_ON_DEVICE_ERROR:
      return function
    ...
  ```

2. `self.parallel_config` is also read inside `except`. If it is ever missing, the `AttributeError` replaces the real exception AND the worker does not exit (the exact hang this PR prevents). `self` is types `Any`, so mypy cannot catch it. Reading it before the `try` fixes that and leaves the except block with no lookups at all:
```
@wraps(function)
def guarded(self, *args, **kwargs):
  fail_fast = self.parallel_config.distributed_executor_backend == "mp"
  try:
    return function(self, *args, **kwargs)
  except Exception as exc:
    if fail_fast:
      abort_worker(exc, where=function.__qualname__)
    raise
```

vllm_rbln/v1/worker/utils.py
```diff
@@ -47,6 +48,53 @@
 
 logger = init_logger(__name__)
 
+_F = TypeVar("_F", bound=Callable[..., Any])
+WORKER_DEVICE_FAILURE_EXIT_CODE = 70
```
rebel-cjlee: Is a single fixed exit code the intent here?

Whatever fails, the worker exits with the same code. A `SysError(125)`, a `SysError(5)` and a plain `ValueError` all give 70, and your tests assert that. The real error code stays in the log text only.

That's fine if 70 just means "the guard ended this process on purpose". But the PR description never explains the choice, so right now I can't tell whether 70 carries any meaning.

vllm_rbln/v1/worker/utils.py
```diff
@@ -47,6 +48,53 @@
...
+
+
+def abort_worker(exc: Exception, *, where: str) -> NoReturn:
+    """Skip device cleanup even if logging raises.
+
+    Logging can still block on handler locks or I/O.
+    """
+    try:
```
rebel-cjlee: No objection to the try/finally itself, it's cheap insurance. My problem is the docstring.

It says "Logging can still block on handler locks or I/O" right next to a try/finally, which reads as if the try/finally covers blocking. It doesn't. `finally` runs when control leaves the try block. If `logger.error` blocks, control never leaves, `os._exit` never runs, and the worker stays alive. That's the hang this PR is fixing.

vllm_rbln/envs.py
```diff
@@ -54,6 +54,7 @@
     VLLM_RBLN_NUM_HIDDEN_LAYERS: int = 0
     VLLM_RBLN_USE_DEVICE_TENSOR: bool = True
     VLLM_RBLN_DISABLE_OFFLOAD: bool = False
+    VLLM_RBLN_FAIL_FAST_ON_DEVICE_ERROR: bool = True
```
rebel-cjlee: The name says "device error" but the code catches every Exception. Your PR description says "No error classification", and the test parameterizes `ValueError("software error")`, so the filter the name promises isn't there.

Also, the default is **on** and the only reason to touch this is to turn it off, so a `DISABLE_` name might fit better.

- `VLLM_RBLN_FAIL_FAST_ON_DEVICE_ERROR` -> `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST`
- `fail_fast_on_device_error` -> `worker_fail_fast`
- `WORKER_DEVICE_FAILURE_EXIT_CODE` -> `_FAIL_FAST_EXIT_CODE`

One nice side effect is the parsing. With a default of `False` you can go back to the form every other boolean in this file uses, `os.environ.get(NAME, "False").lower() in ("true", "1")`, instead of the `not in ("0", "false", "no")` it needs today. A typo then leaves fail fast on rather than off.

vllm_rbln/v1/worker/async_output.py
```diff
@@ -40,7 +43,10 @@ def __init__(
         req_ids: list[str],
         placeholder_pos: dict[str, int],
         logprobs_tensors: LogprobsTensors | None,
+        parallel_config: ParallelConfig,
     ):
+        # Public to satisfy the shared fail_fast_on_device_error contract.
+        self.parallel_config = parallel_config
```
rebel-cjlee: Passing the whole `ParallelConfig` here felt a little odd to me, since the only thing that ever gets read is one string, and only on the error path.

What about resolving it once in `RBLNWorker.__init__` and `RBLNModelRunner.__init__` and passing just a bool?
```
self.fail_fast = issubclass(Executor.get_class(self.vllm_config), MultiprocExecutor)
```

Both classes already have `self.vllm_config`, so it's one line each. Then this object carries a bool instead of a config, the comment above isn't needed, and the decorator just reads `self.fail_fast`.

`Executor.get_class` also covers the cases `== "mp"` misses, like `RayExecutorV2`, which subclasses `MultiprocExecutor`. It does imports internally, so init is the right place for it anyway, not inside except.

vllm_rbln/platform.py
```diff
@@ -551,15 +551,10 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        if parallel_config.distributed_executor_backend not in (None, "mp", "uni"):
             logger.warning(
-                (
-                    "%s is not supported on RBLN, fallback to mp "
-                    "distributed executor backend."
-                ),
+                "%s is not supported on RBLN. Keeping the selected distributed "
+                "executor backend; use 'mp' for supported multi-worker execution.",
                 parallel_config.distributed_executor_backend,
             )
```
rebel-cjlee, line 559: This warns that the backend isn't supported and then runs with it anyway.

That's what makes the `== "mp"` check in the guard reachable around. Nothing stops a user landing on ray, and `RayExecutorV2` subclasses `MultiprocExecutor`, so it has the same hang and no guard. `config/parallel.py:902-912` can even pick ray on its own inside a placement group, so the user may not have asked for it.

So I'm not sure the platform change is as independent from the fail fast work as the description says.

If you widen the guard the way I suggested on `async_output.py:49` (`Executor.get_class`), this becomes cosmetic and the warning is fine as it is. If you'd rather fix it here by raising, note that `external_launcher` is in this list too, and that one runs in process (it subclasses `UniProcExecutor`), so it works today.

tests/native/test_platform.py
```diff
@@ -704,12 +726,6 @@ def test_sliding_window_keeps_prefix_caching_on(self):
         assert RblnPlatform._uses_sliding_window(config.model_config.hf_config)
         assert config.cache_config.enable_prefix_caching is True
 
-    def test_a_non_mp_executor_backend_is_only_warned_about(self, configured):
-        # The warning says "fallback to mp" but nothing assigns, and vLLM's own
-        # default for world_size 1 is "uni" -- so it fires on every single-process
-        # run and Executor.get_class still builds a UniProcExecutor.
-        assert configured.parallel_config.distributed_executor_backend == "uni"
-
```
rebel-cjlee, line 712: nit: deleting this test was right, it pinned the behavior you just fixed. But the assert also pinned a second thing, that vLLM resolves `world_size` 1 to "uni". That's the premise the mp-only scope rests on, and the new test sets the backend explicitly so it never looks at the default. If upstream ever made a single-process run default to "mp", the guard would start exiting single-process servers and nothing here would fail. One line somewhere outside TestKnownGaps would keep it.

vllm_rbln/v1/worker/rbln_worker.py
rebel-scottlee: Fair point. The tests only checked the decorator on a stub and never touched the worker. I added `TestWorkerFailFast` in `test_rbln_worker.py`. It calls the real `execute_model`, `sample_tokens` and `execute_dummy_batch` with a model runner that raises, across the executor variants, and taking any of the three decorators off fails the matching cases. I tried it the same way you did.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: I moved the policy check before the `try`. The env flag stays lazy to follow the existing convention in `envs.py` and so tests can change it without reimporting. A regression test now checks that.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: I used `EX_SOFTWARE` (70) for an unhandled worker error. The original exception stays in the log; the multiprocess executor detects process death without interpreting the exit code. Added a comment on the constant and a line in the description.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: Agreed, the docstring promised more than try/finally can do. It now says the exit happens when logging returns or raises, and that there is no exit deadline if logging blocks.

vllm_rbln/envs.py
rebel-scottlee: Renamed all three as you suggested. The old parser also left fail-fast enabled for misspelled values, so that behavior is preserved. The new form matches the other boolean settings, and I've added a typo case to the test.

vllm_rbln/v1/worker/async_output.py
rebel-scottlee: Done. `RBLNWorker` and `RBLNModelRunner` compute `issubclass(Executor.get_class(vllm_config), MultiprocExecutor)` once in `__init__` and the async output takes the bool. The whole `ParallelConfig` no longer needs to be passed around, and the check also covers `RayExecutorV2` and custom `MultiprocExecutor` subclasses given as a class or a qualified name. The worker test has a case for each.

vllm_rbln/platform.py
rebel-scottlee: I switched the guard to the executor-class check you suggested. That includes `RayExecutorV2`, so I kept the warning and left backend selection unchanged. `external_launcher` still propagates the exception in process.

tests/native/test_platform.py
rebel-scottlee: Added `test_default_single_worker_uses_uni_without_warning`. A default config has to come out as `uni` with no executor-backend warning, so if upstream changes the default this fails first.

rebel-cjlee (approved):

PR #1095, opened 2026-09-11
feature(attn): multi-block SWA kernel 

rebel-eunji:
### 🚀 Summary of Changes

This PR adds support for a new SWA kernel that updates the KV cache in an `append` fashion instead of a `shift`.
The new kernel is enabled when `VLLM_RBLN_USE_MULTI_BLOCK_ATTN=1`.
Note that it is not supported yet, but it will be soon.

## Notes
For now, The `kernel_block_size` stays equal to the `window_size, as it was for the shift kernel. sliding_window_attention_v1 itself does not require it — it takes window_size as an operand and reads the block size off the cache tensor — but decoupling the two is left for a follow-up.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/attention/ops/sliding_window_attention.py
```diff
@@ -101,3 +101,32 @@ def sliding_window_attention_naive_decode(
...
+        return torch.ops.rbln_custom_ops.sliding_window_attention_v1(
+            q,
+            k,
+            v,
+            kv_cache,
+            seq_idx,
+            scale,
+            block_tables,
+            window_size,
+            True,  # is_causal
```
rebel-cjlee, line 127: Is it a fixed value, not determined by VLLM_RBLN_FLASH_CAUSAL_ATTN?

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -407,6 +415,17 @@ def __init__(
             self.sinks is None
         )
 
+        self.swa_appends = (
+            sliding_window is not None and envs.VLLM_RBLN_USE_MULTI_BLOCK_ATTN
+        )
+        if self.swa_appends and envs.VLLM_RBLN_USE_CUSTOM_KERNEL:
```
rebel-cjlee: Just curious, is `VLLM_RBLN_USE_CUSTOM_KERNEL` a valid env var for general?
It seems that using this env var often causes compilation errors.

vllm_rbln/patches/attention.py
```diff
@@ -251,7 +258,12 @@ def patched_get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec:
             raise NotImplementedError(
                 "MLA is not supported with sliding window attention."
             )
-        return RBLNSlidingWindowSpec(
+        spec_cls = (
+            SlidingWindowSpec
+            if envs.VLLM_RBLN_USE_MULTI_BLOCK_ATTN
+            else RBLNSlidingWindowSpec
+        )
```
rebel-cjlee, line 265: `RBLNSlidingWindowSpec.__post_init__` asserts `block_size % sliding_window == 0`. Upstream's SlidingWindowSpec does not, so picking it here drops that check. Is it okay?

vllm_rbln/v1/attention/backends/flash_attention.py
```diff
@@ -407,6 +415,17 @@ def __init__(
             self.sinks is None
         )
 
+        self.swa_appends = (
+            sliding_window is not None and envs.VLLM_RBLN_USE_MULTI_BLOCK_ATTN
+        )
```
rebel-cjlee: This is the third place that answers the same question, and it is the only one that stands on its own.

`patched_get_kv_cache_spec` reads the env vars and picks the spec class. Then `build()` at line 295 looks at the spec class, not at the env var, so those two can never disagree. That was the reason for keying `build()` on the spec type. This line reads the env var again, so it is not tied to them.

It's not problem today, bacause all three read the same env var in the same process. But I think it would be better to have on source for this decision.

rebel-cjlee (approved):

rebel-kblee (approved):
lgtm!

vllm_rbln/v1/attention/ops/sliding_window_attention.py
rebel-eunji: The `is_causal` is just a flag for role of mask tensor. If `is_causal` flag is on, the maks is generated in the custom converter.  We are determined to generate mask in the custom converter, then we need to set is_causal=True.
So VLLM_RBLN_FLASH_CAUSAL_ATTN is not related with this code.

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-eunji: `VLLM_RBLN_USE_CUSTOM_KERNEL` controls whether the Triton kernel is used.
In practice, though, it isn't really being used recently, so the code you're referring to is just a guard.

vllm_rbln/patches/attention.py
rebel-eunji: upstream `BlockTable.__init__` replaces the check.
https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/worker/block_table.py#L69

vllm_rbln/v1/attention/backends/flash_attention.py
rebel-eunji: That's a good point. I moved the env check logic building attn_metadata.

PR #1097, opened 2026-09-11
fix(pp): size the worker response queue from batches in flight

rebel-wonsubkim:
### 🚀 Summary of Changes

Derive the depth of a worker's response `MessageQueue` from `max_concurrent_batches` instead of leaving it at the upstream default.

A KV connector makes the executor pass `output_rank=None` to `collective_rpc`, because `KVOutputAggregator` counts `finished_sending` per rank and therefore needs a response from every worker. `worker_busy_loop` then calls `handle_output` on every rank each step, not only on the output rank.

Under pipeline parallelism that meets the pipeline's staggering. The first stage runs ahead of the last one, and a step's future resolves only once the output rank has reported for that same step, so the first stage's responses stay unread for as many steps as the engine keeps in flight -- that is `max_concurrent_batches`, which equals `pipeline_parallel_size`. The upstream ring is a fixed ten slots. Once the pipeline is deep enough the ring saturates, the first stage blocks in `acquire_write`, and because `worker_busy_loop` is serial it stops draining `rpc_broadcast_mq`; the engine then blocks in `collective_rpc` as well and cannot submit the next step. The pipeline runs shallower than `pipeline_parallel_size` while per-stage device time is unchanged.

The fix ties the depth to the quantity that sets the lag and keeps the upstream depth as a lower bound, so TP-only deployments and shallow pipelines are byte-for-byte unchanged.

Changed paths:

* `vllm_rbln/patches/multiproc_executor.py` (new) -- patches `WorkerProc._init_message_queues`
* `vllm_rbln/patches/__init__.py` -- registers the new patch module

---

* Related to rebellions-sw/fsw-inference#588

---

### ✅ Type of Change

---

### 🧪 How to Test

1. Serve a model with `--pipeline-parallel-size 8` and a KV connector attached, for example `--kv-transfer-config '{"kv_connector":"RblnNixlConnector","kv_role":"kv_producer","kv_buffer_device":"rbln"}'`. A consumer is not required; the regression appears with no transfer at all.
2. Send a single long-prompt request with `--max-concurrency 1` so queueing time is zero, and record TTFT.
3. Compare against the same serve command without `--kv-transfer-config`. Before this change the connector run is markedly slower; after it the two match.
4. Edge case: with `--pipeline-parallel-size 1` and with tensor parallelism only, the derived depth falls back to the upstream value, so behaviour is unchanged.

Per-rank confirmation, if a torch profile is collected: the cross-rank `hop` between consecutive stages and the traversal time across all stages stay the same in both runs, while the per-stage `forward` period and the number of chunks in flight change. That separates a pipeline-occupancy regression from a compute regression.

---

### 📋 Checklist

---

### 💬 Notes

The upstream obligation itself is sound: `finished_sending` is a per-rank fact, arrives asynchronously, and the engine cannot tell an empty response from one that has not arrived yet, since `get_response` performs a blocking `dequeue` per queue. Skipping empty responses would need a protocol change. This PR changes only the buffer depth, so every response is still sent, aggregated, and counted exactly as before.

Both `max_chunk_bytes` and the output rank's queue are left untouched, keeping the change to one variable.

### DIFF
--- a/vllm_rbln/patches/multiproc_executor.py
+++ b/vllm_rbln/patches/multiproc_executor.py
@@ -1,0 +1,68 @@
+# Copyright 2026 Rebellions Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Size the worker response queue from the number of batches in flight.
+
+A KV connector makes ``collective_rpc`` pass ``output_rank=None``, so every worker
+enqueues a response each step rather than only the output rank. Under pipeline
+parallelism the first stage runs ahead of the last one, and a step's future
+resolves only once the output rank has reported for that same step, so the first
+stage's responses stay unread for as many steps as the engine keeps in flight.
+The upstream ring is a fixed ten slots, which is fewer than that once the pipeline
+is deep enough: the first stage blocks in ``acquire_write``, and because
+``worker_busy_loop`` is serial it stops draining ``rpc_broadcast_mq``, so the
+engine cannot submit the next step either and the pipeline runs shallower than
+``pipeline_parallel_size``.
+
+Deriving the depth from ``max_concurrent_batches`` ties it to the quantity that
+sets the lag. The lower bound keeps the upstream depth wherever the derived value
+would be smaller, so deployments without pipeline parallelism are unchanged.
+"""
+
+from vllm.config import VllmConfig
+from vllm.distributed.device_communicators.shm_broadcast import Handle, MessageQueue
+from vllm.v1.executor.multiproc_executor import WorkerProc
+
+from vllm_rbln.patches import register_patch
+
+UPSTREAM_MAX_CHUNKS = 10
+
+
+@register_patch(
+    target="vllm.v1.executor.multiproc_executor.WorkerProc._init_message_queues",
+    reason=(
+        "Derive the worker response queue depth from max_concurrent_batches. A KV "
+        "connector makes every pipeline stage enqueue a response per step, and the "
+        "first stage runs that many steps ahead of the point where the engine reads "
+        "them, so the fixed upstream depth leaves it blocked in acquire_write and "
+        "the pipeline runs shallower than pipeline_parallel_size."
+    ),
+    key="vllm_rbln.patches.multiproc_executor.init_message_queues",
+    owner_module="vllm_rbln.patches.multiproc_executor",
+)
+def patched_init_message_queues(
+    self: WorkerProc, input_shm_handle: Handle, vllm_config: VllmConfig
+) -> None:
+    if vllm_config.parallel_config.nnodes_within_dp > 1:
+        _init_message_queues_upstream(self, input_shm_handle, vllm_config)
+        return
+
+    self.rpc_broadcast_mq = MessageQueue.create_from_handle(
+        input_shm_handle, self.worker.rank
+    )
+    max_chunks = max(UPSTREAM_MAX_CHUNKS, vllm_config.max_concurrent_batches * 2)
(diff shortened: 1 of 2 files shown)


vllm_rbln/patches/multiproc_executor.py
```diff
@@ -0,0 +1,68 @@
...
+    if vllm_config.parallel_config.nnodes_within_dp > 1:
+        _init_message_queues_upstream(self, input_shm_handle, vllm_config)
+        return
+
+    self.rpc_broadcast_mq = MessageQueue.create_from_handle(
+        input_shm_handle, self.worker.rank
+    )
+    max_chunks = max(UPSTREAM_MAX_CHUNKS, vllm_config.max_concurrent_batches * 2)
+    self.worker_response_mq = MessageQueue(1, 1, max_chunks=max_chunks)
+    self.peer_response_handles = []
```
rebel-cjlee: What about calling upstream first and only swapping the response queue when derived depth is larger? That way we don't carry a copy of upstream's body, the default depth comes from the queue upstream just built instead of a hard-coded constant here, and the original capture can sit at the top like the sibling patches.

```suggestion
    _init_message_queues_upstream(self, input_shm_handle, vllm_config)

    if vllm_config.parallel_config.nnodes_within_dp > 1:
        return

    # Upstream already built a default-depth queue; replacing it here unlinks
    # its shm segment at once and costs one discarded queue per worker startup.
    max_chunks = vllm_config.max_concurrent_batches * 2
    if max_chunks > self.worker_response_mq.buffer.max_chunks:
        self.worker_response_mq = MessageQueue(1, 1, max_chunks=max_chunks)
```

I checked the discard path: the old queue has no reference cyle, so its shm segment is unlinked right away, and the engine never sees it because `export_handle()` runs after `WorkerProc.__init__` returns.

vllm_rbln/patches/multiproc_executor.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Size the worker response queue from the number of batches in flight.
+
+A KV connector makes ``collective_rpc`` pass ``output_rank=None``, so every worker
+enqueues a response each step rather than only the output rank. Under pipeline
```
rebel-cjlee: It is one response per step? With one per step PP=8 gives 8 unread responses, which still fits in the 10 slot ring, so this sentence can't explain the stall you saw at PP=8. I think what fills the ring is that `step_with_batch_queue` sends 2 RPCs per step, `execute_model` and `sample_tokens`, and both go to every rank once a connector is set. If that's right, it's worth naming it, something like `RPCS_PER_BATCH = 2` with a one line note where it comes from, so the next reader doesn't have to rediscover it.

vllm_rbln/patches/multiproc_executor.py
rebel-cjlee, line 1: Most of this docstring is already in the PR description and in `reason=`. Could you trim it to what a reader needs to change this safely, and leave the story to the PR description?

vllm_rbln/patches/multiproc_executor.py
rebel-wonsubkim: Applied. Calling upstream first is better: the default depth now comes from the queue upstream built, so the constant is gone, and the multi-node branch needs no copy either. Thanks for checking the discard path.

vllm_rbln/patches/multiproc_executor.py
rebel-wonsubkim: You are right, and this closes an arithmetic gap I had left open. `step_with_batch_queue` issues `execute_model` and then `sample_tokens` per step, and both pass `kv_output_aggregator`, so a rank answers twice per step. That makes the outstanding count `max_concurrent_batches * 2`, which is what actually crosses the upstream depth between PP4 and PP8 -- matching the observation that PP4 never hit this. Named as `RPCS_PER_BATCH` with a note on where it comes from, and the depth is now derived from it rather than carrying an unexplained factor.

vllm_rbln/patches/multiproc_executor.py
rebel-wonsubkim: Trimmed to the invariant a reader needs: why the outstanding count is what it is, and what goes wrong when the queue cannot hold it. The rest stays in the PR description.

rebel-cjlee (approved):

PR #1101, opened 2026-09-14
feature: fixed decode query length with specdec

rebel-kblee:
### 🚀 Summary of Changes

A model-based drafter (eagle, eagle3, mtp) proposes on *every* decode step, so its decode query is always `num_spec_tokens + 1` tokens long. Staging that window on every step drops the `qlen=1` decode: the warm-up decode calls per rank halve -- 12 -> 6 at k=1, 18 -> 9 at k=3 -- and the unique decode shapes go 6 -> 4 and 6 -> 5.

A decode query is written as one contiguous KV range inside one block, so the slack between the scheduled tokens and the fixed query length has to go somewhere that keeps the range in a single block. It is split:
- `front_pad = min(slack, num_computed % block_size)` -- as much as the block's used tail can absorb. These are tokens this request already computed, so re-running them rewrites their KV with the same value.
- `back_pad = slack - front_pad` -- the remainder, on slots the request has not reached, which attention never reads because `seq_lens` stays the logical length.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---

### DIFF
--- a/tests/native/v1/worker/test_rbln_model_runner.py
+++ b/tests/native/v1/worker/test_rbln_model_runner.py
@@ -1030,8 +1030,6 @@ class TestMayReorderBatch:
 
 
 class TestDraftInputsFollowTheScheduledToken:
-    # Both drafter paths walk back from the staged query's last slot, which back
-    # padding leaves un-scheduled, so the runner shifts them by that padding.
     STAGED = 8  # two requests, four staged slots each
     LOGICAL = 99  # what the scheduler advanced; the draft must not use it
 
@@ -1095,9 +1093,6 @@ class TestDraftInputsFollowTheScheduledToken:
 
 
 class TestDummyRunDecodeWindowPadding:
-    # make_model_runner builds a real Attention, so these open the NPU --
-    # the marker is what runs them in a fresh process instead of pinning
-    # this session's device.
     pytestmark = pytest.mark.maybe_use_device
 
     NUM_REQS = 2
--- a/tests/native/v1/worker/test_rbln_model_runner.py
+++ b/tests/native/v1/worker/test_rbln_model_runner.py
@@ -282,7 +282,6 @@ class TestPadDepad:
                 padded.cu_num_draft_tokens,
                 torch.tensor([2, 2], dtype=torch.int32),
                 torch.tensor([0, 2, 4], dtype=torch.int32),
-                torch.zeros(2, dtype=torch.int32),
             )
 
 
@@ -983,85 +982,6 @@ class TestCalcSpecDecodeMetadata:
         assert md.bonus_logits_indices.tolist() == [0, 1]
 
 
-class TestPrepareInputsFixedWindow:
-    # _prepare_inputs stages a fixed num_spec_tokens + 1 decode window and fills
-    # the slack in front of the scheduled tokens as far as the block allows, then
-    # behind them. `logits_indices` has to follow the tokens, not the window.
-    BLOCK = 16
-    NUM_SPEC = 3
-
-    def _runner(self, num_computed):
-        runner = _make_runner_stub(
-            num_spec_tokens=self.NUM_SPEC,
-            speculative_config=SimpleNamespace(use_eagle=lambda: True),
-            cache_config=SimpleNamespace(block_size=self.BLOCK),
-            arange_np=np.arange(8),
-            positions=torch.zeros(64, dtype=torch.int64),
-            input_ids=torch.zeros(64, dtype=torch.int32),
-            query_start_loc=torch.zeros(9, dtype=torch.int32),
-            decode_back_pad=torch.zeros(8, dtype=torch.int32),
-            seq_lens_np=np.zeros(8, dtype=np.int32),
-            discard_request_mask=torch.zeros(8, dtype=torch.bool),
-            device=torch.device("cpu"),
-            requests={"0": SimpleNamespace(num_tokens=num_computed + 1)},
-        )
-        runner.is_prefill = False
-        runner.query_start_loc_np = runner.query_start_loc.numpy()
-        runner.decode_back_pad_np = runner.decode_back_pad.numpy()
-        runner.input_batch = SimpleNamespace(
-            num_reqs=1,
-            num_computed_tokens_cpu=np.array([num_computed], dtype=np.int32),
-            token_ids_cpu=np.zeros((1, 64), dtype=np.int32),
-            token_ids_cpu_tensor=torch.zeros(1, 64, dtype=torch.int32),
-            req_id_to_index={"0": 0},
-            num_prompt_tokens=np.array([1], dtype=np.int32),
-            req_ids=["0"],
-        )
-        return runner
-
-    def _prepare(self, num_computed):
-        runner = self._runner(num_computed)
-        scheduler_output = SimpleNamespace(
-            total_num_scheduled_tokens=1,
-            scheduled_spec_decode_tokens={},
-        )
-        logits_indices, _, query_lengths, total = runner._prepare_inputs(
-            scheduler_output, np.array([1], dtype=np.int32)
-        )
-        return runner, logits_indices, query_lengths, total
-
-    @pytest.mark.parametrize(
-        "num_computed,window_start,sample_slot",
-        [
-            # Mid-block: all three slack slots fit in front of the token.
-            (10, 7, 3),
-            # At a block start: nothing to re-run, so all of it goes behind.
-            (16, 16, 0),
-            # One computed token in the block: one in front, two behind.
-            (17, 16, 1),
-        ],
-    )
-    def test_window_is_fixed_and_stays_in_the_block(
-        self, num_computed, window_start, sample_slot
-    ):
-        runner, logits_indices, query_lengths, total = self._prepare(num_computed)
-        window = self.NUM_SPEC + 1
-
-        assert query_lengths.tolist() == [window]
-        assert total == window
-        positions = runner.positions.numpy()[:window].tolist()
-        assert positions == list(range(window_start, window_start + window))
-        # The whole point of the split: one write range, one block.
-        assert window_start // self.BLOCK == positions[-1] // self.BLOCK
-        # The sampled slot is the scheduled token, wherever the padding put it.
-        assert logits_indices.tolist() == [sample_slot]
-        assert positions[sample_slot] == num_computed
-        # seq_lens stays the logical length; the padding must not inflate it.
-        assert runner.seq_lens_np[0] == num_computed + 1
-        # The drafter path reads this to find the same slot.
-        assert runner.decode_back_pad_np[0] == window - 1 - sample_slot
-
-
 class TestMayReorderBatch:
     # Stable descending sort by num_tokens_no_spec, applied in place. Uses a real
     # InputBatch; scheduler_output is unused by the sort path, so None is passed.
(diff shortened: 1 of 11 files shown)


vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -841,23 +843,31 @@ def _prepare_inputs(
...
 
-            assert np.all(backfill >= 0), (
+            assert np.all(slack >= 0), (
                 f"query_lengths={query_lengths}, "
                 f"logical_num_tokens={logical_num_tokens}"
             )
+            block_size = self.cache_config.block_size
+            num_computed = self.input_batch.num_computed_tokens_cpu[:num_reqs]
+            front_pad = np.minimum(slack, num_computed % block_size).astype(np.int32)
+            back_pad = slack - front_pad
```
rebel-cjlee, line 875: I think `back_pad` is missing ;from the sliding window metadata calculation. The metadata builder calculates the previous token count as `seq_len - query_seq_len`, but `query_seq_len` now includes the back padding.

For example, when `num_computed` is 1024, the logical query length is 1, and the fixed window is 3, the staged window starts at position 1024. However, the metadata builder calculates the previous token count as 1022.

This can shift `cache_seq_len`, `cache_offsets`, and the SWA mask by `back_pad`. Could we calculate the cache start from the first staged position, or pass `back_pad` to the metadata builder? It would also be good to add a block boundary test with `RBLNSlidingWindowSpec`.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3080,6 +3116,26 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+            # window would then have to run past max_model_len.
+            window = self.num_spec_tokens + 1
+            block_size = self.cache_config.block_size
+            if block_size < window:
+                raise ValueError(
+                    f"block_size={block_size} cannot hold the {window}-slot "
+                    f"speculative decode window."
+                )
+            remainder = self.max_model_len % block_size
+            if remainder and remainder < window:
```
rebel-cjlee: This check also applies to speculative methods that do not always use a fixed decode window. For example, ngram and suffix call fall back to `qlen=1` when there are no draft tokens or when they reach a block boundary.

A configuration such as `max_model_len=1025`, `block_size=1024`, and `num_spec_tokens=3` can use that fallback, but this check rejects it during KV cache initialization. Should this remainder check only run when `uses_fixed_decode_window` is true?

The `block_size < window` check can still apply to every speculative method because an actual speculative window must fit in one block.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3186,6 +3242,16 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
     # Only RBLN-Specific Methods
     ####################################################################################################
 
+    @property
+    def uses_fixed_decode_window(self) -> bool:
+        """Whether *every* decode step stages a num_spec_tokens + 1 query."""
+
+        return (
+            self.speculative_config is not None
+            and self.num_spec_tokens > 0
```
rebel-cjlee: Is this condition needed? When `self.speculative_config` is not `None`, vLLM validation already guarantees that `num_speculative_tokens` is greater than zero. In this runner, `self.num_spec_tokens` is also copied from that value. So `self.speculative_config is not None` and `self.num_spec_tokens > 0` are effectively the same check here.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3186,6 +3242,16 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
...
+            and self.speculative_config.method in ("eagle", "eagle3", "mtp")
```
rebel-cjlee: Could we use `self.speculative_config.use_eagle()` here instead of keeping another method list? It covers eagle, eagle3, mtp, and dflash. If DFlash can also use the fixed decode window, this property can be simplified to:

```suggestion
            and self.speculative_config.use_eagle()
```

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -841,23 +843,31 @@ def _prepare_inputs(
...
+        # rest behind.
         use_spec_decode = len(scheduler_output.scheduled_spec_decode_tokens) > 0
-        if use_spec_decode and self.num_spec_tokens > 0 and not self.is_prefill:
-            target_query_len = self.num_spec_tokens + 1
-            query_lengths = np.full(num_reqs, target_query_len, dtype=np.int32)
-            backfill = query_lengths - logical_num_tokens
+        window_fixed = not self.is_prefill and (
+            self.uses_fixed_decode_window
+            or (self.num_spec_tokens > 0 and use_spec_decode)
+        )
```
rebel-cjlee, line 863: `use_spec_decode` means that the scheduler provided draft tokens, so speculative decoding is already enabled and `num_spec_tokens` is greater than zero. I think this can be simplified to:

```suggestion
        window_fixed = not self.is_prefill and (
            self.uses_fixed_decode_window or use_spec_decode
        )
```

tests/native/v1/worker/test_rbln_model_runner_inputs.py
```diff
@@ -93,6 +103,50 @@ def test_no_padding_when_scheduler_kept_no_drafts(
...
+        window = self.NUM_SPEC + 1
+
+        logits_indices, spec_md, query_lengths, total = runner._prepare_inputs(
+            make_scheduler_output(num_scheduled_tokens={"a": 1}),
+            np.array([1], dtype=np.int32),
+        )
+
+        assert query_lengths.tolist() == [window]
+        positions = runner.positions[:window].tolist()
+        assert positions == list(range(window_start, window_start + window))
```
rebel-cjlee, line 148: This test stops after checking the runner buffers, so it does not cover how the fixed winodw is converted into attention metadata. The block boundary case can still produce incorrect `cache_seq_len` and `cache_offsets` for `RBLNSlidingWindowSpec` while this test passes. Could we also build the attention metadata for this case and verify the sliding window cache values?

tests/native/v1/worker/test_rbln_model_runner_inputs.py
```diff
@@ -93,6 +103,50 @@ def test_no_padding_when_scheduler_kept_no_drafts(
...
+        _decode_ready(
+            runner,
+            monkeypatch,
+            num_spec_tokens=self.NUM_SPEC,
+            num_computed=num_computed,
+        )
+        window = self.NUM_SPEC + 1
+
+        logits_indices, spec_md, query_lengths, total = runner._prepare_inputs(
+            make_scheduler_output(num_scheduled_tokens={"a": 1}),
```
rebel-cjlee, line 142: The block boundary case only covers a step without scheduled draft tokens. Could we also add a block boundary case with actual draft tokens? That would cover the new `cu_num_tokens - back_pad` path and verify that `SpecDecodeMetadata.logits_indices` stops before the back padding.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3080,6 +3116,26 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
                 "Set VLLM_RBLN_SUB_BLOCK_CACHE=false to disable."
             )
 
+        if self.num_spec_tokens > 0:
```
rebel-cjlee, line 3143: Could we add tests for these new validation errors? It would be useful to cover a window larger than the block, a fixed window that does not fit in the last block, and a non fixed method with the same last block remainder. The last case would make sure this check does not reject methods that can fall back to `qlen=1`.

tests/native/v1/worker/test_rbln_model_runner.py
```diff
@@ -1062,6 +1062,39 @@ def test_emits_swap_records_for_non_pooling(self):
...
+            (False, True, 1),
+            (True, False, 1),
+        ],
+    )
+    def test_only_a_warm_up_with_a_fixed_window_pads(
+        self, make_model_runner, monkeypatch, fixed_window, warmup, expected_tokens
+    ):
+        runner = make_model_runner()
+        monkeypatch.setattr(runner, "num_spec_tokens", self.NUM_SPEC)
+        monkeypatch.setattr(type(runner), "uses_fixed_decode_window", fixed_window)
```
rebel-cjlee, line 1137: This replaces `uses_fixed_decode_window` with a boolean, so the test does not verify the new property itself. Could we add a small parameterized test for the supported methods? If we use `SpeculativeConfig.use_eagle()`, the expected fixed window methods would be eagle, eagle3, mtp, and dflash.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-kblee: Confirmed, and your example is exact.
I fix it to append the back pad in swa meta

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-kblee: Confirmed. It was duplicated condition. I'll drop it

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-kblee: I agreed 👍

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -963,6 +986,7 @@ def _build_attention_metadata(
         max_query_len: int,
         num_reqs_padded: int,
         logits_indices: torch.Tensor | None = None,
+        back_pad: torch.Tensor | None = None,
     ) -> tuple[PerLayerAttnMetadata, CommonAttentionMetadata | None]:
```
rebel-minkyu, line 1000: Could you check if we should apply this similarly to the drafter's attention metadata?

rebel-kblee: Good catch. Thanks! I apply it 👍

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -3192,6 +3248,14 @@ def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
     # Only RBLN-Specific Methods
     ####################################################################################################
 
+    @property
+    def uses_fixed_decode_window(self) -> bool:
+        """Whether *every* decode step stages a num_spec_tokens + 1 query."""
+
+        if self.speculative_config is None:
+            return False
+        return self.speculative_config.use_eagle()
```
rebel-minkyu: Apologies for another late review. I think we should exclude dflash back here. dflash doesn't use `token_indices_to_sample` for sampling, so subtracting back_pad from `token_indices_to_sample` has no effect on it, which can give wrong drafts.

Fundamentally, we better change dflash to use the given `token_indices_to_sample` the way eagle does, but for now I think dropping dflash here and leaving a comment(for why dflash is excluded and a possible OOM with dflash) would be enough. I'll keep this in my TODO list to handle back_pad properly in dflash.

rebel-kblee: I’ve checked it. As for dflash, I haven’t been able to follow up on everything yet either, so thank you for checking it for me!

rebel-minkyu (approved):
LGTM!

rebel-cjlee (approved):

PR #1102, opened 2026-09-14
fix(optimum): kv cache unused blocks padding style alignment

rebel-jaehob:
### 🚀 Summary of Changes

The optimum model runner filled a request's unused block-table slots with -1 after shifting vLLM's 1-based block ids to compiler ids. The in-memory flash-causal attention kernel (CR13) gathers block_table[row, p] for every row of a group whenever partition p is live for any row, so at batch > 1 with requests of unequal length the -1 became an out-of-range KV-cache DMA: the firmware reported an RBDMA bus_read_err and reset the device, which the host saw as SYS_TASK_ABORTED in decode (vllm-optimum-ci-v2, qwen2.5-vl-7b, -b 4). Repeat the row's last block instead; index_list already masks that row's keys in the partition, so the block is only read. This matches what optimum-rbln's own block-table manager and pad_decoder_items already do for their padding rows. mask_block_table becomes a staticmethod so the contract is unit-testable without a runner.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-eunji (approved):
lgtm

PR #1104, opened 2026-09-14
feature(dynamic-kv): size the KV cache from the compiled placement and a per-chiplet memory snapshot

rebel-jonghewk:
### 🚀 Summary of Changes

Replace profile-based dynamic KV cache sizing with sizing derived from the compiled physical placement and a per-chiplet memory snapshot.

The compiled placement is used to determine KV growth on each chiplet. After warm-up, the compile-time cache is released, retained memory is included in the base usage, and the largest safe KV block count is selected using the runtime allocator's reservation behavior.

This change applies only to the **vLLM-native path**. The **optimum-rbln path** is unchanged.

It also extends dynamic sizing to hybrid attention, MLA/DSA caches, and speculative decoding while preserving compile-only and block-override behavior.

#### Why this approach

The previous implementation merged memory profiles and used whole-device accounting. That could diverge from the placement produced by the compiler and hide uneven memory use across chiplets.

The new implementation treats the compiled placement as the source of truth and measures available memory after releasing the compile-time cache. The final fit follows the caching allocator's reservation behavior instead of treating logical tensor bytes as the physical reservation.

Allocator reservations are not monotonic across all request sizes, so the fit scans downward from a linear upper bound rather than using bisection. Unsupported or ambiguous placements fail at startup instead of producing an unsafe block count.

The pre-compile estimate remains a compilation hint. The measured post-warm-up capacity is the final authority and must satisfy the minimum pool required by the scheduler.

#### Behavior and compatibility

- `VLLM_RBLN_USE_DYNAMIC_KV_CACHE=1` enables post-warm-up sizing and reallocation.
- `VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN=1` reports the proposed count and per-chiplet headroom without changing the allocated block count.
- Dry-run compilation still marks KV inputs dynamic, so its compiled output may differ from a run without dynamic sizing.
- `--num-gpu-blocks-override` continues to pin the actual block count; dry-run mode still reports the proposed dynamic count.
- Device memory information is preferred when available. Compile-only runs and unsupported driver versions use the existing fallback accounting path.
- DSA cache support requires compatible `rebel_compiler` and `torch-rbln` versions.

#### Affected modules

- Worker and scheduler KV cache sizing
- Attention KV cache bindings and MLA indexer cache handling
- vLLM patch registration and platform configuration
- Dynamic KV cache documentation
- Native test suite

---

- Related to #870

---

### ✅ Type of Change

- [ ] 🚀 Release (`release`)
- [x] ✨ Feature (`feature`)
- [x] 🧬 Core engine changes (`core`)
- [x] 🛠 Bug fix (`fix`)
- [ ] ⚙️ Performance improvement (`perf`)
- [x] 📄 Documentation (`docs`)
- [ ] ❓ Other (`other`)

---

### 🧪 How to Test

Run the affected native tests:

```bash
uv run --no-sync pytest \
  tests/native/test_envs.py \
  tests/native/test_platform.py \
  tests/native/patches/test_dynamic_kv.py \
  tests/native/v1/core/test_utils.py \
  tests/native/v1/attention/test_flashattn_mla.py \
  tests/native/v1/attention/test_kv_cache_bindings.py \
  tests/native/v1/worker/test_dynamic_kv_sizer.py \
  tests/native/v1/worker/test_kv_placement.py \
  tests/native/v1/worker/test_rbln_worker.py \
  tests/native/v1/worker/test_utils.py
```

Expected: all selected tests pass.

Exercise the rollout path with an existing vLLM-native device workload:

```bash
VLLM_RBLN_USE_VLLM_MODEL=1 \
VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN=1 \
<existing workload command>
```

Expected: the run reports a proposed block count and per-chiplet headroom without reallocating the KV cache.

Then enable `VLLM_RBLN_USE_DYNAMIC_KV_CACHE=1`. Expected: the resized cache remains within the configured memory budget and the workload passes its existing correctness checks.

---

### 📋 Checklist

- [ ] This PR is linked to an existing development issue
- [x] Relevant documentation has been updated

---

### 💬 Notes

- The allocator-aware fit intentionally models physical reservation rather than logical tensor size.
- The downward scan is required because allocator size-class transitions make the fit non-monotonic.
- The allocator fallback is less complete than the driver snapshot and keeps conservative reserves.
- The copy-stream reserve for sub-block prefix caching still requires calibration.


vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,783 @@
...
+            )
+            estimate = one_request
+        return estimate
+
+    def shrink_for_compile(self, kv_cache_config: KVCacheConfig) -> KVCacheConfig:
+        """A small-KV-cache copy of the config, or it unchanged; its
+        `num_blocks` is the hint warm-up traces the dynamic dim with."""
+        if not envs.VLLM_RBLN_USE_DYNAMIC_KV_CACHE:
+            return kv_cache_config
+        skip_reason = self.worker._compile_and_warmup_skip_reason()
```
rebel-eunji: _compile_and_warmup_skip_reason is called so frequently... If it is not called like this, then self.worker is not required anymore.

rebel-eunji:
```python
class DynamicKvMode(Enum):
    DISABLED = "disabled"   # dynamic env off: every method is an immediate no-op
    INERT = "inert"         # dynamic env on, but compile/warm-up is skipped: nothing to do
    PINNED = "pinned"       # --num-gpu-blocks-override pins the count
    DRY_RUN = "dry_run"     # measure and report only, no resize
    ACTIVE = "active"       # full shrink + resize


def resolve_mode(
    *,
    use_dynamic_kv: bool,
    dry_run: bool,
    num_gpu_blocks_override: int | None,
    compile_skip_reason: str | None,
) -> tuple[DynamicKvMode, str | None]:
    """The mode and its reason. Every input is static config/env, so the
    decision is final at init time."""
    if not use_dynamic_kv:
        return DynamicKvMode.DISABLED, None
    if compile_skip_reason is not None:
        return DynamicKvMode.INERT, compile_skip_reason
    if dry_run:
        return DynamicKvMode.DRY_RUN, None
    if num_gpu_blocks_override is not None:
        return (
            DynamicKvMode.PINNED,
            f"--num-gpu-blocks-override={num_gpu_blocks_override}",
        )
    return DynamicKvMode.ACTIVE, None
```
How about using like this? And just replace the branch with using this mode?

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,783 @@
...
+            )
+    return device_count > 0
+
+
+class DynamicKvSizer:
+    """The worker's dynamic-KV state machine: shrink -> capture -> snapshot ->
+    size -> release -> reallocate -> materialize -> check."""
+
+    def __init__(self, worker: "RBLNWorker") -> None:
+        self.worker = worker
```
rebel-eunji: It looks like they are referencing each other.
- worker has self.dynamic_kv
- dynamic_kv_size has self.worker

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,783 @@
...
+    def cache_config(self):
+        return self.worker.cache_config
+
+    @property
+    def scheduler_config(self):
+        return self.worker.scheduler_config
+
+    @property
+    def model_runner(self):
+        return self.worker.model_runner
```
rebel-eunji: Just passing vllm_config to init func resolves it.. We don't need to do like this.

rebel-eunji: And it looks like accessing model_runner requires some additional binding functions after init func.

vllm_rbln/v1/worker/utils.py
```diff
@@ -284,6 +287,48 @@ def read_rbln_card_dram_used_bytes() -> int:
...
+
+def minimum_kv_blocks(vllm_config: VllmConfig, cfg: KVCacheConfig) -> KvMinimum:
+    """Blocks one max-length request and one full decode batch need, summed over
+    the groups sharing the pool."""
+    max_model_len = vllm_config.model_config.max_model_len
+    one_request = 0
+    per_seq = 0
+    for group in cfg.kv_cache_groups:
+        spec = group.kv_cache_spec
+        one_request += spec.max_memory_usage_bytes(vllm_config) // spec.page_size_bytes
```
rebel-eunji: vllm usually uses `cdiv`, not `//` (floor division)

vllm_rbln/v1/worker/utils.py
```diff
@@ -284,6 +287,59 @@ def read_rbln_card_dram_used_bytes() -> int:
...
+        if admission is None and isinstance(
+            spec, (SlidingWindowSpec, ChunkedLocalAttentionSpec)
+        ):
+            raise AttributeError(
+                f"{type(spec).__name__} no longer exposes "
+                "max_admission_blocks_per_request; the per-sequence minimum "
+                "would silently fall back to one block."
+            )
+        per_seq += (
+            admission(max_num_batched_tokens=1, max_model_len=max_model_len)
```
rebel-eunji: The parameter `max_num_batched_tokens` is renamed to `max_in_flight_tokens` after bump.

https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/kv_cache_interface.py#L502

vllm_rbln/v1/worker/utils.py
```diff
@@ -284,6 +289,60 @@ def read_rbln_card_dram_used_bytes() -> int:
...
+            spec, (SlidingWindowSpec, ChunkedLocalAttentionSpec)
+        ):
+            raise AttributeError(
+                f"{type(spec).__name__} no longer exposes "
+                "max_admission_blocks_per_request; the per-sequence minimum "
+                "would silently fall back to one block."
+            )
+        # Positional: the first parameter is max_num_batched_tokens before the
+        # vllm bump and max_in_flight_tokens after it; the position is the same.
+        per_seq += admission(1, max_model_len) if admission is not None else 1
```
rebel-eunji, line 339: nit: This logic seems a little bit unstable... Let's refactoring this logic later.

vllm_rbln/patches/dynamic_kv.py
rebel-jongholee, line 112: Shouldn't we recompute them with get_kv_cache_capacity() after the resize like [this](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/engine/core.py#L315)??

vllm_rbln/v1/worker/kv_placement.py
```diff
@@ -0,0 +1,436 @@
...
+    reserve_bytes: int,
+) -> dict[Unit, ChipletMemory]:
+    """From `torch.rbln.memory_stats_per_chiplet()`: this process's allocator,
+    plus the other tenants' usage spread evenly and the runtime reserve."""
+    reserved = _per_unit(stats, "reserved.current")
+    if not reserved:
+        raise RuntimeError(
+            "memory_stats_per_chiplet() reported no `reserved.current` per chiplet; "
+            "the allocator has no device context yet."
+        )
```
rebel-eunji, line 328: What is reserved.current?
As I know snapshot_from_allocator is called before pre_compile.
At that time, there's no reserved memory for context

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,801 @@
...
+        # The count vLLM sized, held while the cache is shrunk for the compile.
+        self.kv_blocks_before_shrink: int | None = None
+        self.programs: list[Any] = []
+        self.expected_used: dict[Unit, int] = {}
+        # Other tenants' card DRAM at init; the allocator-snapshot fallback only.
+        self.foreign_dram_used_bytes = foreign_dram_used_bytes
+        logger.debug(
+            "foreign device DRAM at worker init: %d bytes (RBLN_VISIBLE_DEVICES=%s)",
+            foreign_dram_used_bytes,
+            os.environ.get("RBLN_VISIBLE_DEVICES", ""),
```
rebel-eunji, line 189: Is backwardcompatibility of RBLN_VISIBLE_DEVICES okay? (RBLN_DEVICES)

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,801 @@
...
+        )
+
+    def pre_compile_estimate(self, estimate_kwargs: dict[str, Any]) -> int:
+        """The bytes vllm sizes the compile-time cache from: the estimate fed
+        the per-chiplet snapshot on a real device, floored at one request."""
+        if self.mode is DynamicKvMode.DISABLED:
+            return estimate_available_memory(**estimate_kwargs)
+        if not torch.rbln.is_dummy_device():
+            snapshot, source = self.memory_snapshot(self.device)
+            if envs.VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN:
```
rebel-eunji, line 210:
```suggestion
            if self.mode is DynamicKvMode.DRY_RUN:
```

vllm_rbln/v1/worker/kv_placement.py
```diff
@@ -0,0 +1,436 @@
...
+    for program in programs:
+        specs = [
+            spec for spec in program.input_specs if spec.physical_placement is not None
+        ]
+        if not specs:
+            continue
+        key = tuple(sorted(_input_key(spec) for spec in specs))
+        if key in groups:
+            continue
+        fingerprint = tuple(sorted(_tensor_fingerprint(spec) for spec in specs))
```
rebel-eunji, line 153: Partial intersection of kv cache triggers double calculation. Is it okay?

rebel-eunji:
@rebel-cjlee @rebel-minkyu @rebel-huijong @rebel-jongholee 

Could you take a look this PR? It will be merged today.

vllm_rbln/v1/worker/utils.py
rebel-cjlee: Is the `1` here on purpose?

I know that `max_admission_blocks_per_request` is vLLM's answer to "how many KV blocks does one request need before the scheduler can admit it". Its first parameter is the number of tokens that are scheduled but not settled yet. That matters for a sliding window, because blocks failing out of the window are freed on a processed token basis. Tokens that are scheduled but not processed still hold their blocks, so a bigger value means one request holds more blocks at its peak.

vLLM uses this function on both sides of the same promise. It sizes the pool with it at startup, and it admits requests with it at runtime. The docstring says that is deliberate, so a request admitted at startup can also be admitted at runtime. Here we pass `1`, while the runtime passes the real `max_in_flight_tokens`. So our number is smaller than the one that gets enforced later.

To put number on it, with `block_size=4096`, `sliding_window=1024`, `max_model_len=32768` and a prefill chunk of 8192, `admission(1)` is 2 blocks and `admission(8192` is 4. With `max_num_seqs=16` the `decode_batch` term comes out as 32 instead of 64.

That is the term the error message below calls "max_num_seqs decode steps". With `1` we are not really checking that a full decode batch fits. This is not a hard failure the way the `one_request` term is, since the scheduler will just run fewer sequence or preempt. But the floor stops guaranteeing what its own message says.

The two terms in this function also disagree with each other. `one_request` on line 325 goes through `max_memory_usage_bytes`, and that call passes the real `max_in_flight_tokens` internally. So within a few lines we measure the same quantity two ways.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,801 @@
...
+            compile_num_blocks,
+            kv_cache_config.num_blocks,
+        )
+        return shrunk
+
+    def assert_attention_layout(self) -> None:
+        """Every attention layer must dispatch to a paged causal or sliding-window
+        naive kernel (`is_causal`, not `is_normal`). Runs here, not in platform
+        validation: the layers exist only after the model build."""
+        if self.mode is DynamicKvMode.DISABLED:
```
rebel-cjlee: These two (here and R327) only return early on `DISABLED`. `shrink_for_compile` on line 252 tells the user "this feature does nothing for this run" when the mode is `INERT`, but these layout checks still run and can refuse the run. Nothing compiles in `INERT`, so there is no layout to protect. Should `INERT` skip these as well?

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -249,6 +171,9 @@ def _init_device_env(self) -> None:
     def init_device(self) -> None:
         self.device = self.device_config.device
 
+        # Before the CCL backend comes up, so nothing of ours counts as foreign.
+        foreign_dram_used_bytes = read_rbln_card_dram_used_bytes()
```
rebel-cjlee: This reads the card DRAM usage on every RBLN start, including runs where the dyanmic KV feature is off. The value is only used by the sizer's allocator snapshot fallback. Its cost is tiny, but it is new behavior on the default path.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,801 @@
...
+        self.programs.extend(programs)
+        logger.info(
+            "[Dynamic KV] captured %d compiled program(s) during warm-up.",
+            len(programs),
+        )
+
+    def pre_compile_estimate(self, estimate_kwargs: dict[str, Any]) -> int:
+        """The bytes vllm sizes the compile-time cache from: the estimate fed
+        the per-chiplet snapshot on a real device, floored at one request."""
+        if self.mode is DynamicKvMode.DISABLED:
```
rebel-cjlee: This early return is what keeps the default path unchanged. When the feature is off, the worker still calls `pre_compile_estimate`, and this line is the only thing that sends it to the plain estimate instead of the per chiplet shanphost.

There is no test for it. If it breaks, every run with the flag off starts sizing from a device snapshot and nothing catches it.

The clamp a few lines below is untested too. That one raises the estimate on purpose when it falls short of one request, so a silent change there would be hard to notice.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,801 @@
...
+
+
+class DynamicKvSizer:
+    """The worker's dynamic-KV state machine: shrink -> capture -> snapshot ->
+    size -> release -> reallocate -> materialize -> check."""
+
+    def __init__(
+        self,
+        vllm_config: VllmConfig,
+        model_runner: Any,
```
rebel-cjlee: We can use the concrete type without an import cycle, by using `TYPE_CHECKING`.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,801 @@
...
+        self.expected_used = {
+            unit: fits[unit].base + predicted[unit] for unit in predicted
+        }
+        return num_blocks, fits, snapshot, reserve_bytes
+
+    def _size_kv_and_release(self) -> KvSizing:
+        """Release the compile-time cache, then size from what the runtime
+        actually handed back; no KV cache is bound until `apply_num_blocks`."""
+        growth, hint_blocks, device = self._kv_growth_from_programs()
+        self.release_kv_cache_tensors(self.model_runner.kv_cache_config)
```
rebel-jongholee, line 635: Could we snapshot before/after the release and WARN with the retained bytes per chiplet when the drop falls short of growth.allocated_at(hint_blocks)?

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,846 @@
...
+
+
+class DynamicKvSizer:
+    """The worker's dynamic-KV state machine: shrink -> capture -> snapshot ->
+    size -> release -> reallocate -> materialize -> check."""
+
+    def __init__(
+        self,
+        vllm_config: VllmConfig,
+        model_runner: Any,
```
rebel-cjlee: We can use the concrete type without an import cycle, by using TYPE_CHECKING.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,846 @@
...
+        logger.info(
+            "[Dynamic KV] captured %d compiled program(s) during warm-up.",
+            len(programs),
+        )
+
+    def pre_compile_estimate(self, estimate_kwargs: dict[str, Any]) -> int:
+        """The bytes vllm sizes the compile-time cache from: the estimate fed
+        the per-chiplet snapshot on a real device, floored at one request when
+        the shrink makes that estimate a placeholder."""
+        if self.mode is DynamicKvMode.DISABLED:
```
rebel-cjlee, line 223: This early return is what keeps the default path unchanged. When the feature is off, the worker still calls `pre_compile_estimate`, and this line is the only thing that sends it to the plain estimate instead of the per chiplet shanphost.

There is no test for it. If it breaks, every run with the flag off starts sizing from a device snapshot and nothing catches it.

The clamp a few lines below is untested too. That one raises the estimate on purpose when it falls short of one request, so a silent change there would be hard to notice.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,846 @@
...
+            compile_num_blocks,
+            kv_cache_config.num_blocks,
+        )
+        return shrunk
+
+    def assert_attention_layout(self) -> None:
+        """Every attention layer must dispatch to a paged causal or sliding-window
+        naive kernel (`is_causal`, not `is_normal`). Runs here, not in platform
+        validation: the layers exist only after the model build."""
+        if self.mode is DynamicKvMode.DISABLED:
```
rebel-cjlee: These two (here and R350) only return early on `DISABLED`. `shrink_for_compile` on line 252 tells the user "this feature does nothing for this run" when the mode is `INERT`, but these layout checks still run and can refuse the run. Nothing compiles in `INERT`, so there is no layout to protect. Should `INERT` skip these as well?

vllm_rbln/v1/worker/utils.py
```diff
@@ -235,37 +240,37 @@ def _read_card_attr_int(card_index: int, attr: str) -> int | None:
...
+    if torch.rbln.is_dummy_device():
         return None
-    distinct = set(values.values())
-    if len(distinct) != 1:
-        raise RuntimeError(
-            "visible RBLN cards report different dram_total values "
-            f"({values}); a single per-chiplet DRAM budget cannot be derived."
+    try:
+        total = int(torch.rbln.get_device_properties().total_memory)
+    except RuntimeError as exc:
```
rebel-cjlee: This `try` is here so that a `torch_rbln` without `get_device_properties` falls back to the built in DRAM capacity. It catches `RuntimeError`, but a missing symbol raises `AttributeError`, so the fallback never runs. I reproduced it on the pinned `torch_rbln:0.5.0rc0`.

`estimate_available_memory` calls this whenever the flag is on, so the run dies before the model loads instead of falling back.

The test for this case in `tests/native/v1/worker/test_utils.py` simulates it with `side_effect=RuntimeError`, and every patch around it passes `create=True`. That `create=True` is itself the sign the attribute does not exist.

This feature needs a new `torch_rbln` anyway, since `capture_programs` is missing from the same pin, so it may not bite in practice. but the fallback claims to handle old versions and it does not.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,846 @@
...
+        else:
+            logger.warning(
+                "[Dynamic KV] this torch_rbln has no mem_get_info_per_chiplet(); "
+                "sizing from this process's allocator instead."
+            )
+        # Cached-but-free blocks would otherwise count as used.
+        torch.rbln.empty_cache(device)
+        memory_per_chiplet = int(
+            torch.rbln.get_device_properties(device).memory_per_chiplet
+        )
```
rebel-cjlee: We only reach this fallback when `mem_get_info_per_chiplet` is missing, which means an older `torch_rbln`. But `get_device_properties` is missing from those same versions, so the fallback raises on its own first line instead of sizing from the allocator.

`test_a_torch_rbln_without_the_query_falls_back_too` passes because the fake always provides `get_device_properties`.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,846 @@
...
+                "VLLM_RBLN_USE_DYNAMIC_KV_CACHE does not support cross-layer KV "
+                f"sharing, but {len(mr.shared_kv_cache_layers)} layer(s) reuse "
+                "another layer's KV cache.",
+            )
+
+    def capture_programs(self):
+        """Scope that records the programs warm-up builds, when the flag is on."""
+        if self.mode is DynamicKvMode.DISABLED:
+            return nullcontext(None)
+        if not has_torch_rbln:
```
rebel-cjlee: `has_torch_rbln` is set by a plain `import torch.rbln` at module load, so it only tells us the module exists. A version that imports fine but has no `capture_programs` slips past this check, and warm up dies later with a bare `AttributeError` instead of the message written here.

Probing the attribute would make the message do its job.

vllm_rbln/envs.py
```diff
@@ -285,11 +286,21 @@ def use_auto_port() -> bool:
...
     ),
     # --- DYNAMIC KV CACHE ---
-    # Size the KV cache from the compiled artifact instead of the estimate
+    # Size the KV cache from the compiled artifact instead of the estimate.
+    # The dry-run variable below implies it.
     "VLLM_RBLN_USE_DYNAMIC_KV_CACHE": (
         lambda: (
             os.environ.get("VLLM_RBLN_USE_DYNAMIC_KV_CACHE", "False").lower()
             in ("true", "1")
+            or os.environ.get("VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN", "False").lower()
```
rebel-cjlee: This makes `VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN` turn the feature on, which is handy, But it also overrides a value the user set on purpose: `VLLM_RBLN_USE_DYNAMIC_KV_CACHE=0 VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN=1` -> USE = True

A user who explicitly turned the feature off gets it back on. The refusals in `platform.py` then name `VLLM_RBLN_DYNAMIC_KV_CACHE`, a variable that user never set to `1`. Letting an explicit `0` win, or warning on that combination, would be less surprising.

vllm_rbln/envs.py
```diff
@@ -402,6 +413,7 @@ def use_auto_port() -> bool:
         "VLLM_RBLN_NUM_HIDDEN_LAYERS",
         "VLLM_RBLN_USE_DEVICE_TENSOR",
         "VLLM_RBLN_USE_DYNAMIC_KV_CACHE",
+        "VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN",
```
rebel-cjlee, line 421: `RBLN_COMPILE_ENV` is the set of variables that go into the mega cache bundle key. A dry run does not shrink the cache and does not resize it, so the graphs it compiles are identical to a real run's.

 With this variable in the compile set, measuring with a dry run and then enabling the real mode produces a different bundle key for the same graphs, and the whole model recompiles. `RBLN_NON_COMPILE_ENV` looks like the right side. Let me know if there is a compile effect I'm missing.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: Resolved in 6cfc0d5b: the skip reason is read once at init via `resolve_mode()`, and the sizer no longer holds the worker.

rebel-jonghewk: Applied in 6cfc0d5b as proposed, same ordering (so a dry run still wins over an override). Every branch reads `self.mode` now; the last direct env read went in cc60131b.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: Gone in 6cfc0d5b: `DynamicKvSizer(vllm_config, model_runner, foreign_dram_used_bytes)`. Only the worker holds the reference.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: Done. `vllm_config` gives the cache/scheduler/model config, `device_config.device` and `parallel_config.rank`.

rebel-jonghewk: Right, so it is built at the end of `init_device()`, where the model runner already exists. Nothing is bound afterwards.

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: Fixed in 1153cb0d.

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: Fixed in 1153cb0d by calling it positionally: the name changes across the bump but the position does not, so it works on both. Verified against vllm 0.26.0 locally.

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: Agreed, left for a follow-up.

vllm_rbln/v1/worker/kv_placement.py
rebel-jonghewk: `snapshot_from_allocator` is only the fallback for a torch_rbln without `mem_get_info_per_chiplet()`, and its one caller is `pre_compile_estimate`, which runs from `determine_available_memory()` -- after `load_model()`. The weights are on the device by then, so `reserved.current` is populated; the raise covers the case where there genuinely is no device context.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: Behaviour was already fine: `_adopt_deprecated_device_control_env_var()` folds `RBLN_DEVICES` into `device_control_env_var` and unsets it before the worker is built. The log was hardcoding one of the two names -- fixed in cc60131b to print `current_platform.device_control_env_var`.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: Applied in cc60131b.

vllm_rbln/v1/worker/kv_placement.py
```diff
@@ -0,0 +1,436 @@
...
+        mem = snapshot[unit]
+        fits[unit] = UnitFit(
+            total=mem.total,
+            budget=int(mem.total * gpu_memory_utilization),
+            used=mem.used,
+            kv_resident=int(kv_resident.get(unit, 0)),
+            reserve=reserve_bytes,
+            base=mem.used - int(kv_resident.get(unit, 0)) + reserve_bytes,
+            per_block=per_block,
+            num_blocks=linear[unit],
```
rebel-cjlee: `UnitFit.num_blocks` is filled with `linear[unit]`, which is the count before allocator rounding is applied. The count we actually choose comes from the scan on line 393, and rounding can push it well below the linear value.

When that scan reaches 0, `_size_kv_from_snapshot` raises "on some chiplet the non-KV base already exceeds the budget", while `format_fits` in the same message prints `blocks=N` with N above 0 for every unit. Those two statements contradict each other, and an operator will go raise `--gpu-memory-utilization` when the real cause is segment rounding.

Carrying the chosen count as its own field would make the message match.

vllm_rbln/v1/worker/kv_placement.py
rebel-jonghewk: You are right that nothing prevents it. Measured: Llama 3.2 3B gives 1 set of 28 inputs for 28 layers, GPT-OSS 2 sets of 18 for 36 -- and the two GPT-OSS sets are a partial intersection (17 of 18 arg positions shared). Both sum to the layer count today, but that is luck. `InputSpec` carries no tensor identity to dedupe by, and the check I could write (spec count vs `kv_cache_tensors`) would also fire on GPT-OSS, so cc60131b logs what the slope is summed over next to what vllm allocated instead. A real fix needs tensor identity on the spec, i.e. a compiler-side change.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: Done in c8b7dc07: snapshot before the release too, and warn with the retained bytes per chiplet when the drop falls short of `growth.allocated_at(hint_blocks)`.

docs/dynamic_kv_cache.md
```diff
@@ -5,30 +5,78 @@ memory. That estimate is a whole-card figure and has no notion of chiplets, so o
...
+- **Base** -- how many bytes are already spoken for on each chiplet. A per-chiplet
+  memory snapshot is taken after the compile-time cache is released, so what
+  the runtime does not hand back is measured as base rather than assumed away:
+  `torch.rbln.mem_get_info_per_chiplet()` (the driver's view, every process
+  included) when the UMD/KMD provide it, otherwise this process's caching
+  allocator (`torch.rbln.memory_stats_per_chiplet()`) plus a fixed reserve for the
+  runtime's direct allocations and the other tenants' usage sampled at start-up.
+
+The count is the tightest chiplet's
+`floor((total * gpu_memory_utilization - base) / per_block)`. When the
```
rebel-cjlee: This presents `floor((total * gpu_memory_utilization - base) / per_block)` as the answer. In the code that value is the starting point, and `max_num_blocks` scans downward from it because the allocator rounds each shard up to a segment size.

The Growth bullet does mention the rounding, but a reader who stops here will take the formula as final.

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,877 @@
...
+    minimum_kv_blocks,
+    rescale_kv_cache_config,
+)
+
+try:
+    import torch.rbln  # noqa: F401
+
+    has_torch_rbln = True
+except ImportError:
+    has_torch_rbln = False
```
rebel-cjlee: Do we need this check at all? I ended up with a more basic question: is vllm-rbln meant to work without `torch-rbln` installed?

`torch-rbln` is not in `pyproject.toml`, so the question is fair. But inside this file the flag guards only 2 of the 12 `torch.rbln.*` calls. The other ten call is directly. `torch.rbln.is_dummy_device()` is called with no guard.

So one of two things is true. Either `torch-rbln` is required here, and then the guard is dead weight and the error message on line 369 can never be reached. Or it is genuinely optional, and then the ten unguarded calls are the real problem and line 369 is too late to help.

Also worth nothing, this is the third copy of the same try block in `platform.py` and `rbln_worker.py`.

vllm_rbln/patches/dynamic_kv.py
rebel-jonghewk: You're right -- `_initialize_kv_caches` derives `kv_cache_size_tokens` and `kv_cache_max_concurrency` from the shrunk compile-time config, and nothing recomputed them after the resize, so both stayed at the hint's values. Fixed: `get_kv_cache_capacity()` on the resized config, as upstream does.

vllm_rbln/envs.py
rebel-jonghewk: The graphs differ today: a dry run does not shrink, so it traces the KV dim at the count vllm sized, while the real mode traces at the compile hint. Same model, gpt-oss: dry compiled `(2, 1032, 8, 1, 128, 64)`, real `(2, 256, 8, 1, 128, 64)`. Nothing else in the bundle key separates them -- `CacheConfig.compute_hash()` ignores `num_gpu_blocks` as a derived counter -- so dropping this variable would let one replay the other's codegen, which is wrong in both directions.

The reason it shapes the graph at all is that the compiler bakes the extent of the `mark_dynamic`'d dim (the K->V offset is a compile-time immediate = extent * per-block bytes). Once that is fixed the extent stops mattering, the two become the same graph, and this moves to NON_COMPILE with no recompile between modes. Left a comment at the call site saying exactly that; happy to do the move in that PR.

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: 4ac3b4c2

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 4ac3b4c2

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 4ac3b4c2

vllm_rbln/envs.py
rebel-jonghewk: 4ac3b4c2

vllm_rbln/envs.py
rebel-jonghewk: 565f1af2

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 6cfc0d5b

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 6cfc0d5b

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 6cfc0d5b

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: 1153cb0d

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: cc60131b

vllm_rbln/v1/worker/kv_placement.py
rebel-jonghewk: cc60131b

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: c8b7dc07

vllm_rbln/patches/dynamic_kv.py
rebel-jonghewk: aeaf3273

vllm_rbln/v1/worker/kv_placement.py
rebel-jonghewk: Fallback path only, and it runs after `load_model()`, so `reserved.current` is populated.

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: 859c44af

vllm_rbln/v1/worker/rbln_worker.py
rebel-jonghewk: 859c44af

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 859c44af

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: cf63ccf8

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 859c44af

docs/dynamic_kv_cache.md
rebel-jonghewk: 859c44af

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-jonghewk: 859c44af

rebel-cjlee (approved):

vllm_rbln/v1/worker/dynamic_kv_sizer.py
```diff
@@ -0,0 +1,890 @@
...
+
+        gmu = self.cache_config.gpu_memory_utilization
+        num_blocks, fits = max_num_blocks(
+            snapshot,
+            growth,
+            gpu_memory_utilization=gmu,
+            kv_resident=kv_resident,
+            reserve_bytes=reserve_bytes,
+        )
+        logger.info(
```
rebel-eunji, line 602:
```
(Worker_DP4_EP4 pid=2183667) [vllm-rbln] INFO 09-15 20:43:42 [dynamic_kv_sizer.py:580] [Dynamic KV] rank 0: computed_num_blocks=405 gpu_memory_utilization=0.920 per unit: 0:0(total=37580963840 budget=34574486732 used=35414605824 kv_resident=62058921984 reserve=0 base=-26644316160 per_block=150994944 blocks=405) 0:1(total=37580963840 budget=34574486732 used=34105982976 kv_resident=62058921984 reserve=0 base=-27952939008 per_block=150994944 blocks=414) 0:2(total=37580963840 budget=34574486732 used=34105982976 kv_resident=62058921984 reserve=0 base=-27952939008 per_block=150994944 blocks=414) 0:3(total=37580963840 budget=34574486732 used=34105982976 kv_resident=62058921984 reserve=0 base=-27952939008 per_block=150994944 blocks=414)

```

base refers negative number. Is it okay?

rebel-jongholee (approved):

vllm_rbln/v1/worker/dynamic_kv_sizer.py
rebel-eunji: I found the reason of negaive value.
### `base` goes negative because `kv_resident` is double-counted

In the dry-run report, chiplet `0:0` logs `base=-26644316160`. `base` is
`used - kv_resident + reserve`, and the subtrahend is what is wrong: `kv_resident`
comes out larger than the chiplet's entire capacity, so the subtraction cannot
land anywhere but below zero.

**Snapshot inputs — correct.** The driver snapshot itself is sound.

| Quantity | Value |
| --- | ---: |
| chiplet total | 35.0 GiB |
| `used` (0:0) | 33.0 GiB |

**Derived values — off by exactly 2x.**

| Quantity | Logged | Should be | Off by |
| --- | ---: | ---: | ---: |
| `per_block` | 144 MiB | 72 MiB | 2x |
| `kv_resident` (= 411 x `per_block`) | 57.8 GiB | 28.9 Gi
| `base` (= `used` - `kv_resident`) | -24.8 GiB | +4.08 GiB | — |

`kv_resident` exceeding the chiplet's total capacity is th
be resident in more memory than the chiplet physically has. The same factor shows
up in the predicted-bytes line, which reports more KV per chiplet than the card
holds in total.

Checked against the placement: one shard is `(2, 64n, 2, 1, 128, 64)` in bfloat16,
i.e. 4 MiB per block, and vllm allocated 18 KV tensors for the 36 layers — so
`per_block` should be `18 x 4 MiB = 72 MiB`.

### Where the 2x comes from

The slope line already reports both sides of the discrepan

```
[Dynamic KV] the slope is summed over 36 KV input(s) in 2
vllm allocated 18 KV cache tensor(s) for 36 layer(s).
```

Programs `0/0` and `0/1` bind the same 18 KV tensors, but one of them appears
under a different arg index and a different dynamo symbol in each graph:

| | `0/0` | `0/1` |
| --- | --- | --- |
| shared | `args_2(s95)`, `args_10(s10)` … `args_25(s91)` — 17 inputs identical in name, symbol, shape and shards | the same 17 |
| unique | `args_9(s4)` | `args_26(s47)` |

`_input_key` folds `tuple(placement.shape)` into the tensor's identity, and that
tuple carries the symbol name (`('symbol', 's4')` vs `('sy
sets therefore do not compare equal, `select_kv_input_groups` returns two groups,
and `_kv_growth_from_programs` sums both — charging the 17 shared tensors twice.
The fingerprint refusal does not fire for the same reason: fingerprints carry the
symbol name too. This is exactly the case the comment above that log line
anticipates.

rebel-eunji: nit: I think these variables should be renamed.
e.g.
- base to `non_kv_used`.  base sounds like starting point. (a little bit ambiguous)
- resident_kv to `reclaimed_kv` (reclaimed after resize)

non_kv_used = used − reclaimed_kv + reserve

rebel-jonghewk: https://github.com/RBLN-SW/vllm-rbln/pull/1104/commits/913568efe42f83da327909a0665b91ee16ff95ab

rebel-eunji (approved):
It works well 👍

uv.lock
```diff
@@ -1633,9 +1633,9 @@ dependencies = [
     { name = "torchvision", marker = "platform_python_implementation == 'CPython'" },
     { name = "transformers" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/84/35/84cfc698aad10e330a3b87bcfeccfc4a9eddef596b21b31263e1837ae006/optimum_rbln-0.11.3a6.tar.gz", hash = "sha256:30e2a4aa821a073c6e52279746abd5c2edbea667de953ddabee54cc5cbbefe12", size = 658264, upload-time = "2026-09-10T04:11:51.918Z" }
+sdist = { url = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/packages/optimum-rbln/0.11.3a6/optimum_rbln-0.11.3a6.tar.gz", hash = "sha256:30e2a4aa821a073c6e52279746abd5c2edbea667de953ddabee54cc5cbbefe12" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/31/10/3c5c8b270f6284742253e5cce59e52a28e46753da61d73da864491392421/optimum_rbln-0.11.3a6-py3-none-any.whl", hash = "sha256:f850841f619dc36b62b324728d775d0f726cff36faad68539e9338e5dd72645d", size = 678795, upload-time = "2026-09-10T04:11:49.669Z" },
+    { url = "https://nexus.mgmt.rbln.in/repository/pypi-group-nightly/packages/optimum-rbln/0.11.3a6/optimum_rbln-0.11.3a6-py3-none-any.whl", hash = "sha256:f850841f619dc36b62b324728d775d0f726cff36faad68539e9338e5dd72645d" },
 ]
```
rebel-eunji: It doesn't matter.

rebel-jonghewk:
FYI . I updated the version for torch-rbln & rebel-compiler.

PR #1108, opened 2026-09-15
fix(executor): close worker response queues before ending the workers

rebel-jindol21:
### 🚀 Summary of Changes

Supersedes #1100, which was cut before #1094 landed and could no longer be merged: `dev` has since squash-merged #1094 *and* added its own `vllm_rbln/patches/multiproc_executor.py` (worker response queue sizing), while #1100 created that same file from scratch. This PR is the same change re-applied on top of `dev`, with both patches coexisting in one module.

`MultiprocExecutor.shutdown` closes the worker response queues only after `_ensure_worker_termination` has walked the grace / SIGTERM / SIGKILL ladder — up to the configured grace period plus the 4 s SIGTERM wait upstream hardcodes. EngineCore's main thread is parked in `MessageQueue.acquire_read` on one of those queues, so for that whole window it cannot reach its fatal handler, does not publish `ENGINE_CORE_DEAD`, and the API server keeps answering 200 on `/health` — `AsyncLLM.check_health` only consults a local flag that nothing has set yet.

This patch closes those queues before that wait. The wait itself is unchanged and still runs to completion. There is no upstream hook between the two steps, which is why this is a `register_patch` rather than an extension point.

**Why it matters.** Under data parallelism all ranks share one `engine_dead` flag and one supervisor, so a dead rank is supposed to take the whole server down. During the window above, the ranks that did *not* fault keep serving successfully, so a single fault yields a mix of successes and failures instead of failing everything. Where a deployment requires all-or-nothing semantics, that mix is a correctness gap rather than a latency one.

**Re-entrancy.** Because the engine now races ahead of that wait, `shutdown` is made re-entrant-blocking: the dying engine's own `EngineCore.shutdown()` waits for the monitor thread to finish the ladder. A worker stuck in a collective only dies on the SIGKILL, and an EngineCore process that exits before then would orphan it with its NPU context still open. The wait is bounded — `VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS` + the 4 s SIGTERM wait + 5 s slack, computed when the wait begins — and logs a warning if it expires.

A second, smaller change moves the engine failure callback ahead of the blocking teardown in `start_worker_monitor`. An idle EngineCore is parked on its input queue, which only that callback feeds, so upstream's ordering leaves it alive until the workers have ended. The two together cover both cases: the queue close wakes an engine that is mid-step, the callback wakes one that is idle.

**Model path:** neither — this is executor/teardown plumbing, shared by both paths.

---

* Supersedes #1100 (same change, unmergeable base)
* Related to rebellions-sw/lmcache-rbln#164 (defect ②: a DP replica that did not fault keeps serving while a peer is already gone)
* Builds on #1094, now merged as `a2088f6c`

> There is no issue for this in this repository; the investigation and the 46-case fault matrix live in the lmcache-rbln issue above.

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit tests, no NPU needed: `uv run --no-sync pytest tests/native/patches/test_multiproc_executor.py` — 7 tests. Six exercise `shutdown` / `start_worker_monitor` and every one of them fails against upstream's functions run on the same fake; the seventh checks the bound arithmetic. A fake executor records the order in which the patched `shutdown` touches the queues and the termination wait: the response queues must close before `_ensure_worker_termination` starts and after the death writers, a re-entrant `shutdown` must block until that call returns and must give up after a bound that tracks the grace period, the failure callback must run before the blocking teardown, and an engine woken by that callback must still wait for the termination.
2. The whole patch suite, so the new patches and `dev`'s queue-sizing patch are exercised together: `uv run --no-sync pytest tests/native/patches/` — 172 tests.
3. End to end, serve any model with `--tensor-parallel-size 2` and a KV connector, then make one rank's model step raise (we used the rebel-compiler vSSD fault injector in `guilty` mode; killing the NPU context of one rank by any other means gives the same shape). Compare `serve.log` against the base commit alone.
4. Expected: on the base alone, `EngineCore encountered a fatal error` appears only after `sending SIGKILL`; with this commit it appears in the same second as `[shutdown] Executor: waiting for worker exit`, while the ladder continues behind it. In both cases the workers are ended and no NPU context is left behind.

---

A/B on TP2, same fault. The "before" column is the base commit (#1094) alone; the "after" column adds this commit.

```
before (upstream ordering)
  21:28:15  RBLN worker ... Ending this worker process
  21:28:15  Worker proc VllmWorker-1 died unexpectedly, shutting down executor
  21:28:15  [shutdown] Executor: waiting for worker exit count=1
  21:28:20  [shutdown] Executor: workers still running after grace period; sending SIGTERM
  21:28:24  [shutdown] Executor: workers still running after SIGTERM; sending SIGKILL
  21:28:24  EngineCore encountered a fatal error          <-- only now

after (this commit)
  21:31:13  RBLN worker ... Ending this worker process
  21:31:13  Worker proc VllmWorker-1 died unexpectedly, shutting down executor
  21:31:13  [shutdown] Executor: waiting for worker exit count=1
  21:31:13  EngineCore encountered a fatal error          <-- at worker death
```

Both arms: workers ended, `rbln-smi` contexts released, `/sys/class/rblnfs/rblnfs*/objects` empty.

On DP2 the two arms are identical, and that is expected: a DP engine owns a single worker which is already gone by the time the monitor fires, so its own termination wait costs nothing. The patch only changes layouts where the dying engine still holds sibling workers — TP, PP and EP.

Wider validation: the 46-case RDS fault matrix in rebellions-sw/lmcache-rbln#164 was run against the pre-rebase commit with #1094 applied. No NPU context was left behind in any of the 46 cases, including every `guilty` case across TP2, PP2, TP2xPP2, DP2, MoE DP2/DP2+EP/TP4+EP and MiniMax PP4. The matrix was then re-run against this rebased commit on the same stack: 46/46 cases ran, residue clean in every one, 38 of 46 verdict-identical, and none of the 8 differences is a regression — see the campaign comment below for the A/B.

---

### 📋 Checklist

---

### 💬 Notes

* **What changed relative to #1100.** The `shutdown` and `start_worker_monitor` bodies are byte-identical to #1100's. The differences are: the module now also holds `dev`'s `patched_init_message_queues` (queue sizing) and a docstring covering both purposes; the test moved to `tests/native/patches/test_multiproc_executor.py` to match the current layout; and the worker-death log line now carries the exit code, matching upstream's wording on vllm 0.26.
* Ordering only — no upstream behaviour is skipped. The death writers still close first, the grace/SIGTERM/SIGKILL ladder is unchanged, and the tail of `shutdown` still shuts the broadcast and response queues as upstream does.
* **Scope: `MultiprocExecutor` only.** `RayExecutorV2` subclasses it but overrides both `shutdown` and `start_worker_monitor`, so neither patch reaches it; the worker-side `fail_fast` in `rbln_worker.py` does apply there. On Ray a faulted worker exits promptly but the engine keeps upstream's teardown timing. Not validated on Ray.
* The re-entrant wait is the part most worth a careful read. It exists solely so the EngineCore process cannot outlive `_ensure_worker_termination`; if `_ensure_worker_termination` ever became non-blocking upstream, this could be dropped.
* No env toggle. The ordering is either correct or it is not, and an escape hatch would carry the old behaviour into deployments indefinitely. Reverting the commit locally is how to reproduce the "before" column above.
* Not validated: PD-disaggregated and multi-node layouts. The teardown path is shared, but neither was exercised.

---


vllm_rbln/patches/multiproc_executor.py
```diff
@@ -11,23 +11,51 @@
...
+from vllm.v1.executor.multiproc_executor import MultiprocExecutor, WorkerProc
 
+from vllm_rbln.logger import init_logger
 from vllm_rbln.patches import register_patch
 
+logger = init_logger(__name__)
+
+# Bounds the engine's wait on the monitor thread: the configurable grace period
+# plus the 4 s SIGTERM wait upstream hardcodes, with slack.
+_TERMINATION_JOIN_TIMEOUT_S = 30.0
```
rebel-cjlee: The comment above says this is the configurable grace period plus the 4s upstream hardcodes, but the value itself is fixed.

With the default `VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS` of 5 you get 9s, so 30 is plenty today. If someone raises it to 26 or more, the re-entrant wait on line 142 expires before the SIGKILL, the engine process leaves first, and we are back in exactly the case this patch exists to prevent.

The comment already knows the formula, so maybe just compute it at call time: `envs.VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS + 4 + slack`.

vllm_rbln/patches/multiproc_executor.py
```diff
@@ -58,3 +86,117 @@ def patched_init_message_queues(
...
+            # EngineCore reach its fatal handler now rather than after it.
+            for mq in list(getattr(self, "response_mqs", None) or []):
+                mq.shutdown()
+            for w in workers:
+                if w.worker_response_mq is not None:
+                    w.worker_response_mq.shutdown()
+                    w.worker_response_mq = None
+
+            try:
+                MultiprocExecutor._ensure_worker_termination([w.proc for w in workers])
```
rebel-cjlee: Upstream calls this as `self._ensure_worker_termination(...)`. Going through the class skips any subclass override. Nothing overrides it today so this is not a bug.

But this whole function is a copy of upstream, and every line that differs for no reason makes it harder to diff against upstream later. Could we keep `self.`?

vllm_rbln/patches/multiproc_executor.py
```diff
@@ -58,3 +86,117 @@ def patched_init_message_queues(
...
+    if response_mqs := getattr(self, "response_mqs", None):
+        for mq in response_mqs:
+            mq.shutdown()
+        self.response_mqs = []
+
+    logger.debug_once("[shutdown] Executor: complete")
+
+
+@register_patch(
+    target="vllm.v1.executor.multiproc_executor.MultiprocExecutor.start_worker_monitor",
```
rebel-cjlee, line 166: Heads up, `RayExecutorV2` overrides both `shutdown` and `start_worker_monitor`, so neither patch reaches it. Meanwhile `rbln_worker.py:175` turns `fail_fast` on for it as well, because it checks `issubclass(..., MultiprocExecutor)`. So on Ray the worker still exists fast but the engine keeps the old timing.

Not asking to widen this PR. Could you add one line to the PR description says this covers `MultiprocExecutor` only, so this Ray gap is on record?

tests/native/patches/test_multiproc_executor.py
```diff
@@ -0,0 +1,217 @@
...
+        lambda sentinels: [ex.workers[1].proc.sentinel],
+    )
+    mpx.patched_start_worker_monitor(ex, inline=True)
+
+    assert ex.is_failed
+    assert "callback" in log
+    assert log.index("callback") < log.index("terminate:start"), log
+
+
+def test_monitor_does_nothing_when_already_shutting_down(monkeypatch):
```
rebel-cjlee: This test passes with upstream's `start_worker_monitor` too, so it does not fail without this patch. Same for `test_death_writers_close_first` on line 106.

Could you make them fail without the change? The other 4 tests do fail against upstream, I checked those.

vllm_rbln/patches/multiproc_executor.py
```diff
@@ -58,3 +86,117 @@ def patched_init_message_queues(
...
+    owner_module="vllm_rbln.patches.multiproc_executor",
+)
+def patched_shutdown(self: MultiprocExecutor) -> None:
+    with _terminated_lock:
+        terminated: threading.Event = (
+            getattr(self, "_rbln_terminated", None) or threading.Event()
+        )
+        self._rbln_terminated = terminated
+
+    if not getattr(self, "shutting_down", False):
```
rebel-cjlee: Non blocking nit:

The lock above covers creating the Event but no this check and the `= True` on line 113, which is what actually decides who the first caller is. Upstream could not hit that, because the monitor called `shutdown()` before the callback. This PR flips that order on line 194, so the engine thread is now running toward this line while the monitor has not reached it yet.

In practice it should never bite. The gap is one `logger.debug`, about 84ns when the level is INFO, and the engine thread needs milliseconds to get here. The one case I;m not sure about is DEBUG logging during a real fault, when every rank is writing at once and that stderr write can block.

Not asking for defensive code, just for line 110 and 113 to move inside the `with _terminated_lock:` block that is already here. Feel free to skip it if you think it is noise.

vllm_rbln/patches/multiproc_executor.py
rebel-jindol21: Agreed — a fixed 30 s silently breaks once `VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS` ≥ 26. Will compute it at call time:

```python
_SIGTERM_WAIT_S = 4.0  # upstream hardcodes this in _ensure_worker_termination
_JOIN_SLACK_S = 5.0    # the ladder polls at 0.1 s and kill() returns at once

def _termination_join_timeout_s() -> float:
    return envs.VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS + _SIGTERM_WAIT_S + _JOIN_SLACK_S
```

`vllm.envs` resolves attributes lazily, so this tracks the env at the moment the engine enters the wait, not at import. The warning will print the computed value. `test_reentrant_wait_is_bounded` currently monkeypatches the constant; it will monkeypatch the function instead, and I'll add one case that raises `VLLM_WORKER_SHUTDOWN_TIMEOUT_SECONDS` and checks the bound moved with it.

vllm_rbln/patches/multiproc_executor.py
rebel-jindol21: Will keep `self._ensure_worker_termination(...)`. The class call was there for the unit test: the fake executor is not a `MultiprocExecutor`, so `self.` had nothing to resolve to, and the fixture patched the staticmethod on `mpx.MultiprocExecutor` instead. That is a test-shaped reason leaking into production code — the wrong way round. The fixture will install the fake on the fake executor (`_Executor._ensure_worker_termination = staticmethod(fake)`), and that line goes back to being byte-identical to upstream.

vllm_rbln/patches/multiproc_executor.py
rebel-jindol21: Confirmed: `RayExecutorV2(MultiprocExecutor)` overrides both `shutdown` (`ray_executor_v2.py:548`) and `start_worker_monitor` (`:479`), and `register_patch` replaces the base-class attributes, so the subclass overrides win. Meanwhile `rbln_worker.py:175` enables `fail_fast` via `issubclass(..., MultiprocExecutor)`, which is true for `RayExecutorV2`. Adding to the description under Notes:

> **Scope: `MultiprocExecutor` only.** `RayExecutorV2` subclasses it but overrides both `shutdown` and `start_worker_monitor`, so neither patch reaches it; the worker-side `fail_fast` in `rbln_worker.py` does apply there. On Ray a faulted worker exits promptly but the engine keeps upstream's teardown timing. Not validated on Ray.

tests/native/patches/test_multiproc_executor.py
rebel-jindol21: You're right, and thanks for checking the other four. Those two (`test_death_writers_close_first`, `test_monitor_does_nothing_when_already_shutting_down`) assert behaviour the patch inherits from upstream, so they can't fail against it. Replacing both with tests that do — each pins an ordering that only this patch produces, at the same spot in the sequence the old test was looking at.

**`test_death_writers_close_first` → `test_worker_queues_close_between_the_death_writers_and_the_wait`.** Not just "death writers first" but where the queue close lands relative to them:

```python
death = [i for i, e in enumerate(log) if e.startswith("mq:death")]
wresp = [i for i, e in enumerate(log) if e.startswith("mq:wresp")]
assert max(death) < min(wresp) < log.index("terminate:start"), log
```

Upstream closes `worker_response_mq` after the wait, so `min(wresp) < terminate:start` fails there. A patch that closed the queues *before* the death writers (tempting — it wakes the engine even earlier, but the death writer is the child's exit signal, so the workers would then sit through the grace period for nothing) fails on the left half.

**`test_monitor_does_nothing_when_already_shutting_down` → `test_engine_woken_by_callback_waits_for_the_termination`.** Drives the path the engine actually takes: the monitor fires, the callback wakes the engine, and the engine's own `shutdown()` — re-entrant, from another thread — must block until the ladder finishes:

```python
def callback():
    log.append("callback")
    Thread(target=engine_teardown).start()   # EngineCore.shutdown() in its finally
ex.failure_callback = callback
Thread(target=mpx.patched_start_worker_monitor, args=(ex,),
       kwargs={"inline": True}, daemon=True).start()
...wait until "terminate:start" in log...
assert log.index("callback") < log.index("terminate:start"), log
assert not engine_returned.wait(0.3), "engine outran the termination wait"
blocker.set()
assert engine_returned.wait(5)
```

Against upstream's monitor the callback has not run when the wait starts, so the first assertion fails immediately (no 5 s hang — the fake termination is released only after the check). Against a patch without the re-entrant wait the second assertion fails. The two existing tests that cover these halves separately (`test_failure_callback_runs_before_shutdown`, `test_reentrant_shutdown_waits_for_the_termination`) stay; this one is the only place the composition is exercised through the monitor rather than by calling `patched_shutdown` directly.

Suite stays at 6, all 6 now red against upstream — I'll confirm that by pointing the fake at `MultiprocExecutor.shutdown` / `.start_worker_monitor` before pushing, the way you did.

vllm_rbln/patches/multiproc_executor.py
rebel-jindol21: Not noise — will move it. You're right that the PR's own reordering opens the gap: the engine thread now heads for `shutdown()` while the monitor thread may still be between creating the Event and `shutting_down = True`. With both inside the lock the first caller is decided atomically:

```python
with _terminated_lock:
    terminated = getattr(self, "_rbln_terminated", None) or threading.Event()
    self._rbln_terminated = terminated
    first_caller = not getattr(self, "shutting_down", False)
    if first_caller:
        self.shutting_down = True
```

The log line and everything after stay outside — the lock is module-global, so the critical section should stay a few attribute ops. For the record, if it did bite today the engine thread would win the first-caller branch and run the ladder itself: upstream's timing, no orphan. But two threads inside `_ensure_worker_termination` at once is not a state I want to have to reason about.

vllm_rbln/patches/multiproc_executor.py
```diff
@@ -11,23 +11,60 @@
...
-"""Size the worker response queue from the responses a pipeline keeps unread.
-
-A KV connector makes `collective_rpc` pass `output_rank=None`, so every worker
-answers each RPC instead of only the output rank. Under pipeline parallelism the
-first stage's answers stay unread until the output rank reports for the same
-step, so `max_concurrent_batches * RPCS_PER_BATCH` of them are outstanding at
-once. Upstream's fixed depth is smaller than that once the pipeline is deep
-enough, and the first stage then blocks in `acquire_write` instead of taking its
-next batch.
+"""Worker response queue fixes for the multiprocess executor.
```
rebel-cjlee, line 14: Could you the module docstring be trimmed?

The "Shutdown ordering" paragraph says the same things as the `shutdown` patch `reason=` just below it, and the SIGKILL and orphaned NPU part of the next paragraph is already in the comment on line 152. So a reader gets it twice.

The "Sizing" paragraph predates this PR, so I would leave that one alone.

rebel-jindol21: Yes — both duplications are real. The "Shutdown ordering" paragraph restates `patched_shutdown`'s `reason=`, and the SIGKILL / orphaned-context sentence restates the comment at the re-entrant wait. Trimming to the one thing neither of those says — why closing the queues first forces the re-entrancy contract:

```python
"""Worker response queue fixes for the multiprocess executor.

Sizing: <unchanged>

Shutdown ordering: closing the response queues ahead of the termination wait
lets the engine race ahead of that wait, so `shutdown` is made
re-entrant-blocking. Each patch below carries its own reason.
"""
```

"Sizing" stays as is.

rebel-cjlee (approved):

PR #1109, opened 2026-09-15
other: use the fsw team's dedicated nodes for vllm-rbln

rebel-wansoo:
The vllm test in vllm-rbln has been running on the SW team's CR13 resources. Now that the FSW team has received its own CR13 resources, this change switches the test to run on FSW-dedicated nodes.


rebel-cjlee (approved):
@rebel-wansoo Is `dedicated: framework.sw` meant to be combined with the step's `agents.queue: udc` and `resources.npu`, or does it replace  the queue selection?

PR #1112, opened 2026-09-15
feature(metrics): report per workload, not per engine

rebel-jswoo:
## Why

vllm-perf CI measures the same engine at two input lengths — `isl 1024` and `isl 51200` — and today that costs two processes, because the metrics cannot be separated within one:

- `_PerformanceContext` accumulates for the life of the runner and has no reset, so a second workload's report would be the first plus the second;
- `print_stats()` is reachable only from the patched `RBLNWorker.shutdown`;
- the JSON filename is `metrics_<rank>.json`, with no workload dimension, so a second write would overwrite the first.

Measured on that lane (build 110, AxK2 dp8): the second workload pays **15.6 min** of engine setup — 11 min of `init engine (profile, create kv cache, warmup model)`, ~3 min of weight reload — to produce **14.7 min** of actual measurement. That is 44 config pairs a night, ~7.5 node-hours.

## What

`flush_pass(worker, label)` reports what has been measured and drops the context; the next workload builds its own through the existing `_ctx()` path and starts from zero. `label` joins the report name and the JSON filename, so one rank leaves one report per workload.

It returns whether anything was recorded. A workload whose engine ran no pass is **unmeasured, not zero** — a caller publishing performance numbers has to be able to tell those apart, and returning `False` is what lets it.

The caller reaches it with `collective_rpc`, so nothing outside this module touches `_metrics_ctx`.

## Compatibility

Unlabelled callers — the `shutdown` path, i.e. every current one — are unaffected. `metrics.json` / `metrics_<rank>.json` keep their names, the report name is unchanged, and the payload only gains `"label": null`.

## Tests

`TestFlushPass` covers the case this exists for: two passes of 8 ms and 2 ms through one worker leave two reports of one call each, and the second reads 2.0 ms rather than the cumulative mean. Plus the three ways it must answer `False` — nothing recorded, no context, metrics off — and that a label separates two reports of one rank.

I could not run the suite locally: this repo's lockfile needs `uv >= 0.11.25` and this machine has 0.9.2, and I did not want to update the toolchain underneath someone. The changed logic is exercised in isolation and the real run is on CI here.


rebel-cjlee (commented):
Not blocking, this is a debug feature. Just a few questions.

1. Nothing in the repo calls `flush_pass` yet. Which script or follow up PR will call it? Seeing the caller would make it easier to judge the shape.
2. `_slug` folds different labels into one filename. `isl 1024` and `isl/1024` both become `isl_1024`, so the second report quietly overwrites the first. Worth a warning when the file is already there or is it intended?
3. In this module `pass` means one `execute_model` call, but `flush_pass` flushes a report built form many of them. Would `flush_report` read better? Just a nit.

PR #1115, opened 2026-09-16
perf(specdec): optimize eagle3 under PP

rebel-minkyu:
### 🚀 Summary of Changes

This PR optimizes eagle3 path under PP.

#### 1. One handoff tensor for the EAGLE3 aux hidden states.
Under pipeline parallelism, the tensors for aux hidden states are transferred each, resulting in a linear increase in send/recv latency depending on the number of tensors being transferred. The aux tensors now travel concatenated on the feature dim as a single `aux_hidden_states` entry: one transfer whatever the layer count. That concatenated layout is the one the drafter's `fc` already consumes, so nothing downstream changed.

#### 2. No drafting on an intermediate chunk.
An intermediate chunk's sampled token is a placeholder and  the real token arrives only with the last chunk, so every draft built on it is discarded. Previously, the proposer nevertheless ran its full `num_speculative_tokens` loop on each intermediate chunk. It now returns zeros instead. The early return sits *after* the first `model_executable` call. That pass is what walks the drafter's KV cache across the chunk, so only the decoding passes are skipped.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


vllm_rbln/v1/spec_decode/eagle.py
```diff
@@ -175,6 +175,11 @@ def propose(
             draft_tokens_ids = self._to_target_token_ids(draft_ids[:num_reqs])
             return draft_tokens_ids.view(-1, 1)
 
+        # Early exit if it's an intermediate chunked prefill,
+        # since the draft tokens are discarded anyway.
+        if self.runner.is_intermediate_chunked_prefill:
+            return draft_ids.new_zeros((num_reqs, self.num_speculative_tokens))
```
rebel-cjlee: Isn't there any problem in case of DP enabled? `is_intermediate_chunked_prefill` is the state only own rank.

rebel-minkyu: We can skip the rest for intermediate chunks also under DP case, so I guess it's fine. But let me just quickly verify it under DP.

tests/native/patches/test_minimax_m2.py
```diff
@@ -147,21 +156,21 @@ def _run_pipeline(pp_size: int, aux_layers: tuple[int, ...], monkeypatch):
             intermediate_tensors=carried,
         )
         if rank < pp_size - 1:
-            slots_sent.append(_slot_indices(out.tensors))
+            layers_sent.append(_stamped_layers(out.tensors.get(eagle3_pp.AUX_COMBINED)))
+            keys_sent.append(sorted(out.tensors))
```
rebel-cjlee: `sorted` hides the order, and the order is what  this change has to get right. `list(ouit.tensors)` keeps the insertion order, and then line 203 reads `["hidden_states", "residual", eagle_pp.AUX_COMBINED]`, which is the thing we actually rely on.

vllm_rbln/v1/spec_decode/eagle.py
rebel-minkyu: Thank you for the catch. It indeed had a problem under DP when there's a peer in a decoding phase. I fixed it to block the early-exit under DP when a drafter has MoE.

tests/native/patches/test_minimax_m2.py
rebel-minkyu: Fixed.

rebel-cjlee (approved):

PR #53, opened 2025-08-13
fix(core): bug fix for block table logic

rebel-jonghewk:
### 🚀 Summary of Changes

**Fix block-table range & simplify padding**
- Ensure block tables use IDs in 0..<estimated_block_size (regression from PR #49).
- Simplify padding:
  - Pad row size with -1.
  - Pad per-block entries with the first unused block ID (works whether a dummy block 0 exists or not).

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [x] 🛠 Bug fix (`bug-fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [x] PR title follows Conventional Commits format
* [x] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #1119, opened 2026-09-16
feature(nixl): support stripe width in connector configuration

### LINKED ISSUE #1118: feat(nixl): expose stripe width through connector configuration
Allow RBLN NIXL users to select single-rail transfers with `kv_connector_extra_config={"stripe_width": 1}`. Forward the option on both host-staging and direct-device registration paths, preserving the plugin default when omitted.

Depends on the adapter API in https://github.com/rebellions-sw/nixl-rbln/pull/53.

rebel-yskim:
### Summary of Changes

Allow `kv_connector_extra_config` to select the RBLN plugin's stripe behavior. Forward `stripe_width` during host-staging and direct-device backend registration:

- `kv_connector_extra_config={"stripe_width": 0}` enables multi-rail striping. A transfer is striped only when the pool has at least two rails, the transfer has at least two descriptors, and the payload is at or above `stripe_threshold_bytes`. Otherwise, each transfer uses one rail.
- `kv_connector_extra_config={"stripe_width": 1}` selects single-rail transfers.
- Omitting `stripe_width` preserves the installed plugin's default. The plugin pinned by nixl-rbln PR 53 (`0.2.0.dev102+g91b8951`) defaults to `0`, so explicitly selecting `1` changes its behavior. That plugin accepts only `0` and `1`; other values fail backend initialization. The value is a mode selector, not a requested rail count.

These semantics match the [updated wrapper docstring in nixl-rbln #53](https://github.com/rebellions-sw/nixl-rbln/pull/53/changes/874530b7004646b0d8f5d78ea63c2ba0212e2b25). Settings apply when the backend is first created; an existing backend is not reconfigured.\n\nPrefill and decode may use different `stripe_width` values. Each transfer follows the setting of the side that posts it: decode on the pull path and prefill on the push path. The setting therefore does not participate in `rbln_compat_hash`.

Requires an installed `nixl-rbln` package containing [#53](https://github.com/rebellions-sw/nixl-rbln/pull/53), which adds the `stripe_width` arguments used by both registration paths.

Affected module: shared RBLN NIXL connector, used by both the optimum-rbln and vLLM-native model paths. Changes are limited to the worker and its existing tests.

### Related Issues / Tickets

Closes #1118.

Depends on https://github.com/rebellions-sw/nixl-rbln/pull/53.

### Type of Change

- [x] Feature

### How to Test

```bash
uv run --no-sync python -m pytest tests/native/distributed/kv_connector/test_rbln_nixl_base_worker.py -k 'host_bounce_creates_backend_and_delegates or registers_with_vram_segment' -q
```

Expected: an omitted setting is not passed to `nixl-rbln`, while explicit `0` and `1` values are forwarded on both registration paths.

Validation: the original implementation tests passed in an existing local executor environment. For the compatibility follow-up, pre-commit checks passed for both changed files; the focused pytest command could not be collected in the current worktree because `vllm` is not installed. Actual RDMA transfers were not run.

### Checklist

- [x] Linked to an existing issue
- [x] Test method and expected result described


rebel-junamsong (approved):
LGTM

rebel-jonghewk (approved):

rebel-ykchoi:
The body's only example is the default, so it selects nothing.

The plugin advertises `stripe_width` with `"1"` as its default
(`rbln_plugin.cpp:40`; `rbln_backend.h:540` declares `uint32_t stripe_width_ = 1`),
and the striped path is gated on `stripe_width == 0` (`rbln_stripe.h:41`).
So `1` is single rail *and* the default, and `0` — every rail — is the only
value that changes behaviour.

The body offers `{"stripe_width": 1}` "to select single-rail transfers", then
says omission preserves the backend default. Those are the same outcome. An
operator who follows the example sets the value it already had, sees nothing
change, and concludes the knob does nothing here — while `0`, the setting they
wanted, appears nowhere in the description or the issue.

Worth stating alongside it: `0` only stripes when the pool has at least two
rails, the transfer has more than one descriptor, and the payload is at or above
`stripe_threshold_bytes` (`rbln_stripe.h:38-43`). Without that, `0` reads as
broken whenever a transfer falls under the threshold.

The code passes all three values through correctly. This is the description,
not the implementation.

rebel-yskim:
@rebel-ykchoi I've edited PR description. Could you please check and review?
fyi, nixl-rbln package would be generated in today's nightly CI.

rebel-ykchoi:
Correcting myself: the default is `0`, not `1` — I read a stale checkout.

`stripe_width` flipped in the plugin at `421cc76` ("feat: stripe_width default
1 -> 0", 2026-09-14). I was reading `d733e4a` from 09-10, which still had `1`.
Your revision, `91b8951`, is after the flip, so the body is right and my comment
was not.

The rewritten description is clearer than what I asked for — it states all three
values, names the plugin revision, and spells out the conditions under which `0`
actually stripes.

rebel-ykchoi:
A question about the name, not the change.

`stripe_width` takes exactly `0` or `1`, and the description has to spell out that
the value is a mode selector rather than a rail count. A name that read as a mode
would not need that sentence.

`0` is the part that catches me. In the plugin's own params table
`keepalive_interval_ms=0` disables and `listen_port=0` means "kernel picks", but
`stripe_width=0` is the maximum — every rail. Someone who has just read those rows
guesses "off" and gets the opposite. `enable_lane_rate_weighting` sits right next to
it as the same kind of two-state switch, named `enable_*` with `on`/`off`.

This PR is the first place a user types the key, so it seemed worth asking before it
becomes a deployment surface: is the name settled on the plugin side, or worth
raising there? Either way I would not change it here — forwarding the plugin's own
key is right, and a second vocabulary would be worse.

rebel-yskim:
> is the name settled on the plugin side, or worth raising there? 

I'm not sure - you may ask to the RBLN SSW team. I thought that `stripe_width` is still meaningful term (at least, for me), so if you want to raise the discussion, it would be better to ask for `nixl-rbln-plugin` team.

rebel-jongwonlee (commented):
Two things, and one answer worth recording.

## Omitting the option does not leave the plugin requirement where it was

The kwarg goes out unconditionally — `stripe_width=None` when unset — so on a
nixl-rbln predating #53 every deployment fails at `register_kv_caches` with
`TypeError: ensure_rbln_backend() got an unexpected keyword argument
'stripe_width'`, including deployments that never asked for striping. Nothing
in the repo catches it first: `pyproject.toml` has no nixl-rbln pin and the
import is optional, so there is no version floor to fail against. On the D2D
path it lands after warm-up, since registration is deferred there.

"Omitting `stripe_width` preserves the installed plugin's default" is true of
the value and not of the call. Passing it only when set makes the two agree:

```python
extra = {} if self._stripe_width is None else {"stripe_width": self._stripe_width}
nixl_rbln.ensure_rbln_backend(self.nixl_wrapper, device_id=0, **extra)
```

The nightly plugin build covers the rollout; what this covers is a rollback, or
a fleet that is briefly mixed.

## `false` selects striping

JSON booleans arrive as Python `bool`, and `bool` is a subclass of `int`:

```
{"stripe_width": false}  ->  False,  and False == 0, False in (0, 1) -> True
```

So `false` passes whatever the plugin checks and selects `0` — striping on, the
opposite of what was typed. `2` and `"0"` do fail, but at backend init, per
rank, with the plugin's message rather than one naming the knob.

`ec_cfg.get_from_extra_config(key, default)` is how the EC connector in this
repo reads its fourteen knobs, and a default of `0` there gives the read a type
to check against.

## Worth one sentence in the description

P and D do not have to agree on this. From the plugin author in the multi-rail
thread: a transfer follows the setting of whichever side posts it — decode's on
the pull path, prefill's on the push path — so a mixed pair is valid rather than
a misconfiguration.

That is the same shape as `chunk_mode`, and the reason neither belongs in
`rbln_compat_hash`. It is also the first thing an operator will ask when rolling
this out to one side of a P/D pair, and the answer currently lives only in
Slack.

rebel-cjlee (approved):

rebel-yskim:
@rebel-jongwonlee I've applied your comment: [b14f628](https://github.com/RBLN-SW/vllm-rbln/pull/1119/commits/b14f6281732cb3173e836c615cf98cfab0572db2), and edited PR description. please check.

I did not add a second validation layer for JSON booleans. With the current nixl-rbln implementation, False is serialized as "False", and the plugin parser rejects values other than "0" and "1"; it does not select stripe width 0.

Also, get_from_extra_config() in the pinned vLLM version is a plain dict.get() and does not provide type validation. If stricter Python-side validation is desirable for every caller, I think it should be added to nixl-rbln first so that all integrations follow one consistent contract.



I would like to keep this PR as a thin integration of the API introduced by nixl-rbln PR no.53. That change exposes the nixl-rbln-plugin configuration, while this PR only makes the value configurable from vllm-rbln.

Naming, validation policy, and backend-lifecycle API design are owned by nixl-rbln and nixl-rbln-plugin. If that contract changes there, vllm-rbln can consume the updated contract rather than defining a separate policy at this layer.

rebel-yskim:
Since this is a time-sensitive fix, I initially limited the scope to the minimum integration needed between vllm-rbln and nixl-rbln. I have incorporated the concrete compatibility and documentation feedback from this review, while leaving broader API design decisions to the repositories that own those interfaces.

rebel-jongwonlee (approved):
The fix is in and my second point was wrong.

## `b14f628`

Both paths take `**extra`, and the test pins more than I asked for — not just
the value but the key's absence:

    assert ("stripe_width" in called) == (stripe_width is not None)

That is the assertion that would have caught the original shape, so it is the
right one to leave behind.

## The bool concern does not exist — I misread the wrapper

You are right, and the reason is one line in nixl-rbln #53:

    params["stripe_width"] = str(stripe_width)

The value is stringified before the plugin ever compares it, so `bool`'s
int-ness never comes into play:

    False → 'False' → rejected      0   → '0' → accepted
    True  → 'True'  → rejected      '0' → '0' → accepted
                                    2   → '2' → rejected

So `false` is refused rather than read as `0`, and `"0"` — which I also said
would fail — is accepted and correct. Both halves of that point were wrong; I
argued from `isinstance(False, int)` without reading how the value reaches the
plugin.

The scope argument holds too. With #53's docstring now stating the contract
("The bundled plugin rejects values other than 0 and 1 during initialization"),
a second check here would be a second place for that contract to live, and the
two could disagree.

## One line in the description

"Requires an installed `nixl-rbln` package containing #53" is now stronger than
the code, in the opposite direction from before: since `b14f628`, omitting the
option sends no kwarg, so an older wrapper is fine. "Required to use the option"
would match.

Approving.

rebel-ykchoi (approved):

PR #1099, opened 2026-09-13
refactor: move env vars to OptimumRBLNConfig

rebel-eunji:
### 🚀 Summary of Changes

1. Move env vars to OptimumRBLNConfig
2. Change rbln_config in additional_config to optimum_overrides

## Notes
1. Refactor the fields - `optimum-rbln path, written by the platform hook and the config sync`

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
5. Verify output: `...`
6. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #1122, opened 2026-09-16
feature(config): move RBLN options from the environment to `--rbln-*` flags

### LINKED ISSUE #1047: proposal(config): move the `VLLM_RBLN_*` options that change model behavior into `RBLNConfig`

rebel-cjlee:
### 🚀 Summary of Changes

RBLN options now come from `--rbln-*` flags instead of environment variables. The model path is not one of those options: upstream already names it, so it is set with `--model-impl auto|transformers|optimum`.

**61 files, but most of the change is mechanical.** Only three parts need a close read.

| Kind | Size | What |
| -- | -- | -- |
| Mechanical | Most of it | Messages, comments, docs, and CI targets that named a variable now name a flag. Tests set options through the config. |
| Core | Three parts | See below. |

#### Core 1. When patches are applied

- **Problem.** `register_ops()` runs before vLLM parses the arguments, so it cannot know the model path. Patches belong to the vLLM native path only.
- **Change.** Patches are now applied from two places. The process that resolves the path applies them itself, in the post parse window of `pre_register_and_update()`, which is the first statement of `create_engine_config`. Every other process inherits the resolved path and applies them from `register_ops()`. Both places apply the whole registry, patches and registrations alike. A smaller set would leave a process half patched when one process plays several roles, such as with `VLLM_ENABLE_V1_MULTIPROCESSING=0`. The RBLN KV connectors moved into the registry for that reason: they were registered by an import inside `register_ops()`, which only a process that inherited the path reaches, so a run started with the flag had none of them in `KVConnectorFactory`.
- **What makes the split hold.** `register_ops()` reads the resolved path from the environment, and only a caller that actually resolved it publishes it there. Adopting the path at import no longer does, since  the answer available then is a guess while the deprecated variable still selects the path. So in the process that parses the arguments the variable is unset, which is what "not yet" means, and that process waits for the platform hook.
- **Groundwork.** `maybe_transfer_kv_layer` reads `get_attention_context` when it decorates, and keeps it in the closure. Building the wrap at import time would freeze upstream's version there. The registry gained `build=`, which builds the wrap at apply time instead. With that, importing `patches/` no longer changes upstream, which is what keeps the optimum path's upstream clean.

Files to read: `vllm_rbln/patches/registry.py`, `vllm_rbln/patches/attention.py`, `vllm_rbln/patches/kv_connector_factory.py`, `vllm_rbln/platform/vllm_impl.py`

#### Core 2. `--model-impl`

- **Problem.** The path comes from the config, but it is needed before the config exists. `DeviceConfig` reads `device_type` on the line right after `pre_register_and_update()`.
- **Change.** Two bridges.
  - Inside one process, a wrapper on `create_engine_config` reads `--model-impl` as soon as the arguments are parsed, points the platform at the path it names, and hands upstream's field back its default. Nothing is written into `additional_config`: a resolved one already names its path by being one of the two classes, which is what the workers it is pickled to read.
  - Between processes, `_VLLM_RBLN_RESOLVED_MODEL_IMPL` carries it. This is a plain constant, not an entry in `envs.py`, and each process overwrites it for the processes it spawns. Ray copies only the names `get_env_vars_to_copy` selects, and the leading underscore misses its prefixes, so `RblnPlatform.additional_env_vars` lists it. Ray is not a supported executor here yet; this closes the silent way to get the path wrong before it is.
  - `register_ops()` installs the wrapper. `LLM(...)` never calls `add_cli_args`, so installing it from `pre_register_and_update` would miss that call.
  - The same wrapper records `--max-num-batched-tokens` as the user gave it, which the optimum path compiles as the prefill chunk size. It runs before upstream fills in a default for an unset one, so the raw value is still there, and it is the only place that knows the path this early. Upstream's bare string `additional_config` is left as it is: this path cannot be written into it, and `VllmConfig` types the field and refuses it in words.
  - **Three names, one per reader.** `_VLLM_RBLN_RESOLVED_MODEL_IMPL` is what a process hands the once it spawns, `envs.INHERITED_MODEL_IMPL` is that variable read once at import (what the parent resolved, and what a config naming no path falls back to), and `platform._MODEL_IMPL` is what this process runs, which `_impl()` dispatches on. One name for all three made a process answer its own question: a second `LLM(...)` with no path of its own took the path of the engine before it.
  - **What the flag means here.** `vllm` is the vllm model path, `transformers` and `optimum` are the optimum one, and anything else is refused. `auto` is no answer at all, since every `EngineArgs` carries it whether the user typed it or not: left to the ladder it takes the path this process was handed, and then the default, which is `optimum`.
  - **What upstream is told.** Upstream reads the same field, and there `transformers` means its own Transformers backend, which `is_backend_compatible()` refuses for several of the models optimum-rbln supports; `ModelConfig` raises before the platform hook sees it. So the wrapper consumes the value and hands the field back `auto`, and upstream resolves what it would have without us. A built `additional_config` that disagrees with the flag is refused rather than resolved: the class would drop what the caller typed, the flag would hand the resolution a config it cannot read.

Files to read: `_capture_model_impl` and `_apply_model_impl` in `vllm_rbln/platform/__init__.py`, `resolve_model_impl` in `vllm_rbln/config.py`

#### Core 3. The optimum path's config

`OptimumRBLNConfig` follows the field definitions on `refactor/rbln-config-optimum-path` (#1099), including the rename the maintainer made there (`rbln_config` became `optimum_overrides`). `RBLNConfigBase` holds what both paths share, and each path resolves its own class. The `--rbln-*` flags have to be registered before parsing, so both classes' fields are registered together, and the class a config resolves into is what narrows them.

Each class holds one path's options, so the class a caller builds is the choice: `LLM(additional_config=RBLNConfig(...))` reaches the vllm model path with nothing else said. `_CONFIG_CLASSES` lists the optimum class first, because a field both classes declare takes its `--help` default from the first of them and an unset flag leaves the run on the optimum path.

File to read: `vllm_rbln/config.py`

#### The rest

Eighteen environment variables are deprecated. They still work and only print a warning, with one exception: `VLLM_RBLN_USE_VLLM_MODEL` names the model path, and the environment's answer has already been acted on by the time the flag is read, so a value that disagrees with `--rbln-model-impl` is refused rather than quietly overruled. They go away in the future release **(0.12.0 mentioned in the comment)**. The warning comes from one place, `_env_overrides()`, and it builds the flag name from the field name, so a renamed field is named correctly too.

`RBLNConfiag`'s field set changed here, so **the compile cache key changes once.** The path is not one of those fields: each path hashes a class of its own, and that is what keeps their bundle keys apart.

---

* Resolves #1047
* Related to #1110

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

```
uvx pre-commit run --all-files
uv run --no-sync pytest tests/vllm -q
uv run --no-sync pytest tests/torch_compile -q
uv run --no-sync pytest tests/optimum -q --ignore=tests/optimum/v1/sample \
    --ignore=tests/optimum/lora --ignore=tests/optimum/optimum_compile

# The flag as a real command line argument
vllm serve JackFram/llama-68m --rbln-model-impl vllm --enforce-eager \
    --max-model-len 256 --max-num-batched-tokens 128 --block-size 128
```

Notes:

- `VLLM_RBLN_USE_VLLM_MODEL=1` with `--model-impl optimum` fails naming both, instead of running the optimum path on patched upstream.
- `LLM(additional_config=RBLNConfig(compile_model=False))` reaches the native path with no `model_impl` written by hand.
- `--additional-config '"something"'` is refused by `VllmConfig`, not by a `TypeError` from an operator.

---

---

### 📋 Checklist

---

### 💬 Notes

- `user_max_num_batched_tokens` given in `additional_config` is overwritten by the capture above. That is also the behavior before this PR.

---


rebel-eunji:
Could you consider Ray support?
_RBLN Ray executor environment variable propagation issue_

**Problem**

- Ray actors are launched by the raylet, not the frontend, so changes to `os.environ` are not inherited.
- vLLM forwards environment variables into `runtime_env` only when `get_env_vars_to_copy()` selects them. It filters on three criteria:
  1. Keys present in `vllm.envs.environment_variables`
  2. Names starting with prefixes like `VLLM_`, `NCCL_`, or `HF_`
  3. `Platform.additional_env_vars`
- `_VLLM_RBLN_RESOLVED_MODEL_IMPL` matches none of these. It starts with an underscore, so it fails the prefix check; it is omitted from `environment_variables`; and `RblnPlatform` does not define `additional_env_vars`.
- As a result, the Ray worker receives no patched import path, stays on optimum (or whatever the deprecated variable points to), and runs `RBLNWorker` against an unpatched upstream.

**Fix**

- Add one line: `RblnPlatform.additional_env_vars = [envs.RESOLVED_MODEL_IMPL_ENV]`
- This mirrors how TPU inference handles `MODEL_IMPL_TYPE`.

AGENTS.md
```diff
@@ -79,7 +81,8 @@ The registry rules:
 - Never `setattr` an upstream symbol directly. Every replacement goes through the registry, which verifies that it took.
```
rebel-eunji: We need to fix this rule.

rebel-eunji (approved):
lgtm
Could you ping me after moving --rbln-model-impl to --model-impl? (I may not be able to check it right away, as I'm on vacation..)

AGENTS.md
rebel-cjlee: I added an exception

rebel-cjlee:
@rebel-eunji I applied `--model-impl` instead of `--rbln-model-impl`.

.buildkite/pipelines/vllm-rbln/scripts/perf/targets/minimax-m2.7.yaml
```diff
@@ -3,10 +3,10 @@ model: MiniMaxAI/MiniMax-M2.7
 
 env:
   VLLM_DISABLE_COMPILE_CACHE: "1"
-  VLLM_RBLN_USE_VLLM_MODEL: "1"
-  VLLM_RBLN_BATCH_ATTN_OPT: "1"
 
 serve:
+  rbln_model_impl: vllm
```
rebel-jonghewk: Does this also has to be changed?

rebel-cjlee: Right, I missed it. Thanks.

rebel-jonghewk (approved):

rebel-huijong (approved):
LGTM. Once this is merged, could you share an overall walkthrough of how we should use it with vllm serve or an LLM instance?

PR #1124, opened 2026-09-16
fix(test): fix num_gpu_blocks_override in pytest

rebel-eunji:
## Problem
There's a hang when we run gemma3 in pytest. It was triggered by lack of blocks.

## Solution
gemma3 model includes SWA, so the required number of blocks should be changed.

## Summary
New SWA kernel has different logic of block allocation.
VLLM_RBLN_USE_DYNAMIC_KV_CACHE is not turned on in pytest.

## Detail

- **Sizes the pool from the whole minimum, not one request.** The helper returned only
  the one-request term, so a spec raising `max_num_seqs` silently under-sized the pool:
  at `max_num_seqs=4` it still returned 11 where the engine needs 13, and the engine
  then fails to start rather than skipping. It now mirrors `minimum_kv_blocks` --
  `1 + max(one_request, max_num_seqs * per_seq)` -- and is renamed `kv_blocks_needed`
  to match what it returns.

- **Resolves the chip with `host_chip()` rather than `current_platform.is_cr13()`.**
  `is_cr13()` raises on a host with no NPU and no `RBLN_FORCE_NPU_NAME`, and the one
  caller (`test_dp_specs.py`) carries no mark, so it runs in the default lane on every
  PR. `tests/native/utils.py` already has `host_chip()` for exactly this, and it
  returns `None` there instead.

- **Derives the in-flight term from `pipeline_parallel_size`.** The window term scales
  with `max_concurrent_batches`, and that is not always 2:

  | async scheduling | pp | `max_concurrent_batches` |
  | --- | --- | --- |
  | on | 1 | 2 |
  | on | 4 | 4 |
  | off | 1 | 1 |
  | off | 4 | 4 |

  A default native config is the first row, so the hardcoded 2 was right there. It is
  half the real value for the `pipeline_parallel_size=4` spec in
  `tests/native/compile/models.py`, which does not pin `num_gpu_blocks_override`.
  `max(pipeline_parallel_size, 2)` covers every row.


rebel-cjlee (approved):

rebel-eunji:
@rebel-cjlee 
I made the block allocation logic more robust.

PR #1114, opened 2026-09-15
refactor(kv-cache): put num_blocks ahead of the K/V axis

rebel-jonghewk:
## Summary

Change the attention KV-cache layout from
`[kv(2), num_blocks, ...]` to `[num_blocks, kv(2), ...]` for dynamic KV cache support. This keeps each block's K and V data contiguous without changing the allocation size.

This PR also updates:
- KV-cache indexing in the five Triton attention kernels
- Dynamic-axis marking and host/device block-copy paths
- Tests and documentation for the new layout

MLA and DeepSeek indexer caches are already blocks-first and remain unchanged.

## Important

This PR depends on a matching compiler-side update to `rbln_custom_ops`, which still expects the old layout. End-to-end serving will not work until both changes land, so this PR remains a draft.

NIXL attention-cache registration is rejected for now because its descriptors still assume separate K/V regions. MLA and Mamba continue to use the upstream path.

## Review notes

Two pre-existing issues are fixed alongside the layout change:
- Correct contiguous strides in `triton_attention_naive` and `triton_causal_attention_naive`
- Initialize `_is_kv_layout_blocks_first` in `RblnTransferTopology`

Please pay particular attention to the stride and offset changes in the Triton kernels.

## Validation

Passed:

```bash
uvx pre-commit run --files $(git diff --name-only origin/dev)
uv run --no-sync pytest \
  tests/native/distributed/kv_connector/test_rbln_transfer_topology.py \
  tests/native/patches/test_kv_connector_utils.py \
  tests/native/v1/attention/test_kv_cache_bindings.py \
  tests/native/v1/worker/test_utils.py \
  "tests/native/v1/attention/test_flash_attention.py::TestFlashAttentionBackendStatic" \
  "tests/native/v1/worker/test_rbln_model_runner.py::TestProcessKvCacheCopyOps"
```

Not covered:
- Triton kernel execution: no NPU was available
- `test_rbln_model_runner_kv_cache.py`: gated model access was unavailable
- End-to-end serving: blocked by the compiler-side layout update

Model path: vLLM-native only; optimum-rbln is unchanged.

PR #1125, opened 2026-09-16
feature(nixl): transfer a KV cache that packs K and V into one block

rebel-ykchoi:
### 🚀 Summary of Changes

RBLN's attention KV cache has two layouts, chosen per process by the kernel
namespace that reads it: `rbln_custom_ops` puts K and V inside one block,
`rbln_triton_ops` keeps them in separate spans. NIXL registration accepted only
the second, so disaggregated serving on a GQA model needed the flag that selects
it -- the packed layout is what an untouched configuration gets.

This registers the packed cache as one region per layer and teaches the
descriptor path to name K and V separately wherever a side reads *part* of a
block. A whole block is one contiguous range under either layout, so the break
appears only where a side reads part of one. Two of those are widened here:

- **a head band**, when the two ends cut heads differently (unequal TP). The
  band sits once in K and once in V, a half-block apart.
- **a sliding window**, which is a byte prefix of a block. One prefix over a
  packed block runs past K's end and never reaches V.

One integer carries this: how many ranges one piece of a block breaks into.
It is 1 throughout on the split layout, where every new expression reduces to
the shipped one. MLA is key-only, so it stays at 1 there too and nothing in
this PR changes what an MLA model does.

**A third way to read part of a block is refused rather than widened.** A peer
whose block size differs reaches upstream's arithmetic, which cuts a local block
into `block_size_ratio` equal pieces and pairs each with one peer block -- an
assumption that a block is a run of token ranges. On a packed block that cut
falls on the K/V boundary instead: at ratio 2 the pieces are the whole of K and
the whole of V, and the descriptors carry no K/V axis to say which is which, so
a peer block's keys and values both land in our keys. The equal-TP path hands
its descriptors to upstream unchanged and nothing covered the combination -- the
head-matched refusal runs only when the TP degrees differ, the PP assert only
above one stage, and upstream's own `assert block_size_ratio == 1` only for a
hybrid layout. The fourth commit refuses it where that path begins, next to the
existing refusal for a peer on the other layout. It also pins what
`2 // len(cache_list)` encodes, since that value is model-wide while the
expression is evaluated per layer: an attention layer registers K and V as one
region or as two, and the layers have to agree.

**The second commit fixes a defect the first one runs into.** Deferred KV
registration probes the attention backend for the block axis, and that read
resolves the RBLN config through the global vllm config -- which warm-up is
already past. The rank dies on `Current vLLM config is not set`. It predates
this branch: the probe arrived with the layout switch, and the block-count RPCs
were given a scope of their own at the time while this path was not. It rides
here rather than on a branch of its own because nothing in this PR reaches a
server without it.

**The third commit is a separate subject carried here, and nothing in this PR
depends on it: a PP worker gets past upstream's HMA refusal.** MLA is key-only,
so `kv_per_block` stays 1 and the packed path is inert for the model this
commit unblocks; the two are independent and travel together only because they
were ready together. Upstream refuses `pipeline_parallel_size > 1` whenever
the KV layout reads hybrid, because its PP region slicing derives one region
count per layer and takes "not a `FullAttentionSpec`" as the proxy for "not
uniform". A single group merging same-type full attention specs fails that proxy
while still slicing uniformly: an MLA model with a sparse indexer arrives that
way, because the indexer's cache differs from the attention layers' in dtype, so
the specs merge into one `UniformTypeKVCacheSpecs` even though every layer still
contributes the same number of regions. The flag upstream reads is suppressed
across its `__init__` for that shape alone and put back, with `_is_hma_required`
assigned the value upstream would have reached so its later block-size and
permute guards still see it. The shape test unwraps the merged group with
upstream's own `get_representative_spec_type` rather than trusting the wrapper:
Mamba and sliding-window layers merge into the same wrapper and are not uniform,
and every later net -- upstream's `_has_mamba`, our `_has_swa` and
`_check_pp_constraints` -- reads the group spec instead of what it wraps, so a
wrapper-only test would hand those layouts to the region slicing with nothing
left to refuse them.

- `_head_matched_desc` -- the offset that was wrong. It multiplied the head
  index by the whole block length; on a packed block that is twice the head
  stride, so a request for head *n* of K landed on a head of V. Every
  divisibility and bounds check passed.
- `register_local_xfer_handler`'s window branch and `_compute_desc_ids` --
  the two lists are paired element-wise, so the id space had to widen with them.
  Both fall back to upstream when no window is in play, which is the path the
  block-size refusal now guards.
- `_check_d2d_region_pairing` -- two peers off one build can hold different
  layouts, so the layout travels in the handshake and a mismatch is refused.
  The connector metadata version moves with it.

**Tests.** All of it runs in-process against fakes; no assertion here moves a
byte between two engines. The descriptor cases decode each address back into
`(K-or-V, head)` and compare it against the head the pairing asked for, rather
than asserting the address -- an address assertion passes whatever the address
means, which is how the offset above stayed green.

The test diff is larger than the source diff. It is spread over registration,
the handshake, both descriptor lists and the id arithmetic, because a mutation
planted in any one of them survived the suite before these were added.

* Resolves rebellions-sw/icp-serving#597

### ✅ Type of Change

* [x] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

### 🧪 How to Test

1. `pytest tests/native/distributed/kv_connector/`
2. `pytest tests/native/patches/ tests/native/v1/worker/`
3. Edge cases pinned by the new cases: a peer on the other layout (refused by
   name), a peer whose block size differs (refused), a peer block length that
   does not halve, a sliding window on a packed block, a peer that cuts heads
   more coarsely than we do -- which leaves the head split at 1 while the block
   still needs two ranges -- and, for the HMA suppression, a merged Mamba or
   sliding-window group, which must reach upstream with the flag untouched.

### 💬 Notes

- The version ledger comment above `RBLN_NIXL_CONNECTOR_VERSION` is over the
  five-line comment budget. It was already over before this change; the line
  added here is one ledger entry, and trimming it would mean deleting earlier
  entries.
- The topology's `__post_init__` reimplements upstream's, and the test that
  holds the two in step now asserts containment rather than equality: the
  layout has to be decided there and kept, because it comes from the RBLN
  config, which is only guaranteed to be set while the model is built, while
  the regions are cut later.
- **A limit of the HMA tests.** They stub `super().__init__()` and leave
  `_is_hma_required` unset so that only the override can produce it, which is
  what makes them test the override -- and also means they cannot catch
  upstream changing the `any(...)` condition the override reproduces. The
  value would diverge silently and the symptom would be a PP worker that
  should have been refused. Assembling a real `NixlBaseConnectorWorker`
  in-process is not affordable, so the removal TODO is what bounds this.
- The HMA suppression replaces an upstream symbol's effect for the duration of
  one call rather than going through `patches/`, because what it adjusts is the
  argument upstream reads, not upstream itself. It carries a removal condition
  rather than a version: nothing released drops the refusal yet.

---


rebel-jongwonlee (commented):
Read both commits. The identity reduction holds and the two lists agree; one
question about a third place, and one small thing.

## Confirmed independently

**The config-scope fix is the one I traced on #1114** and did not get to post
before it merged: `compile_or_warm_up_model` opens no scope, so
`finalize_kv_cache_registrations` → `RblnTransferTopology.__post_init__` →
`get_kv_cache_shape` → `get_rbln_config()` reaches `get_current_vllm_config()`
outside one. Same line, same reason. Good that it rides here rather than
waiting for a branch of its own.

**`kv_per_block = 1` reduces every new expression to the shipped one.**
Checked each: `_kv_runs` returns 1, `kv_stride` is `kv_block_len`,
`for kv in range(1)` emits one entry at `+0`, `num_full_descs` is unchanged,
and `block_ids_2d * 1 + arange(1)` is `block_ids`.

**The two lists run in the same nesting.** Local is region → block → piece →
fanout → kv; remote builds `pieces` in piece-then-fanout order and loops
block → piece → kv under a caller that loops regions. Same order, and the
count assertion pins it:

```
assert len(out) == (len(logical_pairs) * len(areas_iter) * num_blocks
                    * split * fanout * kv_runs)
```

**The test approach is the part worth keeping.** Decoding each address back
into `(K-or-V, head)` and comparing against the head the pairing asked for is
what an address assertion cannot do — and, as the description says, is why the
old offset stayed green. That is the shape the next change to this arithmetic
should copy.

## One question

The description says the break appears in exactly two places — a head band and
a sliding window. A peer whose block size differs looks like a third: it also
reads part of a block, and it lands on upstream's arithmetic rather than on
either of the two paths this PR widened.

```
_compute_desc_ids              -> super() when `_sw_ratio is None`
register_local_xfer_handler    -> super() when nothing is narrowed
```

Both know nothing of `kv_per_block`, which is right for a whole block — but
`block_size_ratio > 1` has upstream slice one, and a prefix of a packed block
is part of K rather than a smaller block's K and V.

The two refusals I can find do not cover it: `_validate_head_matched_handshake`
rejects `block_size_ratio != 1` for heterogeneous TP, and
`_validate_remote_agent_handshake` asserts it for PP. Equal TP, no PP, unequal
block size is the combination that reaches upstream's path.

I cannot settle this here — no vLLM 0.26 source on this machine to read what
upstream does with the ratio. If it is unreachable, or if upstream's slicing
happens to stay inside K on both ends, one sentence saying so would close it
alongside the other two.

## One small thing

`self._kv_per_block = 2 // len(cache_list)` sits inside the per-layer loop, so
the last layer wins. Today every non-MLA, non-Mamba layer gives the same
answer, so it holds — but this is the shape #1058 fixed one level down: a
per-layer quantity read once and applied to all of them. And the expression
degrades quietly rather than loudly — `len(cache_list)` of 3 gives 0, which
surfaces later as a division by zero rather than at the line that assumed two.

`assert len(cache_list) in (1, 2)` next to it would pin both, and reads as the
fact the `2 //` encodes.

## What I did not do

No NPU and no 0.26 environment, so this is static: the reduction by
substitution, the two nestings by reading them side by side, and the guard
search above. The suite was not run, so the mutation claim in the description
is not something I checked.

rebel-jongwonlee (approved):
Both points are in, and the fourth commit answers the question the way I hoped
— refused rather than widened.

## The block-size refusal

It sits exactly in the gap, and the commit message names a third existing guard
I had missed (upstream's own `assert block_size_ratio == 1` for a hybrid
layout). Every path through `_validate_remote_agent_handshake` now has one:

```
_check_d2d_region_pairing              other layout          [commit 1]
pp_size <= 1, head-matched          -> ratio != 1 refused    [existing]
pp_size <= 1, equal TP              -> ratio != 1 refused    [commit 4]
pp_size > 1                         -> assert ratio == 1     [existing]
```

And the failure is stated more precisely than I had it: at ratio 2 the pieces
are the whole of K and the whole of V, so a peer block's keys and values both
land in our keys.

## The `2 // len(cache_list)` pin

This went past what I asked. I proposed `assert len(cache_list) in (1, 2)`,
which closes the quiet zero; the second assert closes the half I had only noted
in passing — that the value is model-wide while the expression runs per layer.

## The HMA suppression

Checked the spec hierarchy in 0.26, and the test admits exactly the shape it
means to:

```
MLAAttentionSpec(FullAttentionSpec)       -> suppressed   (the case)
SlidingWindowSpec(AttentionSpec)          -> not
MambaSpec(KVCacheSpec)                    -> not
ChunkedLocalAttentionSpec(AttentionSpec)  -> not
```

`get_representative_spec_type` is upstream's own and `UniformTypeKVCacheSpecs`
guarantees the inner specs share a type, so picking any is sound.

`_is_hma_required = not hma_disabled` is what upstream reaches under
`suppress`: its `any(not isinstance(g.kv_cache_spec, FullAttentionSpec) ...)`
over one `UniformTypeKVCacheSpecs` group is True, so the `and` reduces. And
restoring in `finally` is safe because upstream reads the flag once, at the
line that sets `_is_hma_required`; every later use reads the attribute.

Two things about it, neither blocking.

**Nothing compares the override against what upstream would compute.** The test
stubs `super().__init__()` and, as its comment says, deliberately leaves
`_is_hma_required` unset "so only the override can produce it" — which is the
right call for testing the override, and also means a change to upstream's
`any(...)` condition would leave these green while the value silently diverged.
The symptom would be a PP worker that should have been refused. I do not think
that is fixable in-process without a real NIXL agent, so this reads as a limit
worth naming rather than a gap to close — the removal TODO is what bounds it.

**Why does this one ride here?** The second commit carries an explicit reason
("nothing in this PR reaches a server without it"). The third is introduced as
"a separate subject carried here" with no such line. MLA is key-only, so
`kv_per_block` stays 1 and I cannot see a dependency on the packed work. If
there is one, a clause would say so; if there is not, that is fine too — but
the asymmetry with the second commit's justification is what made me ask.

## The description

The new paragraph on the third place is the part of this I would keep: it names
all three pre-existing guards and why none of them covered the combination,
which is the reasoning a future reader would otherwise have to redo. Same for
the clause on why `register_local_xfer_handler` and `_compute_desc_ids` falling
back to upstream is the path the refusal guards.

Approving.

rebel-jongwonlee (approved):
Both new commits are tests, and the product delta is byte-identical to what I
approved: comparing the added and removed lines of `<base>...<head>` restricted
to `vllm_rbln/` before and after gives no difference. So this re-approval rests
on the same code.

**The relocation is lossless.** 22 test names, each appearing once added and
once removed with matching counts, and 27 assertions on both sides. The file is
gone and nothing else moved with it.

Merging classes into files that already have their own is where a relocation
can lose something quietly — a class or a method landing on a name the
destination already binds replaces it with no error, which is the hazard
`test_lifetime_mixins.py` exists for one level down. Checked both destinations:
no duplicate class names, and no duplicate method names inside any class.

**The new case closes a real gap, in the shape where it is hardest to see.**
The existing descriptor tests call `_head_matched_desc` with `kv_runs` handed
in, and the parameter defaults to 1, so both the derivation and the hand-off
were unpinned and dropping either is silent. Following the arithmetic in the
case it picks:

```
cuts_l = 4 x 2 = 8      _head_split(8, 4) -> 1   (cuts_r <= cuts_l)
cuts_r = 1 x 4 = 4      _kv_runs(2, 8, 4) -> 2   (cuts_l != cuts_r)
                        sub_len = 256 // 2 // 1 = 128
                        len(out) = 1 x 4 x 2 x 1 x 1 x kv_runs  ->  8 and 16
```

This is the one shape where `split` and `kv_runs` disagree — a peer cutting
heads more coarsely than we do — so a mutation that conflated them would
survive every case where they happen to be equal. And the two workers differ in
`_kv_per_block` alone, which isolates the variable.

Counting descriptors and their lengths rather than decoding addresses is the
right division here: the address semantics are already pinned by
`TestDescriptorsNameTheRightElement`, and what was missing was whether the
count reaches the builder at all.

One nit, since the two workers are meant to read as a pair: `_worker` already
takes `kv_per_block`, and the sibling case twenty lines above passes it as a
keyword, so `packed._kv_per_block = 2` after construction is the only place in
the class that sets it the other way.

Re-approving.

rebel-jonghewk (approved):

rebel-ykchoi:
@rebel-cjlee @rebel-minkyu — could either of you take a look when you get a chance? P/D transfer does not work at all for models on the packed KV layout (the one the rbln_custom_ops kernels read) without this.

rebel-cjlee (approved):

PR #1128, opened 2026-09-17
fix(spec-dec): decode window cross block boundary

rebel-kblee:
### 🚀 Summary of Changes

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-kblee:
@rebel-cjlee @rebel-minkyu Please do let me know if there is anything I have missed.

rebel-cjlee:
@rebel-kblee Is this PR also valid in not R100 device?

rebel-kblee:
> @rebel-kblee Is this PR also valid in not R100 device?

https://rebellions-ai.slack.com/archives/C08PCMPJRA7/p1789621262554209

rebel-minkyu:
Could you check if the next block is properly allocated in scheduler with non-eagle methods, such as ngram, where num_lookahead_tokens=0?

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -874,9 +874,10 @@ def _prepare_inputs(
                 f"query_lengths={query_lengths}, "
                 f"logical_num_tokens={logical_num_tokens}"
             )
-            block_size = self.cache_config.block_size
             num_computed = self.input_batch.num_computed_tokens_cpu[:num_reqs]
-            front_pad = np.minimum(slack, num_computed % block_size).astype(np.int32)
+            front_pad = np.maximum(
+                0, num_computed + query_lengths - self.max_model_len
+            ).astype(np.int32)
```
rebel-cjlee, line 880: With `front_pad = 0` the slack all goes behind, so the window now ends at `num_computed + num_spec`. But `allocate_slots` only reserves up to `num_computed + logical` when `num_lookahead_tokens` is 0, and it is 0 for ngram, suffix and medusa. eagle and mtp are fine, those reserve `num_spec`.

Real run, `block_size` 16, `num_spec` 4, ngram, 13 computed, 2 drafts: the request owns 1 block (positions 0..15) but the runner stages positions 13..17. `block_table[req][1]` was never filled, and `BlockTable.add_row` does not clear the row tail, so it holds either 0 (the shared null block) or the previous owner's block. I checked on device that the kernel happily splits the write into that entry, so we scatter KV into someone else's page.

Before this PR the scheduler clamp in `rbln_scheduler.py` plus the old `min(slack, num_computed % block_size)` always put the slack in front, on positions that already had blocks. That kept the window inside the reservation, which went away with them.

#1131 hits the same thing on the drafter side and keeps a check for it (`table[arange, last_page] == 0` then drop the drafts). Could we either pass `num_spec_tokens` as the lookahead in `RBLNScheduler` for those methods, or do the same check here?

I think this is the same concern @rebel-minkyu  raised.

tests/native/v1/worker/test_rbln_model_runner_inputs.py
```diff
@@ -109,25 +109,21 @@ def test_drafter_runs_the_logical_length(self, make_model_runner, monkeypatch):
...
-            (3, 1, 2),
-            # At a block start there is no tail to re-run, so the slack has
-            # nowhere to go but behind the scheduled token.
-            (BLOCK, BLOCK, 0),
+            3,
+            BLOCK - 1,
+            BLOCK,
         ],
     )
-    def test_window_is_fixed_and_stays_in_one_block(
```
rebel-cjlee, line 129: These three values all land on the same case. The new formula only reads `num_computed + query_lengths` against `max_model_len`, and `block_size` is gone, so 3, `BLOCK - 1` and `BLOCK` all give `front_pad = 0`. Against `dev`, `BLOCK` passes as well, since `1024 % 1024` is 0 and the old formula also gave 0 there.

Rather than trimming it, could we point the parameterize at the `max_model_len` boundary, which is where `np.maximum` actually switches? With `MAX_MODEL_LEN` 4096 and window 3 I get:

- num_computed 3 / positions [3, 4, 5] /  back_pad 2 / front_pad 0
- num_computed 4093 /  positions [4093, 4094, 4095] / back_pad 2  / front_pad 0
- num_computed 4094 /  positions [4093, 4094, 4095] / back_pad 1  / front_pad 1

4093 is the last value that does not shift, and the window ends exactly at `max_model_len - 1` there. Each value checks something different, and all three fail on `dev`. That also lets `test_a_window_past_max_model_len_moves_the_overshoot_in_front` fold into the same parameterize, so one test tells the whole story of the formula.

One case is missing either way. `front_pad` and `back_pad` are per request arrays while `query_lengths` is `np.full`, so the interesting shape is a batch where only one request is near the end. I ran two requests at `num_computed` 10 and `max_model_len - 2` and got `positions [10, 11, 12, 4093, 4094, 4095]` with `back_pad [2, 1]`. That one fails on `dev` too.

tests/native/v1/core/test_rbln_scheduler.py
```diff
@@ -1082,26 +1068,14 @@ def _running_decoder_with_spec():
         req0.spec_token_ids = [1] * 4
         return sched, req0
 
-    def test_unsafe_decode_ready_join_trims_whole_batch(self):
-        # req1 matches the full 16-token prefix -> joins at the block start, so
-        # its backfill would cross the boundary -> the batch loses its spec.
+    @pytest.mark.parametrize("prefix_len", [16, 8])
+    def test_a_decode_ready_join_keeps_the_batch_on_spec(self, prefix_len):
```
rebel-cjlee, line 1099: These classes lost their docstrings in the rename, so what they guard now is not written down anywhere. `prefix_len` 16 and 8 used to be the unsafe and safe sides of the removed check. Both keep spec now, so the pair only makes sense if you know the old story. Same for the `TestBlockBoundaryJoin` name and for `TestScheduleSpecDecodeCap`, since the cap is gone. Could we put one line back on each class saying what the new behavior is, and say in a comment why 16 and 8 are still there?

tests/native/v1/worker/test_rbln_model_runner_inputs.py
```diff
@@ -145,11 +141,36 @@ def test_window_is_fixed_and_stays_in_one_block(
...
+        )
+
+        assert query_lengths.tolist() == [window]
+        positions = runner.positions[:window].tolist()
+        assert positions[-1] == runner.max_model_len - 1
+        assert positions == [num_computed - 1, num_computed, num_computed + 1]
+        assert logits_indices.tolist() == [1]
+        assert runner.decode_back_pad_np[0] == 1
+
     def test_kept_drafts_at_a_block_start_sample_before_the_back_padding(
```
rebel-cjlee: Its siblings got renamed but this one kept "at a block start". It still passes num_computed = BLOCK, and block position no longer changes anything, so the name points at behavior the runner does not have. What it really covers is that a kept draft samples before the back padding, which the rest of the name already says.

vllm_rbln/v1/worker/rbln_model_runner.py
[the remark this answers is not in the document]
rebel-cjlee: This comment looks wrong.

rebel-kblee: Yes. I had missed the point that mentioned. Thank you! @rebel-cjlee @rebel-minkyu

rebel-cjlee (approved):

PR #1129, opened 2026-09-17
feature(dynamic-kv): dynamic kv as default

rebel-jonghewk:
## Summary

Enable dynamic KV cache sizing by default on the vllm model path. It is selected through `--rbln-use-dynamic-kv-cache` (`RBLNConfig.use_dynamic_kv_cache`, tri-state): unset, configurations the path cannot resize fall back to vLLM's pre-compile estimate; set, such a configuration is refused at start-up; `--no-rbln-use-dynamic-kv-cache` keeps the estimate. `VLLM_RBLN_USE_DYNAMIC_KV_CACHE` stays as the deprecated bridge into the field until 0.14.0.

This builds on [#1138](https://github.com/RBLN-SW/vllm-rbln/pull/1138), which moves RBLN NIXL KV-cache registration after the resize.

## Changes

- Remove `VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN`, the `DRY_RUN` mode, and dry-run reporting.
- Centralize the resolved decision in `dynamic_kv_unsupported_reason()` / `dynamic_kv_enabled()`.
- Use static sizing for device-tensor-off, custom-kernel, non-paged attention, DFlash drafter, and unverified connector configurations.
- Keep `mark_dynamic` and the mega-cache signature aligned with the resolved dynamic/static mode.
- Support the three RBLN NIXL connectors and `RBLNLMCacheConnectorV1` after post-resize KV registration.
- Log unsupported fallback once; `--no-rbln-use-dynamic-kv-cache` is silent.
- Update focused tests and the dynamic-KV documentation.

## Migration

The default flip changes the mega-cache bundle key for every deployment that never set the variable, so each one recompiles cold once on the first start after upgrading. `--no-rbln-use-dynamic-kv-cache` restores the previous sizing but not the previous key.

## Test plan

- `pytest tests/vllm/v1/worker/test_utils.py`
- `pytest tests/vllm/v1/worker/test_dynamic_kv_sizer.py`
- `pytest tests/vllm/v1/worker/test_mega_cache.py`
- `pytest tests/vllm/v1/worker/test_rbln_worker.py tests/vllm/patches/test_dynamic_kv.py`
- `pytest tests/vllm/test_platform.py tests/vllm/test_rbln_config.py tests/vllm/test_envs.py`
- Run pre-commit checks on changed files.

## Follow-up

Extend `DYNAMIC_KV_SUPPORTED_CONNECTORS` after additional connectors are verified against the post-resize registration flow.


vllm_rbln/v1/worker/utils.py
```diff
@@ -303,6 +303,85 @@ def compile_and_warmup_skip_reason(vllm_config: VllmConfig) -> str | None:
     return None
 
 
+DYNAMIC_KV_SUPPORTED_CONNECTORS = (
```
rebel-jongholee, line 306: Could we add an assertion to make sure these four names exist in KVConnectorFactory._registry? If there's a typo, it silently falls back to static with a warning, which is the exact failure mode we want to avoid.

rebel-jonghewk: https://github.com/RBLN-SW/vllm-rbln/pull/1129/commits/1c9f40c0eea47f18a80291ac5134d2830146b997

rebel-jongholee (approved):

rebel-jinhwan:
Hardware verification for the `RBLNLMCacheConnectorV1` entry, since the test plan
above is unit tests only. One RBLN-CR13, Qwen3-1.7B, `block_size=256`,
`LMCACHE_CHUNK_SIZE=256`, lmcache-rbln at `main`.

```
[Dynamic KV] reallocating KV cache: 4 -> 3936 blocks
LMCache INFO: Registered 28 KV caches on rbln:0, allocated shape=(3936, 2, 8, 1, 256, 128)
[rank 0] OK: layer-0 data_ptr matches (0xeb4e3b1a), final_num_blocks=3936
```

Registration reads 3936 blocks and happens once — the connector never sees the
4-block compile cache. Probed through `collective_rpc`: the pointers
`RBLNConnector` holds are a subset of `model_runner.kv_caches`, layer 0 matches,
and they do not move once serving starts. Store then restore returned identical
output with `external_prefix_cache_hits = 256`; with the flag off it comes up at
3925 blocks and produces the same output. The entry holds up.

notes from the same run.

**`MultiConnector` reads as unsupported.** The check compares one name, but the
wrapper registers nothing of its own — `register_kv_caches` and the post-resize
finalize both fan out to its children. So `MultiConnector[RblnNixl…,
RBLNLMCacheConnectorV1]` silently falls back to static sizing under the new
default. #1143 fixes that on top of this branch, and verified on hardware as
well: NIXL registers `dmabuf_size=1031798784` = 3936 blocks × 1 MiB ÷ 4 chiplets.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -462,16 +463,12 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
...
 
-        self.dynamic_kv.assert_attention_layout()
-
         self.model_runner.initialize_kv_cache(
             self.dynamic_kv.shrink_for_compile(kv_cache_config)
         )
         if not self.dynamic_kv.defers_kv_registration:
             self.model_runner.register_kv_caches_with_connector()
 
-        self.dynamic_kv.assert_cache_layout()
```
rebel-cjlee, line 473: This drops the only guard for cross layer KV sharing and nothing replaces it. `dynamic_kv_unsupported_reason()` has no row for it, and the docs row is gone too. But `rbln_model_runner.py:3104-3110` still gives a shared layer the same view object as its target, which is what the comment you deleted said the compiler will not take. The feature is on by default now, so such a model goes down that path with no warning. Could you check if it is safe to drop?

vllm_rbln/v1/worker/utils.py
```diff
@@ -303,6 +304,95 @@ def compile_and_warmup_skip_reason(vllm_config: VllmConfig) -> str | None:
...
+        return (
+            "RBLN_USE_CUSTOM_KERNEL is on, and the rbln_triton_ops kernels take "
+            "no dynamic KV input"
+        )
+    if not rbln_config.use_flash_causal_attn:
+        return (
+            "flash causal attention is off, so the model dispatches to an "
+            "attention kernel that does not accept a dynamic KV input"
+        )
+    if vllm_config.attention_config.use_non_causal:
```
rebel-cjlee: It is per model, not per process. So, the draft config can set it only. Is this intentional?

vllm_rbln/envs.py
```diff
@@ -287,20 +286,11 @@ def use_auto_port() -> bool:
     ),
     # --- DYNAMIC KV CACHE ---
     # Size the KV cache from the compiled artifact instead of the estimate.
-    # The dry-run variable below implies it, but only when this one is unset:
-    # an explicit 0 is a decision, not a default to override.
+    # Configurations the path cannot size turn it off on their own
+    # (`dynamic_kv_unsupported_reason`); 0 turns it off everywhere.
```
rebel-cjlee: Design question. before this PR an explicit `=1` with an unsizable config raised at start up. Now it falls back with one warning.
For the unset case that is right. But the comment you removed said an explicit 0 is a decision and not a default to override, and the same holds for an explicit 1. Could we fall back when the variable is unset, and raise when the user typed 1?

tests/vllm/v1/worker/test_dynamic_kv_sizer.py
```diff
@@ -426,31 +378,69 @@ class TestModeResolution:
...
+            ),
+            attention_config=SimpleNamespace(use_non_causal=False),
+            kv_transfer_config=None,
+        )
+
+        with caplog.at_level("WARNING"):
+            sizer = DynamicKvSizer(config, object(), foreign_dram_used_bytes=0)
+
+        assert sizer.mode is dks.DynamicKvMode.DISABLED
+        assert "[Dynamic KV] off for this run" not in caplog.text
```
rebel-cjlee: This string is logged nowhere. So the assertion can never fail, and the test does not check what its name promises.

tests/vllm/test_envs.py
```diff
@@ -366,17 +366,3 @@ def test_config_fields_are_not_hashed_into_the_bundle_key():
...
-    monkeypatch.setenv("VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN", "1")
-    assert envs.environment_variables["VLLM_RBLN_USE_DYNAMIC_KV_CACHE"]() is True
-    assert envs.environment_variables["VLLM_RBLN_DYNAMIC_KV_CACHE_DRY_RUN"]() is True
-
-
-def test_dynamic_kv_cache_is_opt_in(monkeypatch):
-    # The flag is read during compile, so a non-False default would change the
-    # compiled artifact for every existing deployment.
-    monkeypatch.delenv("VLLM_RBLN_USE_DYNAMIC_KV_CACHE", raising=False)
-    assert envs.environment_variables["VLLM_RBLN_USE_DYNAMIC_KV_CACHE"]() is False
```
rebel-cjlee, line 382: Dropping this is fine, since the reason in its comment is now covered by the mega cache key. But nothing takes its place, so the new default is untested. Could you add:
```python
    def test_dynamic_kv_cache_is_on_by_default(monkeypatch):
        monkeypatch.delenv("VLLM_RBLN_USE_DYNAMIC_KV_CACHE", raising=False)
        assert envs.environment_variables["VLLM_RBLN_USE_DYNAMIC_KV_CACHE"]() is True
```

tests/vllm/v1/worker/test_mega_cache.py
```diff
@@ -87,6 +101,19 @@ def test_rebel_minor_bump_invalidates(self, monkeypatch):
         monkeypatch.setattr(mega_cache, "_rebel_major_minor", lambda: "0.12")
         assert mega_cache.config_signature(_stub_config()) != before
 
+    def test_resolved_dynamic_kv_decision_invalidates(self, monkeypatch):
+        enabled = True
+        monkeypatch.setattr(
+            mega_cache,
+            "dynamic_kv_enabled",
+            lambda config: enabled,
+            raising=False,
```
rebel-cjlee: `raising=False` is not needed, since `mega_cache.dynamic_kv_enabled` exists. It only means the test keeps passing if that import is ever removed, while covering nothing.

vllm_rbln/v1/worker/utils.py
```diff
@@ -303,6 +304,95 @@ def compile_and_warmup_skip_reason(vllm_config: VllmConfig) -> str | None:
...
+    if kv_transfer is not None:
+        unregistered = [
+            name
+            for name in DYNAMIC_KV_SUPPORTED_CONNECTORS
+            if name not in KVConnectorFactory._registry
+        ]
+        assert not unregistered, (
+            "DYNAMIC_KV_SUPPORTED_CONNECTORS names connectors the factory never "
+            f"registered: {unregistered}"
+        )
```
rebel-cjlee: Non-blocking.
I get the intent, but I would rather not have this here. `_registry` is private in upstream, and `assert` disappears under `python -O`. What it guards is a typo in our own tuple, which is a developer mistake and not a runtime state. A test gives the same guarantee with no monkeypatch:
```python
def test_every_supported_connector_is_registered():
    for name in DYNAMIC_KV_SUPPORTED_CONNECTORS:
        assert name in KVConnectorFactory._registry
```
Then an upstream rename shows up as a red test instead of a broken run. The current test patches the tuple and checks the assert fires, so it only tests the assert.

Also this `if kv_transfer is not None` sits right on top of the same condition at line 369.

rebel-jonghewk:
@rebel-jinhwan Thanks for the hardware run and for #1143 — I will merge this first and review #1143 on top.

On the staging buffer: `budget_left` is the distance to `total × gpu_memory_utilization`, not to the physical limit. The `(1 − gmu)` slice is intentionally outside the KV budget, and that is where allocations made after sizing, like this one, land — the same contract as upstream vLLM. Even at the default 0.93 that slice is two orders of magnitude above the
buffer, so 28–62 MiB fits comfortably and I would leave it without a reserve
unless a real OOM shows up

rebel-ykchoi (commented):
**One blocker**, and an answer to one of the questions already open in this thread. The hygiene side is clean — comment budgets, no absolute measurements, no swallowed failures, the removed env var gone from all three places. I have left out what the inline comments already cover.

## 1. Blocker — a DFlash run takes the dynamic path, which this PR's own reason list rules out

`v1/worker/utils.py:344`. The question already raised on that line — whether reading `use_non_causal` off the target config is intentional, since it is per model and only the draft config sets it — has a consequence worth stating in full.

The deleted `DynamicKvSizer.assert_attention_layout()` walked the built layers and required `impl.is_causal is True and impl.is_normal is False`, naming the offending layer. The replacement reads `vllm_config.attention_config.use_non_causal` on the **target** config. Upstream sets that field on a `replace()`d **draft** config only (`vllm/v1/spec_decode/dflash.py:88-93`), and this repo *requires* the drafter to be non-causal — `v1/spec_decode/dflash.py:103` raises `NotImplementedError` when it is not. So every RBLN DFlash run has a non-causal layer, and the new gate cannot see it.

The drafter's KV cache is inside what gets marked, on three counts:

- the proposer is built on the same runner — `v1/worker/rbln_model_runner.py:358`, `RBLNDFlashProposer(self.vllm_config, self.device, self)`
- `mark_dynamic` is applied to every entry of `self.kv_cache_bases`, with no target/drafter distinction — `:3126-3134`
- the sizer already expects a drafter's KV inputs and sums them — `v1/worker/dynamic_kv_sizer.py:460`, `# Disjoint sets (a target's and a drafter's): the sum is the answer.`

And nothing else stops it: `speculative_config`, the drafter and the draft config appear nowhere in `dynamic_kv_unsupported_reason`, `dynamic_kv_sizer.py` or `patches/dynamic_kv.py`. Before this PR the layer walk refused the run; now it is the default path into a kernel this PR's own reason text describes as one that "does not accept a dynamic KV input".

What is not established is the failure *mode* — whether the compiler rejects the graph or the output is wrong. That does not change the shape of the finding.

## 2. On two open threads

**Cross-layer KV sharing is safe to drop.** Both attention implementations on this path reject it at construction — `v1/attention/backends/flash_attention.py:383` and `v1/attention/backends/mla/flashattn_mla.py:149`, each `raise NotImplementedError("KV sharing is not supported in RBLN.")` when `kv_sharing_target_layer_name is not None` — and those two are the only implementations under `v1/attention/backends/`. A model that shares KV cannot build its layers, so `shared_kv_cache_layers` cannot be populated and `rbln_model_runner.py:3102-3110` cannot be reached. The removed guard covered a state that cannot occur, which `AGENTS.md` §Scope asks us not to guard. Removing the documentation row with it is consistent.

**The staging-buffer answer holds**, and one edge is left. `kv_placement.py:381` computes `budget = int(mem.total * gpu_memory_utilization)`, so the `(1 − gmu)` slice is indeed outside the KV budget and later allocations land there. The one case where that slice is not there is `gmu = 1.0`, which `:367` accepts (`0 < gpu_memory_utilization <= 1`). Under the static estimate the whole-card formula happened to leave room anyway; the measured path at `1.0` leaves none. Worth a sentence in the documentation rather than a reserve.

**Raising on an explicit `=1`** — I agree with the question on `envs.py:290`, and 3.2 below is the same theme from the other side: there are configurations the gate cannot classify, where the resize fails after the decision is made.

## 3. Behavior

### 3.1 `MultiConnector` is matched by name and silently falls back — `v1/worker/utils.py:369`

The gate compares one top-level name. Everywhere else the repo unwraps: `iter_kv_connectors` ("Yield leaf connectors, expanding `MultiConnector` children recursively"), and the resize path's own `finalize_kv_cache_registrations` recurses too. So the mechanism already supports the wrapper and only the gate refuses it — and `v1/worker/rbln_worker.py:516` and `tests/vllm/v1/worker/test_kv_connector_preemptions.py:20` both describe that wrapper as the setup RBLN uses. The fix is in the stacked follow-up, but **this PR is the one that turns the default on**, so the exposure arrives here. Adding `"MultiConnector"` to the allowlist leaves the suite green, so neither the current behavior nor the fixed one is pinned by a test — one test in this PR would fix that regardless of where the behavior lands.

Two more cases sit on the same axis. `KVConnectorFactory.get_connector_class` prefers `kv_connector_module_path` over the registry, so a name that happens to match admits a class nobody verified; and the dotted form the LMCache shim's own docstring suggests does not match the allowlist, so a verified class is refused. The discriminator is not the class that loads.

### 3.2 The default turns configurations that used to serve into a start-up refusal — `v1/worker/dynamic_kv_sizer.py:279`

When the estimate is not above the compile hint, `shrink_for_compile` raises. `dynamic_kv_unsupported_reason` cannot see the estimate, so this shape is not among the ones the feature turns itself off for. Previously only an explicit opt-in met it. This is the case the function's own docstring and the documentation set out to avoid — "refusing would stop a run that `VLLM_RBLN_USE_DYNAMIC_KV_CACHE=0` would have served".

### 3.3 MLA is not enumerated by either the old guard or the new gate — `v1/worker/utils.py:315`

`RBLNFlashAttnMLAImpl` carries `is_causal` and no `is_normal`, and whether `MLAAttention` was ever collected by the old `get_layers_from_vllm_config(..., Attention)` walk is unclear. The new gate does not distinguish MLA either. The sparse DeepSeek branch dispatches to an op outside the `paged_*` family, which is the argument the other reasons rest on, yet nothing turns it off. Could the description say which side MLA is on?

## 4. Tests

### 4.1 The wiring this PR exists for is unpinned

Five mutations survive the suite:

| mutation | result |
|---|---|
| `DynamicKvSizer.__init__` receives `unsupported_reason=None` (`dynamic_kv_sizer.py:170`) | survives |
| `rbln_model_runner.py:3025` back to the raw env flag | survives |
| `rbln_model_runner.py:3126` likewise | survives |
| `rbln_worker.py:184` back to the raw flag | survives |
| the foreign-DRAM read made unconditional | survives |

The first means "a configuration that cannot resize also turns the worker off" — the behavior this PR is built around — is not asserted anywhere. The next two mean the stated reason for centralizing the decision, that `mark_dynamic` follows the resolved value rather than the flag, has nothing behind it. Overall: 24 of 32 mutations caught.

### 4.2 Three propositions lost with the deleted tests

Beyond the one already noted on `test_envs.py`:

- **No test builds a speculative-decoding shape at all** (`test_speculative_decoding_passes` was removed). With 1 that is not a coincidence: nothing in the gate and nothing in the suite would notice a DFlash run entering the dynamic path.
- **No test builds an MLA shape** (`test_mla_passes` removed; the new config helper has no `use_mla`). Same relationship to 3.3.
- No test builds a sliding-window layer shape.

### 4.3 On the proposed default test

`test_declared_default_matches_resolved` pins the *agreement* between the `TYPE_CHECKING` declaration and the lambda, not the value: flipping both together still passes. So the suggested assertion is not a duplicate of it.

### 4.4 Three failures in the cpu lane

`--device-tensor 0`: `test_the_scheduler_keeps_the_estimate_while_the_cache_is_shrunk`, `test_dynamic_kv_feeds_the_chiplet_snapshot`, `test_dynamic_kv_raises_a_short_estimate_to_one_request`. The lane was already red at the base (five failures), so this is not a regression, but all three are tests this PR writes or rewrites, and the cause is the same: the sizer's mode now depends on `VLLM_RBLN_USE_DEVICE_TENSOR` and these patch only the flag. The `_device_tensor_on` autouse fixture in the same file is the fix.

Default lane: 608 passed.

## 5. Operational impact, not in the description

Flipping the default invalidates every existing mega-cache bundle. The resolved env hash differs for a deployment that never set the variable, removing the dry-run variable changes the factor keys, and the decision is now a warm-up graph-set factor as well. With `mega_cache_only`, that bundle is the only cache, so every existing deployment recompiles cold once.

The removed `test_dynamic_kv_cache_is_opt_in` carried exactly this as its reason — "a non-False default would change the compiled artifact for every existing deployment". A migration line in the description would carry it forward.

## 6. Repository rules

- **`envs.py:274-277` reformats `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST`, which has nothing to do with this change.** Both the base and the new text are `ruff format` stable, so the hook did not produce it. `AGENTS.md` §Scope: "Do not refactor, rename, or reformat code that is not part of the fix. A refactor is its own PR."
- **`v1/worker/utils.py:320`** — "Every reason here is a property of the deployment or of the kernel the model dispatches to, not a request from the caller" is contradicted by the function's own first three branches, all of which are environment variables the caller sets.
- **One place per fact** — `v1/worker/utils.py:333-334` and `tests/vllm/v1/worker/test_utils.py:1458-1459` carry the same two lines verbatim, and `docs/dynamic_kv_cache.md` a third copy.
- **`v1/worker/dynamic_kv_sizer.py:80`** — the DISABLED `mode_reason` is never read in production (`:262`, `:270`, `:389` are INERT and PINNED), but a test pins its literal.
- **Leftovers** — `tests/vllm/v1/worker/test_rbln_worker.py:1192-1193` still stubs `assert_attention_layout` and `assert_cache_layout`, both removed here; `tests/vllm/patches/test_dynamic_kv.py:31` still describes the dry run.
- **`docs/dynamic_kv_cache.md`** contradicts itself on `VLLM_RBLN_USE_VLLM_MODEL`: `:72-73` says the path needs it and turns itself off without it, `:130-132` says the feature is absent on the optimum path rather than disabled. The gate has no such check, so the second is the accurate one. The document also has no row for `MultiConnector` (3.1).

## 7. Description and commits

- The description does not use `.github/PULL_REQUEST_TEMPLATE.md`; no linked issue and no label, both required by `CONTRIBUTING.md`.
- It says "vLLM-native path". `AGENTS.md` §Terms: **"Not 'native'"** — the term is **vllm model path**.
- `980409a3` has no `type(scope):` and no body, and it owns nearly all of this PR: the default flip, the dry-run removal, both layout guards, the unrelated reformat and the documentation rewrite. `91427fc9` introduces `dynamic_kv_unsupported_reason` with an empty body; `03612916` likewise. The two commits that do carry a body are good.

## 8. Verification

Everything above was checked against the diff, the base revision and the installed upstream sources; the mutation results and lane numbers come from running the seven touched test files. No hardware was used. For 1 that leaves the failure mode open, while the path into it is established from the code cited. How often 3.2 is reached and whether the sparse MLA op is on the compiler's whitelist are unverified. The stacked follow-up was not read — 3.1 describes this branch on its own.

vllm_rbln/v1/worker/rbln_worker.py
rebel-jonghewk: Intended, but the reason I deleted with it was stale: since rebel_compiler's update, a dynamic KV input may reach several paged consumers through views (the shape gpt-oss's deduped base already serves on), and `reallocate` re-runs the sharing loop.

vllm_rbln/v1/worker/utils.py
rebel-jonghewk: Not intentional — as you say, upstream sets `use_non_causal` on the draft config only, so the check on the target config never fired for the case it was written for. Replaced it with `speculative_config.method == "dflash"`: the RBLN drafter is non-causal by requirement, so the method is the signal. A causal drafter (eagle3) stays supported; test added for both.

rebel-cjlee (approved):

rebel-ykchoi:
Re-read at `6c685dce`, against `dev` at `5b92af85`. Everything below was checked at that boundary.

## Closed

**The DFlash gate is the right discriminator.** `RBLNDFlashProposer` is selected exactly on `speculative_config.method == "dflash"` (`v1/worker/rbln_model_runner.py:357`), and upstream normalizes the method to that value. The class is covered as a whole: `use_non_causal` is set in exactly one place in upstream's `v1/spec_decode/` — the DFlash draft config — and `dspark` has no proposer here or upstream. A causal drafter staying supported matches what the old layer walk did.

**The start-up refusal sees it.** `check_and_update_config` runs from `VllmConfig.__post_init__` (call at `config/vllm.py:1454`, the method starts at 936), and the same `__post_init__` reads `speculative_config` well before that, so the refusal is not evaluated against a half-built config.

**Moving the knob behind `--rbln-use-dynamic-kv-cache` holds up.** `RBLNConfig.compute_hash` hashes every field except the three in `ignored_factors`, so dropping the variable from `RBLN_COMPILE_ENV` keeps the artifact keyed; the two sets still partition, with the variable now in `RBLN_NON_COMPILE_ENV`; and `mega_cache` still carries the resolved decision. The deprecation names the version that removes it, which is what keeps a bridge temporary.

**Also closed since the first round**: the registry `assert` became a test and took the duplicated `if kv_transfer is not None` with it; the cross-layer alias unbind is gone with its reason in its own commit; a speculative-decoding shape is now built by a test, which was the coverage gap behind the blocker; and the description says "vllm model path".

`tests/vllm` is green at this revision — the files this PR touches and their neighbours.

## New, minor

**`unset` and `True` hash differently while resolving identically** — `config.py:102`.

```
unset  dd350b521b2a8eed
True   f811f8b2e80f920a
False  1b7e2e92cd66db16
```

`dynamic_kv_enabled` is `use_dynamic_kv_cache is not False and reason is None`, so `None` and `True` are the same decision on any configuration that can be sized. They differ only in whether an unsizable one is refused or falls back, which does not change a compiled graph. `mega_cache` already keys the bundle on the resolved decision, so hashing the request as well splits the cache between a deployment that passes the flag and one that leaves it unset, for no behavioural difference. Adding the field to `ignored_factors`, or normalizing `None` to `True` before hashing, collapses the two.

## Still open

Re-checked at this revision:

- `envs.py` still reformats `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST`, which this change does not otherwise touch.
- `v1/worker/utils.py:319` — "not a request from the caller" has moved further from the code: the value it reads is a CLI flag, and an explicit `True` is exactly a request from the caller, which the new refusal acts on.
- The two verbatim lines in `v1/worker/utils.py` and `tests/vllm/v1/worker/test_utils.py`, with a third copy in the documentation.
- `tests/vllm/patches/test_dynamic_kv.py:31` still describes the dry run.
- The description does not use `.github/PULL_REQUEST_TEMPLATE.md`, and has no linked issue and no label; `CONTRIBUTING.md` requires the last two.
- The documentation has no row for `MultiConnector`, and the allowlist still matches one top-level name.

Carried from the earlier round and not re-measured here: the surviving mutations on the worker-side wiring, the cpu-lane failures, and a migration note for the one-time cold recompile.

rebel-jonghewk:
@rebel-ykchoi I've took care of the minors (https://github.com/RBLN-SW/vllm-rbln/pull/1129/commits/6dfda0d9d7ea1c490e4bc5d4cd9f131ba46eb39e)

PR #1130, opened 2026-09-17
other(platform): default gpu_memory_utilization to 0.93

rebel-jonghewk:
### 🚀 Summary of Changes

Raise the RBLN default `gpu_memory_utilization` from upstream vLLM's `0.92` to `0.93`.

The override is applied through `RblnPlatform.apply_config_platform_defaults`, so it covers both the optimum-rbln and vLLM-native paths.

#### What changed

- Add `RBLN_DEFAULT_GPU_MEMORY_UTILIZATION = 0.93`.
- Apply the RBLN default through vLLM's platform-default hook.
- Set the `vllm serve` parser default to `0.93`.
- Add regression coverage for config resolution, explicit overrides, the serve parser, and upstream default drift.

#### Behavior and known limitation

- An omitted value resolves to `0.93`.
- An explicit value other than upstream's default is preserved.
- `gpu_memory_utilization` has no unset sentinel. Because omitted values and an explicit `0.92` are indistinguishable by the time the platform hook runs, both resolve to `0.93`.
- The upstream help text still mentions `0.92`, although the parser's advertised default is now `0.93`. Updating that text is out of scope for this PR.

---

### ✅ Type of Change

- [ ] 🚀 Release (`release`)
- [ ] ✨ Feature (`feature`)
- [ ] 🧬 Core engine changes (`core`)
- [ ] 🛠 Bug fix (`fix`)
- [ ] ⚙️ Performance improvement (`perf`)
- [ ] 🔁 Refactor or code cleanup (`refactor`)
- [ ] 📄 Documentation (`docs`)
- [x] ❓ Other (`other`): platform default change

---

### 🧪 How to Test

```bash
uv run --no-sync pytest tests/native/test_platform.py::TestSchedulerOverrides -x
```

Coverage verifies that:

- the upstream default resolves to the RBLN default (`0.93`);
- an explicit non-default value is preserved;
- `vllm serve` uses `0.93` as the parser default;
- `LLM.__init__` and `CacheConfig` still share the same upstream default.

---

### 👀 Reviewer Notes

Please focus on:

- whether `apply_config_platform_defaults` is the correct lifecycle hook;
- the `0.92` ambiguity described above;
- the canary test guarding against upstream default drift.

---

### 📋 Checklist

- [x] The test method and expected results are described


rebel-kblee (approved):
👍

tests/native/test_platform.py
```diff
@@ -449,6 +453,33 @@ def test_unset_max_num_seqs_takes_the_rbln_default(self, configured):
...
+        )
+
+    def test_llm_init_still_shares_the_upstream_default(self):
+        """LLM.__init__ carries its own literal, not CacheConfig's; the override
+        keys on CacheConfig's, so the two must not drift apart."""
+        llm_default = (
+            inspect.signature(LLM.__init__).parameters["gpu_memory_utilization"].default
+        )
+        assert llm_default == CacheConfig.gpu_memory_utilization
+
```
rebel-cjlee: These four tests are about `CacheConfig`, but `TestSchedulerOverrides` is about the scheduler. Could t hey go in their own class, for example `TestCacheOverrides`?

vllm_rbln/platform.py
```diff
@@ -295,11 +297,30 @@ def pre_register_and_update(
             if action.dest == "block_size":
                 action.choices = None  # Override choices
 
+        for action in parser._actions:
```
rebel-cjlee: There are already two loops over `parser._actions` right above this one. Could this condition go into one of them?

tests/native/test_platform.py
```diff
@@ -449,6 +453,33 @@ def test_unset_max_num_seqs_takes_the_rbln_default(self, configured):
...
+        assert config.cache_config.gpu_memory_utilization == 0.5
+
+    def test_serve_advertises_the_rbln_gpu_memory_utilization(self):
+        parser = AsyncEngineArgs.add_cli_args(FlexibleArgumentParser())
+        assert (
+            parser.get_default("gpu_memory_utilization")
+            == RBLN_DEFAULT_GPU_MEMORY_UTILIZATION
+        )
+
+    def test_llm_init_still_shares_the_upstream_default(self):
```
rebel-cjlee: What is this test meant to catch? It passes on `dev` as well, so it does not cover anything this PR adds.

rebel-cjlee (approved):
These comments are not blocking. Just check please.

PR #1131, opened 2026-09-17
fix(spec-dec): remove boundary restrictions 

rebel-kblee:
### 🚀 Summary of Changes

following this pr
https://github.com/RBLN-SW/vllm-rbln/pull/1131#issue-5484623407

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

### 📋 Checklist

---

### 💬 Notes

---


tests/native/v1/spec_decode/test_dflash.py
```diff
@@ -207,62 +183,64 @@ def test_a_wider_batch_is_refused(self, max_num_seqs):
...
+            ([[71, 0, 0, 0]], [0]),
+            ([[71, 6, 0, 0]], [SPANS]),
+        ],
+        ids=["stays_on_its_page", "next_page_allocated"],
+    )
+    def test_a_step_proceeds_when_the_page_it_ends_on_is_allocated(
+        self, table, ctx_lens
+    ):
+        with pytest.raises(RuntimeError, match="reached the draft pass"):
+            self._run(table, ctx_lens)
```
rebel-cjlee: This is the one test that covers the change, and it does not check the thing that changed. On `dev` a spanning row had its write redirected to the next page and then lost its drafts (`self._dropped_rows = crossing`), but the step still reached the draft pass. So `pytest.raise(RuntimeError)` passes on the old behavior too. The new part, that a spanning row keeps its drafts, it never asserted. Of the two params only `next_page_allocated` changed behavior, and `stays_on_its_page` never spanned in the first place.

I put `self._dropped_rows = spanning` back in right after the give up branch ;and the whole suite still passed, 90 passed and 5 skipped.

`_run` raises before it can return the proposer, so this needs a small change. Something like a sentinel exception and a `try`, the way `PastTheGuard` was used elsewhere in the repo:
```python
class _ReachedDraftPass(Exception):
    pass

def _run(self, table, ctx_lens):
    ...
    proposer = SimpleNamespace(..., _build_draft_attn_metadata=_raise)
    try:
        RBLNDFlashProposer._run_query_pass(proposer, cad, ...)
        proposer.reached = False
    except _ReachedDraftPass:
        proposer.reached = True
    return proposer
```

Then this test becomes `assert proposer.reached` plus `assert not proposer._dropped_rows.any()`, and the give up test becomes `assert not proposer.reached` plus the existing `.all()`. Both sides covered, and the `spanning` mutant fails.

rebel-kblee: It seems best to amend this section to ensure that the crux of the change is verified. I’ve checked it!

rebel-cjlee (approved):

PR #1091, opened 2026-09-11
model(optimum): Support for Qwen3-ASR

rebel-eunji:
### 🚀 Summary of Changes

It enables the model
- https://huggingface.co/Qwen/Qwen3-ASR-1.7B-hf
- https://huggingface.co/Qwen/Qwen3-ASR-0.6B-hf

## Notes
1. Only hf-format Qwen3 ASR models are supported.
2. This PR includes fix about hf_overrides on optimum-rbln as vllm overwrites hf_config with its custom hf_config. So just passing hf_config raises an error in optimum-rbln. @rebel-thkim 

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [ ] 🛠 Bug fix (`fix`)
* [ ] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

1. Run `...`
4. Verify output: `...`
5. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #1132, opened 2026-09-17
fix(optimum): fix num_hidden_layers in hf_overrides for CI 

rebel-eunji:
## Summary

Restore config-object forwarding to optimum-rbln for config classes defined in `transformers`. This preserves the configuration prepared by vLLM, including the reduced layer count set through `hf_overrides` for CI.

Config classes defined outside `transformers` continue to use config kwargs for compatibility.

## Background

#1091 replaced direct `hf_config` forwarding with a limited set of config kwargs: `num_hidden_layers` and `layer_types`.

In this path, optimum-rbln reloads the checkpoint configuration and applies the kwargs. This does not always preserve the configuration prepared by vLLM, causing two issues when reducing the layer count:

- **Gemma4 fails during config validation.** The checkpoint's `per_layer_config` retains entries for layers beyond the reduced count:
  ```text
  ValueError: `per_layer_config` keys must be integer layer indices in the range [0, 12); got [17, 23, 29].
  ```
- **GPT-OSS silently compiles at full depth.** `RBLNGptOssForCausalLM.get_pytorch_model` replaces the supplied kwargs with the unused kwargs returned by `AutoConfig`, so the layer-count override does not reach the model.

Adding more config kwargs would not resolve both issues: Gemma4 would require manually reconstructing config state, while GPT-OSS would still discard the overrides.

## Changes

During compilation on a cache miss:

- **For config classes defined in `transformers`**, pass `hf_config` directly to `from_pretrained`. This preserves the existing config state for models such as Gemma4 and GPT-OSS.
- **For config classes defined outside `transformers`**, retain the kwargs-based fallback. This avoids passing potentially incompatible config objects, such as vLLM's `qwen3_asr` and `qwen3_5` configs.
  - Forward `num_hidden_layers` and, when present, `layer_types`.
  - Nest these overrides under `text_config` for composite configs.

### Unit tests

```bash
uv run --no-sync pytest tests/optimum/v1/models/optimum/test_model_base_init_model.py
```

The tests verify that:
- A transformers config is forwarded as the original object.
- A flat non-transformers config uses top-level config kwargs.
- A composite non-transformers config uses kwargs nested under `text_config`.

### Model validation

Use a cold compilation cache and `hf_overrides={"num_hidden_layers": 12}`:

- **Gemma4:** Verify that compilation succeeds without a `per_layer_config` validation error.
- **GPT-OSS:** Verify that the compiled model uses the reduced layer count.

## Known Limitation

The GPT-OSS kwargs-handling issue in optimum-rbln is not fixed here. Direct config forwarding avoids it for transformers-defined configs, but it can still affect the fallback path.

The existing `FIXME(optimum-rbln)` remains because the underlying fix belongs in `RBLNGptOssForCausalLM.get_pytorch_model`.

## Related PR

- Related to #1091

## Type of Change

- [x] Bug fix

## Checklist

- [x] The test method and expected results are described


rebel-thkim (approved):
👍

rebel-jonghewk (approved):

PR #1103, opened 2026-09-14
perf(sampler): draw the random bonus token in the rejection graph

rebel-eunik:
### 🚀 Summary of Changes

Follow-up to #1077, which moved the all-greedy bonus token into the rejection-sample graph.

#### Performance

**The random step's bonus token is drawn inside the rejection graph.**
Unless logprobs are requested or an argmax-invariant processor (min-p) is active, the graph receives the bonus rows' logits and a per-row temperature and draws with `rbln::top_k_top_p` under the same per-request `top_k`/`top_p` as the rejection op. A greedy row of a mixed batch carries `top_k == 1` and still argmaxes. The separate bonus sampler launch is gone for these steps.

**Logits rows are gathered with `torch.index_select`, and the postprocess gather becomes a view when the sampled rows lead the batch.**
`index_select` is the cheaper gather on the device. Decode+spec pads every request to K+1 slots, so when every request but the last carries the full draft count the sampled rows are a prefix of the logits and no gather is needed.

#### Bug fixes

- Warm-up covers every rejection graph a step can ask for.
- The rejection graph gets its own `top_k`/`top_p` buffers.
  the sampler read sampling metadata's tensors stale and emitted token id 0 for every row but the first under device tensors with random sampling, spec decode and `max_num_seqs > 1`. The impl now copies `top_k`/`top_p` into its own buffers

---

### 📌 Related Issues / Tickets

* Related to #1077
* Related to [fsw-efficiency#27](https://github.com/rebellions-sw/fsw-efficiency/issues/27)

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [ ] 🧠 Model support (`model`)
* [ ] 🧬 Core engine changes (`core`)
* [x] 🛠 Bug fix (`fix`)
* [x] ⚙️ Performance improvement (`perf`)
* [ ] 🔁 Refactor or code cleanup (`refactor`)
* [ ] 📄 Documentation (`docs`)
* [ ] ❓ Other (`other`): please describe

---

### 🧪 How to Test

See [fsw-efficiency#27's comment](https://github.com/rebellions-sw/fsw-efficiency/issues/27#issuecomment-5659990031). 
When using n-gram (k=3), ITL improved by approximately 7-10% for Llama-3.2-1B and about 2% for MiniMax-M2.7 DP4EP

---

### 📋 Checklist

* [x] PR title follows Conventional Commits format
* [x] This PR is linked to an existing issue
[...]

PR #1134, opened 2026-09-18
feature(sampler): support rbln sampler synthetic mode

rebel-eunik:
### 🚀 Summary of Changes

Follow-up to #1103. Lets `rejection_sample_method="synthetic"` run under `VLLM_RBLN_SAMPLER=1`, so the sampler's cost can be measured at a chosen acceptance rate instead of whatever the target model happens to accept.

#### Feature

- **Synthetic acceptance is decided inside the rejection graph.**
  `RBLNRejectionSampler` used to assert against the mode. It now lets `rbln::rejection_sample` run unchanged — that is the cost being measured — and overrides only the accepted count, which a host `torch.rand` draw produces each step and the compiled graph takes as one more input.

- **The rates are settled once at construction.**
  The draw runs on the host, so the rates are moved there — leaving them on the sampler's device would cost a D2H every step — and padded to the width the graph expects.

- **Warm-up asks for the variant the real steps will ask for.**
  Dynamo specializes on that extra input, so warming up with the mode off would leave the first real step to recompile.

---

* Related to #1103

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [ ] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Unit tests cover the graph path end to end: the host draw beating the count `rbln::rejection_sample` returned, the boundary token, the first rejection ending the run, the cap against the drafted count, the rates landing on the host, and the warm-up forwarding the setting.

```
pytest tests/torch_compile/unit/v1/sample/test_rbln_rejection_sampler.py \
       tests/torch_compile/unit/v1/sample/test_torch_rejection_sampler.py \
       tests/vllm/v1/worker/test_rbln_model_runner.py
```

On device, this script asks for a rate schedule the target model's own acceptance would not produce, so the measured per-position rates have to follow the configuration rather than the target. 

```bash
VLLM_RBLN_USE_VLLM_MODEL=1 VLLM_RBLN_COMPILE_STRICT_MODE=1 \
VLLM_DISABLE_COMPILE_CACHE=1 python check_synthetic_eagle3.py
```

```python
from vllm import LLM, SamplingParams
from vllm.v1.metrics.reader import Counter, Vector

NUM_SPEC_TOKENS = 3
# Unconditional per-position acceptance rates, non-increasing.
RATES = [0.9, 0.5, 0.2]

llm = LLM(
    model="meta-llama/Llama-3.1-8B-Instruct",
    max_model_len=16384,
    block_size=4096,
    enable_chunked_prefill=True,
    max_num_batched_tokens=512,
    max_num_seqs=1,
    additional_config={"use_custom_sampler": True},
    speculative_config={
        "method": "eagle3",
        "model": "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B",
        "num_speculative_tokens": NUM_SPEC_TOKENS,
        "rejection_sample_method": "synthetic",
        "synthetic_acceptance_rates": RATES,
    },
    disable_log_stats=False,
)
llm.generate(
    ["A robot may not injure a human being", "The capital of France is"],
    SamplingParams(temperature=0.1, top_p=0.9, max_tokens=2048, ignore_eos=True),
)

drafts, per_pos = 0, [0] * NUM_SPEC_TOKENS
for m in llm.get_metrics():
    if m.name == "vllm:spec_decode_num_drafts":
        assert isinstance(m, Counter)
        drafts += m.value
    elif m.name == "vllm:spec_decode_num_accepted_tokens_per_pos":
        assert isinstance(m, Vector)
        for i in range(min(len(m.values), NUM_SPEC_TOKENS)):
            per_pos[i] += m.values[i]

print(f"configured rates: {RATES}")
print(f"measured  rates: {[round(c / drafts, 3) for c in per_pos]}  (drafts={drafts})")
```

Measured on this branch:

```
configured rates: [0.9, 0.5, 0.2]
measured  rates: [0.9, 0.511, 0.206]  (drafts=1565)
```

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-cjlee (approved):

rebel-jonghewk:
@rebel-eunji @rebel-daeyang can you take alook?

vllm_rbln/v1/sample/rbln_rejection_sampler.py
rebel-daeyang, line 408: The Torch impl rejects `>= vocab_size` ids, in synthetic mode ([L1022](https://github.com/RBLN-SW/vllm-rbln/blob/dfab4c5f4eaf264e59f02321a841049fe4a8b1d9/vllm_rbln/v1/sample/rbln_rejection_sampler.py#L1022), [L1052](https://github.com/RBLN-SW/vllm-rbln/blob/dfab4c5f4eaf264e59f02321a841049fe4a8b1d9/vllm_rbln/v1/sample/rbln_rejection_sampler.py#L1052)), for unmapped EAGLE3 `d2t` entries. Here a synthetic accept can return out-of-range id, so should this assert check the upper bound as well?

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -69,12 +73,12 @@ def __init__(
...
         )
 
-        if use_rbln_sampler:
-            assert not self.synthetic_mode, (
-                "RBLNRejectionSampler does not support synthetic rejection "
-                "sampling (rejection_sample_method='synthetic'). Use "
-                "`VLLM_RBLN_SAMPLER=0` for this mode."
-            )
+        # NOTE(RBLN): the host draws every step; settle the rates there, padded to K.
+        if self.synthetic_conditional_rates is not None:
```
rebel-daeyang, line 77: nit: the "host draws every step" rationale is NPU-impl-only, but this also moves the Torch impl's rates off-device. Do we need to put it under `if use_rbln_sampler:`?

vllm_rbln/v1/sample/rbln_rejection_sampler.py
```diff
@@ -69,12 +73,12 @@ def __init__(
...
+            rates = torch.zeros(num_spec_tokens, dtype=torch.float32)
+            width = min(num_spec_tokens, self.synthetic_conditional_rates.shape[0])
```
rebel-daeyang, line 79: vLLM already enforces `len(rates) == num_speculative_tokens` ([speculative.py#L264](https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/config/speculative.py#L264)), so the `min(...)` + zero-pad looks unreachable — would `.cpu()` plus a length assert be simpler? `test_npu_impl_keeps_the_synthetic_rates_on_the_host` would then need its 2-rate / 3-spec config lined up as well.

PR #1136, opened 2026-09-18
feature(worker): emit a structured fatal event and fail fast on startup errors

rebel-scottlee:
### 🚀 Summary of Changes

#1094 made a worker whose model step raises end its own process with exit code 70, so the engine fails instead of hanging. We then ran that on a Kubernetes stack with a log collector behind it and tried to answer the question an operator actually has after a restart: which rank failed first, where, and with what. Two gaps showed up, and this PR closes both.

**The exit left nothing a pipeline can index.** `abort_worker()` wrote a prose line and a traceback. In a container that line sits between the other ranks' output, the collector has to find it by matching text, and the container's own exit code is the API server's, not the worker's 70, so nothing outside the log says a worker died at all. The worker now writes one JSON line to fd 2 right before the exit:

```json
{"event":"rbln.worker.fatal","schema_version":1,"ts":"2026-09-18T07:13:54.370Z","source":"vllm-rbln","pid":3289293,"event_id":"1a0ec39f18ec41558247335eb934230d","where":"RBLNWorker.compile_or_warm_up_model","exception_type":"RuntimeError","exception_message":"warm-up failed for the compiled decode graph","exit_code":70,"rank":1,"dp_rank":0,"cause":"unknown"}
```

It goes straight to the descriptor, not through `logging`. Both this line and the readable `logger.error` that follows it are written from a helper thread, and the exit waits for that thread only briefly: in a container stderr is a pipe, and a reader that has stalled would otherwise hold `os._exit` hostage. A stalled reader or a blocked handler lock therefore costs the record, not the exit; two tests cover that, one of them against a genuinely full stderr pipe. The exception message is truncated, `rank` and `dp_rank` come from the worker and are `null` when the receiver has none, and `event_id` lets a collector deduplicate a line it has seen twice.

`cause` is always `"unknown"`, on purpose. The worker sees an exception; whether a device, the storage or the code is at fault is for the layers below the worker to say. The field exists so that collectors can keep one schema across sources, and it must not be read as a verdict. The readable `logger.error` line stays as it was.

The decorator reads `rank` and `dp_rank` from its receiver, which for `AsyncRBLNModelRunnerOutput.get_output` is the output object rather than the worker, so an output-thread failure was going to be recorded with both ranks `null`. The runner now takes the worker's rank at construction and hands both ranks to every output it builds.

One related change in `vllm_rbln/logger.py`: the format now honours `VLLM_LOGGING_PREFIX` the way `vllm.logger` does. A launcher that sets a per-rank prefix was getting tagged vLLM lines and untagged vllm-rbln lines from the same process.

**Startup was outside the fail-fast policy.** The guard from #1094 wrapped the model step only. `determine_available_memory`, `get_kv_cache_spec`, `initialize_from_config` and `compile_or_warm_up_model` reach the worker through `collective_rpc` with no timeout, and the executor collects the replies in rank order. When a rank other than the first raised during warm-up, its busy loop logged the exception and answered FAILURE, but the executor was still waiting on the first rank, which was blocked in a collective with the one that had bailed out. The container sat there until the startup probe ran out, and the log had no line saying which rank had failed. `init_device` and `load_model` run inside `WorkerProc.__init__` and already end the process through `worker_main`; they are decorated too so every startup step reports the same way.

The second commit applies `@worker_fail_fast` to those six methods. The policy itself is unchanged: a worker under the `MultiprocExecutor` family exits with 70 and leaves the event, in-process execution re-raises, and `VLLM_RBLN_DISABLE_WORKER_FAIL_FAST=1` still turns it off. With the change, a rank that fails in warm-up ends within seconds of the exception and the whole container ends with it.

The consumer side of this (collector configuration, dashboards, alert rules on the event) lives in our deployment repository, not here.

---

* Builds on #1094, merged as `a2088f6c`.
* Related to rebellions-sw/icp-serving#595 *(internal tracking issue; cross-repo, so GitHub links it but does not auto-close it)*

> There is no issue for this in this repository; the requirement is tracked in the internal issue above.

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit tests, no NPU needed with `--device-tensor 0`:
   * `pytest -q --device-tensor 0 tests/vllm/v1/worker/test_utils.py -k "fatal or event_write or blocked_record or stalled_stderr"`. One JSON line naming the failure site, unknown ranks as `null`, a write failure that still exits with 70, a record that blocks forever and still lets the exit through, and a subprocess whose stderr pipe is full when it aborts and still leaves with 70.
   * `pytest -q --device-tensor 0 tests/vllm/v1/worker/test_async_output.py -k names_its_rank` and `tests/vllm/v1/worker/test_rbln_model_runner.py -k fail_fast`. An output-thread failure names the rank and dp rank the output was built with, including an output built by the real `RBLNModelRunner.sample_tokens`.
   * `pytest -q --device-tensor 0 tests/vllm/v1/worker/test_rbln_worker.py -k startup_failure`. The six startup methods under `mp`, `mp` with the toggle set, and `uni`. The `mp` cases exit with 70 and leave exactly one event whose `where` names the method; the toggle and `uni` cases re-raise.
   * With `vllm_rbln/` reset to `dev` and the tests kept, the six `mp` startup cases, the three event tests and the rank tests fail, and the full-pipe subprocess test hangs until its timeout. The toggle and in-process cases pass on both trees, which is what they are there to guarantee.
2. End to end, a multi-rank layout under the mp executor (we used PP): SIGKILL one worker process while requests are in flight. The killed process leaves no event, since it never runs Python again. Each surviving rank that was in a collective with it raises, writes one `rbln.worker.fatal` line with its own `rank`, and exits with 70; the engine dies and in-flight requests get 500.
3. Same layout: raise from one rank's `compile_or_warm_up_model` (a one-line patch in the container is enough). Before this PR the container stays up with no indication of which rank failed until the startup probe gives up. With it, that rank writes the event and exits within seconds, and the container ends.
4. Set `VLLM_LOGGING_PREFIX="[rank1] "` on a worker: vllm-rbln's own lines carry the prefix like vLLM's do.

---

The line above is what the collector sees, one per failing worker process. Both cluster runs in step 2 and 3 produced exactly that shape, followed by the unchanged `RBLN worker <pid>: RBLNWorker.<method> raised ...` line and vLLM's `Worker proc VllmWorker-N died unexpectedly, shutting down executor`.

---

### 📋 Checklist

---

### 💬 Notes

* `schema_version` is 1 and moves only when a field is renamed or changes meaning. Adding a field does not bump it.
* Two commits, meant to be read in order: `abort_worker()` and `_fatal_event()` in `vllm_rbln/v1/worker/utils.py` first, then the decorator placements in `rbln_worker.py`.
* Nothing here interprets the exception. If a lower layer later gives us a signal that names a device, that would be a separate event with its own `cause`, not a change to this one.


vllm_rbln/v1/worker/utils.py
```diff
@@ -56,26 +61,103 @@
...
+# version moves only when a field is renamed or changes meaning.
+_FATAL_EVENT = "rbln.worker.fatal"
+_EVENT_SCHEMA_VERSION = 1
+_EVENT_MESSAGE_LIMIT = 500
+# How long the exit waits for the fatal record. stderr is a pipe whose reader
+# may have stalled; past this the record is dropped, not the exit.
+_FATAL_RECORD_TIMEOUT_S = 1.0
+
+
+def _write_event_line(line: str) -> None:
```
rebel-cjlee, line 74: `_write_event_line` writes to fd 2. The readable log line goes through the logger, and the logger stream is `VLLM_LOGGING_STREAM`, which defaults to stdout. So the two really do split. I ran it:
```
stderr: {"event":"rbln.worker.fatal",...,"where":"probe",...}
stdout: [vllm-rbln] ERROR ... probe raised RuntimeError: boom ...
```

The comment at `utils.py:64` says this is for "pipelines that index container logs". Such a pipeline gets the event on one stream and the traceback on the other, with no ordering between them. Can we write the event through the same stream as the logger? If fd 2 is required, could you put the reason in the comment?

vllm_rbln/v1/worker/utils.py
```diff
@@ -56,26 +61,103 @@
...
+    The structured event and the readable log line are written from a helper
+    thread; the exit waits for them at most _FATAL_RECORD_TIMEOUT_S. A blocked
+    stderr pipe or handler lock therefore costs the record, not the exit.
     """
+
+    def record() -> None:
+        with contextlib.suppress(Exception):
+            _write_event_line(
+                _fatal_event(exc, where=where, rank=rank, dp_rank=dp_rank)
+            )
```
rebel-cjlee: If stderr stalls, we lose the log line too, even though stdout is fine. Writing the log line first would keep it.

vllm_rbln/v1/worker/rbln_worker.py
```diff
@@ -172,6 +172,7 @@ def _init_device_env(self) -> None:
             os.environ["RBLN_NPUS_PER_DEVICE"] = str(num_devices)
 
     @instrument(span_name="Init device")
+    @worker_fail_fast
```
rebel-cjlee: `@instrument` is outside and `@worker_fail_fast` is inside, so `os._exit` happens inside the span. The span never records the exception and never closes, even though `record_exception` is `True` by default.

There is a second effect. `instrument` reads `func.__code__` and `functools.wraps` does not copy `__code__`. So the span attributes `code.filepath` and `code.lineno` now point at `guarded` in `utils.py` instead of the read method:
```
dev: sync_wrapper -> init_device   (rbln_worker.py)
PR : sync_wrapper -> guarded       (utils.py)
```

Moving `@worker_fail_fast` above `@instrument` fixes both.

tests/vllm/v1/worker/test_utils.py
```diff
@@ -163,6 +168,115 @@ def broken_log(*args, **kwargs):
...
+            with pytest.raises(SystemExit) as excinfo:
+                worker_utils.abort_worker(RuntimeError("device failed"), where="step")
+        finally:
+            release.set()
+        assert excinfo.value.code == 70
+        assert exit_codes == [70]
+
+    def test_exit_is_not_held_by_a_stalled_stderr_consumer(self):
+        # A container's stderr is a pipe. Fill one, make it fd 2 and abort: the
+        # event write and the log line both land on it, and the exit must not
```
rebel-cjlee: It says "the event write the log line both land on it". With the default stream the log line goes to stdout, not to the blocked pipe. What the test really proves is that the event write is bounded. Could we reword it?

vllm_rbln/logger.py
```diff
@@ -22,7 +22,11 @@
 from vllm import envs
 from vllm.logger import _DATE_FORMAT, _use_color, init_logger
 
-_FORMAT = "[vllm-rbln] %(levelname)s %(asctime)s [%(fileinfo)s:%(lineno)d] %(message)s"
+# Same prefix hook as vllm.logger, so one env var tags every line of a process.
+_FORMAT = (
+    f"{envs.VLLM_LOGGING_PREFIX}[vllm-rbln] %(levelname)s %(asctime)s "
+    "[%(fileinfo)s:%(lineno)d] %(message)s"
```
rebel-cjlee: This is not part of the fix. `[vllm-rbln]` already says where the line comes from and I think we don't need to use `VLLM_LOGGING_PREFIX` to distinguish between vllm and ours.

vllm_rbln/v1/worker/rbln_model_runner.py
```diff
@@ -241,9 +241,12 @@ def __init__(
         self,
         vllm_config: VllmConfig,
         device: torch.device,
+        rank: int | None = None,
```
rebel-cjlee: `WorkerBase.__init__` already does `self.parallel_config.rank = rank` (`worker_base.py:81`), and the runner already holds `self.parallel_config`. The very next line at `:2027` uses `self.parallel_config.data_parallel_rank`, so the two lines(`:2026-2027`) could read the same way:
```python
rank=self.parallel_config.rank
dp_rank=self.parallel_config.data_parallel_rank,
```

That removes the constructor argument, the attribute, the worker change and the test line.

vllm_rbln/v1/worker/utils.py
rebel-cjlee, line 1: `vllm_rbln/v1/worker/utils.py:88`, `:90`, `:92`, `:96`, `:99-102`

`cause` is already "unknown", and it carries a three line comment saying readers must not read it as a cause. A field like that carries no information. `source` repeats what the `event` name already says. `event_id` has nothing to join against. `schema_version` versions a schema that has no consumer yet. `exit_code` is always 70.

The fields that actually vary are `ts`, `pid`, `where`, `exception_type`, `exception_message`, `rank` and `dp_rank`. Could we keep just those?

tests/vllm/v1/worker/test_utils.py
```diff
@@ -163,6 +168,115 @@ def broken_log(*args, **kwargs):
...
+            fail_fast=True,
+            rank=3,
+            parallel_config=SimpleNamespace(data_parallel_rank=1),
+        )
+        with pytest.raises(SystemExit):
+            step(worker)
+        (event,) = self._event_lines(capfd)
+        assert event["event"] == "rbln.worker.fatal"
+        assert event["schema_version"] == 1
+        assert event["source"] == "vllm-rbln"
```
rebel-cjlee, line 283: Some assertion test statically defined values. (line 247, 248, 253, 256, and 257)

vllm_rbln/v1/worker/utils.py
```diff
@@ -56,26 +61,103 @@
...
 
-def abort_worker(exc: Exception, *, where: str) -> NoReturn:
-    """Exit without device cleanup when logging returns or raises.
+def _worker_dp_rank(worker: Any) -> int | None:
+    config = getattr(worker, "parallel_config", None)
+    rank = getattr(config, "data_parallel_rank", None)
+    if rank is None:
+        # An async output carries the ranks it was built with, not a config.
+        rank = getattr(worker, "dp_rank", None)
+    return rank if isinstance(rank, int) else None
```
rebel-cjlee: Each one is called from a single place (`:180` and `:181`). They also carry `isinstance(rank, int)` guards, and `:115-117` sniffs the receiver type to tell an async output from a worker. There are only two receivers and we know both shapes, so two `getattr` lines inside the decorator would do. `AGENTS.md` asks us to inline a helper with one caller, and not to handle states that cannot happen.

vllm_rbln/v1/worker/utils.py
```diff
@@ -56,26 +61,103 @@
...
+    thread; the exit waits for them at most _FATAL_RECORD_TIMEOUT_S. A blocked
+    stderr pipe or handler lock therefore costs the record, not the exit.
     """
+
+    def record() -> None:
+        with contextlib.suppress(Exception):
+            _write_event_line(
+                _fatal_event(exc, where=where, rank=rank, dp_rank=dp_rank)
+            )
+        with contextlib.suppress(Exception):
```
rebel-cjlee: `AGENTS.md` says not to catch `Exception` to get past a problem. The write one could be `OSError`. The logger one is close to unnecessary, because `logging` already swallows handler errors on its own, and the old `try/finally` gave the same guarantee.

tests/vllm/v1/worker/test_rbln_worker.py
rebel-cjlee, line 1: In line 215-216 and line 219-228.
The `mp/uni` and `disabled` matrix is already covered for the same decorator by `test_step_failure_uses_executor_policy.py` at `:315`, and it covers more backends there. The policy itself is one line inside `worker_fail_fast`, so it does not change per method. What is new in this PR is that these six methods now carry the decorator, so `mp` alone with six phases should be enough.

Also the `if/elif` chain at `:232-260` reaches into a different internal for each of the six methods, so the test now knows how all six are implemented. If someone reorders a call inside `init_device` or renames a helper, the test breaks even though the behavior it checks did not change. The existing test at `:315` avoids this by only using methods that delegate to the runner, so it can set `worker.model_runner = SimpleNamespace(**{runner_method: step})` and be done. `get_kv_cache_spec` fits that shape directly, and `load_model` is clode to it, so at least those two could move into the existing parametrization instead of getting their own branch.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: correct, the regular log defaults to stdout while the json goes directly to stderr. I kept the direct write so the json does not depend on logging locks or formatters. I’ve documented that choice and made it clear that the two records have no guaranteed ordering.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: I reproduced this. Reversing the order would introduce the opposite problem: a blocked logger could prevent the JSON from being written. The two records now run on separate threads and share the existing deadline. Tests cover either output blocking, as well as both blocking.

vllm_rbln/v1/worker/rbln_worker.py
rebel-scottlee: Moved @worker_fail_fast outside @instrument.
The tests now check that the span records the exception and finishes before the worker exits, and that its file attribute points to the worker code. The guide also notes that finishing the span does not guarantee it has been exported before exit.

tests/vllm/v1/worker/test_utils.py
rebel-scottlee: Fixed the comment and explicitly configured the regular log to use stdout in this test. It now checks that the readable log still reaches stdout when stderr is full, as well as checking that the process exits.

vllm_rbln/logger.py
rebel-scottlee: I should have explained the purpose more clearly. The launcher sets prefixes such as [rank0] and [rank1] to identify the DP processes running in the same Pod. [vllm-rbln] identifies the library, while the prefix identifies the DP process. I kept it so we can trace the surrounding regular logs alongside the fatal event.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-scottlee: Agreed. I removed the constructor argument and the separate attribute. Both rank and dp_rank now come from self.parallel_config, and the async output test checks that the rank is preserved in the fatal event.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: I checked the consumers and removed event_id, since nothing uses it for correlation or deduplication.
I kept the remaining fields to preserve the format used by the deployment tooling. event identifies the records in queries and alerts, cause is displayed in the dashboard, and exit_code exposes the worker’s exit status separately from the container’s. schema_version and source are also part of the documented common envelope. These values are currently fixed, as you pointed out, but I’ve kept them consistent with the consumer contract for this change.

tests/vllm/v1/worker/test_utils.py
rebel-scottlee: Removed the event_id assertion along with the field. I kept the remaining assertions as a small check on the JSON actually emitted through the fatal path, since those fields are part of the format used by the dashboard and operator guide.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: Removed the helpers and the isinstance checks, and inlined the lookups in the decorator. The worker reads its DP rank from parallel_config; the async output uses the dp_rank stored when it was created.

vllm_rbln/v1/worker/utils.py
rebel-scottlee: Narrowed the JSON write handling to OSError and removed suppress(Exception) around the logger call. Recording happens in separate threads, and the main path still exits through finally, so a logging exception or stalled write cannot prevent that exit.

tests/vllm/v1/worker/test_rbln_worker.py
rebel-scottlee: Removed the repeated mp/uni/disabled matrix from the startup tests. I moved load_model and get_kv_cache_spec into the existing runner-delegation parametrization, as suggested. The remaining four startup phases are checked with mp, including span completion before exit for the instrumented methods.

vllm_rbln/logger.py
rebel-cjlee: Thanks for explaining the context.
However, given that the process identify is already clearly included in the logs (e.g., `EngineCore_DP2`, `Worker_DP2`), I don't feel that fully justifies adding `VLLM_LOGGING_PREFIX`.
Since process tracing is already handled natively by these identifiers, adding another prefix feels redundant. Is there any edge case or specific scenario I might be missing here?

rebel-scottlee: Agreed, the existing process names already identify the DP rank. I’ve removed the VLLM_LOGGING_PREFIX change from this PR. The fatal JSON carries its own pid, rank, and dp_rank, so it doesn’t depend on the extra prefix.

rebel-cjlee (approved):

PR #1138, opened 2026-09-18
feature(dynamic-kv): register the KV caches the resize allocated

rebel-ykchoi:
### 🚀 Summary of Changes

The KV cache's size is not settled at one point. Dynamic KV decides the block
count after warm-up, from the real batch and a memory snapshot, and
**reallocates the cache**. A KV connector registers its regions before that.

Two deferrals that do not know about each other. The connector already waits
for warm-up, because a KV cache's physical view exists only once the compiled
model has run; dynamic KV settles the size *after* that. So the connector
registers a placeholder the resize throws away, the memory it registered is
freed, and the cache that replaces it sits at a different address. Everything
the connector froze goes stale with it -- the block count, each region's address
and length, the transfer handles already built, and the handshake metadata
already serialized and sent to the peer.

Nothing misbehaves today because the platform refuses the combination at
start-up. The guard is there; what is missing is support.

This moves registration from *once during start-up* to *after the cache is
settled*:

- For the one mode that reallocates, the warm-up registration is skipped and
  the whole registration runs once the resize has allocated. Nothing is
  registered early, so nothing has to be unregistered.
- The connector takes its block count from what the worker last allocated for,
  rather than from the estimate it copied before the compile.
- The host-bounce buffer is sized from those same tensors, so it defers with
  them.
- The platform admits the RBLN NIXL read path instead of refusing every
  connector. The ordering is driven from the worker and is connector-agnostic,
  so one outside that set is untried rather than known broken, and keeps the
  refusal.

Every other mode keeps the start-up order untouched.

- `register_kv_caches_with_connector` -- lifted out of `initialize_kv_cache` so
  the worker can call it at either point. It rebuilds the mapping from the
  tensors the runner is holding, not from what the allocation returned, which
  is what makes a call after the resize describe the new cache rather than the
  old one.
- `defers_kv_registration` -- one predicate, read at three call sites, so they
  cannot drift on which mode reallocates.
- The block count the connector adopts in `register_kv_caches`. `__init__`
  copies the count vLLM estimated before the compile; under a resize that
  number is the estimate, not the allocation, and registration is the first
  point that can know the difference.
- The platform's refusal now names which connectors the path is open for
  instead of saying the combination cannot be used.

**Tests.** Nothing here reaches a device or a peer. The ordering cases assert
the call sequence rather than a final state: a mode that reallocates must
produce `initialize → resize → register`, and every other mode
`initialize → register → resize`. A state assertion cannot tell those apart,
which is the whole content of this change. The mode case asserts the deferral
against whether the shrink returned a different config object, so the two facts
are read independently rather than one restating the other.

* Resolves rebellions-sw/icp-serving#601

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

### 🧪 How to Test

1. `pytest tests/vllm/v1/worker/test_rbln_worker.py tests/vllm/v1/worker/test_dynamic_kv_sizer.py`
2. `pytest tests/vllm/v1/worker/test_rbln_model_runner_kv_cache.py tests/vllm/test_platform.py`
3. `pytest tests/vllm/distributed/kv_connector/`
4. Edge cases pinned by the new cases: a mode that only reports and does not
   reallocate keeps the start-up order; a connector outside the opened set is
   still refused, and the refusal says which ones are open.

### 💬 Notes

- Which mode reallocates is a property on the sizer rather than a test repeated
  at each call site. The worker asks the same question in three places, and a
  fourth reader would otherwise have to rediscover that only one mode replaces
  the tensors.
- The shared worker fixture now sets the allocated block count on the config
  before building the connector. `initialize_kv_cache` settles that value
  before the worker registers, so a fixture leaving it unset was building a
  state production cannot reach; the assertion added here is what surfaced it.
- vllm model path only. The optimum model path does not reach any of this.


vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/registration.py
```diff
@@ -528,6 +528,14 @@ def register_kv_caches(self, kv_caches: dict[str, torch.Tensor]) -> None:
                 "cannot describe. Mixed full and sliding-window attention is a "
                 "different thing and is supported."
             )
+        # `__init__` copied the count vllm estimated before the compile. A
+        # dynamic-KV resize settles the real one after warm-up and registers
+        # only then, so take what the worker last allocated for.
+        num_gpu_blocks = self.vllm_config.cache_config.num_gpu_blocks
+        assert num_gpu_blocks is not None
+        self.num_blocks = num_gpu_blocks
+        self._logical_num_blocks = num_gpu_blocks
```
rebel-jonghewk, line 537: Not a blocker, but is it okay to update only num_blocks here? self.kv_cache_config would end up stale. I don't think the current implementation reads the other fields, but if something later reads kv_cache_tensors[i].size, that value would be invalid.

rebel-jonghewk (approved):
LGTM. Thanks for taking care of this. I left one comment, but it doesn't seem like a blocker, so I've approved.

vllm_rbln/distributed/kv_transfer/kv_connector/v1/rbln_nixl/registration.py
rebel-ykchoi: Good question, and the premise is right -- I had the opposite in my head and
went to check.

**The config object is replaced, not updated in place.** `reallocate` builds
`new_cfg = kv_cache_config_at(old_cfg, n)` and hands that to the runner; the
connector keeps the object it was constructed with. So what it holds is
pre-resize.

**On "only num_blocks": those two lines are upstream's two counts, and both
have to follow the allocation.** `num_blocks` is the physical one -- upstream
multiplies it by `_physical_blocks_per_logical_kv_block` at
`nixl/base_worker.py:563` -- and `_logical_num_blocks` is the logical one it
saves off first at `:550`. They are equal on RBLN only because we pin that
ratio to 1. So this is not one number written twice; it is both of the counts
the connector reads.

**What does stay stale is narrower than the whole object.** `kv_cache_config_at`
is a `copy.copy` plus a `deepcopy` of `kv_cache_tensors` alone:

| field | the connector's copy, after a resize |
|---|---|
| `kv_cache_groups` | not stale -- the shallow copy shares the same list |
| `kv_cache_tensors[*].size` | stale |
| `num_blocks` | stale |

**Neither stale field is read.** The connector touches `self.kv_cache_config`
in five places: `num_blocks` at construction, which this change supersedes at
registration; `kv_cache_groups` in three; and `kv_cache_tensors` in exactly one
-- `registration.py:291`, which takes `len(...)`. A `deepcopy` preserves the
length, so that read is invariant under a resize.

So nothing is wrong today, which matches your reading. Keeping the object
itself in step would reach the runner and the registration path, and there is
no reader to justify that yet -- so I am recording it rather than folding it
in, to be closed by whatever change first wants a field off that object.

rebel-jonghewk:
I'll handle any additional work separately if needed, using this as the base. https://github.com/RBLN-SW/vllm-rbln/pull/1129

rebel-cjlee (approved):

rebel-jinhwan (approved):

PR #1139, opened 2026-09-21
fix(nixl): fail the reads a peer that stopped answering leaves in flight

rebel-jindol21:
### 🚀 Summary of Changes

A producer that stops answering after the handshake leaves the reads against it with nothing to end them: the transfers neither complete nor error, so `get_finished` never names those requests and the decode scheduler holds them in `WAITING_FOR_REMOTE_KVS`. The heartbeat is the one call made against that producer every step, and upstream swallows its failure with a `logger.debug`.

`_send_heartbeats` now declares the engine gone on a send failure. `_handle_dead_engine` fails that peer's in-flight reads through the same path a failed transfer takes, releases their handles, and drops the `_remote_agents` entry — without which `_ensure_handshake` skips the reconnect for the life of the process.

A read that `_ready_requests` carries into a later step is failed rather than issued when its engine is no longer in `_remote_agents`. Upstream's only teardown is TTL eviction, which cannot reach a freshly handshaken engine; declaring one gone on a heartbeat failure can, in the window between the handshake publishing the agent and the readiness callback enqueueing the read.

`_handle_failed_transfer` gains one guard: a request whose read moved no block is not reported. An empty local block list means `get_num_new_matched_tokens` returned no async load, so that request was never put in `WAITING_FOR_REMOTE_KVS` — reporting it would trip `assert req_id in self.requests` (an aborted request's release notification) or `assert RequestStatus.is_finished(req.status)` (a live full prefix hit). Upstream would also index `local_block_ids[0]` to invalidate what was read, which raises for such a request. Every failure path already converges on this method, so the handshake done-callback and `_pop_done_transfers` are covered by the same line.

**Affected modules**: Distributed / KV transfer — the `rbln_nixl` connector, decode-side read path. Both model paths use the same connector; nothing model-specific is touched.

**Scope note.** The first revision of this PR also reported notify-only reads and carried a scheduler→worker tag so the aborted ones could be excluded. @rebel-huijong and @rebel-ykchoi showed that both directions of that reporting kill the EngineCore, and the invariant behind their findings — a read with no local blocks is never one the scheduler waits on — removes the need for the feature entirely. It is gone, with the tag round trip: `ABORT_NOTIFY_ATTR`, the `request_finished` / `build_connector_meta` / `start_load_kv` / `get_finished` / `_read_blocks` overrides, `_is_notify_only`, `_settle_zero_read`, `_fail_zero_read` and two worker-side sets. `pull_scheduler.py` and `pull_worker.py` are back at the base revision. 7 files and 744 added lines became 3 files and 209.

---

* Resolves # 
* Related to ICR-47

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. `uv run --no-sync pytest tests/vllm/distributed/kv_connector/test_rbln_nixl_handshake.py -k "DeadPeer or MovedNoBlock"` — 4 tests. Restore the two changed source files to the base revision and all four fail.
2. `uv run --no-sync pytest tests/vllm/distributed` — 477 passed, 4 skipped. Full suite: 2293 passed, 26 skipped, 8 xfailed.
3. 1P1D stack, host-bounce buffer, `--max-num-seqs 1` on D and a short KV lease, one request queued behind a long local generation. SIGKILL P's process tree — children first, or the EngineCore is reparented and keeps answering — then bring P back. Run twice against the base revision with only the two source files differing:
   * `Heartbeat to engine ... failed`: never logged on base → logged once.
   * A read already in flight when the peer dies: left unreported → reported as `peer_unreachable`, and the scheduler fails the request.
   * A read still queued when the peer dies: not reported either way; it fails on the retried handshake, which happens only because the `_remote_agents` entry was dropped.
   * After P restarts: no new agent uuid and the next request fails instantly against the dead one → a new uuid appears and the request is served.
   * An abort notification's leftover `_recving_metadata` entry reaching the dead-peer path is logged and dropped, not named to the scheduler. The EngineCore survives; without the guard this is the `assert req_id in self.requests` death.

---

### 💬 Notes

**What this does not fix.** The CI hang the ticket reports is not resolved here. Its cause is separate: after the connection dies, a transfer handle that was already posted neither completes nor errors, so `_pop_done_transfers` never calls `_handle_failed_transfer` at all. That traces to nixl-rbln's failover semantics — it logs `upper layer must re-load to recover`, and the connector has nowhere to receive that. A separate ticket is proposed for it; the evidence is in the ICR-47 comment.

**Withdrawn from the first revision's description.** It claimed the queued request ends much sooner with the change. That does not reproduce: on the base revision the request ends in comparable time, through the transfer-setup failure that already existed. The reproducible difference is detection and reconnect, not latency.

**A leak this does not close.** `_recving_metadata` is stored for every read and popped only for the ones `get_finished` names, so a read that moves no block keeps its entry for the life of the process — and is reconsidered on every later dead-peer sweep. An earlier revision dropped the entry in the guard; that had to come out, because the handshake done-callback runs on the executor thread and consuming the entry leaves whichever caller arrives second with no block list to judge by, which sends a blockless read to upstream's report. Closing the leak safely means changing what a missing entry means for every caller of `_handle_failed_transfer`, which belongs in its own change.


rebel-jindol21:
@rebel-cjlee Could you go through this PR?

rebel-huijong (commented):
On a full local prefix hit (e.g., same prompt sent twice) the request never enters `WAITING_FOR_REMOTE_KVS` (`count == 0`), yet the empty read still gets `_settle_zero_read`-ed and returned as `done_recving` in `get_finished `. [Then `_update_from_kv_xfer_finished` hits a RUNNING request and `assert RequestStatus.is_finished(req.status)` would kill the engine.](https://github.com/vllm-project/vllm/blob/568afb3a13806beb53bb2e6bd518269357b237c0/vllm/v1/core/sched/scheduler.py#L2644) Could you check?

rebel-ykchoi (commented):
**Verdict: needs changes before merge.** Two defects reach the exact failure this PR set out to prevent — a request named to the scheduler that it no longer holds, tripping `assert req_id in self.requests` and killing the EngineCore. Both come from one root cause. The rest is scope and hygiene.

## 1. Blocking — the abort tag's lifetime

`_abort_notify_reqs` is built as a *per-step* set, but the reads it has to protect outlive a step. That mismatch leaks in two places.

### 1.1 `_handle_dead_engine` reports the abort notifications — `handshake.py:410-445`

The exclusion exists in `_settle_zero_read` and `_fail_zero_read`, but not here. This method walks `_recving_metadata` filtered only by engine and reports every entry:

1. an abort-only read has a `_recving_metadata` entry — upstream `pull_worker.py:66` stores it unconditionally (`# always store metadata for failure recovery`)
2. `_report_failed_recv` → `_failed_recv_reqs.put(req_id)`
3. upstream `get_finished` drains that queue and does `done_recving.update(failed_recv_reqs)` (`base_worker.py:1958-1969`)
4. `scheduler.py:2640` — `assert req_id in self.requests`

Reproduce (the `TestADeadPeerIsReported` helpers build the worker):

```python
w._abort_notify_reqs = {"r_abort"}
w._recving_metadata = {"r_abort": _meta(local_ids=())}
w.nixl_wrapper.send_notif.side_effect = RuntimeError("conn is gone")
w._send_heartbeats(_heartbeat({"r_abort"}))
assert list(w._failed_recv_reqs.queue) == []   # fails: ['r_abort']
```

### 1.2 A read deferred by a pending handshake loses its tag — `pull_worker.py:67-70`

`start_load_kv` replaces the whole set each step. Upstream defers a read when the peer has not been handshaken yet (`pull_worker.py:60-67`: `_background_nixl_handshake(...)` then `continue`) and runs it from `_ready_requests` at the end of a **later** `start_load_kv`. By then the set holds that step's tags, so `_settle_zero_read` no longer skips it and the request is reported — same assert.

This is not a rare path. A first contact with a producer always defers (the handshake is a zmq round trip), and a request the D-side serving layer rejects is typically the first thing said to that producer. It also opens right after 1.1 drops `_remote_agents`.

Reproduce: queue the read while `_remote_agents` is empty, fill `_remote_agents`, push the entry onto `_ready_requests`, then run `start_load_kv` with empty metadata and `get_finished` — the request comes back in `done_recving`.

**Both fall out of one choke point.** `scheduler.py:2634` calls `update_connector_output` immediately before the loop that asserts, and `finished_recving` is a mutable `set[str] | None` (`v1/outputs.py:199`). Dropping the abort ids there makes the worker's report irrelevant and removes `ABORT_NOTIFY_ATTR`, the `build_connector_meta` override, the `start_load_kv` override and the worker-side set. Whether to take that shape is yours; the constraint is that the tag has to live as long as the read, not as long as a step.

## 2. Correctness — non-blocking

### 2.1 A fourth place still goes unreported — `state.py:95`

Upstream's handshake done-callback calls `_handle_failed_transfer(req_id, None)` directly (`base_worker.py:919-926`). A notify-only read has `local_block_ids == []` (`BlockIds` is a list per KV cache group, and the abort path queues `(request, [])`), so `local_block_ids[0]` raises `IndexError` — the very reason `_report_failed_recv` exists, but it is not used there. The commit message says "Three places". Inside a `Future` callback the exception is swallowed, which leaves the request unreported: the symptom this PR fixes elsewhere.

### 2.2 `_recving_metadata` leaks one entry per abort — `pull_worker.py:169-172`

Entries are added for every request in `reqs_to_recv` and removed only for `done_recving` (upstream) and `settled` (here). An abort-only read is in neither, so its entry stays for the life of the process. This is what turns 1.1 from "one request" into "everything that engine ever aborted".

### 2.3 The stated reason for `_failed_zero_reads` does not hold — `pull_worker.py:58-60`

The comment says a second report would "settle some later request early" because the aggregator counts one report per worker. Upstream drains `_failed_recv_reqs` **into a set** (`base_worker.py:1959-1963`), so duplicates within a step already collapse; and the field is cleared every `get_finished`, so it cannot stop a duplicate across steps either. What remains is avoiding a double `record_failed_transfer()`. Either fix the comment or drop the field.

## 3. Tests

**3.1 Nothing checks that the tag reaches the worker.** Replacing the `getattr` in `start_load_kv` with an empty set leaves every test green. The worker tests set `_abort_notify_reqs` by hand and the scheduler test stops at the `setattr`. The whole guard against killing the engine rests on that one line.

**3.2 7 of the 21 new tests pass without the change.** Restore only the four production files to the base revision (`git show <base>:<path>`) and run the three files. Two of the seven do not catch any single-line mutation of the new code — `test_a_normal_finish_is_not_tagged` and `test_a_live_heartbeat_changes_nothing` — so they assert nothing this PR introduced.

**3.3 A stub drops the line that makes the ordering matter** — `test_rbln_nixl_base_scheduler.py:479-482`. Upstream sets `params["do_remote_prefill"] = False` immediately after queueing the read (`pull_scheduler.py:221`). That is precisely why `notify_only` must be captured **before** `super()`. The stub omits it, so moving the capture after `super()` keeps the suite green. Adding that one line to the stub keeps everything passing and makes the ordering test discriminate.

**3.4 Untested branches.** `_is_notify_only`'s `block_size_ratio > 1` branch is never reached (every test fixture returns 1), and removing the `_recving_metadata.pop` in `get_finished` changes no result. The guard in `test_a_stage_that_posts_a_transfer_keeps_the_request` is not actually exercised: the fixture returns `"DONE"`, so upstream reports the request in the same step whether the guard is there or not — `"PROC"` would discriminate.

## 4. Scope

The feature itself — report a read that moves no block — is `_settle_zero_read`, the `get_finished` override and one branch in `_read_blocks_for_req`. What sits on top of that is where the defects came from:

- the scheduler→worker tag round trip (§1)
- `_is_notify_only` (`pull_worker.py:134`) re-derives a decision upstream makes internally. Upstream clips the mapped local list **before** deciding; this reads the unclipped list. Where they diverge, a failed read is reported as complete — the bug this PR fixes. Upstream already calls `xfer_stats.record_failed_notification()`, so observing that counter across the delegated call would remove both this helper and the re-sent notification block.
- `_failed_zero_reads` (§2.3)
- guards for states that cannot occur, which `AGENTS.md` §Failure handling rules out: `pull_scheduler.py:65` (upstream always populates `_reqs_need_recv` on that branch), `pull_worker.py:121` and `:169`, `handshake.py:447` (the only caller indexes `_remote_agents[engine_id]` just before)
- defaults that cannot fire: `getattr(metadata, ABORT_NOTIFY_ATTR, ()) or ()` and `remote_rank: int | None = None` (`pull_worker.py:100`), both always supplied

**Splitting**: the notify-only reporting and the abort exclusion cannot be separated — merging the first alone kills the engine, and the second is its guard. The heartbeat work is independent and shares only `_report_failed_recv`; today it is missing that coupling, which is §1.1.

## 5. Repository rules

- **`AGENTS.md` §What not to publish** — the PR description and the commit body carry absolute latency figures and a host name. A relative change is fine; the numbers are not.
- **§Failure handling** — `handshake.py:439` catches `Exception` and continues after a `logger.debug`, which is the swallowed failure the rule names. (The one at `:451` is different: an invariant in the comment explains why the entry must go regardless.)
- **§Comments** — `_settle_zero_read`'s docstring is 16 lines against the 15-line limit. Its third paragraph repeats what `_fail_zero_read` and `request_finished` already say; moving it out brings the block under the limit without splitting it.
- **§Comments, one place per fact** — the ticket id appears 13 times across source and tests, and the "`get_finished` names only two structures" explanation lives in four places in `pull_worker.py` plus the commit and the description.
- **§Commits and PRs** — the commit body is six paragraphs and five of them restate the description. The squash merge discards the body, so that explanation belongs only in the description.
- `state.py:105` — `meta is not None` is not needed: upstream's `if (meta := self._recving_metadata.get(req_id)) and ...` already handles a missing entry, and that condition is what forces the else branch to copy two upstream lines. Narrowing it to the empty-`local_block_ids` case removes the copy.

## 6. PR description

- None of the template sections are filled in (`gh pr create --body` does not apply the template).
- No linked GitHub issue and no label; `CONTRIBUTING.md` requires both.
- The description does not say which model path is affected (`AGENTS.md` §Two model paths).
- It reports `tests/native` — the tests in this PR are under `tests/vllm/`.

## 7. Questions

- `_send_heartbeats` (`handshake.py:462`) is close to a verbatim copy of upstream's, differing in one `list(...)` and the `except` body. Can the mirrored version be pinned in a line, so a later upstream change is visible rather than silent?
- `_send_heartbeats` and `_handle_dead_engine` live on the shared mixin, so the push worker gets them too (`push_worker.py:192`). There, `_recving_transfers[req_id]` is a completion sentinel rather than handles, so `_handle_dead_engine`'s "in-flight handles are released" does not hold. Is that path intended, and should it have a test?
- When the teardown itself fails, only `_remote_agents` is dropped; `src_xfer_handles_by_remote`, `transfer_topo` and the rest survive, and `register_remote_engine` reuses an existing entry. Would a restarted producer then be read with its old geometry?
- Can the exception set `send_notif` raises be named, so the two `except Exception` sites can narrow to it?

## 8. Verification

Everything in §1, §2.1–2.3, §4 and §5 was checked against the installed upstream sources and the diff. §1.1, §1.2 and §2.1 were reproduced at unit level — the report reaching `finished_recving`, and the `IndexError` — not on a live 1P1D stack; the last step (the engine actually dying, the callback swallowing the `IndexError`) is inference from the code path. The figures in the description were not reproduced. The push path in §7 was read, not run.

rebel-jindol21:
@rebel-huijong You're right, and it turned out to be worse than the case you found — thank you.

The invariant I had missed: a read with empty `local_block_ids` is *never* one the scheduler waits on. `get_num_new_matched_tokens` returns `load_kv_async=True` only together with `count > 0`, the scheduler enters `WAITING_FOR_REMOTE_KVS` only on `load_kv_async`, and `update_state_after_alloc` produces the empty list exactly when `num_external_tokens == 0`. Our own `RblnNixlSchedulerBase.get_num_new_matched_tokens` already leans on that — it returns `(0, False)` rather than `(0, True)` precisely because "an async load of nothing trips the scheduler's own assertion". The other producer of an empty read is the abort path in `request_finished`, and that request is gone from `self.requests` entirely.

So there is no live request behind a notify-only read to report in either direction: reporting it complete hits the `assert RequestStatus.is_finished(req.status)` you linked, reporting it failed hits `assert req_id in self.requests` one line above. Upstream logging it and moving on was correct all along.

Pushed in f7b073c: the whole notify-only reporting path is gone rather than guarded, and with it the scheduler→worker abort tag it needed. `pull_scheduler.py` and `pull_worker.py` are back at the base revision. What remains is the heartbeat fix plus one guard in `_handle_failed_transfer`.

The live run confirms your case is not theoretical. A blockless read did reach the failure path on the 1P1D stack — an aborted request's release notification, `num_local_blocks: 0` — and with the guard it is logged and dropped, engine intact:

```
13:24:08  peer_unreachable  req=cmpl-a8b93805aa5fcb03-0-858c0db6   num_local_blocks: 1
          -> Failing 1 request(s) due to KV load failure            (correct, it was waiting)
13:26:09  peer_unreachable  req=a8b93805aa5fcb03                   num_local_blocks: 0
          -> no report, no assert, engine alive                     (the abort notify, dropped)
```

rebel-jindol21:
@rebel-ykchoi Thanks — §1.1 and §1.2 are both real, and chasing them showed the feature they guard should not exist. Pushed as f7b073c.

## §1, §2.3, §4 — resolved by removal

A read with no local blocks is never awaited by the scheduler: `load_kv_async` is returned only with `count > 0`, `WAITING_FOR_REMOTE_KVS` is set only on `load_kv_async`, and `update_state_after_alloc` yields the empty list exactly when `num_external_tokens == 0` — which our own `get_num_new_matched_tokens` override already depends on. @rebel-huijong found the live-request half of the same thing independently. So `_settle_zero_read` reported only requests the scheduler was not waiting on, in both directions.

The notify-only reporting is deleted rather than fixed at the choke point. Gone with it: `ABORT_NOTIFY_ATTR`, the `request_finished` / `build_connector_meta` / `start_load_kv` / `get_finished` / `_read_blocks` overrides, `_is_notify_only`, `_settle_zero_read`, `_fail_zero_read`, `_zero_read_reqs`, `_failed_zero_reads`. `pull_scheduler.py` and `pull_worker.py` are back at the base revision. That closes 1.1, 1.2, 2.3 and most of §4 at the source. You were right that `update_connector_output` was the place to put the drop — it turned out there was nothing left to drop.

## §1.1 + §2.1 + §2.2 — one fix, at `_handle_failed_transfer`

`_report_failed_recv` was wrong in its fallback: when the metadata holds no local blocks there is nothing to report, not something to queue. The helper is gone; `RblnNixlWorkerState` now overrides `_handle_failed_transfer` itself, which is where every failure path already converges — `_handle_dead_engine`, the handshake done-callback you identified in §2.1, `_pop_done_transfers`, and the transfer-setup failure. One predicate, no caller left out, and the stale abort entries are dropped as they arrive rather than accumulating (§2.2).

Your §1.1 reproduced on the live stack, as the abort notification of a request that had already been failed two minutes earlier:

```
13:24:08  peer_unreachable  req=cmpl-a8b93805aa5fcb03-0-858c0db6   num_local_blocks: 1
          -> Failing 1 request(s) due to KV load failure
13:26:09  peer_unreachable  req=a8b93805aa5fcb03                   num_local_blocks: 0
          -> no report, no assert, engine alive
```

The gap between the two lines is §2.2 measured: the entry survived from 13:24 to 13:26.

## §5 — repository rules

- **What not to publish** — the absolute latency figures and the host name are out of the description; the commit body is one paragraph and no longer restates it.
- **Failure handling** — the `except Exception` around `release_xfer_handle` at the old `:439` is gone. A release that fails on a connection we have just declared dead is the same condition, so there was nothing to get past. The one at `:451` stays, with the invariant that explains it.
- **Comments** — `_settle_zero_read`'s docstring went with the method. The ticket id now appears once, in the description rather than in the source.
- **`state.py:105`** — the copied upstream lines are gone; the override narrows to the empty-`local_block_ids` case and delegates everything else to `super()`.

## §6 — PR description

Template filled, affected module named, model paths addressed, `bug` label added (the repo has no `fix` or `core` label — the `Type of Change` boxes carry that). `tests/native` was a copy-paste error; the tests are under `tests/vllm/`. **One thing I cannot close:** no Development-related issue exists for ICR-47 in this repo's tracker, so `Resolves #` is still empty and marked as such. It needs filing before merge.

I also withdrew a claim from the first description: the queued request does *not* end sooner with this change. On the base revision it ends in comparable time, through the transfer-setup failure that already existed. The reproducible difference is detection and reconnect.

## §3 — tests redone

The suite is rewritten around the reduced scope: 4 tests, and I checked each one fails against the base revision rather than assuming it. The four that passed without the change are deleted — `test_a_live_heartbeat_changes_nothing`, `test_a_pending_handshake_is_left_alone`, `test_a_reader_of_another_peer_is_untouched` (redundant with the engine-filter test) and one that only asserted upstream's own behaviour. The `_abort_notify_reqs` plumbing your §3.1 pointed at no longer exists to test.

## §7 — answers

- **`_send_heartbeats` mirror**: pinned in a line above it — mirrors `NixlBaseConnectorWorker._send_heartbeats` as of vllm 0.26.0, differing only in the `except` body. I also dropped the `list(...)`, which the `break` made unnecessary, so the two really are one clause apart.
- **Push path**: right, and worth being explicit about. `_recving_transfers[req_id]` there is a `setdefault(req_id, [])` sentinel, so the "in-flight handles are released" half is a no-op on that side. Given that the reduced change now routes everything through `_handle_failed_transfer`, the behaviour there is upstream's own report for a read that holds blocks and a dropped entry for one that does not — which is correct for both sides. I have read that path but not run it; if you would rather it carry a test before merge, say so and I will add one.
- **Restarted producer's geometry**: I think yes. The teardown drops `_remote_agents` but leaves `src_xfer_handles_by_remote` and `transfer_topo`, and `register_remote_engine` reuses an existing entry. In the live run the request after the restart succeeded, so it did not bite there, but that is not evidence the geometry matched. It looks like a real second bug and I would rather not fold it into this PR — a separate one, unless you want it here.
- **`send_notif`'s exception type**: it raises `nixl_cu13._bindings.nixlRemoteDisconnectError` for a dead peer, but that is one of several the binding can raise for a failed notification, and catching only it would let the others through in the one place we most need to notice them. I left `Exception` at those two sites for now — tell me if you would rather narrow it and let the rest propagate.

## Verification

Unit: 4 new tests, each failing against the base revision; `tests/vllm` 2293 passed, 26 skipped, 8 xfailed; `pre-commit` clean on the changed files. Live: the 1P1D scenario twice with the change and once against the base, same parameters, with the heartbeat-failure log, the `peer_unreachable` report, the new agent uuid after the restart and the abort-notify drop above all observed. The nixl-rbln failover cause of the CI hang was read, not fixed, and is still proposed as a separate ticket.

rebel-ykchoi:
The rework closes both blockers I raised, and it closes them at the source rather than around them: the discriminator moved from a per-step tag to `meta.local_block_ids`, which is a property of the request itself and is available on every path.

- `_handle_dead_engine` now reports through `_handle_failed_transfer`, so the guard covers it — an abort notification is dropped instead of named.
- The tag round trip is gone, so the one-step lifetime it had cannot arise.
- The handshake done-callback is covered by the same line, which was the fourth unreported place I mentioned.
- The metadata entry is deleted there, closing the leak on the failure side.

Checked: the premise holds — `get_num_new_matched_tokens` returns the async load only together with a positive count (`nixl/pull_scheduler.py`), so a read with no local block is never one the scheduler waits on. Restoring only the two source files to the base revision makes all four new tests fail. Mutating the guard's condition, the metadata delete, the report call in `_handle_dead_engine`, or the `break` in `_send_heartbeats` each turns the suite red; `tests/vllm/distributed/kv_connector` is green as it stands.

Four things left, none blocking.

**1. `del` where upstream uses `pop` — `state.py:105`.** `_handle_failed_transfer` is also called from the handshake done-callback, which runs on the thread that completes the future (`base_worker.py` submits the handshake to a `ThreadPoolExecutor` and adds `request_ready` as its callback). If that path and `_handle_dead_engine` reach the same request, the second `del` raises `KeyError`, and inside a `Future` callback that is swallowed — the request then goes unreported, which is the symptom this PR fixes. Upstream uses `.get()` and `.pop(req_id, None)` at the same lines, which reads as the codebase already assuming that concurrency. I did not reproduce it; `pop(req_id, None)` closes it.

**2. The failure log claims an action that does not happen — `handshake.py:424-430`.** `_log_failure(msg="Marking blocks as invalid")` fires per request before the guard runs. For a read with no local blocks nothing is invalidated and nothing is reported. Logging it is intentional per the description; it is the message text that is off.

**3. `if engine_id not in self._remote_agents: return` — `handshake.py:447`.** The only caller indexes `self._remote_agents[engine_id]` immediately before and breaks out, so the key is always there. `AGENTS.md` §Failure handling — "Do not add error handling for states that cannot occur". Removing it leaves the suite green.

**4. `assert handle is None` — `state.py:104`.** Fine as an invariant, but it disappears under `python -O`, and then a handle reaching that branch would be dropped without being released. Removing it also leaves the suite green.

Two from the earlier round are still open: `Resolves #` in the description is still a TODO, and the new commit's subject carries no `type(scope):`.

rebel-jindol21:
@rebel-ykchoi Thanks for checking the premise independently. Three of the four are in; the fourth I would rather keep, and **§1 needed a different fix than the one suggested** — the reasoning is below, and I was wrong about the shape twice before landing it.

## 1. `del` — accepted, but `pop(req_id, None)` does not close it

The threading premise holds: `_ensure_handshake` submits to `self._handshake_initiation_executor` (a `ThreadPoolExecutor`, `base_worker.py:476`) and `request_ready` is attached with `add_done_callback`, so it runs on that thread and, unlike `done_callback`, never takes `_handshake_lock`. `_handle_dead_engine` runs on the main thread inside `start_load_kv`. Two threads, one request, no lock between them.

But `pop` only removes the `KeyError`. The worse half survives it:

```python
meta = self._recving_metadata.get(req_id)   # None -- the other thread popped it
if meta is not None and not meta.local_block_ids:   # False, so we fall through
    ...
super()._handle_failed_transfer(req_id, handle)     # -> _failed_recv_reqs.put(req_id)
```

Whoever arrives second has no `local_block_ids` to judge by, so a blockless read goes to upstream's report — `assert req_id in self.requests`, the death this PR exists to prevent. Worse than the unreported request you described.

I tried inverting it instead (report only when blocks are positively visible, treat a missing entry as already settled). That is self-consistent, but it changes what a missing entry means for *every* caller, and `TestShardReadPath::test_a_failed_stage_leaves_no_handles_behind` caught it immediately — none of the tests in that file store `_recving_metadata`, and upstream's `.get()` is exactly the tolerance they rely on. Making that test pass would have meant editing a base-revision test to suit my change, which is the wrong direction.

**So the entry is not consumed at all.** Nothing else pops it for a blockless read — `get_finished` pops only what it reports, and such a read is never reported — so the discriminator cannot vanish, and a second call gives the same answer as the first. That is the whole fix: one deleted line.

The cost is that I have to give back the credit for §2.2. The entry now leaks, as it already does on upstream's success path. Closing it requires a delete, a delete makes the discriminator racy, and making the delete safe requires the missing-entry semantics above. That is a deliberate change to a shared method and belongs in its own PR, not tucked into this one.

**One more race, from the same source, that the review did not name.** The sweep in `_handle_dead_engine` was a comprehension over the live `_recving_metadata.items()`. It runs Python between iterations, so the callback thread dropping an entry mid-sweep raises `RuntimeError: dictionary changed size during iteration` on the main thread. It now snapshots with `list(...)` — a single C-level call, so atomic under the GIL — and filters after. Covered by `TestMetadataTakenMidSweep`, which reproduces it deterministically by having the peer lookup drop another entry.

## 2. The log message — accepted

`msg="Marking blocks as invalid"` is gone. `_log_failure`'s `msg` is optional and `failure_type="peer_unreachable"` already says the true thing.

## 3. `if engine_id not in self._remote_agents: return` — accepted

Removed. The one caller indexes `self._remote_agents[engine_id]` on the line before and breaks, so the key is always there; AGENTS.md §Failure handling is explicit about that.

## 4. `assert handle is None` — I would keep it

The `-O` observation is true but I think it argues the other way. **The assert is not what releases the handle** — with it or without it, that branch returns without releasing. It documents an impossible state; removing it makes the state undocumented, not handled. AGENTS.md §Failure handling asks for exactly this: *"Do not route a case that cannot happen into an `else`. Use `assert` or `raise`."*

And the state stays impossible under the race above. A non-`None` handle reaches `_handle_failed_transfer` only from `_pop_done_transfers` and the transfer-setup failure, both of which require a read that posted a transfer, which requires local blocks. A blockless read never puts anything in `_recving_transfers`. Happy to drop it if you still disagree — it is one line either way.

## Still open

- **`Resolves #`** — unchanged and still a TODO. There is no Development-related issue for ICR-47 in this repo's tracker; it has to be filed and I did not want to open one under a guessed title. If you tell me the shape you want, or file it, I will link it.
- **Commit subject** — `CONTRIBUTING.md` exempts these: *"Individual commit messages in PR branches do not need to strictly follow Conventional Commits, but should remain readable and descriptive."* The squash merge takes the PR title, which does follow the format. Rewriting the branch to change a subject that the merge discards costs a force-push for nothing, so I would rather leave it — say the word if you disagree.

## Verification

Pushed as 8ac18d7. `tests/vllm/distributed` is green (479 passed, 4 skipped) and both new tests fail against the previous revision. One note on the full-suite number: a run mid-review reported 72 failures in the device-backed lane, which turned out to be another user's `VLLM::EngineCore` holding NPU 0 on this host — the same failures appear with my changes stashed, and the lane is green again pinned to a free device. Nothing to do with the diff, but worth saying rather than quietly re-running until it looked good.

rebel-ykchoi (approved):
Approving the code. Checked at `8ac18d75`, against the merge-base `8be2f8e0`.

The rework closed both things I raised, and it closed them at the source: the discriminator moved from a per-step tag to `meta.local_block_ids`, which is a property of the request and is available on every path. `8ac18d75` then took the `del` back out — leaving the entry is what lets whichever path arrives second still judge by the block list. That is a sharper reading than the `pop(req_id, None)` I suggested, which would have kept the `KeyError` away and still let the second path fall through to upstream's report. The log message that claimed an invalidation it never performed, and the unreachable `_remote_agents` guard, are gone with it.

What I ran:

- `tests/vllm/distributed/kv_connector` — green.
- Restoring only `handshake.py` and `state.py` to the base revision turns **all six** of this PR's tests red, so none of them pass without the change.
- Three mutations, all caught: re-adding the `del` in the guard, reading `_recving_metadata` live instead of through the snapshot, and dropping the engine filter in the sweep. The decisions this last commit rests on are pinned.

Two small things, neither blocking:

- **The description's "Left to upstream" paragraph is now stale.** It says the change drops the entry when it reaches the failure path; `8ac18d75` removed that drop. The entry now stays on both paths, and that is the cost of the guard working — worth saying it that way, since it is a deliberate trade rather than an oversight.
- `assert handle is None` in `state.py` disappears under `python -O`, and a handle reaching that branch would then be dropped without being released. Fine as an invariant; only noting it.

Two things before merge: `Resolves #` is still the TODO the description carries — the ticket needs a Development-related issue filed and linked, and the PR has no label. And the branch is still on `8be2f8e0` while `dev` has moved three times since, so it is worth rebasing and letting CI run once on the current base.

Not verified: no hardware. The 1P1D scenario in **How to Test**, the thread interleaving itself, and the push path that shares `_send_heartbeats` are read, not run.

rebel-jindol21:
Both Copilot findings hold up. Pushed as 9e2b473.

## High — race between agent publication, readiness enqueue, and teardown

Real, and introduced by this PR. The ordering is as described: `_ensure_handshake` registers `done_callback` first and `_background_nixl_handshake` registers `request_ready` second, so they run in that order on the executor thread, with the main thread free to run between them. `done_callback` publishes `_remote_agents[eid]` and releases `_handshake_lock`; a heartbeat in that gap finds the agent, fails, and `_handle_dead_engine` runs `_cleanup_remote_engine`, which pops `_remote_agents`, `dst_xfer_side_handles`, `dst_num_blocks`, `tp_mappings` and unregisters the engine from `transfer_topo`. `request_ready` then enqueues, and the next `start_load_kv` drains `_ready_requests` into `_read_blocks_for_req`.

What makes it this PR's problem rather than upstream's: upstream's only teardown is TTL eviction, and its own docstring states the invariant it relies on — *"Engines with active transfers or pending handshakes cannot be stale"*. A freshly published engine has `_engine_last_active = now`, so `_evict_stale_engines` can never open this window. Declaring an engine gone on a heartbeat failure ignores that timestamp, so it can.

I did not take the generation/dead-marker route. The window is not the only way a queued read can outlive its engine, and a counter only catches the ones it was designed for, so the guard sits at the point of use instead — `_read_blocks_for_req` fails the request when its engine is no longer in `_remote_agents`, whatever removed it and whenever. That is also the correct outcome: the peer is gone, so the read must be reported, not attempted.

Worth noting what the test showed while verifying it: without the guard the read does **not** raise. It proceeds through the released descriptors and the request is never reported — quieter than a `KeyError` and worse. `TestAReadDeferredPastTheTeardown::test_it_is_failed_rather_than_read` covers it and fails against the previous revision.

The same comment flags lines 439 and 488. Those are the teardown and the heartbeat loop themselves; with the read path guarded there is nothing left for them to hand off in a bad state, so I have not changed them.

## Medium — failed zero-block reads leak metadata entries

Correct on the facts, and the PR description was stale — it still claimed the failure path cleans the entry up. Fixed there.

The entry is left deliberately. An earlier revision of this PR did delete it, and that had to come out for the same concurrency this review is about: the handshake done-callback and the heartbeat sweep can reach one request, and consuming the entry leaves whichever arrives second with no `local_block_ids` to judge by — so a blockless read goes to upstream's report and trips `assert req_id in self.requests`, the death this guard exists to prevent.

A settled/tombstone marker trades one unbounded structure for another — a set of request ids that nothing ever prunes — so it does not close the leak, it renames it. Closing it properly means changing what a missing entry means for every caller of `_handle_failed_transfer`, which is a deliberate change to a method shared with upstream's paths and belongs in its own PR. It is written up as a known limitation instead.

## Verification

`tests/vllm/distributed` 480 passed, 4 skipped; full suite 2289 passed, 26 skipped, 8 xfailed. One disclosure: the new guard reads `_remote_agents`, which a hand-built worker in `test_single_stage_read_delegates_to_upstream` did not set — that attribute always exists in production, so I added it to the fixture; the test's own assertion is untouched.

rebel-cjlee (approved):

PR #1140, opened 2026-09-21
fix(eagle3): support a pipeline-parallel A.X-K2 target

### LINKED ISSUE #1007: axk2: EAGLE3 aux hidden states are dropped at every pipeline boundary
## What

`vllm_rbln/patches/axk2/_skt_model.py` drops EAGLE3 aux hidden states at every
pipeline boundary, and its handoff key set is fixed before the aux layers are known.
Both are latent: with `pipeline_parallel_size=1` neither is reachable. With PP>1 the
last stage — the one the drafter lives on — comes up short and fails with a
`KeyError` on an `aux_hidden_states_*` slot.

`skt/A.X-K2-EAGLE3` exists, so the combination is intended.

These are the same two defects that #1006 fixes for MiniMax-M2. axk2 is outside that
branch's blast radius, so it still has them.

## 1. A non-last stage discards what it captured

`AXK2Model.forward`:

```python
aux_hidden_states = []
for idx, layer in enumerate(islice(self.layers, self.start_layer, self.end_layer),
                            start=self.start_layer):
    if idx in self.aux_hidden_state_layers:
        aux_hidden_states.append(hidden_states + residual)
    hidden_states, residual = layer(...)

if not get_pp_group().is_last_rank:
    return IntermediateTensors({"hidden_states": hidden_states, "residual": residual})
    #                           aux_hidden_states dies here
```

The drafter runs on the last stage but the requested aux layers are spread over all
of them, so on any split at least one index is captured by a stage that then throws
it away.

The capture index itself is correct — `start=self.start_layer` makes it global, so
this model does not have the stage-local indexing bug that MiniMax-M2 had. Only the
carry is missing.

## 2. The handoff key set is fixed at construction

```python
# AXK2Model.__init__
self.make_empty_intermediate_tensors = make_empty_intermediate_tensors_factory(
    ["hidden_states", "residual"], config.hidden_size)   # no aux slots

# AXK2ForCausalLM.__init__
self.make_empty_intermediate_tensors = self.model.make_empty_intermediate_tensors
```

The runner sets the aux layers later, through `set_aux_hidden_state_layers`. By then
the outer class has copied the callable onto itself and the runner calls that copy,
so replacing the inner one afterwards is not seen. The receiving rank never allocates
the aux slots.

## The two have to be fixed together

`make_empty_intermediate_tensors` allocates the buffers the **receiving** rank waits
on. Declaring a slot that the sender never fills is a hang, not a no-op — so fixing
(2) without (1) is worse than fixing neither.

## Not attempted here

I have no A.X-K2 deployment to run at PP>1, so I did not write the fix: shipping an
unverified replacement of a vendored model's `forward` is not something to fold into
an unrelated branch. The mechanism is available — `patches/axk2/model.py` already
replaces three methods on the frozen module by direct assignment, and it imports one
replacement body from another patch module — so the shape of the fix is settled even
though the change is not.

Same-class candidates worth checking: upstream `deepseek_v2`, `kimi_k25`, `qwen3_5`,
`mllama4` and `pixtral` all implement `set_aux_hidden_state_layers`.

Removal condition for all of it: `vllm-project/vllm#50514`.

rebel-chanheo:
### 🚀 Summary of Changes

`skt/A.X-K2` + `skt/A.X-K2-EAGLE3` at `pipeline_parallel_size > 1` was rejected at config validation, so the deployment never started. This opens it. **vllm model path only.**

| # | File | Change |
|---|---|---|
| 1 | `patches/axk2/model.py` | Replace `AXK2Model.forward` so aux hidden states survive the pipeline handoff |
| 2 | `patches/deepseek_eagle3.py` (new) | Name the EAGLE head's layers past the target's full depth, not the per-rank count |
| 3 | `v1/spec_decode/eagle3_pp.py` | Add `AXK2ForCausalLM` to `EAGLE3_PP_TARGET_ARCHS` |

**The shared mechanism in `eagle3_pp.py` is untouched** — #1006 predicted a new architecture costs one `forward` body, and that held.

Two non-obvious points:

* **Capture moved from before layer `i` to after layer `i-1`** (same tensor). This leaves the boundary index `start_layer` to the previous stage, which is the partition `aux_slots_received` / `aux_slots_captured` already use. Capturing before the layer would force a second, incompatible partition into the shared helpers.
* **Change 2 is not A.X-K2-specific.** The head declares `Eagle3DeepseekV2ForCausalLM` → `deepseek_eagle3`, not `llama_eagle3`, so `patches/llama_eagle3.py` never covered it; #1006 already flagged this head as carrying the identical defect. No-op at PP=1. `target_layer_count` does not move with the name — this head builds `DeepseekV2MLAAttention` directly and never reaches the `llama.py` arithmetic that reads it.

---

* Resolves #1007
* Related to #1006 (MiniMax-M2 pair, mechanism reused)
* Related to rbln-serve#448 — leaves the two PP variants red on purpose; they go green once this lands, no change needed on that side

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [ ] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)
* [ ] ❓ Other (`other`)

---

### 🧪 How to Test

```bash
uv run --no-sync pytest tests/vllm/patches/ tests/vllm/v1/spec_decode/test_eagle3_pp.py tests/vllm/test_platform.py -q
```

Expected: all pass (vllm 0.26.0+cpu, rebel-compiler 0.11.3.dev535).

* `test_axk2.py` — stages run in one process with stub layers stamping their own global index, so a stage-local index reads back as a wrong number rather than a shape that still fits. One of three aux sets puts every index on a band boundary.
* `test_eagle3_pp.py` — asserts every allowlisted architecture has a non-upstream `forward`.

Startup guard against the real checkpoints, no device:

```python
EngineArgs(model="skt/A.X-K2", pipeline_parallel_size=8, max_num_seqs=8, trust_remote_code=True,
           speculative_config={"model": "skt/A.X-K2-EAGLE3", "method": "eagle3", "num_speculative_tokens": 3}).create_engine_config()
```

Expected: succeeds. Reverting the allowlist to MiniMax-M2 alone reproduces the error verbatim.

---

Before, at `pipeline_parallel_size=8`:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for VllmConfig
  Value error, EAGLE3 with pipeline_parallel_size=8 is supported on RBLN only for
  target architectures ['MiniMaxM2ForCausalLM'], but got ['AXK2ForCausalLM'].
```

After: config resolves — target `AXK2ForCausalLM`, draft `Eagle3DeepseekV2ForCausalLM`, draft `pipeline_parallel_size` 1, aux ids `(2, 30, 58)` from the head's `eagle_config`.

---

### 📋 Checklist

* [x] This PR is linked to an existing issue (#1007)
* [ ] Relevant documentation has been updated — no doc covers the allowlist; `eagle3_pp.py`'s module docstring now names both patched architectures

---

### 💬 Notes

| | Status |
|---|---|
| Unit tests | ✅ pass |
| Startup guard, real checkpoints, PP8 | ✅ passes; reverting the allowlist reproduces the error |
| Deployment stands up + generates | ✅ reduced: PP4, `VLLM_RBLN_NUM_HIDDEN_LAYERS=8`, aux `(1, 4, 7)` on CR13 |
| Real coordinate: 61 layers, PP8, aux `(2, 30, 58)` | ⏳ running; result to follow in a comment |
| Acceptance / MAL | ❌ not measured |

In the reduced run `4` is stage 2's `start_layer` and stage 1 owns none of the three, so the boundary case and the pass-through path both ran. The last stage's assertion — which fires when the handoff carries a block count other than the requested layer count — did not trip; that is the evidence the boundary index was neither double-counted nor dropped.

Acceptance, not output text, is the signal: rejection sampling makes spec decode lossless, so a mishandled aux carry leaves greedy output unchanged and shows up only as collapsed acceptance. rbln-serve's `spec_decode_floor` gates exactly that.

**Reviewer asks:**

1. `patches/axk2/model.py` now exports `AXK2Model`, only so the test avoids the module's private `_module` handle. Happy to drop it.
2. `Eagle3DeepseekV2ForCausalLM.__init__` still writes the per-rank count into `self.config.target_layer_count`. Nothing on this head's path reads it — inert, not wrong-and-used. Separate change.


rebel-kblee (approved):
lgtm!

rebel-cjlee (approved):

PR #1141, opened 2026-09-21
feature(deps): add a kv_connectors extra for nixl-rbln and lmcache-rbln

rebel-minhopark:
### 🚀 Summary of Changes

Adds a `kv_connectors` optional dependency group next to the existing `runtime` extra:

```toml
kv_connectors = [
    "nixl-rbln>=0.1.0.dev0,<0.1.1",
    "lmcache-rbln>=0.12.0.dev0,<0.12.1",
]
```

`uv.lock` is refreshed accordingly and `DEVELOPMENT.md` documents `--extra kv_connectors`. Both packages are resolved from the nightly Nexus index; the lock pins `nixl-rbln 0.1.0.dev119+g7bbd9adca` and `lmcache-rbln 0.12.0rc1.dev3+g77a7dce`.

Neither model path is affected at runtime: the change is packaging metadata and the lockfile only. Nothing in `vllm_rbln/` imports either package differently than before.

**What the diff cannot show: the base lock is downgraded.** uv resolves the whole graph, extras included, to a single version per package, so the ceilings that `lmcache 0.5.4` declares in its metadata apply to everyone who runs `uv sync`, not only to those who enable the extra:

| Package | lmcache 0.5.4 pin | Before | After |
| :-- | :-- | :-- | :-- |
| numpy | `<=2.2.6` | 2.2.6 / 2.3.5 / 2.4.6 (per Python) | 2.2.6 |
| opentelemetry-api, -sdk, -proto, -exporter-otlp (+ proto-common/grpc/http) | `opentelemetry-api<=1.40.0` | 1.44.0 | 1.40.0 |
| opentelemetry-semantic-conventions | (via api) | 0.65b0 | 0.61b0 |
| opentelemetry-exporter-prometheus | `<=0.61b0` | not locked | 0.61b0 |
| prometheus-client | `<=0.24.1` | 0.25.0 | 0.24.1 |
| protobuf | (via opentelemetry-proto 1.40) | 7.35.1 | 6.33.6 |
| importlib-metadata | (transitive) | 9.0.0 | 8.7.1 |

numpy 2.2.6 ships no cp314 wheel, so a Python 3.14 environment would fall back to an sdist build. CI runs 3.12 and 3.13 only.

The extra itself is heavy but only paid by those who enable it: 31 new locked packages, including `cupy-cuda13x`, `cufile-python`, `nvtx`, the `google-cloud-bigtable` family and `redis` from lmcache, and both `nixl-cu12` and `nixl-cu13` from nixl 1.1.0.

---

* Related to #

---

### ✅ Type of Change

* [x] ❓ Other (`other`): dependency declaration / lockfile

---

### 🧪 How to Test

1. Run `uv lock --check` — passes, the lock matches `pyproject.toml`.
2. Run `uvx pre-commit run --files pyproject.toml uv.lock DEVELOPMENT.md` — passes.
3. On Linux x86_64 (`python:3.12-slim` container, uv 0.12.17), run `uv sync --locked --extra runtime --extra kv_connectors --dry-run` — resolves 202 packages and lists `nixl-rbln==0.1.0.dev119+g7bbd9adca` and `lmcache-rbln==0.12.0rc1.dev3+g77a7dce` in the install plan.

Not done: a real install and `import nixl_rbln` / `import lmcache_rbln`, and running the NIXL or LMCache connector on NPU with this version pair.

---

### 📋 Checklist

---

### 💬 Notes

Version owners, please confirm the pins and the lock result: @rebel-jinhwan for `lmcache-rbln` (and the lmcache 0.5.4 upper bounds above), @rebel-yskim for `nixl-rbln`.

The base-dependency downgrade above is the thing to decide on. If it is not acceptable, the only fix is on the lmcache side (relaxing its upper bounds); vllm-rbln cannot isolate the extra's resolution from the base set with uv.


pyproject.toml
```diff
@@ -75,6 +75,10 @@ dev = [
 runtime = [
     "rebel-compiler~=0.11.3.dev499",
 ]
+kv_connectors = [
+    "nixl-rbln~=0.1.0.dev0",
+    "lmcache-rbln~=0.12.0.dev0",
```
rebel-junamsong: Do we want to allow all `0.12.x` releases here?

`~=0.12.0.dev0` also allows `0.12.1+`. If we only support the `0.12.0` line, should this be `>=0.12.0.dev0,<0.12.1` instead?

rebel-minhopark: Thank you @rebel-junamsong , it's my mistake. I will fix it as you recommended!

rebel-yskim (commented):
LGTM once @rebel-junamsong  has finished reviewing and making any necessary changes. @rebel-junamsong , please approve when done.

rebel-junamsong (approved):
LGTM

PR #1142, opened 2026-09-21
fix(deps): make qwen_vl_utils an optional dependency

rebel-junamsong:
### 🚀 Summary of Changes

`qwen_vl_utils` moves from `[project.dependencies]` to a new `examples` extra.

Requested from the container side: `rbln-serve` image scans (Grype / Trivy) flagged
packages that ship in the image without being used at runtime.

Nothing under `vllm_rbln/` imports it. The three importers are all outside the
package:

```
examples/optimum/multimodal_models/qwen_vl_offline.py
examples/optimum/multimodal_models/exaone4_5_offline.py
tests/optimum/optimum_correctness/multimodal_models/qwen_vl_offline.py
```

#### Why this actually shrinks the image

`vllm-rbln` is the **only** consumer of `qwen-vl-utils` in `uv.lock`, and upstream
vllm declares it nowhere — not in `Requires-Dist`, not in `requirements/common.txt`,
`test.in` or `dev.txt`. Upstream's own multimodal examples and docs import it while
its package code mentions it only in a comment (`qwen3_vl.py`). So this drops the
package from an install rather than moving which requirement pulls it back in.

Verified against the built wheel, which is what a container install reads:

```
before: Requires-Dist: qwen_vl_utils
after:  Requires-Dist: qwen_vl_utils; extra == "examples"
```

#### Why `examples` rather than `test`

The importers are not pytest tests — `tests/optimum/optimum_correctness/` holds
`fire` CLI scripts (AGENTS.md), and `examples/` is not collected at all. Putting it
in `test` would make every lane that syncs `--extra test` install a package none of
them needs.

#### No CI change needed

`.buildkite/` has no reference to `optimum_correctness`, `multimodal_models`,
`qwen_vl_offline` or `exaone4_5_offline` — no lane runs the three importers. The
three pipelines sync `--extra test --extra dev --extra runtime`; an extra they do
not request is simply not installed and does not affect `--locked`.

---

* Raised on Slack by @rebel-minhopark from the `rbln-serve` image vulnerability scan;
  @rebel-eunji confirmed the removal is fine.

---

### ✅ Type of Change

* [x] ❓ Other (`other`): dependency packaging

---

### 🧪 How to Test

```bash
uv lock --check
uv sync --locked --python 3.12 --extra test --extra dev --extra runtime
python -c "import qwen_vl_utils"          # ModuleNotFoundError -- expected
uv sync --locked --python 3.12 --extra test --extra dev --extra runtime --extra examples
python -c "import qwen_vl_utils"          # succeeds
```

| check | result |
|---|---|
| built wheel `Requires-Dist` | `qwen_vl_utils` only under `extra == "examples"` |
| `uv lock --check` | passes |
| CI sync command, no `--extra examples` | `qwen-vl-utils` not installed |
| same sync `+ --extra examples` | installs 0.0.14, imports |
| `run-t0-unit.sh` | 2297 passed, 26 skipped, 8 xpassed |
| packaging guards | 9 passed |
| `pre-commit` on changed files | clean |

`uv.lock` carries the marker move only; no version or source changed.

---

### 📋 Checklist

* [x] Relevant documentation has been updated (`DEVELOPMENT.md`, beside the `runtime` extra note)

---

### 💬 Notes

Affects packaging only; neither model path changes.

Two follow-ups this deliberately leaves alone:

- **`datasets`** is in the same position — 0 imports under `vllm_rbln/`, 20 in
  `examples/`, and upstream vllm requires it only under its `bench` extra, so it too
  would actually leave the image. Its tree (pyarrow, pandas, fsspec, dill,
  multiprocess, xxhash) is much larger than this one. Worth its own PR if the image
  owners want it.
- **`pyyaml`** and **`fastapi[standard]`** are declared here but pulled
  unconditionally by vllm anyway, so removing them would tidy the manifest without
  changing the image.


rebel-seinpark (approved):
LGTM

PR #1143, opened 2026-09-22
feature(dynamic-kv): read MultiConnector through to its children

rebel-jinhwan:
Stacked on #1129 — **base is `dynamic-kv-default`, not `dev`.** Merge that first.

## WHY

#1129 settles the connector question by name:

```python
if kv_transfer.kv_connector not in DYNAMIC_KV_SUPPORTED_CONNECTORS:
    return f"kv_connector={kv_transfer.kv_connector!r} is not among ..."
```

`MultiConnector` is not a connector by that measure. It registers nothing of its
own — `register_kv_caches` and the post-resize finalize both fan out to its
children (`MultiConnector.register_kv_caches`, `iter_kv_connectors`) — so the
name it gets checked under says nothing about whether the resize is open for it.

The effect, with #1129 making dynamic KV the default: `MultiConnector[RblnNixl…,
RBLNLMCacheConnectorV1]` — P/D plus offload, the shape in use — reads as
unsupported and **silently falls back to static sizing**. Nobody asked for that,
and one log line is all there is to show for it.

This is the follow-up #1129 names: *"Extend `DYNAMIC_KV_SUPPORTED_CONNECTORS`
after additional connectors are verified against the post-resize registration
flow."*

## WHAT

`configured_kv_connectors()` replaces a `MultiConnector` with the children it
wraps, and `dynamic_kv_unsupported_reason` checks those. The reason names only
the children that are not open, so the wrapper is not blamed for something it
merely carries.

Two cases fall out of that rather than being special-cased:

- a wrapper with **no children** keeps its own name and is reported — an empty
  list means nothing was *checked*, not that everything passed;
- a child that is itself a `MultiConnector` is left as that name, so the nested
  shape reads as untried like any other name outside the set. Nesting is not
  walked: walking it would admit a shape nobody has tried, which is the opposite
  of what the set is for.

`MULTI_CONNECTOR` deliberately does **not** join
`DYNAMIC_KV_SUPPORTED_CONNECTORS` — that tuple is asserted against
`KVConnectorFactory._registry` and is about connectors, not wrappers.

## Tests

`pytest tests/vllm/v1/worker/test_utils.py tests/vllm/test_platform.py
tests/vllm/v1/worker/test_dynamic_kv_sizer.py tests/vllm/patches/test_dynamic_kv.py`
— 323 passed. Added to `TestDynamicKvUnsupportedReason`: `MultiConnector[nixl,
lmcache]` is supported; one unsupported child is named and the wrapper is not;
a wrapper with no children is reported; a `MultiConnector` child reads as
untried. The existing connector tests move to the same `_ktc()` builder, which
gives the stand-in the `kv_connector_extra_config` a real `KVTransferConfig`
always has.

`ruff check` / `ruff format --check` clean.

## Hardware verification

Done on this stack (#1129 + this commit is the same registration ordering as
#1138, which is what the run exercises). One RBLN-CR13, `Qwen/Qwen3-1.7B`,
`block_size=256`, `LMCACHE_CHUNK_SIZE=256`, `gpu_memory_utilization=0.8`, with
lmcache-rbln at `main`.

**`MultiConnector[RblnNixlConnector, RBLNLMCacheConnectorV1]`, dynamic KV on — PASS**

```
[Dynamic KV] compiling with 4 KV blocks instead of 3918
[Dynamic KV] reallocating KV cache: 4 -> 3936 blocks
RBLN NIXL (D2D): deferring registration of 28 KV cache layer(s) until after warm-up.
LMCache INFO: Registered 28 KV caches on rbln:0, allocated shape=(3936, 2, 8, 1, 256, 128)
NIXL_RBLN: registerMem VRAM ... dmabuf_size=1031798784
```

Both children get the post-resize cache. NIXL's 1,031,798,784 B is exactly
3936 blocks × 1 MiB ÷ 4 chiplets; the 4-block compile cache would have been
~1 MiB. On the LMCache side the registered pointers are a subset of
`model_runner.kv_caches` with layer 0 matching, store → restore returned
identical output, and `vllm:external_prefix_cache_hits = 256`.

**`RBLNLMCacheConnectorV1` alone — PASS.** Same shape (registration reads 3936
blocks, layer-0 `data_ptr` matches). This is the evidence for the
`RBLNLMCacheConnectorV1` entry #1129 adds; that PR's test plan is unit tests
only. Full log in the comment on #1129.

**Baseline with dynamic KV off** comes up at 3925 blocks and produces identical
output and hit counts.

> `MultiConnector` runs segfault **at process exit**, in libnl's
> `rtnl_tc_unregister` under `_dl_fini` — after the run has finished and
> everything has been asserted. Pre-existing NIXL/UCX atexit ordering, not from
> this change; it does not happen with lmcache alone.

### One measurement, and a correction to it

`RBLNConnector._staging` (lmcache-rbln's device-side upload buffer) is allocated
lazily on the first transfer, **after** sizing has handed out the budget:
**29,360,128 B (28 MiB)**, on chiplet 0 only — it is not sharded.

I first read the fit check's `budget_left=+56623104` as the room left for it and
called 54 MiB tight. That was wrong, and #1129 corrected it: `budget` is
`total × gpu_memory_utilization`, so `budget_left` measures the distance to the
policy line, not to the device. Room to the device is
`total − measured = 7,572,815,872` B = **7.05 GiB**. The `(1 − gmu)` slice is
where a post-sizing allocation is meant to land — 2.45 GiB per chiplet even at
the 0.93 default, ~90× the buffer. No reserve; `num_ooms=0`.

> This host runs KMD 3.5.0~rc5, which has no `mem_get_info_per_chiplet()`, so
> sizing fell back to the process allocator snapshot. A host whose driver
> provides the query takes a different path through the budget arithmetic.

## Related

- #1129 — the base. Adds `RBLNLMCacheConnectorV1`; this adds the wrapper over it.
- #1138 — moved registration behind the resize, which is what makes any of this safe.
- rebellions-sw/lmcache-rbln#186 — no lmcache-rbln change needed.
  rebellions-sw/lmcache-rbln#197 was opened to make a second registration land
  completely, then closed: upstream's `assert len(self.kv_caches) == 0` already
  catches that case loudly, before any connector state has moved.
- rebellions-sw/icp-serving#601


vllm_rbln/v1/worker/utils.py
```diff
@@ -310,6 +310,28 @@ def compile_and_warmup_skip_reason(vllm_config: VllmConfig) -> str | None:
     "RblnNixlPushConnector",
     "RBLNLMCacheConnectorV1",
 )
+# vLLM's wrapper for running several connectors at once. It is not itself in the
+# set above because it registers nothing of its own: `register_kv_caches` and the
+# post-resize finalize both fan out to its children, so the resize is open for it
+# exactly when it is open for every child.
+MULTI_CONNECTOR = "MultiConnector"
```
rebel-jindol21, line 317: **`MULTI_CONNECTOR` sits outside the registry self-check the previous commit added.**

This is a second name coupled to vLLM's factory registration by string, but the self-check further down still iterates only the allowlist:

```python
unregistered = [
    name
    for name in DYNAMIC_KV_SUPPORTED_CONNECTORS
    if name not in KVConnectorFactory._registry
]
```

so the new constant gets none of the protection `1c9f40c0` was written for.

If that string ever stops matching — vLLM renames the class, or the constant is mistyped — the first line of `configured_kv_connectors`

```python
if kv_transfer.kv_connector != MULTI_CONNECTOR:
    return [kv_transfer.kv_connector]
```

is always true, the fan-out never happens, and `["MultiConnector"]` is what comes back. That name is not in the allowlist, so `MultiConnector[RblnNixlConnector, RBLNLMCacheConnectorV1]` — the shape this PR exists to unblock — silently reverts to static sizing, and the warning it logs is indistinguishable from a genuinely unsupported connector. That is the same silent-fallback failure the previous commit set out to make loud, re-opened for the one name it does not cover.

Folding it into the same check is enough:

```python
unregistered = [
    name
    for name in (*DYNAMIC_KV_SUPPORTED_CONNECTORS, MULTI_CONNECTOR)
    if name not in KVConnectorFactory._registry
]
assert not unregistered, (
    "the dynamic-KV connector names the factory never registered: "
    f"{unregistered}"
)
```

The message needs widening too, since it no longer names only `DYNAMIC_KV_SUPPORTED_CONNECTORS`. `"MultiConnector"` is registered in vLLM 0.26.0 (`factory.py:194-197`), so the check passes as things stand. A sibling of `test_a_supported_connector_the_factory_never_registered_fails_loudly`, monkeypatching `MULTI_CONNECTOR` out of the registry, would pin it.

One caveat on the fix: the check is a bare `assert`, so `python -O` strips it and the loudness disappears in optimized runs. Raising instead would keep it in both.

rebel-jindol21 (commented):
The premise checks out: in vLLM 0.26.0 `MultiConnector` fans out `register_kv_caches` (line 245), `register_cross_layers_kv_cache` (238) and `set_host_xfer_buffer_ops` (344) to its children, and RBLN's `iter_kv_connectors` / `finalize_kv_cache_registrations` recurse into them, so the deferred-registration ordering really is transparent through the wrapper. Reading the wrapper through to its children is the right call, and the child-config key, the registry name and the `KVTransferConfig` import all line up with upstream.

One thing I would like addressed before merge — inline below.

PR #1144, opened 2026-09-22
feature(compile): accept compile_dtype as an RBLNConfig option

rebel-jonghewk:
#### Summary

rebel-compiler now takes the device dtype as a compile option (`dtype="float16" | "bfloat16"`) instead of only through the `RBLN_COMP_DTYPE` environment variable. This PR exposes that option on the vLLM-native path as `RBLNConfig.compile_dtype`, so the dtype a model compiles in is set per launch with `--rbln-compile-dtype` or `--additional-config '{"compile_dtype": "float16"}'` and travels with the config like every other RBLN option.

Only the plumbing is added here. No model gets a default; the field is empty unless the launch sets it, and empty keeps the compiler's per-target default exactly as before.

#### Details

- **`RBLNConfig.compile_dtype`** — `Literal["", "float16", "bfloat16"]`, default `""`. It is a regular field, so it gets the `--rbln-compile-dtype` flag and the `additional_config` route for free, and pydantic rejects any other spelling. There is deliberately no `VLLM_RBLN_COMPILE_DTYPE` variable.

- **Cache key** — the field is not in `compute_hash()`'s ignore list, so the mega-cache bundle key changes with it. rebel's own per-graph hash also carries `dtype:` in its meta, so a bundle compiled in one dtype is never reused for another.

- **`compiler.compile(dtype=)`** — forwards the value as `options["dtype"]` to the rbln torch.compile backend. Every graph that builds the model goes through it: the model body and `compute_logits` in the model runner, and the eagle / dflash / medusa drafters. The sampler graphs are left alone; they compile with `use_cache=False` and carry no model math.

- **Older rebel-compiler** — a rebel without the option silently ignores unknown option keys, so the graphs would compile in the target default while the launch believes it asked for float16. When `compile_dtype` is set, `check_and_update_config` probes `rebel.compile`'s signature once at startup and raises a `ValueError` if `dtype` is missing, so the launch fails before the model is loaded rather than at the first compile. The current pin (`rebel-compiler~=0.11.3.dev499`) admits pre-releases without the option; `TestCompilerConformance::test_rebel_compile_signature` fails once the pin guarantees it, which is the cue to delete the probe.

#### Next Steps

- A.X K1 needs float16 on the device. This PR does not encode that; the A.X K1 branch will either add `--rbln-compile-dtype float16` to its launch recipe or, if a default is wanted, put it next to the model's own patches once that branch lands on dev.

---

* Related to #

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

1. Run the unit tests for the touched modules:
   ```
   VLLM_RBLN_USE_VLLM_MODEL=1 pytest -q tests/vllm/test_envs.py tests/vllm/test_rbln_config.py tests/vllm/test_platform.py tests/vllm/compilation/test_compiler.py tests/vllm/v1/worker/test_mega_cache.py
   ```
   Expected: all pass (338 locally). The new cases are `TestCompileOptions::test_dtype_is_forwarded` and `test_empty_dtype_is_not_forwarded`, `TestCompilerConformance::test_rebel_compile_signature`, `test_dtype_check_refuses_an_older_rebel` and `test_dtype_check_passes_when_rebel_takes_it`, `test_only_compile_fields_change_the_hash` (compile_dtype changes the hash) and `TestCompileDtype` (an empty `compile_dtype` skips the probe, an older rebel is refused at startup).
2. `pre-commit run --files <changed files>`: ruff check / format, typos, mypy and the license header check pass.
3. On a rebel-compiler that takes the `dtype` compile option, launch any model with `--rbln-compile-dtype float16` and confirm the `meta=npu:...|tp:...|dtype:float16` hash line in the compile debug log. On an older rebel the same launch fails at startup, before the model loads, with `ValueError: compile_dtype needs a rebel-compiler that takes 'dtype' as a compile option; installed ... does not.` Not done on an NPU host: the rebel-compiler installed here predates the option, so only the refusal path ran end to end.

---

### 📋 Checklist


tests/vllm/test_platform.py
```diff
@@ -306,6 +306,20 @@ def test_enforce_fp32_overrides_a_supported_dtype(self):
...
+        assert configured.additional_config.compile_dtype == ""
+
+    def test_the_hook_leaves_a_given_value_alone(self, reconfigure):
+        def mutate(config):
+            config.additional_config = replace(
+                config.additional_config, compile_dtype="float16"
+            )
+
+        config = reconfigure(mutate)
+        assert config.additional_config.compile_dtype == "float16"
```
rebel-cjlee: These two do not fail without this PR, except by `AttributeError` for a field that does not exist yet.

This PR does not touch `platform/`. So one test reads the default back out of the field declaration, and the other checks that no code touches the field, which is true because no such code was added. Compare with `TestDtype` above, which covers real branches in the hook.

I think this class can go away. The real coverage is the three tests in `test_compiler.py` plus the new has line.

vllm_rbln/compilation/compiler.py
```diff
@@ -43,6 +45,16 @@ def _ensure_torch_dynamo_configured() -> None:
     _DYNAMO_CONFIGURED = True
 
 
+def _check_dtype_option_supported() -> None:
+    # rebel.compile gained `dtype` in the same change as the torch.compile option.
+    # Drop this probe with the rebel-compiler pin bump that guarantees both.
```
rebel-cjlee: Nothing will tell us when to do this. After the pin moves, the probe keeps passing and quietly stays forever.

`test_compiler.py` already has the pattern. `TestCompilerConformance.test_rebel_compile_context_signatures` uses `assert "use_global_ctx" not in params` as a drift alarm. Could we add the same here? Then CI fails on the bump and points at the code to remove.

vllm_rbln/compilation/compiler.py
```diff
@@ -118,6 +131,9 @@ def set_option(key: str, value: Any) -> None:
     set_option("use_global_ctx", use_global_ctx)
     set_option("global_device_id", global_device_id)
     set_option("use_static_output", use_static_output)
+    if dtype:
+        _check_dtype_option_supported()
```
rebel-cjlee: This runs at the first compile, so the launch fails after the model is loaded and warm up started. That is minutes of waiting for an error we already knew at startup.

Could the check go in `check_aud_update_config` instead, once, when `compile_dtype` is not empty?
If you want all the rebel dependency knowledge to stay in `compiler.py`, that is fair too. Just raising the tradeoff.

rebel-eunji:
- vllm_rbln/v1/sample/rbln_rejection_sampler.py
- vllm_rbln/v1/sample/rbln_sampler.py

Don't these need compile_dtype as well?

rebel-cjlee (approved):

rebel-jonghewk:
Following up on an offline discussion with @rebel-eunji: we've decided not to include the sampler.

PR #1145, opened 2026-09-22
refactor(config): drop the compile-model option

rebel-cjlee:
### 🚀 Summary of Changes

Removes the compile-model option: the `RBLNConfig.compile_model` field, the `--rbln-compile-model` flag, and the `VLLM_RBLN_COMPILE_MODEL` environment variable. Affects the vllm model path only.

`enforce_eager` and `compile_model` selected the same thing from two places. Every call site read them together (`enforce_eager or not compile_model`), so the second switch added no path of its own. It did add one state that nothing covered: `compile_model=False` with `enforce_eager` unset allocated the KV cache on CPU while the rest of the runner still believed a compile was coming, and the attention ops answered that state with `NotImplementedError`. The op entry points now call their kernel directly, which is what they did whenever they were reached at all.

This also drops the `compile_model` keyword argument the attention ops took, so `RBLNFlashAttentionImpl` and `RBLNFlashAttnMLAImpl` no longer carry it.

---

* Resolves #
* Related to #1122

---

### ✅ Type of Change

---

### 🧪 How to Test

1. Run the suites covering the touched modules:
```
uv run --no-sync pytest
    tests/vllm/test_rbln_config.py
    tests/vllm/v1/attention/test_flash_attention.py
    tests/vllm/v1/attention/test_flashattn_mla.py
    tests/vllm/v1/spec_decode/
    tests/vllm/v1/worker/test_rbln_model_runner.py
    tests/vllm/v1/worker/test_rbln_worker.py
    tests/vllm/v1/worker/test_utils.py
    tests/vllm/v1/worker/test_mega_cache.py -q
```

2. Expected: all pass. `test_mega_cache.py` is the one that fails if the variable is removed from `environment_variables` but left in one of the two partition sets.

3. Edge case: `--rbln-compile-model` and `VLLM_RBLN_COMPILE_MODEL` are now unknown. The flag is rejected by argparse; the environment variable is ignored. Use `--enforce-eager` instead.

---

---

### 📋 Checklist

---

### 💬 Notes

@rebel-jonghewk vllm-rbln used to hold reference implementations for attention, and they are gone now. Looking at the `rebel` package, part of that reference implementation seems to live there. Can the attention kernel ops that `rebel` currently ships be called in eager? Some of the attention ops do seem to run in eager, but I have not checked all of them. If we need them, a follow-up PR should either put the reference implementation back in vllm-rbln, or have the `rebel` package provide them.

---


rebel-jonghewk:
> vllm-rbln used to hold reference implementations for attention, and they are gone now. Looking at the rebel package, part of that reference implementation seems to live there. Can the attention kernel ops that rebel currently ships be called in eager? Some of the attention ops do seem to run in eager, but I have not checked all of them. If we need them, a follow-up PR should either put the reference implementation back in vllm-rbln, or have the rebel package provide them.

Maybe some custom op unit test might use it but not sure. I'll have to check. But I think clearing compile-model should be fine. (I don't know why this option was here in the first place)

rebel-cjlee:
@rebel-jonghewk Sounds good. Since this PR just removes the `compile_modmel` option and controls compilation with `enforce_eager`, I think it's fine even if all attention ops aren't fully eager-supported yet. Later on, I plan to check eager support more closely, including the sampler.

rebel-jonghewk (approved):

PR #1148, opened 2026-09-22
fix(axk2): keep the MLA patch from skipping itself

rebel-cjlee:
### 🚀 Summary of Changes

A.X-K2 served on the vllm model path died in the worker with
```
AssertionError: AXK2 requires MLA (use_mla) for the latent-attention + fused output-gate path; the non-MLA fallback has been removed.
```

The patch that makes axk2 count as MLA was never applied. This removes the condition that was turning it off.

#### Why it broke

`patches/axk2/__init__.py` holds two things. `register_axk2` puts the architecture into `ModelRegistry` and the config class into `_CONFIG_REGISTRY`. The `is_deepseek_mla` patch teaches upstream that axk2 is an MLA model, which upstream decides from a hardcoded `model_type` tuple that does not list it.

The registry applied every registration first, then every patch. The patch carried `condition=not _upstream_has_axk2()`, meaning to stand down once upstream ships the architecture itself. By the time that condition is evaluated, `register_axk2` has already registered it, so the condition reads our own registration as upstream's and the patch skips itself.

`ModelConfig` is then built with `use_mla` False. That config is pickled to the workers, so the assertion fires there while the value was decided in the frontend.

This arrived with #1122, which moved the registration out of the `register_model` entry point and into the patch registry. Before that the two ran at different times and the condition was still true when it was asked.

#### The fix

The condition is gone. `register_axk2` already raises when upstream ships the architecture, telling us to delete the vendored copy, and registrations run before any patch is considered. One guard is enough, and that one cannot mistake our own registration for upstream's.

---

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

Building the engine config for `skt/A.X-K2` from a local snapshot:

| route | before | after |
| --- | --- | --- |
| `--model-impl vllm` | `use_mla=False` | `use_mla=True` |
| `VLLM_RBLN_USE_VLLM_MODEL=1` | `use_mla=False` | `use_mla=True` |

---

---

### 📋 Checklist

---

### 💬 Notes

---


rebel-kblee (approved):
lgtm!

rebel-kblee (approved):

PR #1150, opened 2026-09-22
fix(eagle3): build the draft warm-up on its own KV cache group

rebel-ykchoi:
# fix(eagle3): build the draft warm-up on its own KV cache group

**Base:** `dev`  ·  **Head:** `fix/eagle3-draft-kv-cache-group`

---

### 🚀 Summary of Changes

A hybrid model holds more than one KV cache group, and the groups address the
cache in different kernel blocks -- a sliding window in windows, full attention
in the manager's blocks. So the block table that describes one request is a
different width in each group.

A speculative draft brings its own KV layer, and that layer joins whichever
group its spec matches. In a hybrid target there is no reason for that to be
the first group.

The draft graph is compiled during warm-up and run when a request arrives. The
warm-up took the first group's block table while the propose path takes the
group the drafter belongs to, so the two sites described different geometry.
The compiled graph then rejects the tensor it is handed, and it does so on the
**first request after the server is already serving** -- model load, warm-up
and connector registration all pass first.

This makes the warm-up read the drafter's own group, which is what the propose
path already does.

**Why it had not surfaced.** The speculative targets in use so far are
single-group, and with one group the two sites index the same thing. It was
first seen on gpt-oss-120b with an EAGLE3 draft: its layers alternate
sliding-window and full attention, and the draft joins the full-attention
group, which is not the first. The axis is
whether the target is hybrid, not which EAGLE flavour is in use; that makes it
a different axis from a draft whose KV heads diverge from the target's, which
is about a draft region's size rather than which group's geometry it inherits.

- `_build_dummy_attn_metadata` -- the one changed line. It is the warm-up's
  view; the run-time view is assembled per group in the model runner and
  already selects `kv_cache_gid`.
- `kv_cache_gid` cannot be the `-1` the base class initialises it to. The
  assertion that settles it, and the fact that it runs before the search which
  assigns the field, are both in the same upstream method:
  `grep -n 'validate_same_kv_cache_group' -A 25 $(python -c 'import vllm.v1.spec_decode.llm_base_proposer as m; print(m.__file__)')`.
  No guard is added for it.
- Single-group models read `kv_cache_gid == 0`, so their value does not move.
  `dflash` reuses this helper and goes through the same initialisation, so the
  same argument covers it.

**Tests.** The case that was already there builds a one-element block table,
where index `0` and index `kv_cache_gid` cannot be told apart. The new case
gives two groups of different widths and puts the drafter in the second, so it
fails on the old code with the shapes the runtime reported.

* Related to rebellions-sw/icp-serving#608
* Related to ICR-59

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

### 🧪 How to Test

1. `pytest tests/vllm/v1/spec_decode/test_eagle.py`
2. To see the failure this fixes, revert the production file and run
   `pytest tests/vllm/v1/spec_decode/test_eagle.py::TestBuildDummyAttnMetadata`
   -- the new case reports the two block table widths of a hybrid target.
3. The block table widths only diverge once a target actually has two KV cache
   groups, which no unit case can reach without standing a model up, so that
   part is a testing boundary rather than something this suite covers.

### 💬 Notes

- vllm model path only. The optimum model path does not reach this code.
- `_build_dummy_attn_metadata` has no upstream counterpart. Upstream's dummy
  path takes the block table from the common metadata it is passed rather than
  indexing the batch, so this is ours to fix and not a candidate for a patch
  under `patches/` or an upstream issue.
- **The mega-cache bundle goes stale rather than re-keying.** The key is a
  hash of config and environment, and this change moves neither while it does
  change the draft graph's input shape. A run after it therefore restores the
  bundle, misses one program and refuses. Remove that signature's directory
  under the model's cache root -- not the whole root, which also holds the
  bundles of every other configuration of that model.


rebel-cjlee:
@rebel-ykchoi just curious, what is the model that this PR targets?

rebel-ykchoi:
@rebel-cjlee The configuration it came from is **gpt-oss-120b as the target with an EAGLE3 draft**. That target alternates sliding-window and full-attention layers, so it has two KV cache groups, and the window is much shorter than the block — the two groups address the cache in kernel blocks that differ by a factor of eight. The draft is full attention, so it joins the full-attention group, and that group is not the first one. That is the whole condition: the first group's block table and the drafter's group's block table are different widths.

The fix is not specific to that model, though. Any hybrid target with a speculative draft reaches it, and a single-group target cannot — which is why it had not surfaced, since the speculative targets in use so far have been single-group. EAGLE vs EAGLE3 does not matter here.

[ICR-59](https://rbln.atlassian.net/browse/ICR-59) in the Related Issues section has the fuller write-up: how the two sites came to disagree, and the follow-on it names (a draft layer joining a group raises that group's layer count, so a hybrid ends up with one more KV pool that only the draft layer uses).

I have added the first sentence of that to the description as well, since the question was a gap in it.

rebel-cjlee (approved):

PR #1153, opened 2026-09-22
fix(deps): pick up the lmcache 0.5.4 cp314 wheel in uv.lock

rebel-minhopark:
### 🚀 Summary of Changes

Lock-only change: `uv.lock` now lists the `lmcache-0.5.4-cp314-cp314-manylinux_2_34_x86_64.whl` wheel that was published to Nexus after #1141 merged.

#1141 added the `kv_connectors` extra, but at lock time upstream `lmcache` 0.5.4 only had cp310–cp313 wheels, so the lock fell back to the sdist for Python 3.14. The cp314 wheel is now on `pypi-group-nightly`. A plain `uv lock` does not pick it up: the lock is already consistent with `pyproject.toml`, so uv validates it in milliseconds and never re-queries the index. `uv lock --refresh-package lmcache` re-fetches only the `lmcache` index page and adds the one wheel line; the pinned version is unchanged. `lmcache-rbln` itself is `py3-none-any` and was never the gap.

Model path: neither. Dependencies only, no source change.

---

* Related to #1141

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `uv lock --check` — passes (lock matches `pyproject.toml`).
2. Run `uvx pre-commit run --files uv.lock` — `uv-lock` hook passes.
3. Run `uv export --frozen --no-hashes --extra kv_connectors --format requirements.txt | grep '^lmcache'` — resolves `lmcache==0.5.4` and `lmcache-rbln==0.12.0rc1.dev3+g77a7dce`.
4. On a Python 3.14 Linux x86_64 host, `uv sync --locked --extra kv_connectors` should install the cp314 wheel instead of building the sdist. Not run here (no 3.14 Linux host in this session).

---

### 📋 Checklist

---

### 💬 Notes

Step 4 of the test plan was not executed. The change was verified through lock consistency and export only.


rebel-junamsong (approved):
LGTM

PR #1155, opened 2026-09-22
fix(nixl): carry the stripe width into the D2D registration again

rebel-ykchoi:
# fix(nixl): carry the stripe width into the D2D registration again

**Base:** `dev`  ·  **Head:** `fix/nixl-stripe-width-d2d`

---

### 🚀 Summary of Changes

The connector has two transfer paths and they create the NIXL backend at
different points. Host staging creates it before registering; the direct path
lets the adapter create it from inside the call that registers the KV regions.
A backend-lifecycle setting is applied when the backend is created and not
re-applied to a live one, so each of those calls has to carry it.

When the knob was added it rode both. Our split of the worker into lifetimes
moved the host-staging site with it and left the keyword behind on the direct
one, so the knob has had no effect there since. It fails silently: the value is
passed to the adapter only when it is named, so an unnamed knob and a dropped
keyword look the same from outside.

This restores it, in the shape the pre-split code and the surviving site both
use.

- The guard tests for absence, not for truthiness. `0` is a width the adapter
  takes -- it selects a striping mode rather than meaning "unset" -- so a
  truthy test would make that value unreachable. The host-staging site tests
  the same way, and so did this one before the split.
- The knob is read once, from the connector's extra config, and both sites
  read that one field. Nothing about where it is read changes here.

**Tests.** The suite stayed green because the commit that dropped the keyword
dropped its coverage in the same move: `git grep -c stripe_width <base> --
tests/` comes back empty. Both paths have a case now, each over three values --
unnamed, `0`, and a width. The `0` case is what separates "nobody named it"
from a width of zero, and it is the one that fails if the guard is weakened
back to truthiness. The host-staging case also reads the keyword its fake used
to swallow, which is the same silence this fix is about.

* Resolves rebellions-sw/icp-serving#610

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

### 🧪 How to Test

1. `pytest tests/vllm/distributed/kv_connector/`
2. To see what the new case pins, weaken the guard to a truthy test and run
   `pytest tests/vllm/distributed/kv_connector/test_rbln_nixl_registration.py -k stripe_width`
   -- the `0` case reports the keyword the adapter did not receive.

### 💬 Notes

- vllm model path only. The optimum model path does not reach this code.
- The keyword is passed only when the knob is named, so an adapter build
  without that parameter is unaffected as long as nobody sets it.


rebel-jongwonlee:
Read at `8d4dfe54` against `e8310e7d`. No vllm 0.26 environment here, so this is static review plus history — the suite was not run.

The fix is right and minimal. History backs the causal account: `1fc3d967` (#1119, 9/17) added the keyword to both sites; `fd8b66dc` (#1089, 9/18) removed both and restored only `ensure_rbln_backend`. The shape here matches the surviving site exactly, and the two are mutually exclusive — the D2D branch returns before `ensure_rbln_backend`, so nothing double-applies.

**Verified**

- `make_worker` pins `kv_buffer_device="rbln"`, so the new case really does take the deferred path (`register_kv_caches` stashes → `finalize_kv_cache_registration` → `_register_kv_caches_impl`) and `register_kv_regions` lands exactly once. `== [passed]` is the right assertion, not a membership test.
- The fake keeps the right keywords: `device_id` is positional so it falls into `*a`, and the filter removes only the two structural ones.
- `engine_config` is `@functools.cache`d and `stripe_width=None` is the default, so every existing caller keeps its current entry — three new entries, no key churn across the rest of the suite. `int | None` is hashable.
- `_stripe_width` is read once in `base_worker.__init__`, well before finalize, so both sites see the same value.

**One change — host staging still has the hole this PR closes**

The surviving host-staging case is `test_host_bounce_creates_backend_and_delegates`:

```python
monkeypatch.setattr(
    sys.modules["nixl_rbln"],
    "ensure_rbln_backend",
    lambda wrapper, device_id=0, **kw: ensured.append((wrapper, device_id)),
    raising=False,
)
...
assert ensured == [("wrapper", 0)]
```

`**kw` is swallowed. It pins *that the call happens*, not *what it carries* — the same silence that let the keyword go missing on the direct path. Drop `**extra` from `ensure_rbln_backend` today and the suite stays green.

Two lines close it:

```python
lambda wrapper, device_id=0, **kw: ensured.append((wrapper, device_id, kw)),
...
assert ensured == [("wrapper", 0, {})]
```

— and the same three-value parametrize if you want the two sites symmetric.

**The Tests paragraph does not describe this tree**

> There was a case for this knob already, and it names the host-staging path in its own body, so it never reached the direct one

`fd8b66dc` deleted **both** cases: the host-staging one (`kv_buffer_device="cpu"`) and the D2D one, along with `_prep_impl_worker`'s `stripe_width` parameter. On `e8310e7d` the name does not appear under `tests/` at all:

```
$ git grep -c stripe_width e8310e7d -- tests/
(no matches)
```

So nothing sits beside the new case, and the reason the suite stayed green is stronger than the one given: the commit that dropped the keyword dropped its coverage in the same move. That is also the argument for the change above.

**Recurrence**

The fix reinstates a second literal copy of the guard, which is the mechanism of the original bug. One property makes the next split unable to lose a site quietly:

```python
@property
def _backend_extra(self) -> dict[str, int]:
    """What rides whichever call creates the backend. 0 is a width the
    adapter takes, so the test is on absence."""
    return {} if self._stripe_width is None else {"stripe_width": self._stripe_width}
```

Both sites then read `**self._backend_extra`, and a second backend-lifecycle knob is added in one place. Your call — the duplication is faithful to the pre-split shape, which is a defensible reason to keep it.

**nit** — `TestD2dRegistrationReachesTheAdapterAndThePeer` opens with *"Two things the D2D path does whose absence is silent: neither shows up as a failure anywhere else"*. Three now.

rebel-ykchoi:
All four taken, in `263b4cf5` on top. The branch is also rebased onto current `dev`.

**The property.** Done, and your reason is the one that decided it: the copy *is* the mechanism, so a second split cannot take one call and leave the other. Both now read `_backend_extra`, and the sentence about absence versus truthiness moved there with them, since that is the one place it has to hold.

**The host-staging case.** You were right that the fake made it blind, and it went one layer deeper than that. `build_worker` hands the worker a `MagicMock` config, so `kv_connector_extra_config.get("stripe_width")` answered with a mock rather than `None` — the guard's `is None` was false on that path and the adapter was being handed a mock width. The fake swallowing `**kw` is what kept it invisible. So the fix is two things: the harness now passes a real dict (and takes the width as a parameter), and the case reads the keyword and runs the same three values the direct one does. Dropping the keyword from `ensure_rbln_backend` now turns two of them red; before, nothing moved.

**The description.** Corrected, and the claim I made was wrong in a way worth naming: I wrote that a case existed and named the host-staging path, so it never reached the direct one. `git grep -c stripe_width <base> -- tests/` comes back empty — `#1089` took both cases and the harness parameter with them. The real reason the suite stayed green is stronger than what I wrote: the commit that dropped the keyword dropped its coverage in the same move. I had read the tests on the `#1098` branch and taken them for `dev`'s.

**The lead.** Now "Three things".

rebel-jongwonlee:
Confirmed at `263b4cf5`.

**The MagicMock is the better find, and it is bigger than what I raised.** On the base, `build_worker` was `vllm_config = MagicMock()` with nothing pinned under it (`utils.py:694`), so `kv_transfer_config.kv_connector_extra_config.get("stripe_width")` answered with a mock, `is None` was false, and every `build_worker` test that reached `_register_kv_caches_impl` handed the adapter `stripe_width=<MagicMock>`. Both fakes swallow `**k`, so nothing moved. What I read as "the host case is blind" was really "the `None` row had never run through that harness at all" — on either path.

**And the fix closes the family rather than the member.** `__init__` reads exactly three things off `vllm_config` — `scheduler_config`, `kv_transfer_config.kv_connector_extra_config`, `cache_config.block_size` — and `build_worker` now pins each, plus `parallel_config.pipeline_parallel_size` and `speculative_config`. No second absence test is left reading a mock, so this does not need revisiting knob by knob.

**The rest checks out.**

- Both call sites read `_backend_extra` and the returned dict is what each built before, so the behaviour is unchanged. The sentence about absence versus truthiness sits in the one place it has to hold.
- Dropping the keyword from `ensure_rbln_backend` turns the `0` and `4096` rows red and leaves `None` green — two of three, as you said.
- No collateral from the harness change: the one test that reads `register_kv_regions.call_args.kwargs` (`test_rbln_nixl_registration.py:380`) checks `mem` and `rbln_ctx_ptr` by key, not the dict whole.
- The class comment reads "Three things ... none" now.

**nit** — `_stripe_width` has no class-level annotation anywhere; it is assigned in `RblnNixlWorkerBase.__init__` and read here. Its two neighbours in the block right above `_backend_extra` are the same shape and do have one:

```python
_use_rbln_nixl_backend: bool
_pending_kv_caches: dict[str, torch.Tensor] | None
```

Nothing is newly exposed — the two call sites already read it — but the property makes it the mixin's own, so `_stripe_width: int | None` beside those would match.

Suite not run on my side; no vllm 0.26 environment here.

rebel-jongwonlee (approved):
Approving at `263b4cf5`. Every point from both rounds is in, and the one you found along the way — `build_worker` handing the connector a `MagicMock` config, so the guard's `is None` was false and the adapter took a mock width — was the larger of the two. The harness now pins every attribute `__init__` reads off `vllm_config`, so that closes the family rather than the member.

CI is green on this revision, including `vLLM-RBLN pytest`, which covers the suite I could not run here.

One nit, not blocking:

`_stripe_width` has no class-level annotation. It is assigned in `RblnNixlWorkerBase.__init__` and read in `RblnNixlRegistrationMixin`, and its two neighbours of exactly that shape are declared in the block right above `_backend_extra`:

```python
_use_rbln_nixl_backend: bool
_pending_kv_caches: dict[str, torch.Tensor] | None
```

Nothing is newly exposed — both call sites already read it — but the property makes it the mixin's own, so `_stripe_width: int | None` beside those would match. Fine to fold into whatever touches this next.

rebel-cjlee (approved):

PR #1156, opened 2026-09-22
feature(config): let `--model-impl auto` pick the path from the model

rebel-cjlee:
### 🚀 Summary of Changes

`--model-impl auto` used to end at a fixed default, so a model `optimum-rbln` has no implementation for still started on the optimum model path and failed there. It now reads the model and takes the optimum path where `optimum-rbln` runs the architecture, and the vllm path where it does not.

Both model paths. The selector lives in `config.py`, which both share.

The read is the bottom rung of `resolve_model_impl`. A named `--model-impl`, a built `RBLNConfig` / `OptimumRBLNConfig`, the path a parent handed down and the deprecated `VLLM_RBLN_USE_VLLM_MODEL` all answer above it, so only a run that states nothing pays a config fetch.

What the read does, in order:

1. A directory holding `rbln_config.json` is a compiled `optimum-rbln` artifact, so optimum, with no config read at all. The correctness scripts under `tests/optimum/optimum_correctness/` pass exactly such a directory.
2. Otherwise `get_config()` with the arguments `ModelConfig` itself uses. `hf_overrides` is among them on purpose: an override can rewrite `architectures`, and a verdict read without it can contradict the config built a moment later.
3. An architecture in `optimum-rbln`'s registry goes optimum, anything else vllm.
4. A config that cannot be read falls back to the vllm model path with a warning. The two outcomes that reach that path carry different refusal wording, since an unreadable config and an unimplemented architecture are undone by different things.

Each outcome logs at info, naming the path and the reason, and only on the `auto` rung.

HF traffic does not grow in the test lanes. `tests/vllm` pins its path in the conftest, so the unit lane's hundred-plus config builds never reach the read, and the five engine-building sites in `tests/optimum` now name their path.

---

* Resolves #
* Related to #1047 

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

1. `uv run --no-sync pytest tests/vllm/test_rbln_config.py tests/vllm/test_platform.py`
  : 173 passed. `TestAutoReadsTheModel` covers the architecture verdict both ways, `hf_overrides` changing it, the artifact short circuit, the unreadable fallback with its warning, that a named or inherited path skips the read, that the deprecated variable still answers above the read, and both RBLN-CA* refusals.
2. `uv run --no-sync pytest tests/vllm` → 2451 passed, 26 skipped, 16 xfailed.
3. On hardware, with no `--model-impl` and no `VLLM_RBLN_*` exported:
   `LLM(model="Qwen/Qwen2.5-0.5B-Instruct", block_size=1024, max_model_len=1024, max_num_seqs=2)` compiles through optimum-rbln and generates. Running the same call against the artifact directory it left in `$VLLM_CACHE_ROOT/compiled_models/` resolves through the artifact short circuit instead.
4. A model optimum-rbln does not implement, such as MiniMaxAI/MiniMax-M2.7, takes the vllm path.

---

```
--model-impl auto takes the optimum model path: optimum-rbln runs ['Qwen2ForCausalLM'].
--model-impl auto takes the optimum model path: /…/Qwen_Qwen2.5-0.5B-Instruct_45578f1f holds a compiled optimum-rbln artifact.
--model-impl auto takes the vllm model path: optimum-rbln has no implementation for ['MiniMaxM2ForCausalLM'].
```

---

### 📋 Checklist

---

### 💬 Notes

`auto` cannot read an object-storage model (`s3://`, `gs://`, `az://`): vLLM mirrors it locally only after the path is chosen, so the read fails and the model falls back to the vllm model path. This project does not ship `runai_model_streamer`, so those models do not run here either way; when it does, the read should refuse and ask for an explicit `--model-impl`.

---


vllm_rbln/config.py
```diff
@@ -423,23 +426,110 @@ def resolve_model_impl(
...
+            # one; a caller resolving without an engine takes the documented
+            # default.
+            resolved = "optimum"
+        else:
+            resolved, selected_by, remedy = _model_impl_for(engine_args)
     _reject_disabled_model_impl(resolved, selected_by, remedy)
     return resolved
 
 
+def _model_impl_for(engine_args: Any) -> tuple[ModelImpl, str, str]:
```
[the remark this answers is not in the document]
rebel-cjlee: The rule is righ;t, but inlining ~55 lines would put "which rung answers" and "what the model says" in one function of over 130 lines, so this is a deliberate exception rather than an oversight.

vllm_rbln/config.py
```diff
@@ -423,23 +426,110 @@ def resolve_model_impl(
...
+    """
+    from vllm.transformers_utils.config import get_config
+    from vllm.transformers_utils.utils import maybe_model_redirect
+
+    # Owned by the optimum path, and imported here so that a run which named
+    # its path, and never reaches this, does not import it at all.
+    from vllm_rbln.utils.optimum.paths import is_compiled_dir
+    from vllm_rbln.utils.optimum.registry import is_supported_arch
+
+    model = maybe_model_redirect(engine_args.hf_config_path or engine_args.model)
```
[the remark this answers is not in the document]
rebel-cjlee: `auto` cannot resolve an object-storage model today, since vLLM mirrors it locally only after the path is chosen; this project does not ship `runai_model_streamer` either, so we will handle it when it is needed.

vllm_rbln/utils/optimum/registry.py
```diff
@@ -133,17 +133,22 @@ def is_enc_dec_arch(config: PretrainedConfig) -> bool:
...
+
+
 def is_arch_supported(
     config: PretrainedConfig, model_set: dict[str, tuple[str, str]]
 ) -> bool:
-    architectures = getattr(config, "architectures", [])
+    architectures = getattr(config, "architectures", None) or []
     return any(
         arch in _RBLN_SUPPORTED_MODELS and arch in model_set for arch in architectures
     )
```
rebel-eunji: Having both `is_supported_arch`and `is_arch_supported` looks weird..
I think integrating them or refactoring them is required.

vllm_rbln/config.py
```diff
@@ -423,23 +426,110 @@ def resolve_model_impl(
...
+            resolved = "vllm" if envs.VLLM_RBLN_USE_VLLM_MODEL else "optimum"
             selected_by = "VLLM_RBLN_USE_VLLM_MODEL"
             # Not `--model-impl optimum`: with the variable still set that is
             # the disagreement the refusal above this one is for.
             remedy = "Unset VLLM_RBLN_USE_VLLM_MODEL."
+        elif engine_args is None:
+            # Nothing to read the model off. `create_engine_config` always has
+            # one; a caller resolving without an engine takes the documented
+            # default.
+            resolved = "optimum"
```
rebel-eunji: Do we need this option?
It looks like engine_args is only ever None in the pytest code.

vllm_rbln/utils/optimum/registry.py
rebel-cjlee: It looks weird to me, too.
`model_set` is now optional and defaults to the full registry, so the new callers reads `is_arch_supported(config)` and `is_supported_arch` is gone.

vllm_rbln/config.py
```diff
@@ -428,23 +431,110 @@ def resolve_model_impl(
...
+    is disabled on this host, and they are empty for the optimum path, which no
+    host refuses. They differ per outcome because the two that reach the vllm
+    path are undone by different things.
+
+    Reading the model costs a config fetch that `ModelConfig` pays again a
+    moment later, so only `auto` reaches here. The arguments mirror the ones it
+    uses, `hf_overrides` included: those can rewrite `architectures`, which is
+    what this reads, and a fetch without them can name a path the built config
+    then contradicts.
+    """
```
rebel-eunji: nit: It is a bit long and verbose. Could you make them more concise?

vllm_rbln/config.py
rebel-cjlee: Right, it always has its engine args, so this is now an assert rather than a default and the one test that relied on it moves to passing an engine.

vllm_rbln/config.py
rebel-cjlee: agreed.

rebel-eunji (approved):
lgtm 👍 
However, do we need to prevent using vllm-path in ATOM when we use --model-impl auto?

rebel-jonghewk:
> lgtm 👍 However, do we need to prevent using vllm-path in ATOM when we use --model-impl auto?

As far as I kknow, if `auto` picks `vllm` it will still call `_reject_disabled_model_impl`. right?

rebel-cjlee:
@rebel-eunji @rebel-jonghewk Yes, it checks if device is CA variant right after resolving the model impl.

rebel-eunji:
Thanks. I got it. It was resolved in the previous PR.

PR #1158, opened 2026-09-23
feature(APC): enable sub-block prefix caching on hybrid sliding-window models.

rebel-tskim:
### 🚀 Summary of Changes

Sub-block prefix caching was refused for hybrid models, where a sliding-window layer shares its KV buffer with a full-attention layer (gpt-oss). This PR enables it.
With the sliding group on the window and the full group on the manager block, every full-layer copy fell back to a whole-buffer sync through the host, costing far more than recomputing the prompt. Putting both groups on the manager block made the copies stay on the device engine, at the cost of increased memory read size for sliding groups.

vLLM path only. Two changes:

* `prepare_kernel_block_sizes` gives a sliding-window group on the append kernel the manager block as its kernel block while sub-block caching is on. The shift kernel keeps its window, and so does the append kernel when sub-block caching is off, so nothing changes for a run without the feature.
* `_process_kv_cache_copy_ops` copies only the layers of the op's group and asserts that the group's kernel block is the manager block, which the block ids in a copy op refer to.

### Why the manager block
Upstream gives each sliding/full layer pair one tensor so a single block pool serves both groups. That tensor has one device layout, and the runtime plans a device-to-device copy only for a slice of the tensor the compiled graph reads; a slice through a view with another block size is rejected and the whole allocation is synced to the host instead. With the sliding group on the window (128) and the full group on the manager block (4096), every full-layer copy took that path. Putting both groups on the manager block gives the buffer one geometry, and the existing token-prefix copies then run on the device engine. The append kernel accepts this: its window is a compile-time argument independent of the cache's block size, and it gathers whole blocks around the window.

Measured on gpt-oss-120b, single NPU, sub-block size equal to the prefill chunk:

* Warm TTFT on a sub-block hit dropped by more than two orders of magnitude and is now a small fraction of the cold prefill at every prompt length tried.
* Cold prefill is unchanged.
* Decode is slower while sub-block caching is on: the sliding kernel gathers two 4096-token blocks around its 128-token window per step instead of two 128-token blocks. With decode buckets covering the live batch (`VLLM_RBLN_DECODE_BATCH_BUCKET_LIMIT`), the cost is 1 to 3.5% at batch one and about 5% at batch 8, independent of context length. A kernel that gathers at window granularity would remove it.
* Greedy tokens are identical between cold and warm requests, and identical between sub-block caching on and off.

Models without a shared buffer keep their kernel block and copy path exactly as before. Verified on the tiny MiniMax-M2.7 and DeepSeek-V3.2 variants against the base branch: same KV shapes, identical tokens and logprobs, TTFT and decode
within noise.

---

### ✅ Type of Change

* [x] ✨ Feature (`feature`)

---

### 🧪 How to Test

1. Unit tests:
   `uv run --no-sync pytest tests/vllm/v1/worker/test_utils.py::TestPrepareKernelBlockSizes tests/vllm/v1/worker/test_rbln_model_runner.py::TestProcessKvCacheCopyOps -q`
   Expected: all pass. Against the base branch the new sliding-rule cases and the group-filter test fail.
2. Serve gpt-oss-120b on the vLLM path with `--enable-prefix-caching`, `--block-size 4096`, `--max-num-batched-tokens 512`, `--rbln-sub-block-size 512`, `--max-num-seqs 1`. Send a prompt of a few  thousand tokens twice with `temperature 0` and `logprobs`.
Expected: the second request's TTFT is a small fraction of the first, its tokens and logprobs match the first exactly, and `vllm:prefix_cache_hits_total` grows by a multiple of 512 beyond the whole-block hits.
3. Same serve with `--no-rbln-enable-sub-block-cache` and no sub-block size.
Expected: the KV base shape in the log shows 128-token blocks again, hits land only on whole 4096 blocks, and greedy tokens for the same prompt match step 2.
4. Edge case: a model without a shared buffer (single KV cache group) with sub-block caching on.
Expected: KV shape, hit counts and tokens identical to the base branch.


vllm_rbln/v1/worker/rbln_model_runner.py
rebel-cjlee, line 3205: Not a blocking, just checking.

The description says the shift kernel keeps its window, but I could not find what happens when it keeps the window and sub block caching is on. `RBLNSlidingWindowSpec` still passes `can_use_sub_block_caching`, and this guard was the only thing stopping that pair. On a gpt-oss shaped config (block_size 1024, window 128) `prepare_kernel_block_sizes` gives me `[1024, 128]`, so the new assert fiers on the first copy op for that group.

Do we still want to manage that, or let it go? It only covers CR03 now that #1147 turned the path off on CA.

vllm_rbln/v1/worker/utils.py
```diff
@@ -1108,10 +948,18 @@ def prepare_kernel_block_sizes(
...
+            ):
+                # The shift kernel holds one window per block; the append kernel
+                # gathers whole blocks around the window, whatever their size.
+                kernel_block_sizes.append(kv_cache_spec.sliding_window)
+            else:
+                # A sub-block copy slices the buffer the compiled graph reads,
+                # and this layer shares it with a full-attention layer on the
+                # manager block: the append kernel takes the same block, so the
+                # slice is one geometry for both.
+                kernel_block_sizes.append(kv_cache_group.kv_cache_spec.block_size)
```
rebel-cjlee, line 962: Not blocking, and the "Why the manager block" section answers the choice itself, so just one thing I could not place.

This makes the compiled KV shape of the sliding layers depend on the option, which is also what step 3 of How to Test says: turn it off and the log shows 128 token blocks again. But `compute_hash` still drops `enable_sub_block_cache` from the bundle key, with the comment "Sub-block caching changes what runs, not what is built", and `_warmup_graph_set_factors` does not carry it either.

So a run with the feature on and a run with it off write to the same bundle file. Is it fine?

rebel-cjlee (approved):

vllm_rbln/v1/worker/utils.py
rebel-tskim: Current implementation will cause silent full rebuild on the sub block cache option toggle, only for the hybrid models. Adding the subblock cache option as hash key will cause non-hybrid models to go through unnecessary recompilations. I suggest we leave this as a known issue for now and revisit when needed.

vllm_rbln/v1/worker/rbln_model_runner.py
rebel-tskim: Turning on sub-block caching on CR03 will indeed cause AssertionError after the first cache hit, unlike it was guarded at startup before. However given that CR03 is being deprecated and we still get a AssertionError somewhere rather than silent functionality regression, I suggest that we just let this go.

rebel-eunji:
It's just a suggestion.
Let's go with vllm-path or vllm ~~ , as we decided, and drop vllm-native.

rebel-tskim:
@rebel-eunji Good catch! Updated PR message.

PR #1159, opened 2026-09-23
fix(platform): point optimum-path unsupported asserts at --model-impl vllm

rebel-jonghewk:
#### Summary

The optimum model path refuses tensor parallelism, pipeline parallelism, and speculative decoding. The messages said "not supported in optimum-rbln" or pointed at `--rbln-num-devices-per-local-rank`. None of them told the user that the vllm model path supports all three. Each assert now names the optimum model path and says to set `--model-impl vllm`. The TP assert still offers `--rbln-num-devices-per-local-rank` as the optimum-side option.

#### Details

- **Assert messages** — `check_and_update` in `platform/optimum_impl.py` keeps the same three checks. Only the message strings change. The wording matches the existing async-scheduling warning, which already says "Set --model-impl vllm".

- **TP hint** — The TP message lists two options. Use `--rbln-num-devices-per-local-rank` to compile the model with RSD on the optimum model path, or switch to `--model-impl vllm`.

#### Next Steps

- RBLN-CA* disables the vllm model path (#1147), so on that chip the hint points at a path that is refused. A chip-aware message can come in a follow-up if this matters.

---

* Related to #

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. `vllm serve <optimum model> --tensor-parallel-size 2` (or `--pipeline-parallel-size 2`, or `--speculative-config ...`)
2. Verify that the AssertionError says to set `--model-impl vllm`

---

### 📋 Checklist


rebel-eunji (approved):
lgtm

PR #1157, opened 2026-09-22
fix(core): keep the pipeline-parallel handoff in the model dtype

rebel-dkhong:
### 🚀 Summary of Changes

A pipeline-parallel run of DeepSeek-V3.2 with `VLLM_RBLN_USE_W8A8=1` dies on the first forward pass:
```
AssertionError: previous stage sent [('hidden_states', (1, 512, 7168), torch.float32),
                                     ('residual',      (1, 512, 7168), torch.bfloat16)],
                this stage expects  [('hidden_states', (1, 512, 7168), torch.bfloat16),
                                     ('residual',      (1, 512, 7168), torch.bfloat16)]
```

`recv_intermediate_tensors()` allocates its buffers from `model_config.dtype`, so every stage expects bf16.

The sending stage's `hidden_states` is fp32 at runtime because the compiled W8A8 MoE op returns fp32.

Casting `hs_scale` to `compute_dtype` puts the compute dtype back on the operand the compiler reads, so the op returns bf16 and the handoff matches.

> *What does this PR do? What feature, fix, or improvement does it bring?*

---

### 📌 Related Issues / Tickets

* Resolves #
* Related to #

---

### ✅ Type of Change

* [x] 🧬 Core engine changes (`core`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

### 📸 Screenshots / Logs (if applicable)

---

### 📋 Checklist

* [ ] PR title follows Conventional Commits format
* [ ] This PR is linked to an existing issue
* [ ] The test method is described, and the expected result is clearly stated
* [ ] Relevant documentation has been updated (if applicable)

---

### 💬 Notes

---

PR #1160, opened 2026-09-23
fix(fp8): hand the W8A8 MoE kernel its activation scale in the compute dtype

rebel-thkim:
### 🚀 Summary of Changes

The W8A8 branch of `Fp8MoEMethod.apply` handed `custom_moe_glu_w8a8` a **fp32** per-(token, K-block) activation scale. The op's `register_fake` declares the output in `compute_dtype` (bf16), but the compiler never sees `compute_dtype`: the `moe_glu_w8a8` converter drops it and binds the kernel's `__DTYPE__` (scales, accumulator **and output**) from the `hidden_states_scale` operand. So the compiled kernel emitted fp32 hidden_states while the traced graph believed bf16.

Inside a graph the next norm swallows the fp32 tensor (W8A8 dp8 passes). Under pipeline parallelism the last MoE output of a stage *is* the graph output, so the fp32 leaks into the PP handoff and the next stage rejects it:

```
AssertionError: previous stage sent [('hidden_states', (1, 512, 7168), torch.float32), ('residual', (1, 512, 7168), torch.bfloat16)],
                this stage expects  [('hidden_states', (1, 512, 7168), torch.bfloat16), ('residual', (1, 512, 7168), torch.bfloat16)]
```

Fix: round the scale to `compute_dtype` before quantizing and quantize with that rounded value, so the kernel's `__DTYPE__` matches the fake's declared dtype and dequantizes with exactly the scale the values were divided by. This also restores the kernel's documented fp16/bf16 mode (`input_scale`: "activation scale (fp16)", `layer_out`: fp16 accumulator), matching W8A16 and the eager reference.

---

* Related to vllm-dynamo-ci-v2 build 410, `[3.4] Deepseekv3.2-b1-pp8-W8A8-long-kv8-CR13` (fails identically in builds 391 and 406; never green): https://obedients.k8s.rebellions.in/orgs/main/pipelines/standalone-rebel-compiler-vllm-dynamo-ci-v2/builds/410?job=741d23aa-9aa0-4a4e-9dfc-1dfa5e61be90
* Receiver-side check introduced by #1060

---

### ✅ Type of Change

* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. CI case 3.4: `vllm-decoderonly deepseek-v3 -m deepseek-v3.2 -pp 8 -ep -pcs 512 --block-size 2048 -b 8 --max-tokens 256 -kvd fp8 --prompt-set longbench --min-prompt-len 2048 --max-prompt-len 4096 --run-iter 5` with `VLLM_RBLN_USE_W8A8=1`. Expected: PP1 no longer asserts in `recv_intermediate_tensors`; the run completes and judge.py gates on pearson.
2. Regression: W8A8 dp8 cases (3.3) must still pass the pearson gate — see Notes.

---

### 📋 Checklist

---

### 💬 Notes

* **Numerics change for every W8A8 run, not only PP.** The kernel moves from the accidental fp32 mode to bf16 (scales, dequant, accumulation, output). W8A16 already runs in this mode and passes its gates, and bf16 rounding of the scale is < 0.4 % against the ~6 % e4m3 step, but W8A8 dp8 goldens should be re-checked in the same CI run.
* Follow-ups outside this PR: (a) rebel_compiler's `moe_glu_w8a8` converter should assert `input_types[1] == compute_dtype` at compile time so a fake/kernel dtype disagreement can never pass silently again; (b) after the worker died, EngineCore sat in `shm_broadcast.acquire_write` waiting on dead readers until the 90-minute CI timeout (3 attempts, 4.5 h) — the executor shutdown patch calls `mq.shutdown()` but the writer loop does not observe it.


rebel-cjlee (approved):

rebel-eunji:
https://github.com/RBLN-SW/vllm-rbln/pull/1157 also handles the same issue.
I've asked @rebel-hishin and @rebel-jaehob to sync with the kernel.

PR #1161, opened 2026-09-23
fix(nixl): recover a KV read from a dead RDMA link on either side

rebel-tskim:
### 🚀 Summary of Changes

This PR makes the RBLN NIXL connector survive an RDMA link that dies under a running P/D pair, on either side, to keep up with the HA requirements. Four things that were each fatal or permanent before:

- **(Bugfix) A failed read is reported once.** This is a upstream backport (to be removed later). A request holds one transfer handle per producer stage, and on a dead link they fail in different polls. Upstream reported the request once per failing handle; the second report reached the engine core after the first had popped the request's metadata and tripped its assertion, taking the consumer down. `_pop_done_transfers` now reports only requests whose metadata is still present, and `_handle_failed_transfer` leaves stragglers with the handle cleanup alone. Guarded by `TODO(vllm-project/vllm#54518)`, to be removed with v0.30.0 bump-up.
- **(Bugfix) A full local prefix hit no longer crashes the failure path.** `local_block_ids` is empty then, and indexing it raised inside the handshake future's callback, so the request was never reported and the client got a 500.
- **(Feat) A dead peer is re-handshaked.** Upstream re-dials only an engine it does not know, and a dead peer stays known: every read refreshes its TTL on the way to failing, so the consumer recomputed forever. A peer whose transfer failed is dropped once none of its handles is in flight, and the next request dials it again.
- **(Feat) Every local link down is handled as this side's fault.** The worker polls the physical links in `/sys/class/net` once a second. While all are down nothing is dialled: a `loadRemoteMD` attempted from a dead link leaves nixl core a half-loaded remote that fails every later handshake with `NIXL_ERR_INVALID_PARAM`, so the consumer would never recover once the link returns. The scheduler's handshake listener answers peers with an empty payload frame, which the peer's `_query_agent_meta` reads as a refusal, so a consumer fails at once instead of waiting out the RDMA CM timeout on a dial that cannot succeed. A producer with `kv_connector_extra_config.link_down_exit_s` set exits after that many seconds so the orchestrator replaces it; a consumer rejects the option, since it holds every running generation and serves on recompute. The scheduler reports pending push work while the links are down so an idle producer keeps stepping and actually reaches the exit.

---

* Resolves #
* Related to #1139 

---

### ✅ Type of Change

* [ ] 🚀 Release (`release`)
* [x] ✨ Feature (`feature`)
* [x] 🛠 Bug fix (`fix`)

---

### 🧪 How to Test

1. Unit tests:
   ```bash
   uv run --no-sync pytest tests/vllm/distributed/kv_connector/ -x
   ```
   Every new test fails on `dev` without the source commit and passes with it.
2. Hardware test can be done by dropping NICs in PD disaggregation setups - https://github.com/rebellions-sw/coding-assistant/issues/962, https://github.com/rebellions-sw/coding-assistant/issues/963
3. Edge cases covered by unit tests: a spare live link is not "down"; a host with no physical NIC is never "down"; a veth's state is ignored; a straggling handle whose request already failed releases its handle and reports nothing; a peer with a live handle is not dropped until it drains.


rebel-tskim:
#1139 must land first, and this PR should be rebased on top of the change.

rebel-jongwonlee:
Read at `0f6e9ac8` against `dev`. Static review plus the pinned vLLM's source; no vllm 0.26 environment here, so the suite was not run and nothing below is from a live link.

**Verified first.**

- The `_pop_done_transfers` filter matches what upstream requires. `get_finished` asserts the metadata is still present for everything it reports — `base_worker.py:1985`, `assert meta is not None, f"{req_id} not found in recving_metadata list"` — so a straggler whose entry was already popped must not reach `done_recving`. Moving `_failed_recv_reqs.put` inside the `meta is not None` branch is the same invariant approached from the other end, and the two agree.
- The failed `Future` from `_ensure_handshake` lands where it should. `_background_nixl_handshake` attaches `request_ready`, which calls `f.result()` inside a `try` and routes the exception to `_handle_failed_transfer` (`base_worker.py:913-927`), so the request fails at once instead of dialling, and there is no path that leaves the exception unretrieved.
- The listener's reply matches upstream's shape, `send_multipart((identity, b"", payload, ts))`, and an empty frame cannot be confused with a real payload.

---

## A blockless read is still reported, and that is the half that kills the engine

The `IndexError` is gone. What follows it is not:

```python
if (meta := self._recving_metadata.get(req_id)) is not None:
    if not self._is_hma_required and meta.local_block_ids:
        self._invalid_block_ids.put(set(meta.local_block_ids[0]))
    self._failed_recv_reqs.put(req_id)        # <- still runs
```

`get_finished` merges failures into the set it hands the scheduler (`base_worker.py:1968`, `done_recving.update(failed_recv_reqs)`), and the scheduler walks it like this:

```python
# vllm/v1/core/sched/scheduler.py, _update_from_kv_xfer_finished
for req_id in kv_connector_output.finished_recving or ():      # :2637
    assert req_id in self.requests                             # :2639
    req = self.requests[req_id]
    if req.status == RequestStatus.WAITING_FOR_REMOTE_KVS:
        self.finished_recving_kv_req_ids.add(req_id)
    else:
        assert RequestStatus.is_finished(req.status)           # :2644
```

A full local prefix hit never asked for an async load. Upstream returns `(0, False)` from `get_num_new_matched_tokens` when the count is not positive, so the request is never put in `WAITING_FOR_REMOTE_KVS` — it is `RUNNING` and generating. It takes the `else` and fails `:2644`.

It reaches the failure path at all because upstream keeps it deliberately:

```python
# pull_scheduler.py, update_state_after_alloc
# If remote_blocks and num_external_tokens = 0, we have
# a full prefix cache hit on the local node. We need to call
# send_notif in _read_blocks to free the memory on the remote node.
unhashed_local_block_ids = (
    blocks.get_unhashed_block_ids_all_groups() if num_external_tokens > 0 else ()
)
```

so it is a notify-only entry — no handle, no bytes, `pull_worker.py:277` sends the notification and returns — carried only so the producer can free its blocks. Nothing is waiting on it, which is why the scheduler has no state for it to land in.

So the change turns a 500 into an EngineCore death: later, and further from the cause. `test_a_failed_handshake_with_no_local_blocks_still_reports` asserts `w._failed_recv_reqs.get_nowait() == "r0"`, so the suite pins the report rather than catching it.

## And the placement will disable the guard that fixes it

#1139 has two approvals and fixes exactly this, by returning before the report. It puts `_handle_failed_transfer` on `RblnNixlWorkerState` (`state.py`); this PR puts one on `RblnNixlTransferMixin` (`transfer.py`). All three mixins derive from `RblnNixlWorkerState`, so the worker resolves them in this order:

```python
class RblnNixlWorkerBase(
    RblnNixlHandshakeMixin,
    RblnNixlRegistrationMixin,
    RblnNixlTransferMixin,      # <- this PR
    NixlBaseConnectorWorker,
):
# MRO: ... -> RblnNixlTransferMixin -> RblnNixlWorkerState -> NixlBaseConnectorWorker
```

`RblnNixlTransferMixin` comes first and this override does not call `super()`, so once #1139 is in `dev` a rebase leaves its guard unreachable — no conflict marker, no failing case here, `:2644` back. The two PRs also grew separate machinery for the same goal, dropping a dead peer so it re-handshakes: #1139 from a heartbeat failure, this one from `_engines_to_rehandshake` draining in `get_finished`. Worth landing as one method and one mechanism rather than two that shadow each other.

---

## `every_local_link_down()` can raise out of the paths that exist to survive a link failure

```python
states = [
    (dev / "operstate").read_text().strip()
    for dev in _SYS_CLASS_NET.iterdir()
    if (dev / "device").exists()
]
```

`iterdir`, `exists` and `read_text` are separate syscalls against a tree the kernel mutates. A device that goes away between them — which is what a link teardown can do — raises `FileNotFoundError` out of `get_finished` on the worker and out of `has_pending_push_work` on the scheduler. Both are per-step calls, so the result is a crash rather than a skipped poll, in the one function whose job is to notice that links died. Treating an unreadable device as absent keeps it total.

**The scheduler also polls it unthrottled.** The worker budgets the read:

```python
if now - self._link_checked_at >= LINK_POLL_S:
```

`has_pending_push_work` has no such guard and runs every step, so the scheduler walks `/sys/class/net` and reads one file per NIC per step on its hot path, while the worker — which set the 1 s budget — does it once a second. The listener thread reads it per handshake as well. One throttled reader the three share would match the worker's own intent.

---

## Smaller

- `_nixl_handshake_listener` is a verbatim copy of upstream's with the payload line changed, but it drops the unexpected-message check:

  ```python
  msg, target_pp_rank, target_tp_rank = msgspec.msgpack.decode(msg)
  if msg != GET_META_MSG:
      logger.warning("Connection listener got unexpected message %s", msg)
  ```

  The copy binds that element to `_`. The `_pop_done_transfers` override carries `TODO(vllm-project/vllm#54518)` naming what it backports and when it goes; this copy names no revision, so nothing marks it as upstream's to re-sync. The same treatment would make both checkable.

- `meta.remote.engine_id` is dereferenced unguarded in `_handle_failed_transfer` and in `get_finished`'s `busy` comprehension. `meta.remote` is typed optional and every other site in this package asserts it first (`pull_worker.py:41`, `push_worker.py:105`). An assert would match the file.

- `self._exit_on_link_down = extra.get("link_down_exit_s", 0) > 0` compares whatever the config holds against an int, so a string raises a `TypeError` at scheduler construction rather than a message naming the knob. The worker's role check a few lines away is the right precedent.

PR #1162, opened 2026-09-23
other(test): gate the lock on the OS matrix we ship against

rebel-junamsong:
### 🚀 Summary of Changes

`tests/test_os_support.py` checks that `uv.lock` has an **installable distribution**
for every OS release we ship against — base set and the `kv_connectors` extra —
and joins the always-on packaging-guards gate.

| target | glibc | CPython |
|---|---|---|
| Ubuntu 22.04 | 2.35 | 3.10 |
| Ubuntu 24.04 | 2.39 | 3.12 |
| Ubuntu 26.04 | 2.40 (cap) | 3.14 |
| RHEL 9 | 2.34 | 3.11 |
| RHEL 10 | 2.39 | 3.12 |

#### Why this needs a gate

glibc is the binding axis, and it is invisible from a lane that runs on the build
host's own glibc. `torch-rbln`, `vllm`, `nixl-rbln`, `rebel-compiler`, `lmcache` and
`cryptography` all ship **manylinux_2_34**, which **RHEL 9 meets exactly**. A
dependency that raises its floor one step breaks the oldest target — and breaks it
quietly, as a source-build fallback rather than an error.

#### uv answers, not a reimplementation

```
uv sync --locked --python 3.11 --python-platform x86_64-manylinux_2_34 \
        --extra kv_connectors --no-install-project --no-build --dry-run
```

`--python-platform` and `--python` model the release's glibc and interpreter,
`--no-build` turns "would build from source" into a failure, `--dry-run` downloads
nothing and creates no environment. The resolver answering is the one that performs
the real install, so markers, `resolution-markers`, ABI and platform tags are all
handled by uv.

Hand-rolled tag math was tried first and was wrong twice: it triple-counted `scipy`
(three lock entries, one per Python range) and flagged `pyyaml-ft` as missing on all
five targets when `libcst` only requires it on `python_full_version == '3.13.*'`,
where it does ship a wheel. Both were false alarms; the same class of bug produces
false confidence just as easily.

#### The test bites — both axes checked against a negative control

| probe | result |
|---|---|
| glibc lowered to 2.28 (RHEL 8-class) | fails on `torch-rbln==0.5.0rc1` |
| CPython 3.9 | rejected by `requires-python` |
| CPython 3.14 | fails on `numpy==2.2.6` (no cp314 wheel) |

#### ubuntu-26.04 is a strict xfail

That row fails today: `numpy==2.2.6` publishes no cp314 wheel. 26.04 is also
unreleased, so its glibc and CPython are provisional. `xfail(strict=True)` keeps the
row honest — when the lock carries a 3.14-installable numpy the row turns **red as an
unexpected pass**, forcing the mark off rather than letting a stale exemption sit.

#### Scope

Complements the two guards beside it rather than replacing them:
`test_packaging_bounds.py` covers RBLN version ceilings, `test_python_support.py`
covers every minor `requires-python` admits, and this covers the specific
(glibc, CPython) pairs our OS targets actually present.

This proves a compatible distribution **exists** per target. It does not prove the
package imports and runs there — manylinux guarantees glibc symbols, not what a wheel
`dlopen`s at runtime. A container matrix would be the way to cover that, and is not
in this PR.

---

* Raised alongside the `rbln-serve` image work, where RHEL 9 and the Ubuntu LTS
  releases are the deployment targets.

---

### ✅ Type of Change

* [x] ❓ Other (`other`): test coverage

---

### 🧪 How to Test

```bash
uv run --no-project --python 3.12 --with pytest --with packaging \
  python -m pytest tests/test_os_support.py -v
```

Expected: `13 passed, 2 xfailed`. The full gate
(`test_packaging_bounds.py` + `test_python_support.py` + this) is
`24 passed, 2 xfailed in 0.75s`.

To see it bite, drop `x86_64-manylinux_2_34` to `x86_64-manylinux_2_28` in
`OS_MATRIX` for `rhel-9`: the base row fails on `torch-rbln`.

---

### 📋 Checklist

* [x] Relevant unit tests have been added

---

### 💬 Notes

Affects CI and tests only; neither model path changes.

One cost worth knowing: uv resolves a real interpreter per row, so a runner without
3.10/3.11/3.14 cached downloads them once into `UV_CACHE_DIR` (`/mnt/uv_cache`, shared).
Steady-state the whole file is under a second.

The Ubuntu 26.04 glibc is entered as `2_40` because that is uv's highest
`--python-platform` value. Testing a newer glibc at that cap is sound — a higher glibc
admits strictly more wheels — but it does mean the row cannot detect a wheel that
requires something above 2.40.


tests/test_os_support.py
```diff
@@ -0,0 +1,74 @@
...
+    pytest.param(platform, python, id=name, marks=marks)
+    for name, platform, python, marks in OS_MATRIX
+]
+PYTHONS = [pytest.param(python, id=name) for name, _, python, _ in OS_MATRIX]
+
+
+@pytest.mark.parametrize("platform, python", INSTALLS)
+@pytest.mark.parametrize("extras", [(), ("kv_connectors",)], ids=["base", "kv"])
+def test_locked_stack_installs(platform, python, extras, tmp_path) -> None:
+    done = subprocess.run(
```
rebel-junamsong, line 55: `xfail(strict=True)` lands on both `extras` params, but the reason is base-set only (numpy has no cp314 wheel). A kv_connectors-only break on 26.04 would be masked, and `strict` can't catch it — still failing either way.

Narrow the mark to the base param, or say in `UNRELEASED` that it covers both.

rebel-junamsong (commented):
Letting uv answer instead of reimplementing tag math is the right call, and the two false alarms in the body are the evidence for it.

Verified the mechanism the gate rests on:

```
--no-build on a source-only dist   -> exit 2, "has no binary distribution"
--python 3.10, UV_PYTHON_DOWNLOADS=never, interpreter absent -> resolves fine
requires-python = ">=3.10,<3.15"   -> admits all five matrix entries
```

So `--dry-run` needs no interpreter and no environment, which is what makes this cheap enough for the always-on lane.

One thing inline.

